PR #1, opened 2018-07-12
Bump Cockpit test API to 172

martinpitt:
No API changes.


martinpitt:
The test failure is unrelated. Our recent fedora-28 image refresh brought in a podman version that broke its varlink interface; I can also reproduce this on master. So I'll land this as the damage is already done.

PR #3, opened 2018-07-26
Implement listing images/containers using React JSX

QiWang19:
Signed-off-by: Qi Wang <qiwan@redhat.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,21 +11,23 @@
   },
   "devDependencies": {
     "babel-core": "^6.25.0",
-    "babel-loader": "^7.0.0",
+    "babel-loader": "^7.1.1",
     "babel-plugin-transform-react-jsx": "^6.24.1",
     "babel-preset-env": "^1.5.2",
     "chrome-remote-interface": "^0.25.5",
     "compression-webpack-plugin": "~1.0.0",
     "copy-webpack-plugin": "~3.0.1",
     "css-loader": "^0.28.11",
-    "eslint": "^3.0.0",
-    "eslint-loader": "~1.6.1",
+    "eslint": "^3.19.0",
+    "eslint-loader": "^1.9.0",
     "eslint-plugin-react": "~6.9.0",
     "extract-text-webpack-plugin": "^2.1.0",
     "htmlparser": "^1.7.7",
     "jed": "^1.1.1",
-    "jshint": "~2.9.1",
-    "jshint-loader": "~0.8.3",
+    "jshint": "^2.9.5",
+    "jshint-loader": "^0.8.4",
+    "less": "^3.8.0",
+    "less-loader": "^4.1.0",
     "po2json": "^0.4.5",
     "sass-loader": "^7.0.3",
     "sizzle": "^2.3.3",
@@ -33,7 +35,18 @@
     "webpack": "^2.6.1"
   },
   "dependencies": {
+    "bootstrap": "^4.1.1",
+    "bootstrap-less": "^3.3.8",
+    "moment": "^2.22.2",
     "node-sass": "^4.9.0",
-    "react-lite": "0.15.30"
+    "patternfly": "^3.54.1",
+    "prop-types": "^15.6.1",
+    "react": "16.0.0",
+    "react-addons-css-transition-group": "^15.6.2",
+    "react-addons-transition-group": "^15.6.2",
+    "react-dom": "16.0.0",
+    "react-modal": "^2.2.2",
+    "reactstrap": "^6.2.0",
+    "strict-loader": "^1.2.0"
   }
 }
--- a/test/check-application
+++ b/test/check-application
@@ -14,35 +14,50 @@ import testlib
 
 
 class TestApplication(testlib.MachineCase):
+
     def testBasic(self):
         b = self.browser
         m = self.machine
 
         self.login_and_go("/podman")
+        b.wait_present("#app")
+        b.wait_present(".content-filter div")
+        b.wait_present("#containers-images")
+        b.wait_in_text("#containers-images", "busybox:latest")
 
-        # some version number
-        b.wait_present("#version")
-        b.wait_not_in_text("#version", "unknown")
-        self.assertRegex(b.text("#version"), "podman version: [0-9.]+")
+        # show image listing toggle
+        b.wait_present('#containers-images tr:contains("busybox:latest")')
+        b.wait_visible('#containers-images tr:contains("busybox:latest")')
+        b.click('#containers-images tbody tr:contains("busybox:latest") td.listing-ct-toggle')
+        b.wait_present('#containers-images tbody tr:contains("busybox:latest") + tr button.btn-delete')
 
-        # list images
-        b.wait_present("#images")
-        b.wait_in_text("#images", "busybox:latest")
-        b.wait_in_text("#images", "alpine:latest")
-
-        # no containers
-        self.assertFalse(b.is_present("#containers li"))
-
-        # create a container
-        m.execute("podman run -d --name swamped-crate busybox sleep 1000")
+        # run a container (will exit immediately)
+        m.execute("podman run -d --name swamped-crate busybox sh")
 
         # FIXME: UI needs to listen to change signals and update automatically
         b.reload()
         b.wait_present("#content")
         b.enter_page("/podman")
-        b.wait_present("#containers")
+        b.wait_present("#containers-containers-filter button")
 
-        b.wait_in_text("#containers", "swamped-crate (docker.io/library/busybox:latest): sleep 1000")
+        # show all containers to interact with stopped ones
+        b.click("#containers-containers-filter button")
+        b.wait_visible("#containers-containers-filter .dropdown-menu")
+        b.click("#containers-containers-filter li[data-data='all'] a")
+        b.wait_present("#containers-containers")
+        b.wait_present('#containers-containers tr:contains("busybox:latest")')
+        b.wait_visible('#containers-containers tr:contains("busybox:latest")')
+
+        # show container listing toggle
+        b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
+        b.wait_present('#containers-containers tbody tr:contains("busybox:latest") + tr button.btn-delete')
+
+        b.click("#containers-containers-filter button")
+        b.wait_present("#containers-containers-filter .dropdown-menu")
+        b.click("#containers-containers-filter li[data-data='running'] a")
+        b.wait_present("#containers-containers")
+        b.wait_in_text("#containers-containers", "No running containers")
+
 
 if __name__ == '__main__':
     testlib.test_main()
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -15,7 +15,7 @@ const builddir = (process.env.SRCDIR || __dirname);
 const distdir = builddir + path.sep + "dist";
 const section = process.env.ONLYDIR || null;
 const nodedir = path.resolve((process.env.SRCDIR || __dirname), "node_modules");
-
+const libdir = path.resolve(srcdir, path.sep + "lib");
 /* A standard nodejs and webpack pattern */
 var production = process.env.NODE_ENV === 'production';
 
@@ -112,11 +112,6 @@ module.exports = {
     externals: externals,
     output: output,
     devtool: "source-map",
-    resolve: {
-        alias: {
-            "react$": path.resolve(nodedir, "react-lite/dist/react-lite.js")
-        }
-    },
     module: {
         rules: [
             {
@@ -154,9 +149,14 @@ module.exports = {
             },
             {
                 exclude: /node_modules/,
-                loader: extract.extract('css-loader!sass-loader'),
-                test: /\.scss$/
-            }
+                loader: extract.extract('css-loader!sass-loader') ||
+                        extract.extract("css-loader?minimize=&root=" + libdir),
+                test: /\.s?css$/
+            },
+            {
+                test: /\.less$/,
+                loader: extract.extract("css-loader?sourceMap&minimize=!less-loader?sourceMap&compress=false")
+            },
         ]
     },
     plugins: plugins
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -15,7 +15,8 @@
     "rules": {
         "react/jsx-uses-vars": "error",
         "no-console": "off",
-        "no-undef": "error"
+        "no-undef": "error",
+        "no-unused-vars": "warn"
     },
     "globals": {
         "require": false,
--- a/lib/cockpit-components-listing.jsx
+++ b/lib/cockpit-components-listing.jsx
@@ -1,0 +1,368 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+"use strict";
+import PropTypes from 'prop-types';
+let React = require('react');
+require('./listing.less');
+
+/* entry for an alert in the listing, can be expanded (with details) or standard
+ * rowId optional: an identifier for the row which will be set as "data-row-id" attribute on the <tr>
+ * columns list of columns to show in the header
+ *     columns to show, can be a string, react component or object with { name: 'name', 'header': false }
+ *     'header' (or if simple string) defaults to false
+ *     in case 'header' is true, <th> is used for the entries, otherwise <td>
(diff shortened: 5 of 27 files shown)


QiWang19:
@martinpitt PTAL

src/ContainerDetails.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import cockpit from 'cockpit';
+
+const _ = cockpit.gettext;
+class ContainerDetails extends React.Component {
+    constructor(props) {
+        super(props);
+    }
```
martinpitt: This component has no state, so it should be a  [functional component](https://reactjs.org/docs/components-and-props.html#functional-and-class-components). ([another interesting blog post](https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc)). Also, it's custom to explicitly spell out the properties that you support/expect, like this:
```js
const ContainerDetails({container}) => (
    <div> ...
        <dt>{ container.Created.. }</dt>
    </div>
);
```

package.json
```diff
@@ -11,29 +11,45 @@
...
     "sizzle": "^2.3.3",
     "stdio": "^0.2.7",
     "webpack": "^2.6.1"
   },
   "dependencies": {
+    "bootstrap": "^4.1.1",
+    "bootstrap-less": "^3.3.8",
+    "create-react-class": "^15.6.3",
+    "exenv": "^1.2.2",
+    "jquery": "^3.3.1",
```
martinpitt: Please avoid introducing jQuery in new code, and particularly not with react. I don't see it actually being used here? I think this can just be dropped. Likewise, please comb the list for other unused stuff, like "exenv" -- it just unnecessarily enlarges `npm install` time, and all of these dependencies introduce maintenance. `react-create-class` also sounds like dead weight -- use proper ES6 classes instead.

src/ContainerDetails.js
```diff
@@ -0,0 +1,28 @@
...
+
+
+    render() {
+        let container = this.props.container;
+        return (
+            <div className='listing-ct-body'>
+                <dl>
+                    <dt>{_("Id")}      </dt> <dd>{ container.ID }</dd>
```
martinpitt: Please avoid the extra whitespace in the element. Perhaps they are not rendered, but it might still cause some layout issues. Also please be consistent with putting the `<dd>`s into a new line.

"Id" should be spelled "ID", as it's an acronym.

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+import Listing from '../lib/cockpit-components-listing.jsx';
+import cockpit from 'cockpit';
+
+import ContainerDetails from './ContainerDetails.js';
+const _ = cockpit.gettext;
+
+class Containers extends React.Component {
+    constructor(props) {
+	super(props);
+	console.log(props.containers);
```
martinpitt: Debugging leftover.

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+    //TODO
+    restartContainer(container) {
+	return undefined;
+    }
+
+    getContainerState(state) {
+	if (state.Status)
+	    return state.Status;
+	if (state.Running)
+	    return "running";
```
martinpitt: Needs i18n (`_("")` wrapper)

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    //TODO
+    restartContainer(container) {
+	return undefined;
+    }
+
+    getContainerState(state) {
+	if (state.Status)
+	    return state.Status;
```
martinpitt: This is confusing. Why is it sometimes a free-form `Status` and sometimes a fixed `state.Paused`? This deserves a comment, or only using one of them if it is redundant.

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+	if (state.Running)
+	    return "running";
+	if (state.Paused)
+	    return "paused";
+	if (state.Restarting)
+	    return "restarting";
+	if (state.FinishedAt && state.FinishedAt.indexOf("0001") === 0)
```
martinpitt: This "magic" string and comparison needs a comment.

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+	    return "created";
+	return "exited";
+    }
```
martinpitt: missing empty line between methods (same error several times above).

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+    renderRow(container) {
+	let isRunning = !!container.State.Running;
+	//TODO: check container.Image == container.ImageID
+	let image = container.ImageName;
+	let state = this.getContainerState(container.State);
```
martinpitt: supposedly these should be  `const`

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+
+	let columns = [
+	    { name: container.Name.replace(/^\//, ''), header: true },
```
martinpitt: What's wrong with showing slashes in the name?

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+	    image,
+	    // util.render_container_cmdline(container),
+	    container.Config.Entrypoint === "" ? container.Config.Cmd.join(" "):container.Config.Cmd.join(""),
```
martinpitt: same value in both cases, and missing spaces around operators. You probably should enable eslint for this project. (And I'll look into doing that for the starter kit)

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+	];
+	let tabs = [{
+	    name: _("Details"),
+	    renderer: ContainerDetails,
+	    data: { container: container }
+	}];
+
+	let startStopActions = [];
+	if (isRunning)
+	    startStopActions.push({ label: _("Stop"), onActivate: this.stopContainer.bind(this, container) });
```
martinpitt: Don't bind methods in a `render` function. It's a more common (and efficient?) style to write something like

    onActivate: () => this.stopContainer(container)

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+	else
+	    startStopActions.push({ label: _("Start"), onActivate: this.startContainer.bind(this, container) });
+
+	startStopActions.push({
+	    label: _("Restart"),
+	    onActivate: this.restartContainer.bind(this, container),
+	    disabled: !isRunning
```
martinpitt: Why bother adding a disabled menu entry?

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+
+	return <Listing.ListingRow key={container.Id}
+				   columns={columns}
+				   tabRenderers={tabs}
+				   navigateToItem={ this.navigateToContainer.bind(this, container) }
+				   listingActions={actions} />;
+    }
+
+    render() {
+	console.log(this.props.onlyShowRunning);
```
martinpitt: debugging leftover

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+	const _ = cockpit.gettext;
+	const columnTitles = [_("Name"), _("Image"), _("Command"), _("CPU"), _("Memory"), _("State")];
+	//TODO
+	let emptyCaption = _("No containers");
+
+	let filtered = this.props.containers.filter((container)=>{
```
martinpitt: Missing spaces around operator, and unnecessary braces around a single argument.

src/Containers.js
```diff
@@ -0,0 +1,137 @@
...
+	    if (this.props.onlyShowRunning && !container.State.Running)
+		return false;
+	    return true;
```
martinpitt: write this more succinctly as `filter(container => (!this.props.onlyShowRunning || container.State.Running));

src/ContainersRunImageModal.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import Modal from 'react-modal';
+
+class ContainersRunImageModal extends React.Component {
+    constructor(props) {
+      super(props);
+
+    }
```
martinpitt: Unnecessary if you don't do anything other than calling the parent. Also, this component can be functional, it doesn't have any state.

src/ContainersRunImageModal.js
```diff
@@ -0,0 +1,24 @@
...
+
+
+    render() {
+      return (
+	<Modal
+	  isOpen={this.props.show}
+	>
```
martinpitt: IMHO this looks really ugly - just write it as one line, it's short enough.

src/Containers.js
QiWang19: copied from dockerUI containers-view. Maybe deal with the different version of dockerAPI?

src/Containers.js
QiWang19: Not the same. Add a space in the first case. The other is to join the directory.

src/Containers.js
QiWang19: Haven't implement the function yet. Just want it can be run correctly.

src/Containers.js
QiWang19: Copied from dockerUI. Unnecessary for varlinkAPI, I've deleted it.

package.json
QiWang19: I changed it. But dockerUI uses Id.

QiWang19:
I fixed the review issues you found. @martinpitt PTAL

src/index.html
```diff
@@ -22,7 +22,10 @@
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
+
+
     <link rel="stylesheet" href="../base1/patternfly.css">
+    <link href="docker.css" type="text/css" rel="stylesheet" id="term-style">
```
martinpitt: Please rename this to podman.css (same for the file) to avoid confusion.

src/Containers.js
```diff
@@ -0,0 +1,115 @@
...
+
+	let columns = [
+		{ name: container.Name, header: true },
+		image,
+		container.Config.Entrypoint === "" ? container.Config.Cmd.join(" ") : container.Config.Cmd.join(""),
+		// TODO: CpuUsage
+		"",
+		// TODO: MemoryUsage, MemoryLimit
+		"",
+		state,
```
martinpitt: This should still be i18n'ed, but I'm happy for this to be a `// TODO: i18n` comment for now.

src/Containers.js
```diff
@@ -0,0 +1,115 @@
...
+const renderRow = (containersStats, container) => {
+	const isRunning = !!container.State.Running;
+	//TODO: check container.Image == container.ImageID
+	const image = container.ImageName;
+	const state = container.State.Status;
+
+	let columns = [
+		{ name: container.Name, header: true },
+		image,
+		container.Config.Entrypoint === "" ? container.Config.Cmd.join(" ") : container.Config.Cmd.join(""),
```
martinpitt: The second join is still confusing, please add a comment what it does.

webpack.config.js
```diff
@@ -156,6 +156,35 @@ module.exports = {
...
+                loader: 'strict-loader' // Adds "use strict"
+            },
+            {
+                test: /\.jsx$/,
+                loader: "babel-loader"
+            },
+            {
+                test: /\.es6$/,
+                loader: "babel-loader"
+            },
```
martinpitt: These all already exist above and are duplicated.

src/ContainerDetails.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
martinpitt: This file must be named .jsx, to apply correct babel and eslint rules. It's not "classic" non-ES6 JavaScript. Same applies to the other *.js files in this PR.

webpack.config.js
```diff
@@ -156,6 +156,35 @@ module.exports = {
...
+            {
+                test: /\.css$/,
+                loader: extract.extract("css-loader?minimize=&root=" + libdir)
+            },
```
martinpitt: You can also rename your .css to .scss so that over time it can uses the newer "S"css features.

webpack.config.js
```diff
@@ -156,6 +156,35 @@ module.exports = {
...
+            {
+                test: /\.less$/,
+                loader: extract.extract("css-loader?sourceMap&minimize=!less-loader?sourceMap&compress=false")
+            },
+            {
+                test: /views\/[^\/]+\.html$/,
+                loader: "ng-cache?prefix=[dir]"
+            },
```
martinpitt: This looks unrelated and should be a separate commit with a proper justification.

webpack.config.js
```diff
@@ -156,6 +156,35 @@ module.exports = {
...
+            {
+                test: /[\/]angular\.js$/,
+                loader: "exports?angular"
             }
         ]
```
martinpitt: There's no angular in this code, and you should stay really far away from introducing it :-) Please drop this.

src/ContainerHeader.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import cockpit from 'cockpit';
+import Select from '../lib/cockpit-components-select.jsx';
+
+const _ = cockpit.gettext;cockpit.gett
```
martinpitt: There's some junk here

src/ContainerHeader.js
```diff
@@ -0,0 +1,53 @@
...
+
+class ContainerHeader extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            filter: 'running',
+            filterText: ''
+        };
+
+        this.filterChanged = this.filterChanged.bind(this);
```
martinpitt: This doesn't seem necessary, you only call this from a method that already has a correct `this`.

src/ContainerHeader.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    filterChanged() {
+        if (this.props.onFilterChanged)
+            this.props.onFilterChanged(this.state.filter, this.state.filterText);
+    }
+
+    handleFilterChange (value) {
+        this.setState({ filter: value });
+        this.props.onChange(value);
```
martinpitt: Should check if the `onChange` property is actually defined. Or declare it a mandatory property.

QiWang19:
@martinpitt I have made the changes you requested.  Sorry, I am still getting used to the github workflow.

webpack.config.js
```diff
@@ -156,7 +156,24 @@ module.exports = {
...
+                loader: "babel-loader"
+            },
+            {
+                test: /\.css$/,
+                loader: extract.extract("css-loader?minimize=&root=" + libdir)
+            },
+            {
+                test: /\.less$/,
+                loader: extract.extract("css-loader?sourceMap&minimize=!less-loader?sourceMap&compress=false")
+            },
```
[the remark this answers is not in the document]
QiWang19: Because I used cockpit-components-listing.jsx from docker UI, it import the less

QiWang19:
@martinpitt I've added some tests. PTAL.

garrett:
@QiWang19: Hey! I saw your request in IRC. I'm replying here to let you know you were heard. :wink:

If you have design questions or some HTML and/or CSS questions, I'm happy to help. I even know JavaScript pretty OK. However, I don't know React/JSX... outside of modifying it to make HTML/CSS/JS a bit better. 

As your PR looks like a lot of JSX, I probably can't help at this time in this particular PR.

That all said, I've been a fan of Podman (and everything related, like Buildah) for a while and I'm really looking forward to having Podman support in Cockpit!

martinpitt:
Back from PTO, having a look. Tests on master started failing with

    Failed to enable unit: Unit file io.projectatomic.podman.socket does not exist.

as podman apparently renamed the socket. This will also break this PR.

.gitignore
```diff
@@ -13,3 +13,5 @@ test/common/
 test/images/
 *.pot
 POTFILES*
+tmp/*
+/src/test.jsx
```
martinpitt: This looks like a debugging leftover. *Don't* add such local files to .gitignore, you are going to forget about them.

test/check-application
```diff
@@ -14,35 +14,46 @@ import testlib
...
-        b.wait_in_text("#images", "busybox:latest")
-        b.wait_in_text("#images", "alpine:latest")
-
-        # no containers
-        self.assertFalse(b.is_present("#containers li"))
-
+        # FIXME: UI needs to listen to change signals and update automatically
         # create a container
-        m.execute("podman run -d --name swamped-crate busybox sleep 1000")
+        # m.execute("podman run -d --name swamped-crate alpine sleep 10000")
```
martinpitt: Leftover stuff, should be dropped.

test/check-application
```diff
@@ -14,35 +14,46 @@ import testlib
...
 
-        # FIXME: UI needs to listen to change signals and update automatically
         b.reload()
```
martinpitt: This reload should now be dropped as well.

test/vm.install
```diff
@@ -15,3 +15,4 @@ fi
 # grab a few images to play with; tests run offline, so they cannot download images
 podman pull busybox
 podman pull docker.io/alpine
+# podman rm eager_murdock
```
martinpitt: Leftover.

test/check-application
```diff
@@ -14,35 +14,46 @@ import testlib
...
+        b.wait_in_text("#containers-images", "busybox:latest")
+
+        # show image listing toggle
+        b.wait_present('#containers-images tr:contains("busybox:latest")')
+        b.wait_visible('#containers-images tr:contains("busybox:latest")')
+        b.click('#containers-images tbody tr:contains("busybox:latest") td.listing-ct-toggle')
+        b.wait_visible('#containers-images tbody tr:contains("busybox:latest") + tr button.btn-delete')
+
+        # show all containers to interact with stopped ones
+        m.execute("podman run -d --name swamped-crate busybox sh")
```
martinpitt: This cannot work, the container will exit immediately. This is why in my initial tests I ran `sleep 10000` in it.

martinpitt: Oh, I see, it's supposed to be in the "exited" state, as you switch to "everything".

martinpitt:
Tests succeeded, landing. Thanks!

PR #5, opened 2018-08-21
Implement container stop, start, restart, delete. Add Manual testing for delete.

QiWang19:


QiWang19:
@martinpitt  PTAL

src/ContainerDeleteModal.jsx
```diff
@@ -6,22 +6,22 @@ import * as utils from './util.js';
...
+				<ModalBody>
+					{_("Deleting a container will erase all data in it.")}
+				</ModalBody>
+				<ModalFooter>
+					<Button color="secondary" onClick={props.handleCancelContainerDeleteModal}>Cancel</Button>
+					<Button id="container-delete-toggle-btn" color="danger" onClick={props.handleRemoveContainer}>{_("Delete")}</Button>{' '}
+				</ModalFooter>
+			</Modal>
+		</div>
+	);
```
martinpitt: Can you please fix your editor to not destroy the indentation in that way? 4 spaces, no tabs, please. And mass-replace `\t` with four spaces in the other files here, too.

package.json
```diff
@@ -35,7 +35,8 @@
     "webpack": "^2.6.1"
   },
   "dependencies": {
-    "bootstrap": "^4.1.1",
+    "@babel/runtime": "7.0.0-beta.55",
+    "bootstrap": "^3.3.6",
```
martinpitt: Why is it necessary to go back here?

src/Containers.jsx
```diff
@@ -40,19 +41,126 @@ class Containers extends React.Component {
...
-    //TODO
-    stopContainer(container) {
-        return undefined;
+    stopContainer(container, timeout) {
+        document.body.classList.add('busy-cursor');
+        const id = container.ID;
+        if (!timeout) {
+            timeout = 10;
+        }
+        utils.varlinkCall(utils.PODMAN, "io.podman.StopContainer", JSON.parse('{"name":"' + id + '","timeout":' + timeout + '}' ))
```
martinpitt: Don't go through efforts to construct a JSON string and then immediately parse it back. This is error prone (e. g. if a container name contains `'`, and hard to read. Just give a JS object, i. e. the argument is `{name: id, timeout}`. This applies to the other calls as well, of course.

src/Containers.jsx
```diff
@@ -40,19 +41,126 @@ class Containers extends React.Component {
         }));
     }
 
-    //TODO
-    stopContainer(container) {
-        return undefined;
+    stopContainer(container, timeout) {
+        document.body.classList.add('busy-cursor');
+        const id = container.ID;
```
martinpitt: This is just being used once, please don't introduce such unnecessary aliasing.

src/Containers.jsx
```diff
@@ -40,19 +41,126 @@ class Containers extends React.Component {
...
+            .then(reply => {
+                const idStop = reply.container;
+                //update container info after stop
+                utils.varlinkCall(utils.PODMAN, "io.podman.InspectContainer", JSON.parse('{"name":"' + idStop + '"}'))
+                    .then(reply => {
+                        const newElm = JSON.parse(reply.container)
+                        //replace list with updated info
+                        let oldContainers = this.props.containers;
+                        let idx = oldContainers.findIndex((obj => obj.ID == idStop));
+                        oldContainers[idx] = newElm;
```
martinpitt: This is invalid. `oldContainers` is the same object as `this.props.containers` and you must not update your own properties. Also, this is generally a bad and brittle architecture: Don't *assume* what the method you are calling (`StopContainer()` are doing - they might fail on any number of things. Just call `StopContainer` and then get the updated state for it with the already existing code.

If `class Containers` is actually supposed to manage and change the "containers" list, then this needs to be state, not a property. A property is owned and managed by the containing element, and then you just present it without ever touching it.

src/Containers.jsx
```diff
@@ -65,10 +173,10 @@ class Containers extends React.Component {
         let columns = [
             { name: container.Name, header: true },
             image,
-            container.Config.Cmd.join(" "),
+            container.Config.Cmd ? container.Config.Cmd.join(" ") : "",
```
martinpitt: `null` and `undefined` should be perfectly valid for showing in a table. Why is that necessary?

src/Containers.jsx
```diff
@@ -65,10 +173,10 @@ class Containers extends React.Component {
...
             container.State.Running ? utils.format_cpu_percent(container.HostConfig.CpuPercent) : "",
-            containerStats ? utils.format_memory_and_limit(containerStats.mem_usage, containerStats.mem_limit) : "",
-            state /* TODO: i18n */,
+            container.State.Running && containerStats ? utils.format_memory_and_limit(containerStats.mem_usage, containerStats.mem_limit) : "",
+            state/* TODO: i18n */,
```
martinpitt: The space got accidentally dropped

src/Containers.jsx
```diff
@@ -79,14 +187,13 @@ class Containers extends React.Component {
...
-            startStopActions.push({ label: _("Stop"), onActivate: () => this.stopContainer(container)});
+            startStopActions.push({ label: _("Stop"), onActivate: () => this.stopContainer(container, undefined)});
         else
             startStopActions.push({ label: _("Start"), onActivate: () => this.startContainer(container)});
 
         startStopActions.push({
             label: _("Restart"),
-            // onActivate: this.restartContainer,
-            onActivate: this.restartContainer,
+            onActivate: this.restartContainer(container, undefined),
```
martinpitt: `restartContainer()` does not take a second argument. Also, you don't actually need to pass trailing `undefined` arguments, they will be undefined by default.

test/check-application
```diff
@@ -34,7 +34,6 @@ class TestApplication(testlib.MachineCase):
         # run a container (will exit immediately)
         m.execute("podman run -d --name swamped-crate busybox sh")
 
-        # FIXME: UI needs to listen to change signals and update automatically
```
martinpitt: This is still true, please keep it. It is really wrong to have to reload the page after state changes, but until podman has a way to notify clients about changes, it's a necessary workaround.

test/check-application
```diff
@@ -58,6 +57,60 @@ class TestApplication(testlib.MachineCase):
         b.wait_present("#containers-containers")
         b.wait_in_text("#containers-containers", "No running containers")
 
+        # delete container busybox:latest from container listing toggle
+        b.click("#containers-containers-filter button")
+        b.wait_visible("#containers-containers-filter .dropdown-menu")
+        b.click("#containers-containers-filter li[data-data='all'] a")
+        b.wait_visible("#containers-containers")
+        b.wait_visible('#containers-containers tr:contains("busybox:latest")')
+        b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
```
martinpitt: Please adjust the comment. This doesn't actually delete anything, just selects the display mode and a container details tab (.. it seems?)

test/check-application
```diff
@@ -58,6 +57,60 @@ class TestApplication(testlib.MachineCase):
...
+
+        # click Start container button from container listing toggle
+        b.wait_visible('#containers-containers tbody tr:contains("busybox:latest") + tr div.dropdown')
+        b.wait_visible('#containers-containers tbody tr:contains("busybox:latest") + tr:contains("Start")')
+        b.click('#containers-containers tbody tr:contains("busybox:latest") + tr:contains("Start")')
+
+        # delete container from listing toggle
+        b.wait_visible('#containers-containers tbody tr:contains("busybox:latest") + tr button.btn-delete')
+        b.click('#containers-containers tbody tr:contains("busybox:latest") + tr button.btn-delete')
+        with b.wait_timeout(20):
```
martinpitt: Please keep the standard 60 second timeout (also below). It should not hurt, and on our loaded CI infra things are really slow sometimes.

test/check-application
```diff
@@ -58,6 +57,60 @@ class TestApplication(testlib.MachineCase):
...
+            b.wait_present(".modal-dialog div")
+        b.wait_present(".modal-dialog div #container-delete-toggle-btn")
+        b.click(".modal-dialog div #container-delete-toggle-btn")
+        with b.wait_timeout(200):
```
martinpitt: Why would that take so enormously long?

test/check-application
```diff
@@ -58,6 +57,60 @@ class TestApplication(testlib.MachineCase):
...
+            try:
+                b.wait_visible("#containers-containers")
+                b.wait_not_in_text("#containers-containers", "busybox:latest")
+            except:
```
martinpitt: No blanket exceptions, please. In general, it's unusual to have exceptions in test code, as you should already precisely know what you are looking for. What is this supposed to do?

test/vm.install
```diff
@@ -15,3 +15,4 @@ fi
 # grab a few images to play with; tests run offline, so they cannot download images
 podman pull busybox
 podman pull docker.io/alpine
+# podman rm eager_murdock
```
martinpitt: Please don't commit such stuff.

src/Containers.jsx
QiWang19: Sorry, a little bit confusing. this.props.containers is from parent component in app.jsx. It is from a state of parent component. And I called function updateContainers to update parent's state. I think it works.

martinpitt: This is still conceptually writing/changing a property of the `Containers` class. Note that it's *also* invalid to directly manipulate the state object (of yourself or your containing object),  this requires calling `setState()`. So you need to make a copy, with `newContainers = Object.assign({}, this.props.containers)`, or if you are into fancy new ES6, `newContainers = {...this.props.containers}` and call `setState()` on these.

However, this still has the problem that it makes assumptions on the effect of the called API function, and doesn't handle errors properly. It's much better to read the actual state from scratch after every API call that changes anything, until podman gets a proper notification API.

package.json
QiWang19: Since cockpit uses redhat patternfly, which only supports bootstrap3

src/Containers.jsx
QiWang19: Both containers and updateContainers() are props from parent component. I use updateContainers() to setState in parent component. Can child component use setSate to update parent's state from spread attributes {...} ? Could you give me some tutorial link to help me ?

src/Containers.jsx
QiWang19: If inspectContainer return a container without Cmd attribute it can not use join.

src/Containers.jsx
martinpitt: No, a component must never meddle with the state of any other component directly, only your own. If the `Containers` class triggers changes to the list of currently known containers, it must own the state. The valid way is to by and large do what you already do, and call a callback method (through a property like `this.props.updateContainers()`, but you must pass a new object (what I've shown with `Object.assign()` or the spread operator), not change your property object in-place.

But again, this is all too brittle. At the moment `podman.ListContainers()` and `GetContainerStats()` lives in app.jsx in `componentDidMount()`, i. e. just happens once. From then on your UI just tries to *predict* how the containers and states changed based on the actions from the UI. But these sometimes fail, containers can spontaneously stop or be started/stopped from other places than Cockpit, so the UI will quickly become out of date.

The proper way to fix that is to teach podman about a proper notification API. While it doesn't have one, I propose this architecture:

 * At the top level (app.jsx or maybe a container-state.es6 helper) you maintain the state of all containers, and a `updateContainers()` method (no arguments), which basically does what app.jsx' `componentDidMount()` does.

 * All components pass the app's `updateContainers()` down as a callback property to components.

 * Whenever any component calls a podman API function, it calls `this.props.updateContainers()` after it returns (on both success and failure!), so that the state and UI will update to reality.

src/Containers.jsx
martinpitt: Right, I just meant the `""` part. But nevermind, it's just a tiny nitpick, it's fine as-is.

src/Containers.jsx
martinpitt: Oh, should be obviousy, but just in case: of course components would also pass down the actual container state object (or a relevant part of it)  as properties to its contained components. Then the right bits of the UI will update at the right time.

test/check-application
QiWang19: Yes, wrong comment. It just opened the collapsed table before clicking the button inside the table.

test/check-application
QiWang19: If deleting a running container, varlink API returns an error. Need to confirm force delete. I want to click force delete button in except.

test/vm.install
martinpitt: Right, you can just squash that into the existing commit on the next force-push. This  PR needs some more involved rework anyway (see above).

test/check-application
martinpitt: Right, but what kind of exception do you expect here? Catching only `testlib.Error` would be less bad, but still tests should generally not be written in that way. If you expect the error and then click force-delete, then explicitly write it in that way -- i. e. after the "delete" click, wait for the "busy" dialog and click the "force delete" there. If it's not predictable what happens here, there's something wrong with the UI code.

QiWang19: But this PR has not implement run image yet. How can I run a container with commend like "sleep 6000" use testlib.py functions and test the delete button with it?

martinpitt: Start it from the command line, like the [existing test](https://github.com/cockpit-project/cockpit-podman/blob/master/test/check-application#L34) does. Just instead of `sh`, start `sleep 10000`. My original PoC even [did that](https://github.com/cockpit-project/cockpit-podman/commit/de285297c4dcffb6d6d881499d3ae78e09682999#diff-c23b7eb2c7c51d1e825ac452faf7eaa1R36), but your previous PR changed the test to launch a container that immediately stops.

QiWang19: testlib looks like a test interface writes by yourself. Has any documentation can tell me how to use it?

QiWang19: And I saw the example in check-docker in docker UI. It just use ```try..except```, without catching error. Can you explain the situation for me?

martinpitt: testlib.py's Browser class is here: https://github.com/cockpit-project/cockpit/blob/master/test/common/testlib.py#L88 . There is no separate documentation how to use it. Some methods have docstrings, many others should be self-explanatory. If a particular one is unclear, please file an issue or ping on IRC, we should improve the docstrings then.

Indeed `del_container_from_details()` from check-docker is really bad style, please don't copy that. If the function does not know whether or not a container needs to be force-deleted, it should get that information as a parameter, and then do the expected steps without this opportunistic (and way too broad) exception handling.

QiWang19:
@martinpitt PTAL

martinpitt:
@QiWang19 : Can you please fix the tab/indentation damage here? This causes a lot of diff noise, making the PR harder to review and the code harder to read. (Remember, 4 spaces indentation, not tabs)

QiWang19:
@martinpitt PTAL

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_visible(".modal-dialog div .btn-danger")
+        b.click(".modal-dialog div .btn-danger")
+        with b.wait_timeout(60):
+            try:
+                b.wait_visible("#containers-containers")
+                b.wait_not_in_text("#containers-containers", "alpine:latest")
+            except:
+                self.confirm()
+                b.wait_visible("#containers")
+                b.wait_not_in_text("#containers-containers", "alpine:latest")
```
martinpitt: As [discussed before](https://github.com/cockpit-project/cockpit-podman/pull/5/files/2970d1e98783d172769129486ef529c0809da310#diff-c23b7eb2c7c51d1e825ac452faf7eaa1). This is a really bad way to write a test (and yes, the existing check-docker is broken in the same way). Please avoid the exception handling entirely here; or if it is necessary for some reason, only catch the precise one you expect, and document why this is necessary. Like that it's totally unclear (also to me) what actually is supposed to happen here.

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
         # show container listing toggle
```
martinpitt: The "show all containers to interact with stopped ones" needs to be adjusted to the `podman run` changes above. Now `busybox` is the one that runs, and `alpine` the one that immediately stops. If that passes, it shows that this is missing a check that the status is actually shown as "stopped", and another check that it did *not* appear in the initial "running containers" list (which is why you have to select "All").

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
...
         b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
         b.wait_present('#containers-containers tbody tr:contains("busybox:latest") + tr button.btn-delete')
-
+        #show running container
         b.click("#containers-containers-filter button")
         b.wait_present("#containers-containers-filter .dropdown-menu")
         b.click("#containers-containers-filter li[data-data='running'] a")
+        b.wait_present('#containers-containers tr:contains("busybox:latest")')
+        b.wait_visible('#containers-containers tr:contains("busybox:latest") + tr:contains("sleep 1000")')
```
martinpitt: These should also check the name of the container.

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
...
+
+        # switch to all containers
+        b.wait_present("#containers-containers-filter button")
+        b.click("#containers-containers-filter button")
+        b.wait_visible("#containers-containers-filter .dropdown-menu")
+        b.click("#containers-containers-filter li[data-data='all'] a")
+        b.wait_present("#containers-containers")
+        b.wait_present('#containers-containers tr:contains("busybox:latest") td:contains("running")')
+        b.wait_present('#containers-containers tr:contains("alpine:latest")')
```
martinpitt: This is now actually doing some of the above, but there are now two sections that switch to "all" containers. Please only have one of them (looks like you can just delete the one above).

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
...
+
+        # delete a stopped container
+        b.click('#containers-containers tbody tr:contains("alpine:latest") td.listing-ct-toggle')
+        b.wait_visible('#containers-containers tbody tr:contains("alpine:latest") + tr button.btn-delete')
```
martinpitt: This is a race condition with React, as the element will not exist right away -- use `wait_present()`.

`wait_visible()` is more appropriate for jQuery pages, where the HTML usually has all possible elements, and the code just shows/hides them as appropriate. But with React the elements actually appear/disappear in the DOM.

Same in more cases below. I. e. as a rule of thumb, always use `wait_present` unless your code explicitly shows/hides stuff.

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
...
+
+    # FIXME: UI needs to listen to change signals and update automatically
+
+    def confirm(self):
+        b = self.browser
+        b.wait_present(".modal-dialog div")
+        b.wait_visible(".modal-dialog div .btn-danger")
+        b.click(".modal-dialog div .btn-danger")
+
+    def loadAllContainerlist(self):
```
martinpitt: This should have a docstring that this includes the `reload()` hack. I. e. it should only be used after running a CLI command, not after UI actions.

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
...
+            try:
+                b.wait_visible("#containers-containers")
+                b.wait_not_in_text("#containers-containers", "alpine:latest")
+            except:
+                self.confirm()
+                b.wait_visible("#containers")
+                b.wait_not_in_text("#containers-containers", "alpine:latest")
+
+        # delete a running container
+        self.loadAllContainerlist()
```
martinpitt: This should not need a page reload here.

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
...
+        m.execute("podman run -d --name swamped-crate busybox sleep 1000")
+        self.loadAllContainerlist()
+        b.wait_present('#containers-containers tr:contains("busybox:latest")')
+        b.wait_visible('#containers-containers tr:contains("busybox:latest")')
+
+        # stop a running container
+        b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
+        b.wait_visible('#containers-containers tbody tr:contains("busybox:latest") + tr:contains("Stop")')
+        b.click('#containers-containers tbody tr:contains("busybox:latest") + tr:contains("Stop")')
+        self.loadAllContainerlist()
```
martinpitt: This should not need a page reload. If you commonly use that case, split the function into two: one for an explicit "page reload and wait for it to be built up again", and one for "load and expand the container list with all containers".

test/check-application
```diff
@@ -51,13 +51,88 @@ class TestApplication(testlib.MachineCase):
...
+        # TODO: wait for container to stop
```
martinpitt: Please implement this - it's the main functionality you are testing here.

src/ContainerDeleteModal.jsx
```diff
@@ -17,7 +17,7 @@ const ContainerDeleteModal = (props) => {
                 </ModalBody>
                 <ModalFooter>
                     <Button color="secondary" onClick={props.handleCancelContainerDeleteModal}>Cancel</Button>
-                    <Button color="danger" onClick={props.handleRemoveContainer}>{_("Delete")}</Button>{' '}
+                    <Button id="container-delete-toggle-btn" color="danger" onClick={props.handleRemoveContainer}>{_("Delete")}</Button>{' '}
```
martinpitt: I don't see this ID being used anywhere. Also, please avoid these. If you need to select these in tests, select the modal dialog first, and then the containing button. If it's just one, just `button` is enough, otherwise you have to be  more specific, like `button.btn-primary`.

src/ContainerRemoveErrorModal.jsx
```diff
@@ -17,7 +17,7 @@ const ContainerRemoveErrorModal = (props) => {
                 </ModalBody>
                 <ModalFooter>
                     <Button color="secondary" onClick={props.handleCancelRemoveError}>{_("Cancel")}</Button>
-                    <Button color="danger" onClick={props.handleForceRemoveContainer}>{_("Force Delete")}</Button>{' '}
+                    <Button id="confirmation-dialog-confirm" className={props.setWaitCursor} color="danger" onClick={props.handleForceRemoveContainer}>{_("Force Delete")}</Button>{' '}
```
martinpitt: This ID is actively dangerous. You set exactly the same one in another component further down, and thus would have two different elements with the same ID. You also don't use it anywhere, so please remove these two.

src/Containers.jsx
```diff
@@ -15,8 +15,10 @@ class Containers extends React.Component {
         this.state = {
             selectContainerDeleteModal: false,
             setContainerRemoveErrorModal: false,
-            containerWillDelete: {}
-        }
+            containerWillDelete: {},
+            setWaitCursor:"",
```
martinpitt: missing space after `:`

src/Containers.jsx
```diff
@@ -40,19 +42,84 @@ class Containers extends React.Component {
...
 
-    //TODO
-    stopContainer(container) {
-        return undefined;
+    stopContainer(container, timeout) {
+        document.body.classList.add('busy-cursor');
+        if (!timeout) {
+            timeout = 10;
+        }
+        utils.varlinkCall(utils.PODMAN, "io.podman.StopContainer", {name: container.ID, timeout: timeout})
```
martinpitt: Hint: You could write this more succinctly as `timeout: timeout || 10`, and drop the above `if`. (you don't have to though, matter of preference)

src/Containers.jsx
```diff
@@ -40,19 +42,84 @@ class Containers extends React.Component {
...
+            .then(reply => {
+                if (!reply.container) {
+                    return;
+                }
```
martinpitt: When does that happen? That looks like an error condition that isn't being handled further?

src/Containers.jsx
```diff
@@ -40,19 +42,84 @@ class Containers extends React.Component {
...
+                    .then((reply) => {
+                        // console.log(reply.newContainers);
+                        this.props.updateContainers(reply.newContainers);
+                        document.body.classList.remove('busy-cursor');
+                    })
+                    .catch(ex => {
+                        console.error("Failed to do Update Container:", JSON.stringify(ex))
+                        document.body.classList.remove('busy-cursor');
+                    })
+            })
```
martinpitt: This code gets repeated several times. It seems more elegant if you let the top-level `app.jsx` component's `updateContainers()` do the `then` and `catch` handling, and then each subcomponent just has to do

    utils.varlinkCall(...).then(this.props.updateContainers)

There's going to be a lot more of these in the future, and at some point this is hopefully being replaced with a proper notification mechanism. So this should be centralized as much as possible.

src/ImageRemoveErrorModal.jsx
```diff
@@ -17,7 +17,7 @@ const ImageRemoveErrorModal = (props) => {
                 </ModalBody>
                 <ModalFooter>
                     <Button color="secondary" onClick={props.handleCancelImageRemoveError}>{_("Cancel")}</Button>
-                    <Button color="danger" onClick={props.handleForceRemoveImage}>{_("Force Delete")}</Button>{' '}
+                    <Button id="confirmation-dialog-confirm" color="danger" onClick={props.handleForceRemoveImage}>{_("Force Delete")}</Button>{' '}
```
martinpitt: This is the duplicate ID, and also being unused.

src/util.js
```diff
@@ -38,7 +38,7 @@ function varlinkCallChannel(channel, method, parameters) {
                 let msg = varlinkCallError(json);
                 reject(msg);
             } else if (json.parameters) {
-                // debugging
+            // debugging
```
martinpitt: please revert, accidental indentation breakage

src/util.js
```diff
@@ -63,9 +63,12 @@ function varlinkCallError(error) {
  * `varlinkCallChannel()` but allows multiple parallel calls.
  */
 export function varlinkCall(channelOptions, method, parameters) {
-    var channel = cockpit.channel(Object.assign({payload: "stream", binary: true, superuser: "require" }, channelOptions));
+        var channel = cockpit.channel(Object.assign({payload: "stream", binary: true, superuser: "require" }, channelOptions));
```
martinpitt: wrong indentation, please revert

src/util.js
```diff
@@ -63,9 +63,12 @@ function varlinkCallError(error) {
...
+
     return varlinkCallChannel(channel, method, parameters).finally(() => {
-        channel.close()
+        setTimeout(()=>{
+            channel.close()
+        }, 500)
```
martinpitt: Why the delay? A hardcoded timeout is almost always a hack, and at least needs an explanation.

src/util.js
```diff
@@ -106,3 +109,86 @@ export function format_memory_and_limit(usage, limit) {
...
+                        newContainersMeta.map((container) => {
+                            varlinkCall(PODMAN, "io.podman.GetContainerStats", {name: container.id})
+                                .then(reply => {
+                                    if (reply.container) {
+                                        newContainersStats[container.id] = reply.container;
+                                        console.log(Object.keys(newContainersStats).length);
+                                    }
+                                    if (newContainers.length === len && Object.keys(newContainersStats).length === len) {
+                                        resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                                    }
```
martinpitt: This is brittle. If any single one of the `InspectContainer()` or `GetContainerStats()` calls fails, the promise will never be resolved. Instead, collect the promises of all calls in an array, and after this `newContainersMeta` loop use [Promise.all()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) to  pick up "everything succeeded" vs. "something failed".

src/util.js
```diff
@@ -106,3 +109,86 @@ export function format_memory_and_limit(usage, limit) {
...
+                if (!newImagesMeta) {
+                    resolve(newImages)
+                    return;
+                }
+                newImagesMeta.map((img)=>{
+                    varlinkCall(PODMAN, "io.podman.InspectImage", {name: img.id})
+                        .then(reply => {
+                            newImages.push(JSON.parse(reply.image));
+                            if (newImages.length === len) {
+                                resolve(newImages);
```
martinpitt: Same `Promise.all` thing here.

test/check-application
QiWang19: I want to delete a stoped container under all containers and to check it has been deleted, not shown under all container list. So I switched to all container here.

test/check-application
QiWang19: then what should be here? without 
```
b.reload() 
b.enter_page("/podman")
```
test failed.
but in my code for podman UI, I only rerender the component to update after deletion, I didn't use reload.

test/check-application
martinpitt: Yes, that makes sense. But there is a similar section of code above which now does nothing, that's the one to get rid of.

test/check-application
martinpitt: Doesn't that point out a bug in the code then? For all actions that you do in the UI, the UI should update automatically. You only need to reload the page for actions that you do outside, like with executing `podman` commands.

martinpitt:
As this already takes rather long again, please re-read https://github.com/cockpit-project/cockpit-podman/pull/3#pullrequestreview-143133842 about making smaller PRs. Implement *one* action, or add *one* new test etc. in separate PRs, so that they have a chance of landing fast,  reviews and fixups don't take (literally) hours and repeat a lot of efforts, etc.

test/check-application
QiWang19: I can't wait for the stop because after I run ```TEST_AUDIT_NO_SELINUX=1 test/check-application -s -t -v``` and log into 127.0.0.2:9091, the UI looks like generateed by outdated code, not that I'm working on. The test uses a component class I used before, which the stop button has no onclick event, but I changed it later. The old component class is Dropdown.jsx, but my code currently uses DropdownContainer.jsx not Dropdown.jsx, I didn't use Dropdown.jsx anymore. So I delete Dropdown.jsx and did 

```
git rm <path>
git add .
```
It still uses the old version code. Can you help me figure it out?

martinpitt: You most certainly don't want `git add .`, but add the new file you are working on. Did you update all the `import` statements for Dropdown.jsx as well? If so, try `git clean -ffdx` and rebuild from scratch, just in case the build system doesn't properly pick up this change.

QiWang19:
I've passed all tests in my local when I created this PR, but it failed in your side. This morning, after you pushed back this branch to me, I failed the tests in my repo. When I ```make check```, after installing some tools it entered ```TEST_AUDIT_NO_SELINUX=1 test/check-application```, and get stuck in checking avaliable updates, finally returned error, like 

```
wait_js_cond(ph_is_present("#containers-containers")): condition did not become true
cdp: {'source': 'network', 'level': 'error', 'text': 'Failed to load resource: the server responded with a status of 401 (Authentication failed)', 'timestamp': 1535554171494.63, 'url': 'http://127.0.0.2:9091/cockpit/login', 'networkRequestId': '13302.8'}
warning: Checking for available updates failed: [object Object]
warning: Checking for available updates failed: [object Object]
```  
You helped me with this during the whole morning, I did not change anyting, but in the evening, I ```make check```again, ```# TESTS PASSED [41s on qiwan]```, lt looks like tests passed in my local, in  my own repo, both in master and containerStop branch.

But I also cloned the official repo cockpit-podman,  and ```make check```, it failed quickly in the install process, the error is
```
man-1-1.fc28.x86_64 'INSTALL=/usr/bin/install -p'
make[1]: Entering directory '/home/qiwan/Documents/pr1/cockpit-podman-master/rpmbuild/cockpit-podman'
make[1]: *** No rule to make target 'install'.  Stop.
make[1]: Leaving directory '/home/qiwan/Documents/pr1/cockpit-podman-master/rpmbuild/cockpit-podman'
error: Bad exit status from /var/tmp/rpm-tmp.yfMT3t (%install)


RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.yfMT3t (%install)
make: *** [Makefile:108: rpm] Error 1
``` 
This really makes me confused, I didn't change anything, but tests works later. Although my master is the same with the origin/master, my master passed tests but the repository I cloned from cockpit-podman cannot pass the test. And this is not the 1st time I met those situation. What might be the reason? I think they run in the same environment. 

And there is a problem with varlink happend one week ago, I used podman 0.8.3-dev at that time. If I want to use varlink in my machine, I must run
```
$ sudo podman varlink --timeout 0 unix:/run/podman/io.podman
```
in the terminal, and keep the window open. In your machine there exists the same problem? I've already convert to podman 0.8.5-dev, the problem is the same.

test/check-application
QiWang19: I added ```testlib.wait(lambda: m.execute("podman ps -a | grep -q Exited"), delay=5)```this line, and tests hangd here. I enterned the shell, 
```
[root@localhost ~]# podman ps -a | grep -q Exited
[root@localhost ~]# podman ps -a | grep Exited
6664a9e8e72d   docker.io/library/busybox:latest   sleep 1000   8 minutes ago   Exited (137) 8 minutes ago
```
It looks like the containre really stoped. but testss till hang there.
the error after Enter is 
```
  File "test/check-application", line 110, in testBasic
    testlib.wait(lambda: m.execute("podman ps -a | grep -qExited"), delay=5)
  File "test/common/testlib.py", line 1312, in wait
    raise Error(msg or "Condition did not become true.")
testlib.Error: Condition did not become true.
```

martinpitt: There is a space missing between `-q` and `Exited`. So grep (and thus the whole command) would fail on something like "unknown option".

martinpitt:
Not sure about your local test failures. But the VM should always have the podman version that gets shipped in Fedora (unless you changed test/vm.install to do something different, of course). The "no rule to make target install" sounds like the Makefile got damaged, or moved away?

QiWang19:
Makefile exists. And I can find the 'install' part there.

test/check-application
QiWang19: sorry, it's just a typo when I deal with markdown in github.  I used
```
testlib.wait(lambda: m.execute("podman ps -a | grep -q Exited")
```, with space. The problem is not beacuse the space.

Comment (QiWang19) [reply test/check-application]:
```
lambda: m.execute("podman ps -a | grep -q Exited"
```
has no reply, I changed to ```lambda: m.execute("podman ps -a | grep -c Exited"```

File: test/check-application (hunk 18 above)
Comment (QiWang19) [reply test/check-application]: And show all container above is test for the dropdown filter.

File: test/check-application (hunk 21 above)
Comment (QiWang19) [reply test/check-application]:
```
        # self.pageReload()
        self.loadAllContainerlist()
        b.wait_present('#containers-containers tbody tr:contains("busybox:latest")')
        b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
        b.wait_present('#containers-containers tbody tr:contains("busybox:latest") + tr button.btn-delete')
```
without self.pageReload(), tests hangs in wait_present ```tr button.btn-delete```, what kind of bug might be?

PR #7, opened 2018-08-29
Merge and adapt recent changes from starter-kit

martinpitt:
CC @QiWan19, FYI. The indentation fixes will likely conflict quite a bit with your current work on rebasing. I suggest the following workflow to resolve that easily:

 - Rebase to current origin/master interactively
 - then as the first commit, revert "Fix wrong indentation, spacing, and  missing semicolons" so that your changes apply cleanly
 - Continue the rebase
 - Build will most likely fail with lots of eslint errors. Fix them with `node_modules/.bin/eslint --fix <filename>`, that should get rid of almost all of them, and the rest by hand.
 - Squash your work and the revert of "Fix wrong indentation, spacing, and  missing semicolons" together, so that as a sum you have a clean change of your work.

I can help you with that if you want to, and force-push your branch.


martinpitt:
Note that I didn't yet apply the Babel 7 change, as that leads to problems with the imported cockpit components. They need to be fixed for proper ES6 first.

Issue #6, opened 2018-08-29
Add missing tests


martinpitt:
This was some leftover from PR #3, but this got landed with incomplete tests to not further drag it out. At least these tests are missing:

- [x]  Start a running container and make sure that it appears so.
 - [x] Check that container names and container commands are shown correctly. Particularly with commands that have more than one word, to cover correct spacing.
- [x] Cover the "Commit", "Delete", and "Start" buttons in container details. The latter currently causes a crash, this needs to be fixed as well.
- [x] Image details and deletion.

martinpitt:
@QiWang19

marusak:
I believe all of these are currently tested.

PR #10, opened 2018-09-05
Change list updating, add container stop&start dropdown menu.

QiWang19:


src/app.jsx
```diff
@@ -91,18 +99,25 @@ class Application extends React.Component {
                     this._asyncRequestContainers = null;
                     this.setState({containersMeta: reply.containers || []});
                     this.state.containersMeta.map((container) => {
-                        utils.varlinkCall(utils.PODMAN, "io.podman.InspectContainer", JSON.parse('{"name":"' + container.id + '"}'))
+                        utils.varlinkCall(utils.PODMAN, "io.podman.InspectContainer", {name: container.id})
+
                                 .then(reply => {
-                                    const temp_containers = this.state.containers;
+                                    let temp_containers = this.state.containers;
                                     temp_containers.push(JSON.parse(reply.container));
```
martinpitt: I didn't spot that in the previous review: This modifies `this.state` directly, which is invalid. This needs to make a copy, with e. g. `temp_containers = this.state.containers.slice()`. Also, I don't get the simple `push()` - when you call it multiple times, you would never remove the old containers from the state; might be better to make the state an id → props mapping instead of a simple list?

src/app.jsx
```diff
@@ -91,18 +99,25 @@ class Application extends React.Component {
...
                                     this.setState({containers: temp_containers});
                                 })
-                                .catch(ex => console.error("Failed to do InspectImage call:", ex, JSON.stringify(ex)));
+                                .catch(ex => console.error("Failed to do InspectContainer call:", ex, JSON.stringify(ex)));
                     });
-                    this.state.containersMeta.map((container) => {
-                        utils.varlinkCall(utils.PODMAN, "io.podman.GetContainerStats", JSON.parse('{"name":"' + container.id + '"}'))
+                    this.state.containersMeta.filter((ele) => {
```
martinpitt: Can you please document (in the constructor) what exactly `state.containers`, `state.containersMeta`, and `state.containersStats` are? an example how the state looks like would be helpful for following what's going on between these three. Also, it's not helpful for them to be lists if you have to modify them - I would expect these would all just be set from `utils.updateContainers()` (or `updateImages()` etc.) exclusively.

src/app.jsx
```diff
@@ -91,18 +99,25 @@ class Application extends React.Component {
...
+                        if (ele.status === "running") {
+                            return true;
+                        } else {
+                            return false;
+                        }
```
martinpitt: This is a very long-winded way to say `filter(ele => ele.status === "running")`

src/app.jsx
```diff
@@ -91,18 +99,25 @@ class Application extends React.Component {
...
+                    }).map((container) => {
+                        utils.varlinkCall(utils.PODMAN, "io.podman.GetContainerStats", {name: container.id})
                                 .then(reply => {
-                                    const temp_container_stats = this.state.containersStats;
+                                    let temp_container_stats = this.state.containersStats;
                                     if (reply.container) {
                                         temp_container_stats[container.id] = reply.container;
                                     }
```
martinpitt: Same in-place modification of state. Keep in mind that `temp_container_stats = this.state.containersStats` does *not* create a copy, it's just an alias name or reference;  or in C terms, a pointer to the same object.

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
...
+    let newContainersStats = [];
+    return new Promise((resolve, reject) => {
+        varlinkCall(PODMAN, "io.podman.ListContainers")
+                .then(reply => {
+                    let newContainersMeta = reply.containers;
+                    if (!newContainersMeta) {
+                        resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                        return;
+                    }
+                    const len = newContainersMeta.length;
```
martinpitt: This isn't being used anywhere near. If it is two hundred lines below, everyone will have forgotten whose length `len` is. So please drop that and spell it out where you need it, that's safer.

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
...
+                                    newContainers.push(inspectRet);
+                                    if (runEle.length === 0 && newContainers.length === len) {
+                                        resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                                    }
+                                });
+                            })
+                            .catch(ex => {
+                                console.error("Failed to do InspectContainer call:", ex, JSON.stringify(ex));
+                            });
+                    let crtStatsRet = [];
```
martinpitt: is this mean to be `ctrStatsRet`? Might be better to spell out the names, to avoid having to guess what acronyms mean. :)

test/check-application
```diff
@@ -27,37 +27,110 @@ class TestApplication(testlib.MachineCase):
...
         b.click('#containers-images tbody tr:contains("busybox:latest") td.listing-ct-toggle')
         b.wait_present('#containers-images tbody tr:contains("busybox:latest") + tr button.btn-delete')
 
         # run a container (will exit immediately)
-        m.execute("podman run -d --name swamped-crate busybox sh")
+        m.execute("podman run -d --name test-sh alpine sh")
+        # run a container
+        m.execute("podman run -d --name swamped-crate busybox sleep 1000")
 
-        # FIXME: UI needs to listen to change signals and update automatically
```
martinpitt: That comment again moved far down accidentally, it should stay where it is.

src/app.jsx
QiWang19: I also tried to make it a map at first ```temp_containers[container.id] = JSON.parse(reply.container);``` but in JS the array length is 0 because the nonsequence index.  not rendered. I think the reason is in ``` cockpit-components-listing.jsx``` Ln299, it checked ```if (!props.children || props.children.length === 0)```. So I just use array and use ```findIndex()``` if needed. Do you have any ideas about this?

If keep it as an array, can I use this instead of slice()?
```
this.setState(preState => ({
    containers: [...preState.containers, JSON.parse(reply.container)]
}));
```

martinpitt: If you turn this into a map, then of course you have to change the render() function of the component that receives this as property (Containers.jsx?) accordingly, to iterate over the dictionary instead of the map. If you need the length, you can use `Object.keys(state.containers).length`.

You can use the spread operator instead of `.slice()` to copy an array, sure. But as I said, merely appending this is not enough - you will have to remove the old entries for the same ID first.

src/util.js
```diff
@@ -39,22 +39,22 @@ function varlinkCallChannel(channel, method, parameters) {
...
-    })
+    });
 }
 
 function varlinkCallError(error) {
     let str = "";
-    error.error ? str += error.error.toString() : str;
-    error.parameters.reason ? str += " " + error.parameters.reason.toString() : str;
+    str = error.error ? str + error.error.toString() : str;
+    str = error.parameters ? str + " " + error.parameters.reason.toString() : str;
```
martinpitt: I fixed that crash independently in PR #13, as it was making tests fail (and the page Oopsing) on master. So you can take this out on your next rebase.

QiWang19:
@martinpitt PTAL

test/check-application
```diff
@@ -27,37 +27,111 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_present("#containers-containers-filter .dropdown-menu")
+        b.click("#containers-containers-filter li[data-data='all'] a")
+        b.wait_present("#containers-containers")
+
+    def containerStopDropdown(self):
+        b = self.browser
+        b.wait_present('#containers-containers tbody tr:contains("busybox:latest") + tr button span.caret')
+        b.click('#containers-containers tbody tr:contains("busybox:latest") + tr button span.caret')
+        b.wait_present('#containers-containers tbody tr:contains("busybox:latest") + tr ul:contains("Stop")')
+        b.wait_present('#containers-containers tbody tr:contains("busybox:latest") + tr ul:contains("Restart")')
```
martinpitt: This is just being used once, so IMHO better moved inline in the actual test, to make it easier to read.

test/check-application
```diff
@@ -27,37 +27,111 @@ class TestApplication(testlib.MachineCase):
...
+        # test the container stop dropdown menu under the listing toggle
+        self.containerStopDropdown()
 
+        #show running container
         b.click("#containers-containers-filter button")
         b.wait_present("#containers-containers-filter .dropdown-menu")
         b.click("#containers-containers-filter li[data-data='running'] a")
+        b.wait_present('#containers-containers tr:contains("busybox:latest")')
+        b.wait_present('#containers-containers tr:contains("swamped-crate")')
+        b.wait_present('#containers-containers tr:contains("busybox:latest") + tr:contains("sleep 1000")')
```
martinpitt: This should also have a `wait_not_present` for the stopped alpine container.

test/check-application
```diff
@@ -27,37 +27,111 @@ class TestApplication(testlib.MachineCase):
...
+
+        # delete a stopped container
+        b.click('#containers-containers tbody tr:contains("alpine:latest") td.listing-ct-toggle')
+        b.wait_present('#containers-containers tbody tr:contains("alpine:latest") + tr button.btn-delete')
+        b.click('#containers-containers tbody tr:contains("alpine:latest") + tr button.btn-delete')
+        b.wait_present(".modal-dialog div")
+        b.wait_present(".modal-dialog div .btn-danger")
+        b.click(".modal-dialog div .btn-danger")
+        b.wait_not_present(".modal-dialog div")
+        self.pageReload()
```
martinpitt: If this is required, it hides a bug. Operations that are done through the UI should never require a page reload, only operations that are done on the command line or elsewhere. If this is a necessary workaround for now, it needs a "FIXME" comment with a detailled explanation.

test/check-application
```diff
@@ -27,37 +27,111 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_present(".modal-dialog div .btn-danger")
+        b.click(".modal-dialog div .btn-danger")
+        self.confirm()
+        b.wait_not_in_text("#containers-containers", "busybox:latest")
+
+    def confirm(self):
+        b = self.browser
+        b.wait_present(".modal-dialog div")
+        b.wait_present(".modal-dialog div .btn-danger")
+        b.click(".modal-dialog div .btn-danger")
```
martinpitt: This should wait until the dialog goes away.

test/check-application
```diff
@@ -27,37 +27,111 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_present('#containers-containers tr:contains("alpine:latest") td:contains("exited")')
+
+        # delete a stopped container
+        b.click('#containers-containers tbody tr:contains("alpine:latest") td.listing-ct-toggle')
+        b.wait_present('#containers-containers tbody tr:contains("alpine:latest") + tr button.btn-delete')
+        b.click('#containers-containers tbody tr:contains("alpine:latest") + tr button.btn-delete')
+        b.wait_present(".modal-dialog div")
+        b.wait_present(".modal-dialog div .btn-danger")
+        b.click(".modal-dialog div .btn-danger")
+        b.wait_not_present(".modal-dialog div")
```
martinpitt: This should call `self.confirm()` instead of duplicating the code.

test/check-application
```diff
@@ -27,37 +27,111 @@ class TestApplication(testlib.MachineCase):
...
         b.enter_page("/podman")
         b.wait_present("#containers-containers-filter button")
 
-        # show all containers to interact with stopped ones
+        # show all containers and check status
         b.click("#containers-containers-filter button")
-        b.wait_visible("#containers-containers-filter .dropdown-menu")
+        b.wait_present("#containers-containers-filter .dropdown-menu")
         b.click("#containers-containers-filter li[data-data='all'] a")
         b.wait_present("#containers-containers")
```
martinpitt: This should re-use `loadAllContainerlist()`.

test/check-application
```diff
@@ -27,37 +27,111 @@ class TestApplication(testlib.MachineCase):
...
         b.click("#containers-containers-filter li[data-data='running'] a")
+        b.wait_present('#containers-containers tr:contains("busybox:latest")')
+        b.wait_present('#containers-containers tr:contains("swamped-crate")')
+        b.wait_present('#containers-containers tr:contains("busybox:latest") + tr:contains("sleep 1000")')
+
+        # switch to all containers
+        b.wait_present("#containers-containers-filter button")
+        b.click("#containers-containers-filter button")
+        b.wait_present("#containers-containers-filter .dropdown-menu")
+        b.click("#containers-containers-filter li[data-data='all'] a")
```
martinpitt: This again should re-use `loadAllContainerlist()`.

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
...
+                    let inspectRet = [];
+                    newContainersMeta.map((container) => {
+                        let proEle = new Promise((resolve, reject) => {
+                            varlinkCall(PODMAN, "io.podman.InspectContainer", {name: container.id})
+                                    .then(reply => {
+                                        resolve(JSON.parse(reply.container));
+                                    })
+                                    .catch(ex => {
+                                        reject(new Error("Failed to do InspectContainer call:", ex, JSON.stringify(ex)));
+                                    });
```
martinpitt: style nitpick: single-line function bodies should not have a `{}`.

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
...
+                .then(reply => {
+                    let newContainersMeta = reply.containers;
+                    if (!newContainersMeta) {
+                        resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                        return;
+                    }
+                    const containerCount = newContainersMeta.length;
+                    let inspectRet = [];
+                    newContainersMeta.map((container) => {
+                        let proEle = new Promise((resolve, reject) => {
```
martinpitt: style nitpick: write this a bit more elegantly with

    `let inspectPromises = newContainersMeta.map(container => new Promise(...));

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
         return _("");
     }
 }
+
+// TODO
+export function container_stop(container, timeout) {
+}
+
+export function updateContainers() {
```
martinpitt: Please add a comment what precisely this function returns, and an example that shows the structure.

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
...
+                        });
+                        inspectRet.push(proEle);
+                    });
+                    Promise.all(inspectRet)
+                            .then((inspectRet) => {
+                                let runEle = newContainersMeta.filter((ele) => {
+                                    return ele.status === "running";
+                                });
+                                inspectRet.map((inspectRet) => {
+                                    newContainers[inspectRet.ID] = inspectRet;
```
martinpitt: Here you are using `newContainers` as an object (dictionary), but you initialized it as an array (`[]`). While that happens to work somewhat, it is confusing to read, and it's all too easy to call list functions like `.map()` on them (which doesn't work).

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
...
+                                    if (runEle.length === 0 && Object.keys(newContainers).length === containerCount) {
+                                        resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                                    }
```
martinpitt: This is a strange special case. I see what you want to do with the `length === containerCount` (resolve the promise after the last iteration), but this is absolutely not necessary to do it in this convoluted way. Just move the `resolve()` after the `map()` call.

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
...
+                                });
+                            })
+                            .catch(ex => {
+                                console.error("Failed to do InspectContainer call:", ex, JSON.stringify(ex));
+                            });
+                    let containerStatsRet = [];
+                    let runCtrArr = newContainersMeta.filter((ele) => {
+                        return ele.status === "running";
+                    }).map((container) => {
```
martinpitt: The same comments from above also apply to the stats section here.

src/util.js
```diff
@@ -106,3 +107,127 @@ export function format_memory_and_limit(usage, limit) {
...
+                                    .catch(ex => {
+                                        reject(new Error("Failed to do InspectImage call:", ex, JSON.stringify(ex)));
+                                    });
+                        });
+                        inspectRet.push(proEle);
+                    });
+                    Promise.all(inspectRet)
+                            .then((inspectRet) => {
+                                inspectRet.map((inspectRet) => {
+                                    newImages[inspectRet.Id] = inspectRet;
```
martinpitt: Much the same comments here; array vs. dictionary conflict, please add a function comment about the return value structure, single-line function bodies, don't do the `inspectRet.push` call, etc.

src/Containers.jsx
```diff
@@ -176,12 +177,12 @@ class Containers extends React.Component {
...
         const containersStats = this.props.containersStats;
         // TODO: check filter text
-        let filtered = this.props.containers.filter(container => (!this.props.onlyShowRunning || container.State.Running));
-        let rows = filtered.map(function (container) {
-            return renderRow(containersStats, container);
+        let filtered = [];
+        Object.keys(this.props.containers).filter(id => { if (!this.props.onlyShowRunning || this.props.containers[id].State.Running) { filtered[id] = this.props.containers[id] } });
+        let rows = Object.keys(filtered).map(function (id) {
+            return this.renderRow(containersStats, this.props.containers[id]);
         }, this);
```
martinpitt: Replace the `function` with an arrow, drop the `this` argument. Also, `filtered` is already a list, no need for `Object.keys` (and you set `filtered` elements as object keys again, which is inconsistent) - it should be a list as you want to show it as a table, and almost certainly want to sort it, too.

src/app.jsx
```diff
@@ -31,17 +31,17 @@ class Application extends React.Component {
...
         this.state = {
             version: { version: "unknown" },
-            images: [],
-            containers: [],
-            imagesMeta: [],
-            containersMeta: [],
-            containersStats:[],
+            images: [], /* detail info of each image from InspectImage */
+            containers: [], /* detail info of each container from InspectContainer */
+            containersStats:[], /* memory usage of running containers */
```
martinpitt: Are these really lists or object maps? Please add examples that show the structure.

src/app.jsx
```diff
@@ -62,6 +62,35 @@ class Application extends React.Component {
         });
     }
 
+    updateContainerStats (newContainerStats) {
+        this.setState({
+            containersStats: newContainerStats
+        });
+    }
```
martinpitt: This is just being used once, and a wrapper around a simple function call. Please eliminate this function, to make the caller easier to read. Every React developer knows what `setState()` does, but when seeing `updateContainerStats()` in code, it's not immediately clear what this does. Also, the "container" vs. "containers" inconsistencies really don't help :)

src/app.jsx
```diff
@@ -73,12 +102,13 @@ class Application extends React.Component {
         this._asyncRequestImages = utils.varlinkCall(utils.PODMAN, "io.podman.ListImages")
```
martinpitt: Don't do the ListImages/ListContainers/InspectImage etc. calls here again. Use the utils.js functions for that.

src/Containers.jsx
QiWang19: here ```filtered``` and ```this.props.containers``` are map-like array, ```filtered[id] = this.props.containers[id]```. That's why ```Object.key``` be use. I'd better keep  ```this.props.containers``` as a map and keep ```filtered``` as an array?

test/check-application
QiWang19: Put test for the stopped alpine container in [#6](https://github.com/cockpit-project/cockpit-podman/issues/6)

src/util.js
QiWang19: Add more promise chain in [#18](https://github.com/cockpit-project/cockpit-podman/pull/18) and move it out of map()

rhatdan:
@QiWang19 Since you are breaking this up into a series of smaller PRs can you close this one?

martinpitt:
Right, let's.

PR #11, opened 2018-09-06
lib: Update Cockpit Listing and Select components

martinpitt:
- Use `import` instead of `require`
 - Drop unnecessary keys (they are only necessary for lists)
 - Fix Listing title default property
 - Support disabled elements in Select
 - Allow any data type in Select


martinpitt:
@mvollmer: This should be familiar to you, you did these changes on the Cockpit side. Can you please review?

PR #12, opened 2018-09-06
test: Properly clean up test VM on image preparation

martinpitt:
Remove the entire previous overlay instead of just removing the built
package. This ensures that there are no leftovers from previous
interactive debugging sessions.

This has already shown to lead to confusion in practice in
cockpit-podman.

Cherry-picked from starter-kit commit 6e05f5b4837a4

 - [x] Fix crash on varlink errors (PR #13)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,8 @@ rpm: dist-gzip $(RPM_NAME).spec
 
 # build a VM with locally built rpm installed
 $(VM_IMAGE): rpm bots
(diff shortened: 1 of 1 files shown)


martinpitt:
This triggers a really curious bug, will debug later.

martinpitt:
OK, what I thought - this failure also happens on master (would have been too weird to be caused by this), and it's also easy to reproduce locally. This needs to  be fixed separately.

martinpitt:
Fixed the crash in PR #13

PR #13, opened 2018-09-07
Fix crashes on error reporting

martinpitt:
If the error returned by varlink did not have a `.parameters` field, the
page oopsed with

    RuntimeError: TypeError: Cannot read property 'reason' of undefined

This also gets picked up by the test cases often and makes them fail.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -54,7 +54,7 @@ function varlinkCallChannel(channel, method, parameters) {
 function varlinkCallError(error) {
     let str = "";
     error.error ? str += error.error.toString() : str;
-    error.parameters.reason ? str += " " + error.parameters.reason.toString() : str;
+    (error.parameters && error.parameters.reason) ? str += " " + error.parameters.reason.toString() : str;
     return str;
 }
 


src/util.js
```diff
@@ -54,7 +54,7 @@ function varlinkCallChannel(channel, method, parameters) {
 function varlinkCallError(error) {
     let str = "";
     error.error ? str += error.error.toString() : str;
-    error.parameters.reason ? str += " " + error.parameters.reason.toString() : str;
+    (error.parameters && error.parameters.reason) ? str += " " + error.parameters.reason.toString() : str;
```
larskarlitski: Why is the error object converted to a string? The calling code needs to get the full parameters object from the error in order to make the right decisions. Varlink errors are defined in `.varlink` interface files and are specific to each interface. There are no special fields (I guess `reason` comes from podman?)

larskarlitski: This of course is not introduced in this PR and not a regression.

If you'd like to get this in quickly to be unblocked, I'm fine with this change.

martinpitt: Indeed there is currently no non-trivial error handling in podman. The errors pretty much just get logged as-is (thus they want a string), or shown in the UI. In the future, particular errors must be handled individually of course, then this isn't good enough any more. I filed issue #14 abuot it.

Right now this PR just fixes the Oops, it doesn't rework the error handling. I don't want to change the code too much right now, both to avoid conflicting with #10, and also because of time constraints.

larskarlitski: Agreed. Your patch fixes the crash and doesn't introduce this kind of error handling.

PR #16, opened 2018-09-18
change images and containers into map structure

QiWang19:


src/Containers.jsx
```diff
@@ -176,13 +177,11 @@ class Containers extends React.Component {
...
         let emptyCaption = _("No running containers");
-        const renderRow = this.renderRow;
         const containersStats = this.props.containersStats;
         // TODO: check filter text
-        let filtered = this.props.containers.filter(container => (!this.props.onlyShowRunning || container.State.Running));
-        let rows = filtered.map(function (container) {
-            return renderRow(containersStats, container);
-        }, this);
+        let filtered = [];
+        Object.keys(this.props.containers).filter(id => { if (!this.props.onlyShowRunning || this.props.containers[id].State.Running) { filtered.push(this.props.containers[id]) } });
```
martinpitt: This works, but it's more elegant/simple to actually use the return value of `filter()`, as the original code did. I. e.

    let filtered = Object.keys(this.props.containers).filter(id => { if (!this.props.onlyShowRunning || this.props.containers[id].State.Running));

This applies to Images.jsx as well.

src/util.js
```diff
@@ -28,7 +28,7 @@ function varlinkCallChannel(channel, method, parameters) {
 
             // FIXME: support answer in multiple chunks until null byte
             if (data[data.length - 1] != 0) {
-                reject("protocol error: expecting terminating 0");
+                reject(new Error("protocol error: expecting terminating 0"));
```
martinpitt: Note: throwing proper Error classes, and adding missing semicolons etc. are unrelated changes, and should be in separate commits (or even PRs). It's okay for now, but for the future.

src/app.jsx
```diff
@@ -73,12 +71,13 @@ class Application extends React.Component {
         this._asyncRequestImages = utils.varlinkCall(utils.PODMAN, "io.podman.ListImages")
                 .then(reply => {
                     this._asyncRequestImages = null;
-                    this.setState({ imagesMeta: reply.images });
```
martinpitt: Nice, this gets rid of the weird *Meta states.

src/app.jsx
```diff
@@ -73,12 +71,13 @@ class Application extends React.Component {
...
-                    this.state.imagesMeta.map((img) => {
+                    let imagesMeta = reply.images || [];
+                    imagesMeta.map((img) => {
                         utils.varlinkCall(utils.PODMAN, "io.podman.InspectImage", {name: img.id})
                                 .then(reply => {
-                                    const temp_imgs = this.state.images;
-                                    temp_imgs.push(JSON.parse(reply.image));
+                                    let temp_imgs = [];
+                                    Object.keys(this.state.images).filter(id => { temp_imgs[id] = this.state.images[id] });
```
martinpitt: This does't filter anything. I think the intention is to fill the `temp_imgs` dictionary, so please use `map()` in these cases. However, this transforms a dictionary (`this.state.images`) to a list, and back into a dictionary. At first (and also second) it looks like the whole effect of this line is just to copy the `this.state.images` dictionary, and then the next line changes one item.

This should rather be done with 

    this.setState({images: Object.assign({}, this.state.images, {img.id: JSON.parse(reply.image)})

or, if you want to use the l33t spread operator, you can do the copy and update like this:

    this.setState({images: {...this.state.images, img.id: JSON.parse(reply.image)})

src/app.jsx
```diff
@@ -31,11 +31,9 @@ class Application extends React.Component {
...
         this.state = {
             version: { version: "unknown" },
-            images: [],
-            containers: [],
-            imagesMeta: [],
-            containersMeta: [],
-            containersStats:[],
+            images: [], /* images[Id]: detail info of image with Id from InspectImage */
+            containers: [], /* containers[Id] detail info of container with Id from InspectContainer */
+            containersStats:[], /* containersStats[Id] memory usage of running container with Id */
```
martinpitt: These should be objects then (`{}`), not arrays. Otherwise it's all too easy to iterate over them with `map()` or similar, and wondering why nothing comes out of it.

src/util.js
QiWang19: Ok, good to know :)

QiWang19:
@martinpitt PTAL

src/Containers.jsx
```diff
@@ -179,9 +179,9 @@ class Containers extends React.Component {
         let emptyCaption = _("No running containers");
         const containersStats = this.props.containersStats;
         // TODO: check filter text
-        let filtered = [];
-        Object.keys(this.props.containers).filter(id => { if (!this.props.onlyShowRunning || this.props.containers[id].State.Running) { filtered.push(this.props.containers[id]) } });
-        let rows = filtered.map((ctr, id) => this.renderRow(containersStats, ctr));
+        let filtered = Object.keys(this.props.containers).filter(id => !this.props.onlyShowRunning || this.props.containers[id].State.Running)
+                .reduce((obj, id) => { return {...obj, [id]: this.props.containers[id]} }, {});
```
martinpitt: The first line is totally fine to build a list `filtered` that has the IDs of everything you want to show. You can directly use it in `let rows = filtered.map(id => ...)`. But now this second line converts the `filtered` list into an object (in a *very* expensive way!), and then the `Object.keys()` below turns it back into a list.

Please don't do this kind of black magic. If you have to copy and convert between data structures a lot, then this is a sign that your data structures are wrong. In this case they are actually okay (you keep the state information as a dictionary, and for rendering you filter them into a list), you are just doing some extra work and then more to undo it :)

src/Containers.jsx
```diff
@@ -26,6 +26,7 @@ class Containers extends React.Component {
         this.handleRemoveContainer = this.handleRemoveContainer.bind(this);
         this.handleCancelRemoveError = this.handleCancelRemoveError.bind(this);
         this.handleForceRemoveContainer = this.handleForceRemoveContainer.bind(this);
+        this.renderRow = this.renderRow.bind(this);
```
martinpitt: You don't actually need that, but it also doesn't hurt much.

src/Images.jsx
```diff
@@ -198,9 +199,9 @@ class Images extends React.Component {
                 [<a key={"searchImages"} role="link" tabIndex="0" onClick={this.handleSearchImageClick} className="card-pf-link-with-icon pull-right">
                     <span className="pficon pficon-add-circle-o" />{_("Get new image")}
                 </a>];
-            // TODO: filter images via filterText
-        let filtered = this.props.images;
-        let imageRows = filtered.map(this.renderRow, this);
+        // TODO: filter images via filterText
+        let filtered = Object.keys(this.props.images).filter(id => id === this.props.images[id].Id);
```
martinpitt: This is a curious way of saying "don't filter at all" :smiley:

QiWang19:
Events like delete and stop has the error because the lists are changed to dictionary. I plan to fix updateContainers and updateImages in the next patch.  After that, the dictionary in the state can be updated correctly.  😊

PR #17, opened 2018-09-19
Add tests for container states and commands

QiWang19:


QiWang19:
@martinpitt PTAL. Add some tests according to first two in [#6](https://github.com/cockpit-project/cockpit-podman/issues/6)

test/check-application
```diff
@@ -27,37 +27,48 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_present("#containers-containers-filter .dropdown-menu")
         b.click("#containers-containers-filter li[data-data='all'] a")
         b.wait_present("#containers-containers")
+        # check containers name
+        b.wait_present('#containers-containers tr:contains("swamped-crate")')
         b.wait_present('#containers-containers tr:contains("busybox:latest")')
-        b.wait_visible('#containers-containers tr:contains("busybox:latest")')
-
+        b.wait_present('#containers-containers tr:contains("test-sh")')
+        b.wait_present('#containers-containers tr:contains("alpine:latest")')
```
martinpitt: This is admittedly a bit picky, but this doesn't assert that the pod names match the image names, i. e. they are in the line where they belong to. A common pattern is to give the `<tr>` some unique attribute, like `data-name="swamped-crate"`, so that you can first select the row and then assert that it contains the container name, command, etc. E. g. 

     b.wait_in_text('#containers-containers tr[data-name="swamped-crate"]', 'swamped-crate')
     b.wait_in_text('#containers-containers tr[data-name="swamped-crate"]', 'busybox:latest')
     b.wait_in_text('#containers-containers tr[data-name="swamped-crate"]', 'running')

An even stricter way is to check the n-th row, like

     b.wait_in_text('#containers-containers tr:nth-of-type(1)', 'swamped-crate')

This of course requires that the order of the rows is predictable. This is usually a good idea anyway for UIs, i. e. usually one sorts them alphabetically by name, or first by image name then by container name, or by age or something. However, the current code doesn't do this (would be good to add, though, and it's trivial to do), so you can't use this variant yet.

Your variant below with `tr:contains()` is also very interesting, and avoids the need for data-* attributes. So I suggest to consistently use that in the above checks as well.

Also: It's also  usually easier to read and maintain to check all properties of one container (like in my example above), then the next one container, i. e. row-by-row. The above checks "column by column", and IMHO that's harder to follow mentally. (But that's a matter of taste, of course).

test/check-application
```diff
@@ -27,37 +27,48 @@ class TestApplication(testlib.MachineCase):
...
         # show container listing toggle
         b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
         b.wait_present('#containers-containers tbody tr:contains("busybox:latest") + tr button.btn-delete')
 
+        # show running container
         b.click("#containers-containers-filter button")
         b.wait_present("#containers-containers-filter .dropdown-menu")
         b.click("#containers-containers-filter li[data-data='running'] a")
-        b.wait_present("#containers-containers")
-        b.wait_in_text("#containers-containers", "No running containers")
```
martinpitt: Please keep this case as well, it's important to ensure that the UI reacts properly to the "no containers" case. I suggest the following:

  1. Ensure that "No running containers" is shown.
  2. Start a running one, reload, ensure it's shown.
  3. Start the other one that immediately exits, reload, ensure it's *not* shown.
  4. Change the dropdown to "everything", ensure that the exited container is then shown and has expected properties.
  5. Change the dropdown back to "running only" and ensure that the latter goes away again.
  6. Stop the running container, reload, ensure that the running one goes away and "no running containers" is shown again.

I think this should cover all possible transitions. Feel free to introduce a helper like

    def check_container(row_name, expected_strings)

which you call like

     check_container("alpine-latest", ["running", "sleep 1000"])

this avoids some repetition.

QiWang19:
@martinpitt  PTAL : )

test/check-application
```diff
@@ -27,37 +27,66 @@ class TestApplication(testlib.MachineCase):
...
         # show image listing toggle
         b.wait_present('#containers-images tr:contains("busybox:latest")')
-        b.wait_visible('#containers-images tr:contains("busybox:latest")')
         b.click('#containers-images tbody tr:contains("busybox:latest") td.listing-ct-toggle')
         b.wait_present('#containers-images tbody tr:contains("busybox:latest") + tr button.btn-delete')
 
+        # make sure no running containers shown
+        b.click("#containers-containers-filter button")
+        b.wait_present("#containers-containers-filter .dropdown-menu")
+        b.click("#containers-containers-filter li[data-data='running'] a")
```
martinpitt: This should use your shiny new `filter_continers()` function. Also, it's not strictly necessary as "running" ought to be the default setting. But it's ok to be explicit about it, too.

test/check-application
```diff
@@ -27,37 +27,66 @@ class TestApplication(testlib.MachineCase):
...
+        # check exited alpine not in running list
+        b.wait_not_in_text("#containers-containers", "alpine:latest")
+
+    def check_container(self, row_name, expected_strings):
+        """Check the container with row_name has the expected_string shown in the row"""
+        b = self.browser
+        for str in expected_strings:
+            b.wait_in_text('#containers-containers tr:contains(' + row_name + ')', str)
+
+    def filter_continers(self, menu_text):
```
martinpitt: I buy an 'a' :smile:

test/check-application
```diff
@@ -27,37 +27,66 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_present('#containers-containers tr:contains("busybox:latest")')
+        self.check_container('swamped-crate', ['swamped-crate', 'busybox:latest', 'sleep 1000', 'running'])
+        # check exited alpine not in running list
+        b.wait_not_in_text("#containers-containers", "alpine:latest")
+
+    def check_container(self, row_name, expected_strings):
+        """Check the container with row_name has the expected_string shown in the row"""
+        b = self.browser
+        for str in expected_strings:
+            b.wait_in_text('#containers-containers tr:contains(' + row_name + ')', str)
```
martinpitt: Completely unimportant style nitpick/hint: The usual python way is to use string interpolation:

    'foo %s bar %s baz' % (firstvar, secondvar)

But it's fine like that, too.

QiWang19: :+1:

QiWang19:
@martinpitt PTAL : )

test/check-application
```diff
@@ -27,37 +27,64 @@ class TestApplication(testlib.MachineCase):
...
+        """Use dropdown menu in the header to filter containers"""
+        b = self.browser
+        b.wait_present("#containers-containers-filter button")
         b.click("#containers-containers-filter button")
         b.wait_present("#containers-containers-filter .dropdown-menu")
-        b.click("#containers-containers-filter li[data-data='running'] a")
-        b.wait_present("#containers-containers")
-        b.wait_in_text("#containers-containers", "No running containers")
-
+        b.click("#containers-containers-filter li[data-data= %s] a" % menu_text)
```
martinpitt: that looks a little curious with the space and unquoted string, but if it works, then fine :-) I'll wait for the tests.

Issue #14, opened 2018-09-07
Fix error handling


martinpitt:
In utils.js `handleVarlinkCallError()`, don't stringify the error, but return it as-is, so that callers can handle different kinds of errors differently. There could be a separate util method to stringify an error for logging or showing to the user, of course.

marusak:
@martinpitt can you elaborate what exactly is this issue about?
If it is about returning caller the error not as string, but as object, than that is happening:
```js
.catch(ex => {                                                      
  handleVarlinkCallError(ex);                                     
  return Promise.reject(ex);                                      
});
```
`handleVarlinkCallError` only logs the error, and then the promise is rejected with the original error object, so callers can deal with it separately.

martinpitt:
Oh nice, and now it's even just being used in one place. It isn't even exported any more. This was a lot more messy when I filed this, great to see that it got fixed now.

PR #18, opened 2018-09-20
Update containers in util & add util function call in app.jsx

QiWang19:


martinpitt:
This crashes the UI. Tests fail with `TypeError: Cannot read property 'bind' of undefined`.

martinpitt:
@QiWang19 : FYI, I just added you to cockpit's "Contributors" team. With that the tests will start automatically for your PRs now, instead of me having to authorize them.

QiWang19:
@martinpitt PTAL : )

src/util.js
```diff
@@ -107,3 +107,80 @@ export function format_memory_and_limit(usage, limit) {
...
+                        return new Promise((resolve, reject) => {
+                            varlinkCall(PODMAN, "io.podman.InspectContainer", {name: container.id})
+                                    .then(reply => {
+                                        resolve(JSON.parse(reply.container));
+                                    })
+                                    .catch(ex => reject(new Error("Failed to do InspectContainer call:", ex, JSON.stringify(ex))));
+                        });
+                    });
+                })
+                .then(function() {
```
martinpitt: I find it hard to mentally follow what's going on in this "ladder" of multiple chained `.then()`s. The block above  does not by itself return a promise, so I believe this isn't really a chain that waits on subsequent promises, but all of these `.then()` blocks run directly after one another.

I think it would make things much easier to understand if all of that code runs in a single `.then()` handler of the `ListContainers` call. (For the record, you are allowed to use empty lines to set apart different steps in the code, to make it easier to read ☺)

Also, style nitpick: Please use `function` or `=>` consistently. I suggest to use `() =>` here; in general, arrow functions are what you want in such promise handlers, as it has a much more intuitive semantics of `this`.

src/util.js
```diff
@@ -107,3 +107,80 @@ export function format_memory_and_limit(usage, limit) {
...
+    let inspectRet = [];
+    let containerStatsRet = [];
+    let countRun;
+    return new Promise((resolve, reject) => {
+        varlinkCall(PODMAN, "io.podman.ListContainers")
+                .then(reply => {
+                    newContainersMeta = reply.containers;
+                    if (!newContainersMeta) {
+                        console.log(newContainersMeta);
+                        resolve({newContainers: newContainers, newContainersStats: newContainersStats});
```
martinpitt: Note that this does *not* magically end the handler function, i. e. this is not an implied `return`. All of the code below will still run and resolve the promise again, with (hopefully :smile: ) the same result. I suggest to just drop this `if` condition - everything above should get along with the list being empty.

src/util.js
```diff
@@ -107,3 +107,80 @@ export function format_memory_and_limit(usage, limit) {
...
+                        });
+                    });
+                })
+                .then(function() {
+                    const containerCount = newContainersMeta.length;
+                    Promise.all(inspectRet)
+                            .then((inspectRet) => {
+                                let runEle = newContainersMeta.filter((ele) => {
+                                    return ele.status === "running";
+                                });
```
martinpitt: Drop the `{ return [...]; }` for single-expression handlers, please: `filter(ele => ele.status === "running");`

src/util.js
```diff
@@ -107,3 +107,80 @@ export function format_memory_and_limit(usage, limit) {
...
+                                inspectRet.map((inspectRet) => {
+                                    newContainers[inspectRet.ID] = inspectRet;
+                                });
+                                if (runEle.length === 0 && Object.keys(newContainers).length === containerCount) {
+                                    resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                                }
```
martinpitt: This condition is not easy to understand. As for the length comparison, how would that ever be false? `Promise.all.then()` already guarantees that all promises succeeded and thus `newContainers` has the same number of elements as `containerCount`. I think you can just drop that condition and the `containerCount` variable. 

In general, if you have to do this kind of explicit length comparison with data type iteration, then this smells like there's something wrong with your data structures or your iteration. These seem alright here, and the condition is just superflous.

Likewise, if there are no running containers, you resolve the returned promise here, but *don't* return the handler. The code below will still run, no? (See the above "hard to understand the structure and then ladder"). 

I think the code should still be correct and much easier to understand if the `Promise.all.then` handler consists only of the above `inspectRet.map()` call.

src/util.js
```diff
@@ -107,3 +107,80 @@ export function format_memory_and_limit(usage, limit) {
...
+                            })
+                            .catch(ex => console.error("Failed to do InspectContainer call:", ex, JSON.stringify(ex)));
+                })
+
+                .then(function() {
+                    containerStatsRet = newContainersMeta.filter((ele) => {
+                        return ele.status === "running";
+                    }).map((container) => {
```
martinpitt: Please don't compute such values twice. I know, I just told you that the other one can go entirely, but if you ever need it twice, compute it at the top and assign it to a `const`.

src/util.js
```diff
@@ -107,3 +107,80 @@ export function format_memory_and_limit(usage, limit) {
...
+                                        console.error("Failed to do GetContainerStats call:", ex, JSON.stringify(ex));
+                                        reject(new Error("Failed to do GetContainerStats call:", ex, JSON.stringify(ex)));
+                                    });
+                        });
+                    });
+                    countRun = containerStatsRet.length;
+                })
+                .then(() => {
+                    Promise.all(containerStatsRet)
+                            .then((containerStatsRet) => {
```
martinpitt: Please name this argument differently. Up to here, `containerStatsRet` is a list of promises, now it's a list of promise results.

src/util.js
```diff
@@ -107,3 +107,80 @@ export function format_memory_and_limit(usage, limit) {
...
+                                containerStatsRet.map((containerStatsRet) => {
```
martinpitt: Same here, now this is overloaded a third time for a single container stat. Just call it `stats` or so :)

src/util.js
```diff
@@ -107,3 +107,80 @@ export function format_memory_and_limit(usage, limit) {
...
+                                    newContainersStats[containerStatsRet.ctrId] = containerStatsRet.ctrStats;
+                                });
+                                if (countRun > 0 && Object.keys(newContainersStats).length === countRun) {
```
martinpitt: Same story for the length comparison (should just go).

So the `countRun` here would make sure to not resolve the returned promise again if it already happened above. But all of this code just works for no running containers -- `Promise.all([]).then()` immediately resolves and runs the then handler.

So do all of the work above unconditionally, and unconditionally resolve the returned promise here, and do it all in one .then handler. The code will be much smaller, simpler, and avoid branches.

src/util.js
QiWang19: I was wondering if using only one .then(), the functions inside then() will execute not in sequence and app.jsx can not get the complete data, it is possible?
```js
new Promise(() => {
    ListContainers()
        .then() {
            InspectContainer();
            getStats();
            resolve()
        }
})
```

martinpitt: It's actually the other way around. In every imperative programming language, statements in a function are always done in sequence - JavaScript is no different.

With multiple `.then()`s, there is the possibility of chaining them or running them in parallel, that's where things get complicated (see https://javascript.info/promise-chaining). But  we don't need that here.

QiWang19:
@martinpitt  PTAL :)

src/util.js
```diff
@@ -107,3 +107,55 @@ export function format_memory_and_limit(usage, limit) {
...
+        varlinkCall(PODMAN, "io.podman.ListContainers")
+                .then(reply => {
+                    let newContainersMeta = reply.containers;
+                    let inspectRet = newContainersMeta.map((container) => {
+                        return new Promise((resolve, reject) => {
+                            varlinkCall(PODMAN, "io.podman.InspectContainer", {name: container.id})
+                                    .then(reply => resolve(JSON.parse(reply.container)))
+                                    .catch(ex => reject(new Error("Failed to do InspectContainer call:", ex, JSON.stringify(ex))));
+                        });
+                    });
```
martinpitt: Note that wrapping `varlinkCall` into a `Promise` doesn't give you anything here. `inspectRet` can just directly be a list of the `varlinkCall` promises. This would further simplifiy the code.

It's also worthwhile to not "destroy" the original exception into a flag string this early -- at some point you want to handle individual exceptions differently in the caller (see  issue #14), so just propagate it as-is to the caller, and let them decide what to do with them (for now, it's ok to just print them).

Also notice that a three-argument Error with (string, exception, string) is rather weird.

src/util.js
```diff
@@ -107,3 +107,55 @@ export function format_memory_and_limit(usage, limit) {
...
+
+                    Promise.all(inspectRet)
+                            .then((inspectRet) => {
```
martinpitt: Naming overloading again -- `inspectRet` is a list of promises, now it turns into a list of call results. Please choose a different name, like `results`.

src/util.js
```diff
@@ -107,3 +107,55 @@ export function format_memory_and_limit(usage, limit) {
...
+                                inspectRet.map((inspectRetEle) => {
+                                    newContainers[inspectRetEle.ID] = inspectRetEle;
+                                });
+                                resolve({newContainers: newContainers, newContainersStats: newContainersStats});
```
martinpitt: It's too early to resolve the top-level promise here, as you still need to process the ContainerStats() results. Just drop this.

src/util.js
```diff
@@ -107,3 +107,55 @@ export function format_memory_and_limit(usage, limit) {
...
+                                containerStatsRet.map((containerStatsRetEle) => {
+                                    newContainersStats[containerStatsRetEle.ctrId] = containerStatsRetEle.ctrStats;
+                                });
+                                resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                            })
+                            .catch(ex => console.error("Failed to do GetContainerStats call:", ex, JSON.stringify(ex)));
+                })
+                .catch(ex => {
+                    console.error("Failed to do ListContainers call:", JSON.stringify(ex), ex.toString());
+                    reject(new Error("Failed to do ListContainers call"));
```
martinpitt: This error is rather useless for the caller, as it loses any information about the reason why it failed. I suggest to just propagate it on, i. e. `reject(ex)`.

src/util.js
```diff
@@ -107,3 +107,55 @@ export function format_memory_and_limit(usage, limit) {
...
+                    });
+
+                    Promise.all(inspectRet)
+                            .then((inspectRet) => {
+                                inspectRet.map((inspectRetEle) => {
+                                    newContainers[inspectRetEle.ID] = inspectRetEle;
+                                });
+                                resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                            })
+                            .catch(ex => console.error("Failed to do InspectContainer call:", ex, JSON.stringify(ex)));
```
martinpitt: Here you also need to reject the returned top-level promise.

src/util.js
```diff
@@ -107,3 +107,55 @@ export function format_memory_and_limit(usage, limit) {
...
+                            });
+
+                    Promise.all(containerStatsRet)
+                            .then((containerStatsRet) => {
+                                containerStatsRet.map((containerStatsRetEle) => {
+                                    newContainersStats[containerStatsRetEle.ctrId] = containerStatsRetEle.ctrStats;
+                                });
+                                resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                            })
+                            .catch(ex => console.error("Failed to do GetContainerStats call:", ex, JSON.stringify(ex)));
```
martinpitt: Same here, this needs to reject the top-level promise and propagate the exception.

QiWang19:
@martinpitt PTAL : )

src/util.js
```diff
@@ -107,3 +107,51 @@ export function format_memory_and_limit(usage, limit) {
...
+                            .map((container) => {
+                                return varlinkCall(PODMAN, "io.podman.GetContainerStats", {name: container.id});
+                            });
+                    Promise.all(containerStatsRet)
+                            .then((replies) => {
+                                replies.map((reply) => {
+                                    let ctrStatsRet = reply.container;
+                                    newContainersStats[ctrStatsRet.id] = ctrStatsRet;
+                                });
+                                resolve({newContainers: newContainers, newContainersStats: newContainersStats});
```
martinpitt: I just noticed that this is a race condition. It could be that all the `GetContainerStats()` calls finish before all the `InspectContainer()` calls -- in that case the promise would be resolved too early, with `newContainers` still being empty or incomplete. So instead, below this you need another

    Promise.all(inspectRet.concat(containerStatsRet)).then(() => resolve(..));

You don't need to handle errors there, these are already handled here.

src/util.js
```diff
@@ -107,3 +107,51 @@ export function format_memory_and_limit(usage, limit) {
...
+export function updateContainers() {
+    let newContainers = {};
+    let newContainersStats = {};
+    return new Promise((resolve, reject) => {
+        varlinkCall(PODMAN, "io.podman.ListContainers")
+                .then(reply => {
+                    let newContainersMeta = reply.containers;
+                    let inspectRet = newContainersMeta.map((container) => {
+                        return varlinkCall(PODMAN, "io.podman.InspectContainer", {name: container.id});
+                    });
```
martinpitt: Style nitpick: You can drop the `{ return }`, it's just a single expression.

QiWang19:
@martinpitt PTAL :) 
For the meta-question, your comments are pretty good for me. They are not too verbose and I learned a lot from you. You told me what to change and also why I need to do that. And also the links of some characteristics of the language are really helpful. Thank you for your patience. :smile:

src/util.js
```diff
@@ -107,3 +107,64 @@ export function format_memory_and_limit(usage, limit) {
...
+                                let lastIdx = inspectRet.length - 1;
+                                replies.map((reply, idx) => {
+                                    if (idx <= lastIdx) {
+                                        let ctrInspectRet = JSON.parse(reply.container);
+                                        newContainers[ctrInspectRet.ID] = ctrInspectRet;
+                                    } else {
+                                        let ctrStatsRet = reply.container;
+                                        newContainersStats[ctrStatsRet.id] = ctrStatsRet;
+                                    }
+                                });
```
martinpitt: Don't duplicate all this code, all of this has already been done above. You just need to resolve the promise (the line below) here.

QiWang19:
@martinpitt  PTAL :)

PR #19, opened 2018-09-26
Fix container delete and add test for it

QiWang19:


QiWang19:
@martinpitt PTAL :)

src/util.js
```diff
@@ -120,6 +120,10 @@ export function updateContainers() {
         varlinkCall(PODMAN, "io.podman.ListContainers")
                 .then(reply => {
                     let newContainersMeta = reply.containers;
+                    if (newContainersMeta === null) {
+                        resolve({newContainers: newContainers, newContainersStats: newContainersStats});
+                        return;
+                    }
```
martinpitt: Out of interest, is that related to this PR? It seems odd that ListContainers would ever answer with a null value, instead of an empty list. It's okay to handle it like this, although it would be simpler to provide a fallback value:

    let newContainersMeta = reply.containers || [ ];

test/check-application
```diff
@@ -72,6 +72,23 @@ class TestApplication(testlib.MachineCase):
         # check exited alpine not in running list
         b.wait_not_in_text("#containers-containers", "alpine:latest")
 
+        # delete running container busybox using force delete
+        b.click('#containers-containers tbody tr:contains("busybox:latest") + tr button.btn-delete')
+        self.confirm_modal("btn-danger")
+        self.confirm_modal("btn-danger")
```
martinpitt: Having to confirm two consecutive dialogs about the same thing feels strange to me. Of course this behaviour can be fixed in a follow-up PR, i. e. only have a "delete" or "force delete" dialog, not both.

However, in this case the test should make some assertions about the dialog, like verifying the primary button label ("Delete" or "Force Delete"). You could pass the expected button label as an argument to `confirm_modal()`.

test/check-application
```diff
@@ -72,6 +72,23 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_not_present(".modal-dialog div")
```
martinpitt: Right now you have a race condition between the two `confirm_modal()` calls. The function itself doesn't wait for the dialog to close, and you also don't call this in between the two dialogs. So the second call would still see the first dialog before it closes.

I suggest to move this `wait_not_present` into `confirm_modal()`, to avoid forgetting about it.

src/util.js
```diff
@@ -110,7 +110,7 @@ export function format_memory_and_limit(usage, limit) {
 
 // TODO: handle different kinds of errors
 function handleVarlinkCallError(ex) {
-    console.error("Failed to do varlinkcall:", JSON.stringify(ex));
+    console.error("Failed to do varlinkcall:", ex);
```
KKoukiou: This is not good change for too reasons:
1rst when you are running the tests, and you are seeing the javascript logs in your terminal, you need JSON.stringify to see the actual object.
2nd this change is not relevant to this PR, it's good approach to keep the PRs do exactly what they mention in the commit message, so that in case something break, it will be easier to find which commit broke it from the history.

QiWang19: thanks, I will change it back.

QiWang19:
@martinpitt PTAL

test/check-application
```diff
@@ -72,6 +72,21 @@ class TestApplication(testlib.MachineCase):
...
         b.wait_not_in_text("#containers-containers", "alpine:latest")
 
+        # delete running container busybox using force delete
+        b.click('#containers-containers tbody tr:contains("busybox:latest") + tr button.btn-delete')
+        self.confirm_modal("btn-ctr-delete")
+        self.confirm_modal("btn-ctr-focedelete")
+        b.wait_not_in_text("#containers-containers", "busybox:latest")
+
+        # delete the exited alpine
+        self.filter_continers("all")
```
martinpitt: Can you please do a separate PR to fix that typo?

PR #20, opened 2018-09-26
Update images in util & add util function call in app.jsx

QiWang19:


QiWang19:
@martinpitt PTAL :)

src/util.js
```diff
@@ -156,3 +156,34 @@ export function updateContainers() {
...
+export function updateImages() {
+    let newImages = {};
+    return new Promise((resolve, reject) => {
+        varlinkCall(PODMAN, "io.podman.ListImages")
+                .then(reply => {
+                    let newImagesMeta = reply.images;
+                    if (newImagesMeta === null) {
+                        resolve(newImages);
+                        return;
+                    }
```
martinpitt: First, this API really ought to not return `null`. But of course it can't hurt to be robust against that. But, please don't write this as a special-case: Get rid of the entire `if` and just have `let newImagesMeta = reply.images || []`. The code below will correctly handle the "empty list" case as well.

QiWang19:
@martinpitt PTAL :)

PR #22, opened 2018-09-27
Allow running eslint with `npm run eslint`

KKoukiou:
Document this command in HACKING.md file

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
   "author": "",
   "license": "LGPL-2.1",
   "scripts": {
-    "build": "webpack"
+    "build": "webpack",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,7 +7,8 @@
   "author": "",
   "license": "LGPL-2.1",
   "scripts": {
-    "build": "webpack"
+    "build": "webpack",
+    "eslint": "eslint --ext .jsx --ext .es6 src/"
```
martinpitt: Can you please also add this

    "eslint:fix": "eslint --fix --ext .jsx --ext .es6 src/"

This is rather useful for fixing whitespace, semicolons, etc.

Also, can you please add `--ext .js` as well, just in case?

PR #26, opened 2018-10-01
Add dropdown menu for start&stop container and test the menu

QiWang19:


QiWang19:
@martinpitt  PTAL

src/Dropdown.jsx
```diff
@@ -1,40 +1,48 @@
...
+
+        return (
+            <div className="btn-group" >
+                <Select.Select initial={this.state.title} onChange={(event) => this.handleClick(event)}>
+                    {menuItems}
+                </Select.Select>
+            </div>
+        );
+    }
+}
```
martinpitt: I really wonder if it's that beneficial to wrap an already existing `Select` React component into this new one -- there's really no extra logic here, just some glue.

QiWang19: looks like ```div``` can keep the distance between this dropdown button and the commit button in its left. Delete this div now?

KKoukiou:
This functionality is already implemented.

PR #27, opened 2018-10-01
Use spread operator in Images and Containers & delete useless props

QiWang19:


QiWang19:
@martinpitt PTAL

QiWang19:
Ok, I will change it back. Just because you pointed out my old method to update props may directly change the state rather than a copy of it. I think this may be a safe way to pass props anyway. Thanks

QiWang19:
@martinpitt  PTAL

PR #28, opened 2018-10-01
Implement text filter in ContainerHeader

QiWang19:


QiWang19:
@martinpitt  PTAL

QiWang19:
```
        # test filter
        b.wait_present(".content-filter div input")
        b.set_val(".content-filter div input", "swamp")
        b.wait_not_in_text("#containers-containers", "alpine:latest")
```
Is this the right method to test the filter, I got the screenshot because test failed, but cannot replicate the this problem in `http://127.0.0.2:9091`
testapplication-testbasic-fedora-28-127 0 0 2-2201-fail

screenshot from 2018-10-02 20-31-15

martinpitt:
Use `.set_input_text` instead of `.set_val()`. The latter doesn't work with React, as it circumvents the key event handlers.

QiWang19:
I also failed to show the containers come back after set input value to ""  
```
b.set_input_text(".content-filter div input", "")
b.wait_in_text("#containers-containers", "alpine:latest")
```
testapplication-testbasic-fedora-28-127 0 0 2-2201-fail

PR #29, opened 2018-10-01
Add modal for container commit

QiWang19:


src/ContainerCommitModal.jsx
```diff
@@ -0,0 +1,174 @@
...
+            message: "",
+            command: this.props.containerWillCommit.Config ? this.props.containerWillCommit.Config.Cmd.join(" ") : "",
+            pause: true,
+            setonbuild: false,
+            onbuild: [""],
+            format: "oci",
+            selectedFormat: "oci",
+            onbuildDisabled: true,
+        };
+        this.state = this.initialState;
```
martinpitt: You need to deep-copy the initial state here.

src/ContainerCommitModal.jsx
```diff
@@ -0,0 +1,174 @@
...
+        this.handleInputChange = this.handleInputChange.bind(this);
+        this.handleCancel = this.handleCancel.bind(this);
+        this.handleCommit = this.handleCommit.bind(this);
+        this.handleOnbuildsInputChange = this.handleOnbuildsInputChange.bind(this);
```
martinpitt: Please name these  like "OnBuilds", it's two words.

src/ContainerCommitModal.jsx
```diff
@@ -0,0 +1,174 @@
...
+                    </form>
+                </div>
+            ));
+
+        let commitContent =
+            <div>
+                <table className="form-table-ct">
+                    <tbody>
+                        <tr>
+                            <td align="right"><label className="control-label" translatable="yes">Container Name</label></td>
```
martinpitt: The `translatable=` tag only works in static .html files, not in React. You need to use `<label>{_("Container Name"}}</label>`.

src/ContainerCommitModal.jsx
```diff
@@ -0,0 +1,174 @@
...
+                                <label htmlFor="format-docker">
+                                    <input type="radio" id="format-docker" value="docker" checked={this.state.selectedFormat === 'docker'} onChange={(event) => this.handleFormatChange(event)} />
+                                    docker
+                                </label>
+                            </td>
+                        </tr>
+                        <tr>
+                            <td align="right"><label className="control-label" translatable="yes">Image Name</label></td>
+                            <td colSpan="3">
+                                <input name="imageName" className="form-control-commit container-imageName" type="text" onChange={this.handleInputChange} required="true" />
```
martinpitt: Please leave out the input field specific class names. You don't define CSS for them, and also really shouldn't - these input fields are all alike, they shouldn't all have a different styling.

KKoukiou:
Pasting the screenshot of the modal window here.
commit-image

Can't you copy the css from the commit modal of docker plugin?

KKoukiou:
@QiWang19 here are some issues I see. Let me know if you planned to fix them in another PR.

* Why image name and tag forms are highlighted with red when other attributes not?
* Author should be auto filled with the logged in users name. (This can be part of another PR)
* Some of the parameters look that they are not aligned with the inputs.
ex. 

Pause | pause the container
On Build | Set container on build variables

Resending screenshot of what I see. Please drop it in PR yourself next time you make some UI change :)

podman-commit

One last thing I noticed that you decided to skip the repository parameter.
$podman commit --help
USAGE:
   podman commit [command options] CONTAINER [REPOSITORY[:TAG]]

I know it's not required, but doesn't it make sense to expose it?

KKoukiou:
@QiWang19 also the commit message is misleading, you should mentioned that you added the commit modal in this PR, instead of only mentioning the css and tests.

src/ContainerCommitModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+        this.setState({onbuild: this.state.onbuild.concat([""])});
+    }
+
+    handleRemoveOnBuild(idx) {
+        this.setState({onbuild: this.state.onbuild.filter((bud, sidx) => idx !== sidx)});
+    }
+
+    // TODO
+    handleInputChange(event) {
+
```
KKoukiou: I think you 'll need here to pass also some key to identify which one from the input forms had changed since you 're using this function as callback in multiple places.
But since it's not implemented you can do it when you implement it.

src/ContainerCommitModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+                            </td>
+                        </tr>
+                        <tr>
+                            <td valign="top" align="right"><label className="control-label">{_("On Build")}</label></td>
+                            <td colSpan="3">
+                                <label>
+                                    <input name="setonbuild" className="container-label" type="checkbox" disabled={this.state.onbuildDisabled} onChange={this.handleInputChange} />
+                                    <span>{_("Set container on build variables")}</span>
+                                </label>
+                                {(this.state.setonbuild && <div>{onbuilds}</div>) }
```
KKoukiou: It's a bit weird that you have implemented the build variables forms, but we can't actually see it at this point in the UI, since the handleInputChanged for the checkbox is not implemented.
I think that it makes sense to implement at a certain point only what can be actually reached in UI and not more.

QiWang19:
I can't directly match the parameters with podman command, I used varlink [commit](https://github.com/containers/libpod/blob/master/API.md#func-commit) function. Other arguments will be put in the modal that is used to run an image.

QiWang19:
commitmodal

QiWang19:
@KKoukiou

src/ContainerCommitModal.jsx
```diff
@@ -0,0 +1,191 @@
...
+                        <button type="button" className="btn btn-default fa fa-plus" onClick={this.handleAddOnBuild} />
+                        <button type="button" className="btn btn-default pficon-close" disabled={idx === 0} onClick={() => this.handleRemoveOnBuild(idx)} />
+                        <div className="form-group">
+                            <input type="text" name="onbuildvar_key" className="form-control" onChange={(evt) => this.handleOnBuildsInputChange(idx, evt)} />
+                        </div>
+                    </form>
+                </div>
+            ));
+
+        let commitContent =
```
KKoukiou: What about implementing the whole modal content with patternfly-react? It will make the code simpler and also get rid of the css.
This modal for example https://rawgit.com/patternfly/patternfly-react/gh-pages/patternfly-3/index.html?selectedKind=patternfly-react%2FForms%20and%20Controls%2FModal%20Overlay&selectedStory=Modal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

and `show info` to check the code.

I did it recently for some other dialog so I can help you with adjusting it if you need.

QiWang19: Thanks, I'll take a look and change it later.

KKoukiou: Looks much better, thanks!
You can also change the forms to patternfly-react if you want but take that just I a ssuggestion,
Each form group could be something like
``` js
<FormGroup controlId='message' disabled={false}>
      <Grid.Col componentClass={Form.ControlLabel} sm={3}>
        {_("Message")}
      </Grid.Col>
      <Grid.Col sm={9}>
         <input name="message" className="form-control" type="text" onChange={this.handleInputChange} /> 
      </Grid.Col>
</FormGroup>
```

package.json
```diff
@@ -50,10 +50,12 @@
     "moment": "^2.22.2",
     "node-sass": "^4.9.0",
     "patternfly": "^3.54.1",
+    "patternfly-react": "^2.20.3",
     "prop-types": "^15.6.1",
     "react": "16.0.0",
     "react-addons-css-transition-group": "^15.6.2",
     "react-addons-transition-group": "^15.6.2",
+    "react-bootstrap": "^0.32.4",
```
KKoukiou: Not used any more.

src/ContainerCommitModal.jsx
```diff
@@ -0,0 +1,191 @@
...
+            message: "",
+            command: this.props.containerWillCommit.Config ? this.props.containerWillCommit.Config.Cmd.join(" ") : "",
+            pause: true,
+            setonbuild: false,
+            onbuild: [""],
+            format: "oci",
+            selectedFormat: "oci",
+            onbuildDisabled: true,
+        };
+        this.state = {...this.initialState};
```
KKoukiou: Do you really need the initialState variable?
I see you use it in the cancel handler to reset the components state, but this would have the same results as just keeping this object in the state variable, which will get reinitialized next time the component will re-render. Or I am missing something?

QiWang19: The initialState is used for changing the checkbox or input field back to default when to reopen the same modal. If the first time I open the modal and make some changes to checkbox or input filed and closed the modal using cancel. Then I use the commit button to open the modal again, the modal should go back to the default.

KKoukiou:
@QiWang19 it looks ok to me, but I see this error in console.
``` js
warning.js:33 Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in.
...
./src/Containers.jsx | @ | index.js:132969
...
```
I complains about the ContainerCommitModal import. It looks correct to me however. Do you see this error as well?

Second, you don't use PropTypes at all. It's a good practice, it helps preventing undefined behaviours and also makes is visible which props are required from your class. But since it's under early development I don't have anything against adding PropTypes later on, in another PR.

PR #34, opened 2018-10-23
Replace <form/> with <FormGroup/> inside <Form/>

QiWang19:
Replace `<form/> `with `<FormGroup/>` inside `<Form/>`, otherwise there will be a warning.


QiWang19:
@KKoukiou PTAL

PR #35, opened 2018-10-25
Implement Commit and tests

QiWang19:


src/ContainerCommitModal.jsx
```diff
@@ -67,9 +67,10 @@ class ContainerCommitModal extends React.Component {
         });
     }
 
-    // TODO
     handleCommit() {
-
+        const commitMsg = {...this.state};
```
KKoukiou: commitMsg is used only here. I think it makes sense to use this.state directly here.

src/Containers.jsx
```diff
@@ -79,9 +79,43 @@ class Containers extends React.Component {
...
+        }
+        commitData.changes.push(...onbuildsArr);
 
+        utils.varlinkCall(utils.PODMAN, "io.podman.Commit", commitData)
+                .then(reply => {
+                    this.props.updateImagesAfterEvent();
+                    this.props.updateContainersAfterEvent();
+                })
+                .catch(ex => {
+                    console.error("Failed to do Commit call:", ex, JSON.stringify(ex));
```
KKoukiou: If Commit call failed the user should see error notification in the open dialog.
In other words the dialog wondow should not close IMO if:
* Commit varlinkCall failed
* Some of the options are invalid (we can perform some empty string checks there for example)

test/check-application
```diff
@@ -93,9 +93,26 @@ class TestApplication(testlib.MachineCase):
...
+        b.set_input_text("#author", "tester")
+        b.set_input_text("#command", "sleep 6000")
+        b.click(".modal-dialog div .btn-ctr-commit")
+        b.wait_not_present(".modal-dialog div")
+        b.wait_present('#containers-images tr:contains("testimg:testtag")')
+        b.click('#containers-images tbody tr:contains("testimg:testtag") td.listing-ct-toggle')
+        # open the listing toggle of testimg and check the commit paramerters
+        b.wait_present('#containers-images tbody tr:contains("testimg:testtag"):has(dd:contains("localhost/testimg:testtag"))')
+        b.wait_present('#containers-images tbody tr:contains("testimg:testtag"):has(dd:contains("sleep 6000"))')
+        b.wait_present('#containers-images tbody tr:contains("testimg:testtag"):has(dd:contains("tester"))')
```
KKoukiou: There should be also test for failed image creation from commited container.
Notification in the UI should be checked in this case.

QiWang19:
@KKoukiou PTAL

src/Containers.jsx
```diff
@@ -79,9 +83,60 @@ class Containers extends React.Component {
...
+            onbuildsArr = utils.getCommitArr(commitMsg.onbuild, "ONBUILD");
+        }
+        commitData.changes.push(...onbuildsArr);
+
+        utils.varlinkCall(utils.PODMAN, "io.podman.Commit", commitData)
+                .then(reply => {
+                    this.props.updateImagesAfterEvent();
+                    this.props.updateContainersAfterEvent();
+                    this.setState((prevState) => ({
+                        setContainerCommitModal: !prevState.setContainerCommitModal
```
KKoukiou: Is there reason that you take here the prevState and not just set it to false?

QiWang19:
:+1: I like it. Do you know how to set the default value of input field in `<FormControl/>`? I wanted to set command value there but failed.

PR #37, opened 2018-11-13
fix image delete & add tests for deletion

QiWang19:


QiWang19:
@KKoukiou PTAL

src/ImageDeleteModal.jsx
```diff
@@ -17,7 +17,7 @@ const ImageDeleteModal = (props) => {
                 </Modal.Body>
                 <Modal.Footer>
                     <Button onClick={props.handleCancelImageDeleteModal}>Cancel</Button>
-                    <Button bsStyle="danger" onClick={props.handleRemoveImage}>{_("Delete")}</Button>{' '}
+                    <Button className="btn-img-delete" bsStyle="danger" onClick={props.handleRemoveImage}>{_("Delete")}</Button>{' '}
```
KKoukiou: Use id for navigatingin UI from the tests. classNames are supposed to use for css for multiple elements.

src/ImageRemoveErrorModal.jsx
```diff
@@ -17,7 +17,7 @@ const ImageRemoveErrorModal = (props) => {
                 </Modal.Body>
                 <Modal.Footer>
                     <Button onClick={props.handleCancelImageRemoveError}>{_("Cancel")}</Button>
-                    <Button bsStyle="danger" onClick={props.handleForceRemoveImage}>{_("Force Delete")}</Button>
+                    <Button className="btn-img-deleteerror" bsStyle="danger" onClick={props.handleForceRemoveImage}>{_("Force Delete")}</Button>
```
KKoukiou: Same here.

src/Images.jsx
```diff
@@ -78,10 +78,8 @@ class Images extends React.Component {
         });
         utils.varlinkCall(utils.PODMAN, "io.podman.RemoveImage", {name: image})
                 .then((reply) => {
-                    const idDel = reply.image ? reply.image : "";
-                    const oldImages = this.props.images;
-                    let newImages = oldImages.filter(elm => elm.Id !== idDel);
-                    this.props.updateImages(newImages);
+                    this.props.updateImagesAfterEvent();
+                    this.props.updateContainersAfterEvent();
```
KKoukiou: Maybe I am missing something, but why the updateContainersAfterEvent is needed here? You delete some image, it suceeds, why you need to update containers?

src/Images.jsx
```diff
@@ -78,10 +78,8 @@ class Images extends React.Component {
         });
         utils.varlinkCall(utils.PODMAN, "io.podman.RemoveImage", {name: image})
                 .then((reply) => {
```
KKoukiou: reply is not used anymore, you can use ```()```.

test/check-application
```diff
@@ -132,6 +132,23 @@ class TestApplication(testlib.MachineCase):
         b.click(".modal-dialog div .alert .close")
         b.wait_not_present(".modal-dialog div .alert")
 
+        # delete image busybox that hasn't been used
```
KKoukiou: You should really consider splitting the testBasic into multiple test functions. testDelete, testAdd etc. It will be easier to maintain! Not a must for this PR, just let's not grow this function more in the future.

src/Images.jsx
QiWang19: if the image has been deleted is used by a running container, that container will be deleted at the same time, so need to update the containers list.

KKoukiou: For this case you have to do force deletion and you have different handlers for that.
There is not posibility that some container will be using the image to be deleted and the deletion of the image will succeed without force option, so that you will get to these handlers here.

src/Images.jsx
```diff
@@ -98,10 +96,8 @@ class Images extends React.Component {
                     this.setState({
                         setImageRemoveErrorModal: false
                     });
-                    const idDel = reply.image ? reply.image : "";
-                    const oldImages = this.props.images;
-                    let newImages = oldImages.filter(elm => elm.Id !== idDel);
-                    this.props.updateImages(newImages);
+                    this.props.updateImagesAfterEvent();
+                    this.props.updateContainersAfterEvent();
```
KKoukiou: Here the updateContainersAfterEvent is needed, but in the previous case no.

QiWang19: yes, you are right.

test/check-application
QiWang19: :+1:

QiWang19: Is it iok if I make it in different test files?

KKoukiou: If you end up creating different file for each feature you 'll end up with dozens of files. I think keeping them in one file under one class but in different function is the way to go for now. But it's not a must for this PR. Feel free to change that in future PRs. :)

QiWang19:
@KKoukiou For the varlink call failure of image deletion, I followed the dockerUI at that time. But I just noticed that docker UI changed the way to deal this. Maybe I can change in podman to stay the same as docker UI in later PRs.

Thank you for giving me the reference for the modals.

PR #43, opened 2019-02-03
Cleanup

larskarlitski:
See individual commits.


larskarlitski:
Added a commit to pin `terser` - same as @martinpitt did for cockpit. Thanks!

PR #44, opened 2019-02-03
Varlink improvements

larskarlitski:
Move varlink-related functions into their own file and improve the API a bit, so that it's possible to reuse the same connection in the future.

Also some varlink calls by using the information we get from `ListContainers` and `ListImages` instead of calling `Inspect*()`.


martinpitt:
Seems podman is victim to the same terser regression that cockpit is -- please see https://github.com/cockpit-project/cockpit/pull/11105 for a quick workaround.

larskarlitski:
> Seems podman is victim to the same terser regression that cockpit is -- please see [cockpit-project/cockpit#11105](https://github.com/cockpit-project/cockpit/pull/11105) for a quick workaround.

I've added a commit on https://github.com/cockpit-project/cockpit-podman/pull/43. Let's see if it works and then I'll rebase this one.

src/varlink.js
```diff
@@ -0,0 +1,97 @@
...
+const decoder = cockpit.utf8_decoder();
+
+/*
+ * Connect to a varlink service.
+ *
+ * Returns a connection object:
+ *
+ * - connection.call(method, parameters) calls "method" (prefixed by the
+ *   interface) with "parameters". Multiple calls can be made without waiting
+ *   for replies, but calls are made sequentially (similar to http). Returns a
```
martinpitt: https://varlink.org/ "Protocol" says the opposite, or it is misleading. At least when I tried to do multiple parallel varlink calls to podman,  things went horribly wrong.

src/varlink.js
```diff
@@ -0,0 +1,97 @@
...
+            connection.onclosed(options.problem);
+    });
+
+    connection.call = function (method, parameters) {
+        parameters = parameters || {};
+
+        const data = encoder.encode(JSON.stringify({ method, parameters }));
+
+        channel.send(data);
+        channel.send([0]);
```
martinpitt: Since this now knows about `pending`, this could throw  an exception if you are trying to make a call with the previous call's answer still outstanding? This is not a regression, and a follow-up for this is fine, just in case you feel like it while you are cleaning this up.

src/util.js
```diff
@@ -49,47 +49,30 @@ function handleVarlinkCallError(ex) {
 }
 
 export function updateContainers() {
-    let newContainers = {};
-    let newContainersStats = {};
-    return new Promise((resolve, reject) => {
-        varlink.call(PODMAN_ADDRESS, "io.podman.ListContainers")
-                .then(reply => {
-                    let newContainersMeta = reply.containers || [];
-                    let inspectRet = newContainersMeta.map(container => varlink.call(PODMAN_ADDRESS, "io.podman.InspectContainer", {name: container.id}));
```
martinpitt: My impression was that this wanted to show much more information in the future. But we can bring it back if/when needed indeed, and maybe we'll only need it on a details page, not for all containers.

src/varlink.js
larskarlitski: Requests can be pipelined, but they will still responded to in order. I just verified again that podman does this correctly.

Does changing the wording on varlink.org to this help clarify: https://github.com/varlink/varlink.github.io/pull/13 ?

src/varlink.js
larskarlitski: See above. You can queue as many calls as you'd like, without an error. It's just that they'll be handled in sequence, not in parallel.

src/varlink.js
martinpitt: A bit. TBH I'm surprised that this is even a legal mode of operation, as it makes state keeping dramatically harder on both sides, and it's not well defined any more when multiple responses to a call come into play. But, this isn't the place to discuss the protocol. :-)

martinpitt:
@larskarlitski : Tests fail. Does this need #45?

larskarlitski:
> @larskarlitski : Tests fail. Does this need #45?

No, these should just be warnings.

I missed that it's `container.names` and not `container.name`. Don't know why the plural.

martinpitt:
@larskarlitski : Your latest force-push just changed a comment, thus it didn't fix the test failure. Was that deliberate?

larskarlitski:
> @larskarlitski : Your latest force-push just changed a comment, thus it didn't fix the test failure. Was that deliberate?

The one before changed the test failure, but there's another one. I'll investigate in a bit.

PR #48, opened 2019-02-05
Fix eslint warnings

martinpitt:
See individual commits.


martinpitt:
I rebased this on top of #47, just to make sure that didn't change anything that the new eslint objects to. Plus adding a Closes #

PR #50, opened 2019-02-07
Implement run image modal dialog

KKoukiou:


KKoukiou:
run-image-busybox

src/ImageRunModal.jsx
```diff
@@ -0,0 +1,177 @@
...
+                           value={dialogValues.memory}
+                           step={1}
+                           min={0}
+                           disabled={!this.state.memoryConfigure}
+                           onChange={e => this.onValueChanged('memory', e.target.value)} />
+                    <Select.Select id='memory-unit-select'
+                                   initial={this.state.memoryUnit}
+                                   enabled={this.state.memoryConfigure}
+                                   onChange={value => this.onValueChanged('memoryUnit', value)}>
+                        <Select.SelectEntry data={units.MiB.name} key={units.MiB.name}>
```
martinpitt: Any reason why you left out KiB? There's a `unit` for it?

src/ImageRunModal.jsx
```diff
@@ -0,0 +1,177 @@
...
+                        </Select.SelectEntry>
+                    </Select.Select>
+                </div>
+
+                <label className='control-label'> {_("With terminal")} </label>
+                <label className="checkbox-inline">
+                    <input id="run-image-dialog-tty" type="checkbox"
+                           checked={this.state.hasTTY}
+                           onChange={e => this.onValueChanged('hasTTY', e.target.checked)} />
+                    {_("Allocate a pseudo-TTY")}
```
martinpitt: I can't say I like this label much. It's very technical and not interesting (implementation detail, not an user-visible feature) and also redundant with the "With terminal" label.

This is not a veto though, we can change this later.

src/Images.jsx
```diff
@@ -161,26 +153,22 @@ class Images extends React.Component {
...
+        return (
+            <React.Fragment key={image.id}>
+                <Listing.ListingRow
+                        rowId={image.id}
+                        columns={columns}
+                        tabRenderers={tabs}
+                        navigateToItem={this.navigateToImage(image)}
+                        listingActions={actions} />
+            </React.Fragment>
+        );
```
martinpitt: What does this do? It just wraps a single element, as before?

test/check-application
```diff
@@ -253,6 +253,49 @@ class TestApplication(testlib.MachineCase):
...
+    def testRunImage(self):
+        b = self.browser
+        m = self.machine
+
+        self.login_and_go("/podman")
+        b.wait_present("#containers-images")
+        b.wait_in_text("#containers-images", "busybox:latest")
+
+        # Open run image dialog
+        b.wait_present('#containers-images tr:contains("busybox:latest")')
```
martinpitt: This should wait on the button to appear, just in case.

test/check-application
```diff
@@ -253,6 +253,49 @@ class TestApplication(testlib.MachineCase):
...
+        b.click('#containers-images tbody tr:contains("busybox:latest") td.listing-ct-actions button')
+        b.wait_visible('div.modal-dialog div.modal-header h4.modal-title:contains("Run Image")')
```
martinpitt: Is that dialog already present, but hidden? The react-y way would be for it to be not rendered at all when closed. Then this should be `wait_present`.

test/check-application
```diff
@@ -253,6 +253,49 @@ class TestApplication(testlib.MachineCase):
...
+
+        # Inspect and fill modal dialog
+        b.wait_present('div.modal-body label:contains("Image")')
```
martinpitt: No strong objection, but this just validates more or less static HTML and duplicates strings. I don't think it's important or helpful to have these label checks.

test/check-application
```diff
@@ -253,6 +253,49 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_in_text("#run-image-dialog-image", "busybox:latest")
+
+        b.wait_present('div.modal-body label:contains("Name")')
+        b.set_input_text("#run-image-dialog-name", "busybox-with-tty")
```
martinpitt: Could this check for a non-empty default value? To make sure that the docker name generation stuff works.

lib/form-layout.less
```diff
@@ -162,8 +162,8 @@
     display: grid;
     grid-gap: @padding-small-vertical;
     justify-content: start;
-    // Only support 2 splits for now (can change to 3 later, if needed)
-    grid-template-columns: repeat(2, auto);
+    // Only support 3 splits for now (can change to 3 later, if needed)
```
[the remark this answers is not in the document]
KKoukiou: Done

src/ImageRunModal.jsx
KKoukiou: Docker supports only MB. I decided to add GB since (at least I) more usually configure systems memory with GB. But enabling the KB select entry is very easy as well. I can add it.

src/ImageRunModal.jsx
KKoukiou: Removed.

src/ImageRunModal.jsx
KKoukiou: Added

src/Images.jsx
KKoukiou: My fault. Removed and added the key property to the containing element.

test/check-application
KKoukiou: Done

test/check-application
KKoukiou: Good point. Removing label checks.

test/check-application
KKoukiou: It should be hidden.

test/check-application
KKoukiou: Sure.

PR #51, opened 2019-02-08
Implement image search and download modal dialog

KKoukiou:


KKoukiou:
**This PR does not have tests.**

Adding tests for this feature is complicated for the time being.

I deployed a local private registry as described here, https://docs.docker.com/registry/deploying/ but this is by default **not** using TLS, thus it is not secure.
Unfortunately the SearchImage API allows only secure connections.
I already opened an issue on libpod, https://github.com/containers/libpod/issues/2293
which means we are not able to use http registries through cockpit podman.

Let me know if you think it's ok to merge this feature without tests till we get the tls option in the API, or enable networking in the test VM, or  somehow create a local offline registry with https support.

search-image-waiting

search-image

pulling-image

KKoukiou:
So, I found out that listing a registry as insecure in the conf file will bypass the tls verify test. Reposting with test suite shortly :)

KKoukiou:
Search image with no results
search-image-no-results

KKoukiou:
Image search failure dialog error
failure-search-api


Image download failure pop up notification
image-download-fail

KKoukiou:
The above two errors where triggered by changing the api method name, not real case error., but should work for verifying that UI works.

src/ImageSearchModal.jsx
```diff
@@ -0,0 +1,134 @@
...
+    onValueChanged(key, value) {
+        this.setState({ [key]: value });
+
+        if (key == 'imageIdentifier') {
+            this.setState({ searchInProgress: true });
+            // Clears the previously set timer.
+            clearTimeout(this.typingTimeout);
+
+            // Reset the timer, to make the http call after 1000MS
+            this.typingTimeout = setTimeout(this.onSearchTriggered, 1000);
```
larskarlitski: Not sure what we usually do here, but 1s seems too much. Maybe 500ms? (Unless we decide for an explicit `Return`)/

@garrett Is there consensus on a good timeout for being sure that text entry has stopped? We're triggering a search here.

src/ImageSearchModal.jsx
```diff
@@ -0,0 +1,134 @@
...
+                        dialogErrorDetail: cockpit.format("$0: $1", ex.error, ex.parameters && ex.parameters.reason)
+                    });
+                });
+    }
+
+    onValueChanged(key, value) {
+        this.setState({ [key]: value });
+
+        if (key == 'imageIdentifier') {
+            this.setState({ searchInProgress: true });
```
larskarlitski: This makes the previous result list disappear as soon as I change the string, which can be quite jarring. The docker page requires an explicit `Return` key stroke.

src/ImageSearchModal.css
```diff
@@ -0,0 +1,13 @@
+.list-group{
+    max-height: 35rem;
+    overflow-y:scroll;
```
larskarlitski: I'm going to be *that* person: there's a space missing here.

src/ImageSearchModal.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+        varlink.call(utils.PODMAN_ADDRESS, "io.podman.SearchImage", { name: this.state.imageIdentifier })
+                .then(reply => this.setState({ imageList: reply.images || [], searchInProgress: false }))
+                .catch(ex => {
+                    this.setState({
+                        searchInProgress: false,
+                        dialogError: _("Failed to search for new images"),
+                        dialogErrorDetail: cockpit.format("$0: $1", ex.error, ex.parameters && ex.parameters.reason)
+                    });
+                });
```
larskarlitski: Unfortunately, this is where our simple handling of varlink calls fails. Cancelling the dialog doesn't cancel this call, which leads to
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in ImageSearchModal (created by Images)
```

We have this problem in other components as well, but they are on the main page and don't get unmounted during normal operation.

We could
* Check for `this._isMounted` in the callbacks, which is annyoing
* Keep track of a varlink connection and `.close()` it on unmount. This would trigger all `.catch` handlers though. Not sure if we want that?
* Use something like redux to get all of the state handling out of components. Phew.

src/ImageSearchModal.jsx
martinpitt: Perhaps not consensus, but precedent. The [realmd join dialog](https://github.com/cockpit-project/cockpit/blob/master/pkg/realmd/operation.js#L69) uses 1s, like this PR. The [docker UI search dialog](https://github.com/cockpit-project/cockpit/blob/master/pkg/docker/search.js#L78) uses 2 s even.

larskarlitski: Ah, indeed. Thanks. I guess doing the same here is fine, then (if hitting `Return` works as well).

garrett: 1. Are we querying a remote server?
   - I assume the Docker hub? If so, I looked for stats.
2. How long does a request from the server take to complete a round trip (contact, process, results)?
   - Doing an autocomplete search on hub.docker.com for "fedora" took 253 ms & 251 ms, then 128 & 130 ms (it seems variable between 130 - 250 ms). There are two requests at the same time, which probably affects timing. One search filters by "store,official" and the other by "community". Both of the autocomplete searches limit by 4.
   - When performing a full search, it took 145 ms, which included 25 results with metadata.
   - Autocomplete and a full search both use the same API, but with different limits (`image_filter` and `page_size` for limiting auto-complete and `type=image` with `page_size` as 25 for the full search).
   - hub.docker.com is hosted on us-east-1 on AWS, so it is faster in North America; the stats above are from a (not fantastic) residential cable modem line in Europe, so the timing is probably more on the conservative side
3. Is the server rate limited?
   - Seems like it is not: https://stackoverflow.com/questions/46424132/docker-pull-request-limits-for-docker-hub-registry

We will want to limit to > 2 characters (unless enter is pressed), debounce, throttle, strip trailing spaces, [XMLHttpRequest.abort()](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort) unneeded pending requests, and cache (based on stripped query string) to improve the user experience.

_Interestingly enough_, 250ms is the rough median reaction time of humans (oft-cited in UX literature and [backed up by a game](https://www.humanbenchmark.com/tests/reactiontime/statistics) when you subtract the 30 ms screen delay) and usually the upper bound to use for making interfaces feel responsive. As the response is typically 130 but can get up to 250 (at least from my home network) with multiple calls at the same time, we probably want to make 1 call if we can, assume it will return within 150ms, and pad to 250ms (that is 100ms of delay) if we want it to feel _instant_. Search results seem to weight official images at the top and have metadata included about official-ness, so we can probably just do 1 call.

_However_, with a spinner, we can bump up the wait time a little. Still, if it is >= 1 second, people will think it's slow. A medium ground could be to aim for 400ms, perhaps. We should implement something, check metrics and how it feels, and adjust accordingly.

As average typing time is roughly 40 words per minute, 40/60 == ⅔ sec. And people will be thinking as they type, so they'll be a little slower. That's how I wound up with 400 ms as a goal, as a compromise between typing speed and responsiveness.

---

This all said: 1 second is definitely too long, however. 2 is glacial. (The Docker page always felt really slow when searching.) So let's try for 400 ms total time (that is: 250 ms debounce + 150 ms query time) on a first pass, with reasoning above and adjust as needed (if there are "too many" queries or if it feels slow).

It's probably better to start with less UI latency and adjust to add more than the other way around.

src/ImageSearchModal.jsx
garrett: Keystrokes that trigger autocomplete searches should not include spaces... and we should show the results _possibly greyed out_ with a progress indicator (spinner) to show that new results are being fetched, instead of immediately clearing everything away.

src/ImageSearchModal.jsx
garrett: BTW: Local registries should have a much smaller delay than Docker hub.

garrett:
> Try to search non https registry which is not listed in insecure registries section.
> search-error

The error message here is not clear. Since we know what the problem is (insecure repo without the allow-insecure flag turned on), we should state the error message in plain sounding text.

Here are suggestions for handling insecure registries:

1. Attempt to automatically add the registry in a secure manner, if possible.
2. Show an error when adding the registry, with solutions either to not add the registery or allow an override.
3. Provide a way to override the error (change the config to allow it being an insecure repo).
4. When attempting to search an insecure registry, show a human friendly string with a way to fix it (remove the registry or mark it as insecure and then try the search again).
5. In search results, mark any containers coming from an insecure registry. 
6. Also mark imported containers that come from an insecure registry in the list too.

Or, more simply: we could not handle insecure registries and have friendly error messages that allow a person to remove them. (At least for a first pass, and then possibly add them later.)

src/ImageSearchModal.jsx
KKoukiou: I honestly don't know how/if it's possible to use  [XMLHttpRequest.abort()](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort) to abort unneeded pending requests.
So, to sum up changes for now,
I'll limit the search for >2 chars expect if Enter is hit.
I will strip the input string.
I 'll set the trigger wait time to 400ms instead of 1000ms. But if aborting the ongoing searches is not possible, this will deteriorate performance.
Regarding the caching I also am not sure what you mean. If you mean just store it locally in some dict, I honestly doubt this is a good idea but I 'll leave it up to @martinpitt  and @larskarlitski

src/ImageSearchModal.jsx
KKoukiou: I wouldn't go with redux just because of this PR, it's bug change.
I would do the this._isMounted trick. I already do it anyway.
But good catch thanks.

KKoukiou:
> > Try to search non https registry which is not listed in insecure registries section.
> > search-error
> 
> The error message here is not clear. Since we know what the problem is (insecure repo without the allow-insecure flag turned on), we should state the error message in plain sounding text.
> 
> Here are suggestions for handling insecure registries:
> 
>     1. Attempt to automatically add the registry in a secure manner, if possible.

A registry either it's configured to serve over https or http. We can't just change that.

> 
>     2. Show an error when adding the registry, with solutions either to not add the registery or allow an override.

We don't have in  cockpit-podman and afaik on cockpit-docker support for adding/removing/editing registries.

> 
>     3. Provide a way to override the error (change the config to allow it being an insecure repo).

I like this idea but I would like to not do this in this PR, since it's rather a new small feature by itself.
But it's a nice idea, to try adding the registry in the  insecure section and retry re-searching/

> 
>     4. When attempting to search an insecure registry, show a human friendly string with a way to fix it (remove the registry or mark it as insecure and then try the search again).

Overwritting this error message is of course feasible but as in cockpit-machines similar discussions, I am against overwritting errors from the backend. Keeping the error message intact can alllow user to search it over his search engine and find solutions that other podman users posted, since the issue is generic and not cockpit specific.
And for this specific error, it's quite clear to me. Client expect HTTPS but registry send over HTTP.
But I am open to duscussion.

> 
>     5. In search results, mark any containers coming from an insecure registry.

The reply from the SearchImage API does not contain these information. Maybe @larskarlitski knows if we can expose this attribute?

> 
>     6. Also mark imported containers that come from an insecure registry in the list too.

Same as above, currently I don't know to to get this information.

> 
> 
> Or, more simply: we could not handle insecure registries and have friendly error messages that allow a person to remove them. (At least for a first pass, and then possibly add them later.)

Removing/adding/editing registries is not supported in the UI, yet. And also neither in docker UI afaik. So explicitly not handling registries that are listed in the configuration file I think is a no-go for now.

src/ImageSearchModal.jsx
KKoukiou: I fixed Lars's comment but I don't really know how to show results greyed out and in front of them a progress indicator. I 'll be glad to implement it however if you can provide some hints.

KKoukiou:
Reposted with fixes for:
- timeout for search to be triggered
- not triggering on backspace
- unmounted components error fixed

Rest should probably be included in the follow up as discussed over IRC.

garrett:
>> 1. Attempt to automatically add the registry in a secure manner, if possible.
> A registry either it's configured to serve over https or http. We can't just change that.

I meant seeing if HTTPS works instead of HTTP, and redirect, like a browser.

But yeah, everything should be later in different PRs.

KKoukiou:
The tests are not triggering the onKeyDown handler when pressing keys with key_press whilst real life input triggers it :/ I need to debug it

KKoukiou:
@martinpitt @larskarlitski this is ready for review again. The rest of enhancements that garrett suggested for registries will come in future PRs as agreed.

martinpitt:
Great work, thank you! Some notes from testing:

 - It now seems to immediately start searching when typing the first letter, and doing a new search on any new keystroke without any delay. IMHO this is too aggressive and abusive to the registries.
 - The image search results should show the registry that contains the image. I. e. it's interesting and often rather important to know whether an image comes from docker.io, fedoraproject.org, or a local one.
 - There's no way to see or specify tags. It seems it will only ever download `:latest`?  (This is material for a follow-up PR, though!)

test/check-application
```diff
@@ -158,6 +165,60 @@ class TestApplication(testlib.MachineCase):
...
 
+    def testDownloadImage(self):
+        b = self.browser
+        m = self.machine
+
+        # Create and start registry container
+        m.execute("podman run -d -p 5000:5000 --name registry registry:2")
+
+        # Add local insecure registry into resgitries conf
+        m.execute("echo \"{0}\" > /etc/containers/registries.conf && systemctl restart io.podman.socket".format(REGISTRIES_CONF))
```
martinpitt: Restarting the socket is rather pointless. This should do `try-restart io.podman.service`, or just `stop` as it will just be started again on demand.

KKoukiou:
@martinpitt It indeed searching from the first letter. Garrett suggested to start from 3rd, I can fix it fast. But the the behaviour didn't change from before, just the delay between the keystroke to time when the search will start. This was before 1sec, now it's 400msec. Also whenever a start a new search I close the connection to the previous search, to we should not suffer from the overhead of many connections being open simultaneously.

The search image results do not contain the registry an image is coming from for now. I think this should require a change in the API itself.
This is for example an image object from the SearchImage reply.
```
​​
0: {…}
​​​​
description: "Official Gentoo stage3-amd64 image."
​​​​
is_automated: true
​​​​
is_official: false
​​​​
name: "gentoo/stage3-amd64"
​​​​
star_count: 74
```

And regarding the tag, I plan to do in in a follow up once this lands! Good point!

KKoukiou:
@martinpitt Maybe you felt that search starts immediately because I immediately show the spinner. But I then wait for timeout to expire to start the search. I should probably delay showing the spinner as well.

martinpitt:
> Maybe you felt that search starts immediately because I immediately show the spinner

Right, that would be it. Sorry for the confusion!

garrett:
400 ms was supposed to be the target time, not a delay time. The delay was supposed to be 250 ms (based on Docker Hub returning at 150 ms). I've since talked with Lars about this and realized we're using Podman's API, which is much, much slower.

Using podman at the command line to search for Fedora takes 9 seconds here. It's querying docker.io,  fedoraproject.org, and quay.io. I tried to see if limiting it (with `--limit`) would make it go faster. It doesn't seem to, sadly. Filtering might help, but it only seems to have stars, is-automated, and is-official.

As it's so slow, I'm guessing that it searches all the sources once after another — and that the developer had a fast connection to all the sites. (Probably in the USA, where query times is probably a lot faster.) It would be better if we could search each repo separate, but at the same time, and merge the results ourselves.

If limiting would've worked to speed it up, I would have suggested doing a quick threshold search for inital results and loading the rest in the background, to speed up result caching. Sadly, it doesn't seem to impact performance whatsoever. This leads me to believe it's just a slow round-trip for one or more of the sources. :unamused:

Also, I noticed the dialog doesn't focus the search entry automatically, so one has to focus the field manually.

## Suggestions for this PR

1. Focus the search field automatically when the dialog pops up
2. Reduce the delay time to 250ms, as suggested above

## Suggestions for follow-up PRs

1. Improve performance (somehow)
2. Fix error messages
3. Fix insecure repositories & containers (as I mentioned in a previous comment)

KKoukiou:
@garrett fixed the autoFocus and delay time for triggering the search.

garrett:
Aha! I figured out Podman timing issues and came up with a way for us to parallelize it. It works from the command line at least, and hopefully will work with the API too.

I've opened up issue #55 with the details.

garrett:
We should have an empty state for the search (including initial empty state). I've added #56 to discuss this.

src/ImageSearchModal.jsx
```diff
@@ -0,0 +1,169 @@
...
+            this.setState({ [key]: value });
+    }
+
+    onKeyPress(e) {
+        if (e.key != ' ') { // Space should not trigger search
+            // Clears the previously set timer.
+            clearTimeout(this.typingTimeout);
+
+            // Reset the timer, to make the http call after 250MS
+            this.typingTimeout = setTimeout(this.onSearchTriggered(e.key == 'Enter'), 250);
```
larskarlitski: This is calling `onSearchTriggered` immediately instead of after 250ms. That's why @martinpitt and @garrett were seeing searches after the first key press.

Also, it leads to a CSP error:  `setTimeout` tries to call the return value of `onSearchTriggered` and since that's not a function, calls `eval()` on it. Yay JavaScript.

This patch below makes it work. I have the extra variable `forceSearch` because we can't access `event` after the handler has run. React reuses the event object and complains if we do.

```diff
     onKeyPress(e) {
         if (e.key != ' ') { // Space should not trigger search
+            let forceSearch = e.key == 'Enter';
+
             // Clears the previously set timer.
             clearTimeout(this.typingTimeout);
 
             // Reset the timer, to make the http call after 250MS
-            this.typingTimeout = setTimeout(this.onSearchTriggered(e.key == 'Enter'), 250);
+            this.typingTimeout = setTimeout(() => this.onSearchTriggered(forceSearch), 250);
```

KKoukiou: Ups, thanks for noticing it @larskarlitski!

martinpitt:
The [test fails](https://fedorapeople.org/groups/cockpit/logs/pull-51-20190215-090202-0ee36de1-cockpit-fedora-29/log) due to what looks like a race condition: It expects a spinner, but the [screenshot](https://fedorapeople.org/groups/cockpit/logs/pull-51-20190215-090202-0ee36de1-cockpit-fedora-29/TestApplication-testDownloadImage-fedora-29-127.0.0.2-2501-FAIL.png) shows that the search result is already present. Probably searching the local registry is so fast that it's hard to catch the spinner?

KKoukiou:
@martinpitt right, and also before I showed the spinner before actually starting the search. So 250ms before last type and spinner is too little for test to catch it. Removing the wait conditions for spinners.

KKoukiou:
@martinpitt you are ok to merge this without the command to wait for the spinner? The search query finished so fast for local registries that it does not have time to show at all.

martinpitt:
@KKoukiou : Yes, that's fine of course.

PR #52, opened 2019-02-11
Enhance html in container commit dialog

KKoukiou:


KKoukiou:
New container commit dialog

container-commit-dialog-new

Old container commit dialog

commit-dialog

PR #54, opened 2019-02-12
varlink: Return a promise from varlink.connect()

larskarlitski:
Return a promise from `varlink.connect()` that resolves when the channel
signals that it is ready. This is not technically needed, because
cockpit allows sending messages to channels before they're ready. It is
a bit cleaner though, because clients can be sure that connections have
been established before starting to send messages.

This was already expected in `varlink.call()`, which is also the only
consumer of this call.

### DIFF
--- a/src/varlink.js
+++ b/src/varlink.js
@@ -4,6 +4,15 @@ import cockpit from "cockpit";
 const encoder = cockpit.utf8_encoder();
 const decoder = cockpit.utf8_decoder();
 
+class VarlinkError extends Error {
+    constructor(name, parameters) {
+        super(JSON.stringify(parameters));
+
+        this.name = name;
+        this.parameters = parameters;
+    }
+}
+
 /*
  * Connect to a varlink service.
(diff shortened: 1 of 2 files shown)


src/varlink.js
```diff
@@ -80,7 +80,9 @@ function connect(address) {
         channel.close();
     };
 
-    return connection;
+    return new Promise((resolve, reject) => {
+        channel.addEventListener("ready", () => resolve(connection));
```
martinpitt: In order for a client to actually detect if the connection failed, should the `close` signal be wired to `reject()`? If one gets a "close" before a "ready", then there's surely something wrong, and the `ready` part could then just remove the event listener for `close`.

martinpitt: Also, doesn't this need to remove its own addEventListener(), otherwise we'd forever "leak" this listener?

larskarlitski: Indeed. Fixed.

PR #57, opened 2019-02-15
Add publishing ports option to the run image dialog

Gundersanne:
TODO:
- [x] actually show ports in containerdetails too

### DIFF
--- a/src/ContainerDetails.jsx
+++ b/src/ContainerDetails.jsx
@@ -12,6 +12,18 @@ const render_container_state = (container) => {
     return cockpit.format(_("Exited"));
 };
 
+const render_container_published_ports = ({ ports }) => {
+    let result = [];
+    if (!ports)
+        return result;
+    for (let i = 0; i < ports.length; ++i)
+        result.push(
+            <React.Fragment key={ ports[i].protocol + ports[i].host_port + ports[i].container_port }>
+                { ports[i].host_ip || '0.0.0.0' }:{ ports[i].host_port } &rarr; { ports[i].container_port }/{ ports[i].protocol }{ i < ports.length - 1 && ', ' }
+            </React.Fragment>);
+    return result;
+};
+
 const ContainerDetails = ({ container }) => (
     <div className='listing-ct-body'>
         <dl>
@@ -25,6 +37,8 @@ const ContainerDetails = ({ container }) => (
             <dd>{container.command ? container.command.join(" ") : ""}</dd>
             <dt>{_("State")}</dt>
             <dd>{render_container_state(container)}</dd>
+            <dt>{_("Ports")}</dt>
+            <dd>{render_container_published_ports(container)}</dd>
         </dl>
     </div>
 );
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -46,20 +46,20 @@ export class ImageRunModal extends React.Component {
     }
 
     getCreateConfig() {
-        let props = {};
+        let createConfig = {};
 
-        props.image = this.state.image.repoTags ? this.state.image.repoTags[0] : "";
-        props.resources = {};
+        createConfig.image = this.state.image.repoTags ? this.state.image.repoTags[0] : "";
+        createConfig.resources = {};
         if (this.state.containerName)
-            props.name = this.state.containerName;
+            createConfig.name = this.state.containerName;
         if (this.state.command)
-            props.command = [this.state.command];
+            createConfig.command = [this.state.command];
         if (this.state.memoryConfigure && this.state.memory)
-            props.resources.memory = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
+            createConfig.resources.memory = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
         if (this.state.hasTTY)
-            props.tty = true;
+            createConfig.tty = true;
 
-        return props;
+        return createConfig;
     }
 
     onRunClicked() {
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { Button, Modal } from 'patternfly-react';
 import * as dockerNames from 'docker-names';
 
@@ -27,6 +28,101 @@ const units = {
     },
 };
 
+class PublishPortsForm extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            publish: [{ key: 0, containerPort: null, hostPort: null, protocol: 'TCP' }],
+        };
+        this.keyCounter = 1;
+        this.removePort = this.removePort.bind(this);
+        this.addPort = this.addPort.bind(this);
+        this.onPortChange = this.onPortChange.bind(this);
+    }
+
+    removePort(idx, field, value) {
+        this.setState(state => {
+            let ports = state.publish.concat();
+            ports.splice(idx, 1);
+            if (ports.length === 0)
+                ports.push({ key: this.keyCounter++, containerPort: null, hostPort: null, protocol: 'TCP' });
+            return { publish: ports };
+        }, () => this.props.onChange(this.state.publish.concat()));
+    }
+
+    addPort() {
+        this.setState(state => {
+            return { publish: [...state.publish, { key: this.keyCounter++, containerPort: null, hostPort: null, protocol: 'TCP' }] };
+        }, () => this.props.onChange(this.state.publish.concat()));
+    }
+
+    onPortChange(idx, field, value) {
+        this.setState(state => {
+            let ports = state.publish.concat();
+            ports[idx][field] = value || null;
+            return { publish: ports };
+        }, () => this.props.onChange(this.state.publish.concat()));
+    }
+
+    render () {
+        const { id } = this.props;
+        const dialogValues = this.state;
+        return (
+            <React.Fragment>
+                {
+                    dialogValues.publish.map((port, idx) =>
+                        (
+                            <div className='publish-port-form ct-form-layout' key={ port.key }>
+                                <div role='group' className='ct-form-layout-split'>
+                                    <input className='form-control'
+                                           id={ (idx === 0 && id) || undefined }
+                                           type='number'
+                                           step={1}
+                                           min={1}
+                                           max={65535}
+                                           placeholder={_("Container port")}
+                                           value={port.containerPort || ''}
+                                           onChange={e => this.onPortChange(idx, 'containerPort', e.target.value)} />
+                                    <Select.Select className='form-control'
+                                                   initial={port.protocol}
+                                                   onChange={value => this.onPortChange(idx, 'protocol', value)} >
+                                        <Select.SelectEntry data='TCP' key='TCP'>
+                                            {_("TCP")}
+                                        </Select.SelectEntry>
+                                        <Select.SelectEntry data='UDP' key='UDP'>
+                                            {_("UDP")}
+                                        </Select.SelectEntry>
+                                    </Select.Select>
+                                    <input className='form-control'
+                                           type='number'
+                                           step={1}
+                                           min={1}
+                                           max={65535}
+                                           placeholder={_("Host port")}
+                                           value={port.hostPort || ''}
+                                           onChange={e => this.onPortChange(idx, 'hostPort', e.target.value)} />
+                                </div>
+                                <div role='group' className='ct-form-layout-split'>
+                                    <Button bsStyle='default' className='pficon-close'
+                                            disabled={dialogValues.publish.length === 1 &&
+                                                      (!dialogValues.publish[0]['containerPort'] ||
+                                                       !dialogValues.publish[0]['hostPort'])}
+                                            onClick={() => this.removePort(idx)} />
+                                    <Button bsStyle='default' className='fa fa-plus' onClick={this.addPort} />
+                                </div>
+                            </div>
+                        )
+                    )
+                }
+            </React.Fragment>
+        );
+    }
+}
+PublishPortsForm.propTypes = {
+    onChange: PropTypes.func.isRequired,
+    id: PropTypes.string.isRequired,
+};
+
 export class ImageRunModal extends React.Component {
     constructor(props) {
         super(props);
@@ -34,6 +130,7 @@ export class ImageRunModal extends React.Component {
             command: 'sh',
             containerName: dockerNames.getRandomName(),
             hasTTY: true,
+            publish: [],
             image: props.image,
             memory: 512,
             memoryConfigure: false,
@@ -58,6 +155,10 @@ export class ImageRunModal extends React.Component {
             createConfig.resources.memory = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
         if (this.state.hasTTY)
             createConfig.tty = true;
+        if (this.state.publish.length > 0)
+            createConfig.publish = this.state.publish
+                    .filter(port => port.hostPort && port.containerPort)
+                    .map(port => port.hostPort + ':' + port.containerPort + '/' + port.protocol);
 
         return createConfig;
     }
@@ -152,6 +253,9 @@ export class ImageRunModal extends React.Component {
                            checked={this.state.hasTTY}
                            onChange={e => this.onValueChanged('hasTTY', e.target.checked)} />
                 </label>
+
+                <label className='control-label' htmlFor='run-image-dialog-publish'>{ _("Ports") }</label>
+                <PublishPortsForm id='run-image-dialog-publish' onChange={value => this.onValueChanged('publish', value)} />
             </div>
         );
 
--- a/test/check-application
+++ b/test/check-application
@@ -284,7 +284,29 @@ class TestApplication(testlib.MachineCase):
         b.set_checked("#run-image-dialog-memory-limit-checkbox", False)
         b.wait_present('div.modal-body label:contains("Memory Limit") + div.form-inline > input[type=number]:disabled')
 
+        # Enable tty
         b.set_checked("#run-image-dialog-tty", True)
+
+        # Configure published ports
+        b.set_input_text('.publish-port-form input:first-child', '5000')
+        b.set_input_text('.publish-port-form input:last-child', '6000')
+        b.click('.publish-port-form .fa-plus')
+        b.wait_present('.publish-port-form + .publish-port-form')
+        b.set_input_text('.publish-port-form:last-of-type input:first-child', '5001')
+        b.set_input_text('.publish-port-form:last-of-type input:last-child', '6001')
+        b.click('.publish-port-form:last-of-type button.dropdown-toggle')
+        b.wait_visible('.publish-port-form:last-of-type .dropdown-menu a:contains("UDP")')
+        b.click('.publish-port-form:last-of-type .dropdown-menu a:contains("UDP")')
+        b.wait_visible('.publish-port-form:last-of-type button.dropdown-toggle:contains("UDP")')
+        b.click('.publish-port-form:last-of-type .fa-plus')
+        b.wait_present('.publish-port-form + .publish-port-form + .publish-port-form')
+        b.set_input_text('.publish-port-form:last-of-type input:first-child', '7001')
+        b.set_input_text('.publish-port-form:last-of-type input:last-child', '7001')
+        b.click('.publish-port-form:last-of-type .pficon-close')
+        b.click('.publish-port-form:last-of-type .fa-plus')
+        b.wait_present('.publish-port-form + .publish-port-form + .publish-port-form')
+        b.set_input_text('.publish-port-form:last-of-type input:first-child', '8001')
+        b.set_input_text('.publish-port-form:last-of-type input:last-child', '8001')
 
         b.click('div.modal-footer button:contains("Run")')
         b.wait_not_present("div.modal-dialog")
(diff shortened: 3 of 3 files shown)


Gundersanne:
image
image

src/ContainerDetails.jsx
```diff
@@ -25,6 +33,8 @@ const ContainerDetails = ({ container }) => (
             <dd>{container.command ? container.command.join(" ") : ""}</dd>
             <dt>{_("State")}</dt>
             <dd>{render_container_state(container)}</dd>
+            <dt>{_("Ports")}</dt>
+            <dd>{render_container_published_ports(container)}</dd>
```
larskarlitski: Do we want to hide those when there are no published ports?

src/ContainerDetails.jsx
```diff
@@ -12,6 +12,14 @@ const render_container_state = (container) => {
     return cockpit.format(_("Exited"));
 };
 
+const render_container_published_ports = (container) => {
```
martinpitt: Style hint: For presentational components you can write `({ prop1, prop2 }) =>` which makes the supported properties more explicit.

src/ContainerDetails.jsx
```diff
@@ -12,6 +12,14 @@ const render_container_state = (container) => {
...
+    if (container.ports)
+        return container.ports
+                .map(port => port.host_ip ? port.host_ip : '0.0.0.0' + ':' + port.host_port + '->' + port.container_port + '/' + port.protocol)
```
martinpitt: Please use `a || b` instead of `a ? a : b`.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+        };
+        this.removePort = this.removePort.bind(this);
+        this.addPort = this.addPort.bind(this);
+        this.onPortChange = this.onPortChange.bind(this);
+        this.validateAndChange = this.validateAndChange.bind(this);
+    }
+
+    removePort(idx, field, value) {
+        let ports = this.state.publish;
+        ports.splice(idx, 1);
```
martinpitt: You must not modify state in any way other than `setState()`, and this does (through aliasing). Make a copy with `this.state.publish.concat()`.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+        if (ports.length === 0) {
+            this.addPort();
```
martinpitt: This needs to be structured differently then. Maybe pass a "local state" object to `addPort()`, and if not given, use `this.state.publish` there? But it still feels a bit too magical and complicated.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+        } else {
+            this.setState({ publish: ports });
```
larskarlitski: You might loose some updates here and in `addPort()` below. It's unlikely in this case, but it is preferred to use the functional version of `setState` in the cases were you need to update a state variable based on previous values. You pass a function which receives the actual previous state in an argument.

It's pretty well explained here: https://reactjs.org/docs/react-component.html#setstate

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+            this.validateAndChange(ports);
+        }
+    }
+
+    addPort() {
+        let ports = this.state.publish;
+        ports.push({ containerPort: null, hostPort: null, protocol: 'TCP' });
```
martinpitt: Same here. Note that for adding you can use the spread operator for concisely creating a copy and modify it:

```js
this.setState({ publish: [...ports,  containerPort: null, hostPort: null, protocol: 'TCP' }] });
```

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+        this.setState({ publish: ports });
+        this.validateAndChange(ports);
+    }
+
+    validateAndChange(ports) {
+        if (typeof this.props.onChange !== 'function')
+            return;
+        let validated = ports.filter(port => parseInt(port.containerPort) >= 1 && parseInt(port.containerPort) <= 65535 &&
+                                     parseInt(port.hostPort) >= 1 && parseInt(port.hostPort) <= 65535);
+        this.props.onChange(validated.map(port => port.hostPort + ':' + port.containerPort + '/' + port.protocol));
```
martinpitt: It feels a bit strange to silently drop invalid port specs here. But we can start with that and refine error handling later.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+    }
+
+    render () {
+        const dialogValues = this.state;
+        return (
+            <div>
```
martinpitt: Please make that a `<React.Fragment>`, the `<div>` isn't needed in the final rendered HTML.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+                {
+                    dialogValues.publish.map((port, idx) => {
+                        return (
```
martinpitt: You can simplify `=> { return (...) }` to just `=> (...)`.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+                            <div className='publish-port-form ct-form-layout' key={'published-port-' + idx}>
```
martinpitt: It feels a bit abuse-ish to use `ct-form-layout` for a non-`<form>`. Was this due to your struggles with layout and putting a table/form (port list) into a table/form (the whole dialog)?

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+                                <div role='group' className='ct-form-layout-split'>
+                                    <input className='form-control'
+                                           type='number'
+                                           step={1}
+                                           min={1}
+                                           max={65535}
+                                           placeholder={_("container port")}
```
martinpitt: Pretty sure these should use Title Case; @andreasn?

src/ContainerDetails.jsx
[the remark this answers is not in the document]
Gundersanne: Will do!

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+                                           value={port.containerPort || ''}
```
martinpitt: Interesting that `''` works for a number input.

src/ContainerDetails.jsx
Gundersanne: Hm well we leave fields empty if they're not filled, so I went with that. 

But we can start doing this of course.

larskarlitski: Yeah sorry I meant to ping @garrett for a design perspective.

src/ImageRunModal.jsx
```diff
@@ -58,6 +157,8 @@ export class ImageRunModal extends React.Component {
             props.resources.memory = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
         if (this.state.hasTTY)
             props.tty = true;
+        if (this.state.publish.length > 0)
+            props.publish = this.state.publish;
```
martinpitt: A React component's `props` should never be modified from inside the class. These are read-only, and meant to only be changed from outside (in which case the component re-renders). Also, why duplicate it if you already track it in state?

src/ImageRunModal.jsx
Gundersanne: Ah cool! I really should read that first.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+
+    onPortChange(idx, field, value) {
+        let ports = this.state.publish;
+        ports[idx][field] = value || null;
+        this.setState({ publish: ports });
+        this.validateAndChange(ports);
+    }
+
+    validateAndChange(ports) {
+        if (typeof this.props.onChange !== 'function')
```
KKoukiou: You can force types of props with PropTypes.
You want something like onChange: PropTypes.func

src/ImageRunModal.jsx
KKoukiou: This div looks redundant?

src/ImageRunModal.jsx
KKoukiou: We tend to capitalize the first letters of string in labels, place holders etc.

src/ImageRunModal.jsx
Gundersanne: this is rather confusing but props here is a local variable passed to the container create.

src/ImageRunModal.jsx
Gundersanne: Maybe i shold change that to null or something, because this input actually returns a string as value (or is that normal?).

src/ImageRunModal.jsx
Gundersanne: Yup, this is pretty much half a trial&error half @KKoukiou 's help. Didn't really manage to do it any other way, though there might be.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+                                           type='number'
+                                           step={1}
+                                           min={1}
+                                           max={65535}
+                                           placeholder={_("host port")}
+                                           value={port.hostPort || ''}
+                                           onChange={e => this.onPortChange(idx, 'hostPort', e.target.value)} />
+                                </div>
+                                <div role='group' className='ct-form-layout-split'>
+                                    <Button className='btn btn-default pficon-close'
```
KKoukiou: patternfly offers the bsStyle=default atribute if you want instead of the className=btn btn-default.
If you don't you it you don't need to use pf Buttons.

src/ImageRunModal.jsx
Gundersanne: It's more abut not letting anything which has an undefined field through. So I can remove the whole int bounds checking.

src/ImageRunModal.jsx
```diff
@@ -27,13 +27,112 @@ const units = {
...
+                                            disabled={dialogValues.publish.length === 1 &&
+                                                      (dialogValues.publish[0]['containerPort'] === null ||
```
KKoukiou: === null can be replaced with !dialogValue.publish[0]['containerPort'], I think it looks better.

src/ContainerDetails.jsx
Gundersanne: I'll just remove the map maybe.

src/ImageRunModal.jsx
KKoukiou: This is a nop here. You should not write to props.

KKoukiou: If you want to update the parents variable you need to pass a handler.

src/ImageRunModal.jsx
```diff
@@ -152,6 +253,9 @@ export class ImageRunModal extends React.Component {
                            checked={this.state.hasTTY}
                            onChange={e => this.onValueChanged('hasTTY', e.target.checked)} />
                 </label>
+
+                <label className='control-label'> {_("Ports")} </label>
```
KKoukiou: you 're missing htmlFor. It's a must for screen readers.

src/ImageRunModal.jsx
KKoukiou: We use ct-form-layout on non forms extensively, at least in machines. But maybe there is better solution indeed.

src/ImageRunModal.jsx
Gundersanne: Should I also avoid aliasing here? So that `getCreateConfig()` actually returns a snapshot?

src/ImageRunModal.jsx
Gundersanne: React complains when it's `null` :(

Also I think all html inputs `value` attribute are strings, regardless of type.

src/ImageRunModal.jsx
Gundersanne: Can I just let this point to the `PublishPortsForm` ? Or does this need to point to actual `input` elements?

garrett:
Overall, it's looking good.

---

> image

1. Are all the port redirects going to be from 0.0.0.0 (or some other form of localhost)?
2. Should we use `→` (`&rarr;`) instead of `->`? (Both of these arrows might be problematic in RTL languages, however.)
3. Does it make sense to collapse the ports when they're the same? (That is, just show the port being exposed and only show the arrow when the port is mapped differently.)
4. Similarly, should /tcp not be shown by default and only /udp and /tcp+udp be displayed?

Doing some combination of the above points should make the list shorter and probably easier to understand at a glance.

This could all go into a separate PR as enhancements.

---

> image

The widgets need more vertical spacing. This might be my bug, and it might change once a newer version of ct-form-layout is pulled into cockpit-podman. It might even be addressed in https://github.com/cockpit-project/cockpit/pull/11154 (which needs a little more work before landing).

Don't let this stop the PR though. :wink:

Gundersanne:
@ garret I tried to mimic the cli interface for ports as much as possible:

```
CONTAINER ID  IMAGE                            COMMAND  CREATED         STATUS             PORTS                                           NAMES
af8ed45780fd  docker.io/library/fedora:latest  sh       1 second ago    Up 1 second ago    0.0.0.0:3500->3500/tcp, 0.0.0.0:2913->3800/udp  clever_mestorf
```

Ports that are defined through this run dialog will always have 0.0.0.0 be the host ip, but on the cli you can also define different host ip's.

garrett:
@Gundersanne: Aha! That makes sense. Thanks!

src/ImageRunModal.jsx
KKoukiou: It need to really point to the element you want to select/focus on when clicking on the label. So I think you want input and not the wrapper. But try it, maybe I am wrong.

src/ImageRunModal.jsx
```diff
@@ -27,13 +28,107 @@ const units = {
...
+                                </div>
+                            </div>
+                        )
+                    )
+                }
+            </React.Fragment>
+        );
+    }
+}
+PublishPortsForm.propTypes = {
```
KKoukiou: You 're missing here id: PropTypes.string.isRequired

But no need to repost for this I guess.

src/ContainerDetails.jsx
martinpitt: I also think it makes more sense to show an empty input by default than an extra checkbox that  will show/hide them, that feels too clumsy.

src/ContainerDetails.jsx
```diff
@@ -12,6 +12,14 @@ const render_container_state = (container) => {
     return cockpit.format(_("Exited"));
 };
 
+const render_container_published_ports = ({ ports }) => {
+    if (ports)
+        return ports
+                .map(port => cockpit.format(_("$0:$1->$2/$3"), (port.host_ip || '0.0.0.0'), port.host_port, port.container_port, port.protocol))
+                .join(', ');
+    return '';
```
martinpitt: `''` is okay, but for the learning experience: The standard way to say "don't render anything" is `null`, not an empty string.

Can this please use a more tasteful arrow and some spaces? Also, I don't think this really needs to be i18n'ed. Thus

    port => cockpit.format("$0:$1 → $2/$3", ...)

src/ImageRunModal.jsx
```diff
@@ -27,13 +28,107 @@ const units = {
...
+    }
+
+    removePort(idx, field, value) {
+        this.setState(state => {
+            let ports = state.publish.concat();
+            ports.splice(idx, 1);
+            if (ports.length === 0)
+                ports.push({ containerPort: null, hostPort: null, protocol: 'TCP' });
+            return { publish: ports };
+        }, () => this.props.onChange(this.state.publish.concat()));
```
martinpitt: wow, 31337 way of updating the state :grin: Simply passing in the updated object would do as well, but we can get more familiar with the functional style.

But why does this need to copy the state in the onChange call?

src/ImageRunModal.jsx
```diff
@@ -27,13 +28,107 @@ const units = {
...
+
+    render () {
+        const { id } = this.props;
+        const dialogValues = this.state;
+        return (
+            <React.Fragment>
+                {
+                    dialogValues.publish.map((port, idx) =>
+                        (
+                            <div className='publish-port-form ct-form-layout' key={'published-port-' + idx}>
```
martinpitt: Using index as key [goes against its purpose](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318). You need to give an intrinsic property of what you render, not an external index. I suggest using the container port (possibly concatenated with the protocol, see above), that's guaranteed to be unique and also intrinsic. Also, you don't need the string prefix, the key must be unique only within the enclosing `{ }` as far as I know (@KKoukiou, is that right?)

src/ImageRunModal.jsx
```diff
@@ -27,13 +28,107 @@ const units = {
...
+                                <div role='group' className='ct-form-layout-split'>
+                                    <input className='form-control'
+                                           id={ (idx === 0 && id) || '' }
```
martinpitt: Let's please not use an empty ID, that smells odd. Dropping the fallback should use `undefined`, if React complains let's use `null`?

src/ImageRunModal.jsx
```diff
@@ -27,13 +28,107 @@ const units = {
...
+        }, () => this.props.onChange(this.state.publish.concat()));
+    }
+
+    onPortChange(idx, field, value) {
+        this.setState(state => {
+            let ports = state.publish.concat();
+            ports[idx][field] = value || null;
+            return { publish: ports };
+        }, () => this.props.onChange(this.state.publish.concat()));
+    }
```
martinpitt: Can you map a container's port X/TCP to one host port, and container port X/UDP to a *different* host port? Intuitively this shouldn't work. If not, then these functions should check this case. I take it you want to keep the array explicitly ordered, otherwise I'd suggest a containerPort → { hostPort, protocol } mapping as state. But this would lead to the UI jumping around, so let's keep the array.

src/ImageRunModal.jsx
```diff
@@ -58,6 +153,10 @@ export class ImageRunModal extends React.Component {
             createConfig.resources.memory = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
         if (this.state.hasTTY)
             createConfig.tty = true;
+        if (this.state.publish.length > 0)
+            createConfig.publish = this.state.publish
+                    .filter(port => port.hostPort && port.containerPort)
```
martinpitt: I verified that 0 is not a valid port number, otherwise this would be wrong (you'd need to check for `!== null`).

src/ImageRunModal.jsx
```diff
@@ -152,6 +251,9 @@ export class ImageRunModal extends React.Component {
                            checked={this.state.hasTTY}
                            onChange={e => this.onValueChanged('hasTTY', e.target.checked)} />
                 </label>
+
+                <label className='control-label' htmlFor='run-image-dialog-publish'> {_("Ports")} </label>
```
martinpitt: Please drop the extra space outside the { }, as they render in HTML. Instead, move it inside the { } (we usually do that).

test/check-application
```diff
@@ -294,6 +307,12 @@ class TestApplication(testlib.MachineCase):
         hasTTY = m.execute("podman inspect --format '{{.Config.Tty}}' busybox-with-tty")
         self.assertEqual(hasTTY, 'true')
 
+        b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
+        b.wait_present('#containers-containers tr:contains("0.0.0.0:6000->5000/tcp, 0.0.0.0:6001->5001/udp")')
+        ports = m.execute("podman inspect --format '{{.NetworkSettings.Ports}}' busybox-with-tty");
+        self.assertIn('6000 5000 tcp', ports)
+        self.assertIn('6001 5001 udp', ports)
```
martinpitt: Can you please also cover removal of port redirections? All the way up to having none, so that we keep covering that case as well?

src/ImageRunModal.jsx
Gundersanne: right now (almost) all the validation is actually done by podman itself. Do we want to start doing extensive validation ? The `.filter` based on ports is just to make sure we don't send any empty rows really, `null` there would be more explicit I guess.

Gundersanne: Also it's possible to have multiple host ports mapped to a single container port, so 127.1:8080->80/tcp and 127.1:8081->80/tcp for instance.

martinpitt: Ah, very well. Does the dialog show a proper error on a port conflict? That's worth adding to the integration test, as that's the sort of thing that tends to silently break with changing versions.

src/ImageRunModal.jsx
Gundersanne: Well technically it could do without, but it propagates this change to the dialog itself, which then stores it in it's own state. And if the dialog would modifiy this part of the state, wouldn't it propagate to the state here?

src/ImageRunModal.jsx
Gundersanne: Can I use the index together with the ports? Because a user might click the  plus button several times, and then you'll have multiple rows with the same key.

Gundersanne: Also using a key based on the values of the form causes an unfocus each time you enter a single number.

src/ImageRunModal.jsx
Gundersanne: The id is only required on the first row/instance.

src/ContainerDetails.jsx
Gundersanne: I removed the checkbox, I'll upload new screenshots

Gundersanne:
https://github.com/cockpit-project/cockpit-podman/pull/57#issuecomment-464117416 new screens

src/ImageRunModal.jsx
martinpitt: Ack, so let's keep this for safety.

src/ImageRunModal.jsx
martinpitt: Right, I didn't consider multiple empty inputs. So host port + index sounds fine; it will make the non-empty entries a  bit more ambiguous, but at worst we'll get a few extra renders there.

src/ContainerDetails.jsx
```diff
@@ -12,6 +12,18 @@ const render_container_state = (container) => {
     return cockpit.format(_("Exited"));
 };
 
+const render_container_published_ports = ({ ports }) => {
+    let result = [];
+    if (!ports)
+        return result;
+    for (let i = 0; i < ports.length; ++i)
+        result.push(
+            <React.Fragment key={ ports[i].protocol + ports[i].host_port + ports[i].container_port }>
```
martinpitt: So you said you want to use the index here as well, right? I suppose this could use the intrinsic values if they are set, and the index if any of them are empty?

src/ImageRunModal.jsx
martinpitt: Right, but it should still be always given with a PublishPortsForm instantiation, no?

martinpitt:
I think I resolved all obsolete review comments now, the open ones are still current.

src/ContainerDetails.jsx
Gundersanne: In details the index is not necessary, because here the values are already set for sure, as these are running containers.

src/ImageRunModal.jsx
Gundersanne: oh woops, you're right. sorry about that :/

Gundersanne: Fixed it!

src/ContainerDetails.jsx
martinpitt: Summary from IRC: Let's keep a `this.id_counter`, increase it when creating a new ports input line, and use that as an id/key for the `{containerPort, hostPort, protocol, id}` object. This will be an invariant intrinsic property that is suitable as key, and survives reordering from deletion.

src/ImageRunModal.jsx
```diff
@@ -27,13 +28,109 @@ const units = {
     },
 };
 
+class PublishPortsForm extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            publish: [{ key: 0, containerPort: null, hostPort: null, protocol: 'TCP' }],
+            keyCounter: 1,
```
martinpitt: It seems a little odd to keep the counter inside the state, instead of a simple `this.keyCounter` object property. But you always update it together with `publish`, so it's ok if you prefer this.

PR #58, opened 2019-02-21
Add environment variables to the run image dialog

Gundersanne:


src/ImageRunModal.jsx
```diff
@@ -273,6 +305,13 @@ export class ImageRunModal extends React.Component {
                                  onChange={value => this.onValueChanged('publish', value)}
                                  default={{ containerPort: null, hostPort: null, protocol: 'TCP' }}
                                  itemcomponent={ <PublishPort />} />
+
+                <label className='control-label' htmlFor='run-image-dialog-env'>{ _("Environment") }</label>
+                <DynamicListForm id='run-image-dialog-env'
+                                 formclass='env-form'
```
KKoukiou: I tend to not use classes that don't introduce any css but are just used for css selectors in the test. IDs are better fit for this purpose. But it's not important to block this PR from being merged.

martinpitt: From my POV it's mostly a matter of taste. Sometimes classes are preferrable if the thing you select is not unique. The one thing to avoid is to attach CSS to IDs.

src/ImageRunModal.jsx
```diff
@@ -28,88 +28,101 @@ const units = {
     },
 };
 
-class PublishPortsForm extends React.Component {
+const PublishPort = (props) => {
```
martinpitt: Style nitpick: This can be shortened to `=> (\n<React.Fragment...)`, no need for the {} and return.

Also, it's more explicit and easier to read to decompose props into the expected props (kind of like "poor (wo)man's propTypes" for functional components):

    ({ foo, bar, baz }) => { ...

src/ImageRunModal.jsx
```diff
@@ -118,7 +131,7 @@ class PublishPortsForm extends React.Component {
         );
     }
 }
-PublishPortsForm.propTypes = {
+DynamicListForm.propTypes = {
     onChange: PropTypes.func.isRequired,
     id: PropTypes.string.isRequired,
```
martinpitt: Let's add the optional `formclass` and the required `itemcomponent` and `default`.  *if* a component has propTypes, it should be complete.

src/ImageRunModal.jsx
```diff
@@ -28,88 +28,101 @@ const units = {
...
+        </React.Fragment>
+    );
+};
+
+class DynamicListForm extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
-            publish: [{ key: 0, containerPort: null, hostPort: null, protocol: 'TCP' }],
+            list: [Object.assign({ key: 0 }, props.default, {})],
```
martinpitt: All your `Object.assign()` calls now have an extra `{}` at the end. This seems to be a no-op, or am I missing anything?

test/check-application
```diff
@@ -349,21 +349,38 @@ class TestApplication(testlib.MachineCase):
...
         b.wait_present('.publish-port-form + .publish-port-form')
-        b.set_input_text('.publish-port-form:last-of-type input:first-child', '5001')
-        b.set_input_text('.publish-port-form:last-of-type input:last-child', '6001')
-        b.click('.publish-port-form:last-of-type button.dropdown-toggle')
-        b.wait_visible('.publish-port-form:last-of-type .dropdown-menu a:contains("UDP")')
-        b.click('.publish-port-form:last-of-type .dropdown-menu a:contains("UDP")')
-        b.wait_visible('.publish-port-form:last-of-type button.dropdown-toggle:contains("UDP")')
-        b.click('.publish-port-form:last-of-type .fa-plus')
+        b.set_input_text('.publish-port-form + .publish-port-form input:first-child', '5001')
+        b.set_input_text('.publish-port-form + .publish-port-form input:last-child', '6001')
```
martinpitt: This duplication is a bit ugly. This is because each port *entry* its own form, right?

Could this use a more specific selector instead, such as `.publish-port-form[containerPort='']` to select the first empty one, or in other cases with the expected port to select a filled one?

src/ImageRunModal.jsx
Gundersanne: yup you're right, this doesn't do anything. I'll remove it

src/ImageRunModal.jsx
Gundersanne: oki

test/check-application
Gundersanne: the outer div doesn't really have any  of the attributes of the input fields. An alternative is `nth-child` from the parent of this div.

src/ImageRunModal.jsx
martinpitt: Please actually remove them then :-)

src/ImageRunModal.jsx
martinpitt: This comment is still current.

src/ImageRunModal.jsx
```diff
@@ -28,88 +28,101 @@ const units = {
...
-            let ports = state.publish.concat();
-            ports.splice(idx, 1);
-            if (ports.length === 0)
-                ports.push({ key: this.keyCounter++, containerPort: null, hostPort: null, protocol: 'TCP' });
-            return { publish: ports };
-        }, () => this.props.onChange(this.state.publish.concat()));
+            let items = state.list.concat();
+            items.splice(idx, 1);
+            if (items.length === 0)
+
```
martinpitt: spurious newline

src/ImageRunModal.jsx
```diff
@@ -68,13 +68,37 @@ const PublishPort = (props) => {
             </div>
         </React.Fragment>
     );
-};
+
+const EnvVar = (props) =>
```
martinpitt: Same here, please decompose props so that it's easy to see at a glance what this expects.

src/ImageRunModal.jsx
```diff
@@ -28,8 +28,8 @@ const units = {
     },
 };
 
-const PublishPort = (props) => {
-    return (
+const PublishPort = (props) =>
+    (
```
martinpitt: Ah, so this changes it -- that should go into the previous commit.

src/ImageRunModal.jsx
```diff
@@ -88,14 +112,14 @@ class DynamicListForm extends React.Component {
             items.splice(idx, 1);
             if (items.length === 0)
 
-                items.push(Object.assign({ key: this.keyCounter++ }, this.props.default, {}));
+                items.push(Object.assign({ key: this.keyCounter++ }, this.props.default));
```
martinpitt: Same here, these changes belong in the previous commit.

test/check-application
```diff
@@ -384,6 +401,12 @@ class TestApplication(testlib.MachineCase):
         self.assertIn('8001 8001 tcp', ports)
         self.assertNotIn('7001 7001 tcp', ports)
 
+        env = m.execute("podman inspect --format '{{.Config.Env}}' busybox-with-tty");
```
martinpitt: This is fine. Just as a thing to consider, to test this all the way through, this could also do `podman exec busybox-with-tty env`, and skip making assumptions about the `podman inspect` API.

src/ImageRunModal.jsx
```diff
@@ -28,88 +28,99 @@ const units = {
...
-                                </div>
-                                <div role='group' className='ct-form-layout-split'>
-                                    <Button bsStyle='default' className='pficon-close'
-                                            disabled={dialogValues.publish.length === 1 &&
-                                                      (!dialogValues.publish[0]['containerPort'] ||
-                                                       !dialogValues.publish[0]['hostPort'])}
-                                            onClick={() => this.removePort(idx)} />
-                                    <Button bsStyle='default' className='fa fa-plus' onClick={this.addPort} />
-                                </div>
+
```
martinpitt: spurious newline

src/ImageRunModal.jsx
```diff
@@ -28,88 +28,99 @@ const units = {
...
+                            <div className={ (formclass || '') + ' ct-form-layout' } key={ item.key }>
+                                {
+                                    React.cloneElement(this.props.itemcomponent, {
+                                        idx: idx, item: item, id: (idx === 0 && id) || undefined,
```
martinpitt: style for the future: we usually use `? :` instead of `&& ||` for this pattern.

PR #59, opened 2019-02-28
Add volumes to the run image dialog

Gundersanne:
Current state:

image

- [x] integrated FileAutoComplete component

### DIFF
--- a/lib/cockpit-components-file-autocomplete.jsx
+++ b/lib/cockpit-components-file-autocomplete.jsx
@@ -1,0 +1,307 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2017 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Original source:
+ * https://github.com/cockpit-project/cockpit/blob/master/pkg/lib/cockpit-components-file-autocomplete.jsx
+ */
+
+import cockpit from "cockpit";
+import React from "react";
+import PropTypes from "prop-types";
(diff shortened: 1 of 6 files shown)


mheon:
Wait, our docs say that volumes are RO by default? That sounds wrong. Can you point out where we say that?

larskarlitski:
> Wait, our docs say that volumes are RO by default? That sounds wrong. Can you point out where we say that?

No, they say that they're mounted read-write by default in `podman-run(1)`.

Gundersanne:
@larskarlitski I'm find with dropping the default option, I copied it over from out docker page but really it's equal to ReadWrite.

larskarlitski:
> We need a bit of design input maybe? section stuff off, and I'd really like to align the `x` and `+` buttons on the right.

Yep, I think @garrett might be able to help with the form layout stuff.

garrett:
> We need a bit of design input maybe? section stuff off, and I'd really like to align the x and + buttons on the right.

Thoughts:

1. The buttons look aligned on the right? Perhaps that was an earlier comment?
2. I see there's a ct-form-layout inside another. Interesting!
3. Things don't work well on mobile, but as long as we file another issue or open another PR for it, that's OK.
4. "With terminal"'s checkbox feels a bit weird.
5. It feels like there are _a lot_ of widgets packed together. I wonder if we should have some vertical breathing space, like with `<hr>`s above Ports, Volumes, and Environment... and perhaps also below Name.

Gundersanne:
@garrett 

maximum hr's:
image

under name and under terminal:
image

What do you prefer?

garrett:
While I like the second option more visually as a default, the first probably makes sense when there are more than one ports, volumes, and/or environments, so + and × buttons are more clearly associated with the section. (In your screenshot at the very top of this PR, it's hard to see which action goes with which section.)

So: Perhaps the first option?

Also, I think it makes sense here to remove or hide the x buttons when they're not available. (`display: none` is fine.)

Gundersanne:
https://github.com/cockpit-project/cockpit-podman/pull/59#issue-257134510

Gundersanne:
@garrett  updated!

lib/cockpit-components-file-autocomplete.jsx
```diff
@@ -0,0 +1,304 @@
...
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
larskarlitski: Is this taken from Cockpit? If so, please add a comment with a link to it, so that we can sync easier.

Gundersanne: done!

PR #61, opened 2019-03-07
spec: Align property values

martinpitt:
Use the same 17-column row for the property keys and aligned values as
in cockpit and composer. This will provide a consistent result from
cockpituous' release-srpm when it adds/modifies Version: and Release:.


martinpitt:
See related https://github.com/cockpit-project/cockpituous/pull/245

PR #66, opened 2019-03-15
Move commit modal handler code into the CommitModalComponent and enhance error messages

KKoukiou:
Container.jsx file is home to much code that should be split into different files. This makes this file a total mess.

By moving all the code related to ContainerCommit to the ContainerCommitModal we can get rid of all these useless handlers that we need to pass around the react components otherwise.

Also error messages previously were printed like JSON.stringify(ex) which is definitely not a user friendly way to tell 'you container commit failed'. So I changed a bit the presentation of error messages as well.

### DIFF
--- a/src/ContainerCommitModal.jsx
+++ b/src/ContainerCommitModal.jsx
@@ -1,6 +1,10 @@
 import React from 'react';
-import { Alert, Modal, Button, FormGroup, FormControl } from 'patternfly-react';
+import { Modal, Button, FormGroup, FormControl } from 'patternfly-react';
 import cockpit from 'cockpit';
+
+import * as utils from './util.js';
+import varlink from './varlink.js';
+import { ErrorNotification } from './Notification.jsx';
 
 import '../lib/form-layout.less';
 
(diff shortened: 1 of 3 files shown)


src/ContainerCommitModal.jsx
```diff
@@ -58,6 +61,50 @@ class ContainerCommitModal extends React.Component {
...
+        let onbuildsArr = [];
+        if (this.state.setonbuild) {
+            onbuildsArr = utils.getCommitArr(this.state.onbuild, "ONBUILD");
+        }
+        commitData.changes.push(...onbuildsArr);
+
+        varlink.call(utils.PODMAN_ADDRESS, "io.podman.Commit", commitData)
+                .then(() => {
+                    this.props.updateImagesAfterEvent();
+                    this.props.updateContainersAfterEvent();
```
Gundersanne: These props are missing I think.

KKoukiou:
@Gundersanne finally I was able to test locally as well. So it's good form me know, if you are ok with this as well.

PR #68, opened 2019-03-26
Use podman events to trigger state updates

KKoukiou:
Blocked till libpod release v1.1.3 https://github.com/containers/libpod/releases

### DIFF
--- a/src/ContainerCommitModal.jsx
+++ b/src/ContainerCommitModal.jsx
@@ -92,11 +92,7 @@ class ContainerCommitModal extends React.Component {
         commitData.changes.push(...onbuildsArr);
 
         varlink.call(utils.PODMAN_ADDRESS, "io.podman.Commit", commitData)
-                .then(() => {
-                    this.props.updateImagesAfterEvent();
-                    this.props.updateContainersAfterEvent();
-                    this.props.onHide();
-                })
+                .then(() => this.props.onHide())
                 .catch(ex => {
                     this.setState({
                         dialogError: cockpit.format(_("Failed to commit container $0"), this.props.container.names),
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -256,10 +256,7 @@ export class ImageRunModal extends React.Component {
 
         varlink.call(utils.PODMAN_ADDRESS, "io.podman.CreateContainer", { create: createConfig })
                 .then(reply => varlink.call(utils.PODMAN_ADDRESS, "io.podman.StartContainer", { name: reply.container }))
-                .then(() => {
-                    this.props.close();
-                    return this.props.updateContainersAfterEvent();
-                })
+                .then(() => this.props.close())
                 .catch(ex => {
                     this.setState({
                         dialogError: _("Container failed to be created"),
--- a/src/varlink.js
+++ b/src/varlink.js
@@ -18,7 +18,7 @@ class VarlinkError extends Error {
  *
  * Returns a connection object:
  *
- * - connection.call(method, parameters) calls "method" (prefixed by the
+ * - connection.call(method, parameters, more) calls "method" (prefixed by the
  *   interface) with "parameters". Multiple calls can be made without waiting
  *   for replies, but calls are made sequentially (similar to http). Returns a
  *   promise that resolves with the return value of the method or a tuple
@@ -53,15 +53,20 @@ function connect(address) {
 
         for (let i = 0; i < chunks.length; i += 1) {
             const message = JSON.parse(chunks[i]);
+            const { resolve, reject, replyCallback } = pending.shift();
 
-            let { resolve, reject } = pending.shift();
-            if (message.error)
+            if (message.error) {
                 reject({
                     error: message.error,
                     parameters: message.parameters
                 });
-            else
+            } else if (replyCallback) {
+                replyCallback(message);
+                if (message.continues)
+                    pending.push({ resolve, reject, replyCallback });
+            } else {
                 resolve(message.parameters);
+            }
         }
     });
 
@@ -81,6 +86,17 @@ function connect(address) {
         channel.send([0]);
 
         return new Promise((resolve, reject) => pending.push({ resolve, reject }));
+    };
+
+    connection.monitor = function (method, parameters, replyCallback) {
+        parameters = parameters || {};
+
+        const data = encoder.encode(JSON.stringify({ method, parameters, more: true }));
+
+        channel.send(data);
+        channel.send([0]);
+
+        return new Promise((resolve, reject) => pending.push({ resolve, reject, replyCallback }));
     };
 
     connection.close = function () {
--- a/test/check-application
+++ b/test/check-application
@@ -6,6 +6,8 @@
 import os
 import sys
 import time
+from distutils.version import StrictVersion
+import unittest
 
 # import Cockpit's machinery for test VMs and its browser test API
 TEST_DIR = os.path.dirname(__file__)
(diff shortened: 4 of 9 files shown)


src/varlink.js
```diff
@@ -53,15 +53,23 @@ function connect(address) {
...
-                reject({
-                    error: message.error,
-                    parameters: message.parameters
-                });
-            else
-                resolve(message.parameters);
+            const { resolve, reject, eventCallback } = pending.shift();
+
+            if (eventCallback) {
+                if (message.parameters && message.parameters.events) {
```
larskarlitski: `message.parameters.events` is specific to podman's API, isn't it? I'd prefer if `varlink.js` stays a generic varlink client implementation so that we can split it out at some point.

src/varlink.js
```diff
@@ -83,6 +91,17 @@ function connect(address) {
         return new Promise((resolve, reject) => pending.push({ resolve, reject }));
     };
 
+    connection.monitor = function (method, parameters, eventCallback) {
```
larskarlitski: Varlink calls everything "reply", not "event". Please stay with that terminology and use `eventCallback` instead of `replyCallback`.

src/varlink.js
larskarlitski: The error case is not handled for `more` calls.

src/varlink.js
KKoukiou: Fixed. (I am sure you meant to use replyCallback btw instead of eventCallback)

src/varlink.js
KKoukiou: Done, thanks for the catch

martinpitt:
The fedora failure indicates that this PR should bump the podman dependency version, to ensure the API  is available.

KKoukiou:
> "req changes" for the missing dependency bump

I also implemented the test skip in case podman version is not sufficient, so that we will not fail on fedora-29.

test/check-application
```diff
@@ -22,6 +24,13 @@ registries = ['localhost:5000']
 """
 
 class TestApplication(testlib.MachineCase):
+    def setUp(self):
+        super(TestApplication, self).setUp()
+
+        podman_version = self.machine.execute("rpm -qi podman  | grep Version | cut -d ':' -f2 | cut -c2-")
+        tests_enabled = StrictVersion(podman_version) >= StrictVersion('1.2.0')
+        if not tests_enabled:
+            raise unittest.SkipTest("podman version if not sufficient")
```
martinpitt: I don't understand why this is necessary. The package should fail to install if the podman version is not sufficient, i. e. the fedora-29 test should fail during image preparation. I wonder why it doesn't?

KKoukiou: As discussed on IRC it was the missing epoch in the spec file. Thanks for solving this.

test/check-application
```diff
@@ -46,9 +48,6 @@ class TestApplication(testlib.MachineCase):
         # run a container
         m.execute("podman run -d --name swamped-crate busybox sleep 1000")
 
-        # FIXME: UI needs to listen to change signals and update automatically
-
-        b.reload()
         b.enter_page("/podman")
```
martinpitt: These `enter_page` should now get dropped as well, they were just a follow-up of the reload.

src/ContainerCommitModal.jsx
```diff
@@ -93,8 +93,6 @@ class ContainerCommitModal extends React.Component {
 
         varlink.call(utils.PODMAN_ADDRESS, "io.podman.Commit", commitData)
                 .then(() => {
-                    this.props.updateImagesAfterEvent();
-                    this.props.updateContainersAfterEvent();
```
martinpitt: Could now be simplified to just `then(this.props.onHide)`

src/ImageRunModal.jsx
```diff
@@ -258,7 +258,6 @@ export class ImageRunModal extends React.Component {
                 .then(reply => varlink.call(utils.PODMAN_ADDRESS, "io.podman.StartContainer", { name: reply.container }))
                 .then(() => {
```
martinpitt: Same here, could now be `then(this.props.close)`

test/check-application
```diff
@@ -132,8 +130,10 @@ class TestApplication(testlib.MachineCase):
         b.wait_present('.modal-dialog div:contains("io.podman.ErrorOccurred")')
         b.click(".modal-dialog div .alert .close")
         b.wait_not_present(".modal-dialog div .alert")
+        self.confirm_modal("btn-ctr-cancel-commit")
```
martinpitt: That looks a bit weird, how does the event handling have an effect on suddenly showing a dialog here?

test/check-application
```diff
@@ -132,8 +130,10 @@ class TestApplication(testlib.MachineCase):
...
 
         # delete image busybox that hasn't been used
+        b.click('#containers-images tbody tr:contains("busybox:latest") td.listing-ct-toggle')
```
martinpitt: Is that an accidental copy&paste, or deliberate? There's now a wait_present *after* this click.

test/check-application
KKoukiou: done

src/ContainerCommitModal.jsx
KKoukiou: done

src/ImageRunModal.jsx
KKoukiou: done

test/check-application
KKoukiou: This is totally a wrong paste or something. And it's indeed weird, not sure where that modal did appear from. I removed the line and it succeeds as well so if you don't want further investigation what was going on I will just ignore it.

test/check-application
KKoukiou: Without it it fails, because the image row will not be expanded for the delete button to be visible (line 139), but I noticed now that there is a second toggle before the button click (line 138), so I will remove this and the second and the state of the row should be expanded

test/check-application
```diff
@@ -135,7 +130,6 @@ class TestApplication(testlib.MachineCase):
 
         # delete image busybox that hasn't been used
         b.wait_present('#containers-images tr:contains("busybox:latest")')
-        b.click('#containers-images tbody tr:contains("busybox:latest") td.listing-ct-toggle')
```
martinpitt: Seems weird (or at least unrelated) to drop this line now? Isn't the delete button inside the details? If not, this should become a separate commit.

test/check-application
```diff
@@ -295,22 +289,14 @@ class TestApplication(testlib.MachineCase):
...
         self.assertNotEqual(old_pid, new_pid)
 
-        # Reload page after restarting the container since we don't implement proper monitoring and guessing the state
-        # can lead to unexpected behavior.
-        b.reload()
-        b.enter_page("/podman")
         self.filter_containers('all')
         b.wait_present("#containers-containers")
-        b.wait_present('#containers-containers tr:contains("swamped-crate")')
-        b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
```
martinpitt: Why was this dropped?

test/check-application
KKoukiou: It's weird but the test was already opening that row much more above, https://github.com/cockpit-project/cockpit-podman/pull/68/files/ab7e4acf1ab8c3eac08bdb1ccaba3d74108ac168#diff-c23b7eb2c7c51d1e825ac452faf7eaa1R39

The only reason we needed to open it again here is because we reloaded the page in the past.

I agree that this test needs refactoring to be more readable.

test/check-application
KKoukiou: The toggle is dropped because we don't use reload any more.

The wait_present line above is exactly the same check as the one wait_present bellow that was added, but indeed this change is not any actual change and it's confusing as part of this commit.

martinpitt: Oh, of course! Thanks

KKoukiou:
@martinpitt I pushed one more commit to get rid of one of the 2 flakes since it was quite persistent. Please take a look again.

martinpitt:
Changes look fine (thanks!), but still a test failure :-/

PR #73, opened 2019-04-04
package.json: Pin dependencies to exact versions

martinpitt:
We now let the Cockpit npm-update bot update these in a controlled
manner.


martinpitt:
Note that I left react-dom at the old version. So the next round of npm-update should pick that up.

PR #74, opened 2019-04-04
Flake fixes

Gundersanne:
number tests succeeded: 6
failures: 1, on clicking run which would indicate a podman error

### DIFF
--- a/lib/cockpit-components-file-autocomplete.jsx
+++ b/lib/cockpit-components-file-autocomplete.jsx
@@ -55,7 +55,7 @@ export class FileAutoComplete extends React.Component {
         this.debouncedChange = debounce(300, (value) => {
             this.pendingCallback = false;
             if (!this.updateIfDirectoryChanged(value)) {
-                var stateUpdate = this.filterFiles(value);
+                let stateUpdate = this.filterFiles(value);
                 this.setState(stateUpdate);
                 this.onChangeCallback(value, { error: stateUpdate.error });
             }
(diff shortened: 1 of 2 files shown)


lib/cockpit-components-file-autocomplete.jsx
```diff
@@ -53,12 +53,13 @@ export class FileAutoComplete extends React.Component {
...
-            if (!this.updateIfDirectoryChanged(value)) {
+            this.updateIfDirectoryChanged(value, () => {
+                this.pendingCallback = false;
                 let stateUpdate = this.filterFiles(value);
-                this.setState(stateUpdate);
-                this.onChangeCallback(value, { error: stateUpdate.error });
-            }
+                this.setState(stateUpdate, () => {
+                    this.onChangeCallback(value, { error: stateUpdate.error });
+                });
```
martinpitt: No need for the { }.

lib/cockpit-components-file-autocomplete.jsx
```diff
@@ -153,16 +154,18 @@ export class FileAutoComplete extends React.Component {
...
         const changed = directory !== this.state.directory;
         if (changed && this.state.directoryFiles !== null) {
-            this.setState({
-                displayFiles: [],
-                directoryFiles: null,
-                directory: directory,
-                open: false,
-            });
+            this.setState(() => {
+                return {
```
martinpitt: No need for the `{ return; }`

martinpitt:
Please also add a Closes #.

This looks like it would equally apply to cockpit itself, can you please send a PR? Thanks!

lib/cockpit-components-file-autocomplete.jsx
Gundersanne: In this case it's needed i think, as that's the functional version.

martinpitt: Right, but `() => { foo: null }` is still a function. :-)

martinpitt: In general, `() => expression` is fully equivalent to `() => { return expression }`

PR #82, opened 2019-04-17
Test cleanups

martinpitt:


KKoukiou:
The fedora-30 failed on the test download which is expected failure, we need to copy the naughties to fedora-30

PR #91, opened 2019-04-24
cockpituous: Automatically release to Fedora 29 and 30

martinpitt:


martinpitt:
I verified that the [Fedora package](https://src.fedoraproject.org/rpms/cockpit-podman/settings#usersgroups-tab) has "cockpit" as committer.

PR #100, opened 2019-05-16
Cmd with arguments

marusak:
Using command like `sh -c "while date; do sleep 2; done"` to create new
container would fail with executable not found as it took the whole line as an executable.

This commit makes it possible to use arguments in the command.

Also show as default command image cmd.

### DIFF
--- a/src/ContainerDetails.jsx
+++ b/src/ContainerDetails.jsx
@@ -34,7 +34,7 @@ const ContainerDetails = ({ container }) => (
             <dt>{_("Image")}</dt>
             <dd>{container.image}</dd>
             <dt>{_("Command")}</dt>
-            <dd>{container.command ? container.command.join(" ") : ""}</dd>
+            <dd>{container.command ? util.quote_cmdline(container.command) : ""}</dd>
             <dt>{_("State")}</dt>
             <dd>{render_container_state(container)}</dd>
             <dt>{_("Ports")}</dt>
--- a/src/ImageDetails.jsx
+++ b/src/ImageDetails.jsx
@@ -23,7 +23,7 @@ const ImageDetails = (props) => {
                 <dt>{_("Entrypoint")}</dt>
                 <dd>{image.entrypoint ? image.entrypoint.join(" ") : ""}</dd>
                 <dt>{_("Command")}</dt>
-                <dd>{image.command ? image.command.join(" ") : "" }</dd>
+                <dd>{image.command ? util.quote_cmdline(image.command) : "" }</dd>
                 <dt>{_("Created")}</dt>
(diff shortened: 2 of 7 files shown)


src/ContainerDetails.jsx
```diff
@@ -34,7 +34,7 @@ const ContainerDetails = ({ container }) => (
             <dt>{_("Image")}</dt>
             <dd>{container.image}</dd>
             <dt>{_("Command")}</dt>
-            <dd>{container.command ? container.command.join(" ") : ""}</dd>
+            <dd>{container.command ? util.quote_cmdline(container.command) : ""}</dd>
```
larskarlitski: This should be `utils`, right?

src/ImageDetails.jsx
```diff
@@ -23,7 +23,7 @@ const ImageDetails = (props) => {
                 <dt>{_("Entrypoint")}</dt>
                 <dd>{image.entrypoint ? image.entrypoint.join(" ") : ""}</dd>
                 <dt>{_("Command")}</dt>
-                <dd>{image.command ? image.command.join(" ") : "" }</dd>
+                <dd>{image.command ? util.quote_cmdline(image.command) : "" }</dd>
```
larskarlitski: `utils` here as well.

src/util.js
```diff
@@ -134,3 +134,92 @@ export function getCommitArr(arr, cmd) {
...
+ * user edit a sequence of words as a single string.
+ *
+ * When parsing, words are separated by whitespace.  Single and double
+ * quotes can be used to protect a sequence of characters that
+ * contains whitespace or the other quote character.  A backslash can
+ * be used to protect any character.  Quotes can appear in the middle
+ * of a word.
+ */
+
+export function quote_cmdline(words) {
```
larskarlitski: Do these come from Cockpit? Please mention that in the comment above so that people know where to apply fixes and pull in new versions from.

larskarlitski:
Ugh, sorry, got it the wrong way around: it should be `util` everywhere, not `utils` (this is confusing!)

marusak:
> Ugh, sorry, got it the wrong way around: it should be util everywhere, not utils (this is confusing!)

Sometimes it is included as `util` sometimes as `utils` :) 
https://github.com/cockpit-project/cockpit-podman/blob/master/src/ContainerDetails.jsx#L3
https://github.com/cockpit-project/cockpit-podman/blob/master/src/ImageRunModal.jsx#L9

src/util.js
marusak: Done

larskarlitski:
> Sometimes it is included as `util` sometimes as `utils` :)

oh my! Sorry, I didn't know that. How confusing.

PR #105, opened 2019-05-23
Update to the new create type

marusak:
Since version 1.3.0 (with epoch 2) there has been some changes in API,
cockpit-podman has been affected with the create type.

There were four changes, which needed to be addressed:
1) env is not map but array of 'key=value'
2) image is passed in array `args`
3) command is passed in array `args`
4) `volumes` option has been renamed to `volume`

These changes were introduced to better match commandline options from
podman itself.

Note about tests:
f30 - this image has this new version and naughty, since it was failing. Expected result is that all tests pass without the naughty being used.

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -219,12 +219,12 @@ export class ImageRunModal extends React.Component {
     getCreateConfig() {
         let createConfig = {};
 
-        createConfig.image = this.state.image.repoTags ? this.state.image.repoTags[0] : "";
+        createConfig.args = this.state.image.repoTags ? [this.state.image.repoTags[0]] : [""];
         createConfig.resources = {};
         if (this.state.containerName)
             createConfig.name = this.state.containerName;
         if (this.state.command)
-            createConfig.command = utils.unquote_cmdline(this.state.command);
+            createConfig.args = createConfig.args.concat(utils.unquote_cmdline(this.state.command));
         if (this.state.memoryConfigure && this.state.memory)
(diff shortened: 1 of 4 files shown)


marusak:
After discussion with @martinpitt we decided to support both APIs for now. I pushed second commit that does that. It is purposefully two commits since when we don't want to support 1.2 version we can just revert the second one with ease.

src/ImageRunModal.jsx
```diff
@@ -218,13 +218,22 @@ export class ImageRunModal extends React.Component {
 
     getCreateConfig() {
         let createConfig = {};
+        // HACK: checking version is because API got broken from 1.2 -> 1.3
+        // Once we support only 1.3 podman, commit which introduces these changes can be reverted
```
larskarlitski: Please include which call was broken and in what way.

Super tiny nitpick: I prefer using active instead of passive voice: the API didn't "get broken", but "podman broke API". Passive voice makes it sound like it was an accident.

src/ImageRunModal.jsx
```diff
@@ -218,13 +218,22 @@ export class ImageRunModal extends React.Component {
...
 
-        createConfig.image = this.state.image.repoTags ? this.state.image.repoTags[0] : "";
+        if (this.props.version < "1.3.0")
```
larskarlitski: You can't compare version numbers lexicographically. It will break as soon as you have components in the version that are larger than 10. For example, `"1.10.0" < "1.3.0"` evaluates to `true`, even though the first version is larger.

There's a `semver` package to do this, but that might be overkill for this little check. You could either split the string and compare the components in order, or skip the version check completely and check the type of the `repoTags` you receive directly (the API difference is that it used to return an array, right?).

PR #106, opened 2019-05-23
Introduce terminal

marusak:
When container is not running not terminal is shown.
When container is started while container details are opened, terminal
shows up. When container stopped while details are opened, "disconnected"
message is shown and terminal is kept open (however you cannot write
into it).

When container is interactive (such as executed `sh`) a prompt is shown.
This is done by sending SIGWINCH signal (podman does the same on
`attach`).

TODO:

- [x] Opening terminal tab of stopped container and then starting container attaches container outside of parent
- [x] Resizing works only when making window bigger - when making smaller it stick outside of viewport
- [ ] Resizing is set through control channel which is FIFO - `cockpit.file` did not work (very likely I used it incorrectly). Using `cockpit.spawn` is not elegant neither it can read from the pipe.
- [x] Tests needs to be adjusted to a new location (just opening another tab and then changing selectors a bit)
- [ ] When container is not running show(?) also button to start container as in the design (I find it bit redundant)

Used to look like this:
terminaleins
Stopping container shows disconnected:
stopedterminal

There is one small issue I am aware of and that being that when terminal is not interactive and is just logging (like `sh -c "while echo hello; do :; done"`) then SIGWINCH signal leaves `^L` on first line. But for now I take this as for follow up. Ideas welcomed :)

Note about tests:
This won't work because it needs https://github.com/cockpit-project/cockpit/pull/11872 which will reach this image only when released. So hopefully sometime after next release. I tested this with F30 image which I manually prepared to contain these changes.

### DIFF
--- a/src/ContainerTerminal.jsx
+++ b/src/ContainerTerminal.jsx
@@ -1,0 +1,184 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2019 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import cockpit from 'cockpit';
+import { Terminal } from "xterm";
+
+import * as utils from './util.js';
+import varlink from './varlink.js';
+
+import "./ContainerTerminal.css";
+
+const _ = cockpit.gettext;
+
+class ContainerTerminal extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.onChannelClose = this.onChannelClose.bind(this);
+        this.onChannelMessage = this.onChannelMessage.bind(this);
+        this.disconnectChannel = this.disconnectChannel.bind(this);
+        this.connectChannel = this.connectChannel.bind(this);
+        this.resize = this.resize.bind(this);
+
+        let term = new Terminal({
+            cols: 80,
+            rows: 24,
+            screenKeys: true,
+            cursorBlink: true,
+            fontSize: 12,
+            fontFamily: 'Menlo, Monaco, Consolas, monospace',
+            screenReaderMode: true
+        });
+
+        this.state = {
+            term: term,
+            container: props.containerId,
+            channel: null,
+            control_channel: null,
+            opened: false,
+            errorMessage: "",
+            cols: 80,
+        };
+    }
+
+    componentDidMount() {
+        this.connectChannel();
+    }
+
+    componentDidUpdate(prevProps, prevState) {
+        if (!this.state.channel && this.props.containerStatus === "running" && prevProps.containerStatus !== "running")
+            this.connectChannel();
+        if (prevProps.width !== this.props.width) {
+            this.resize(this.props.width);
+        }
+    }
+
+    resize(width) {
+        var padding = 11 + 5 + 50;
+        var realWidth = this.state.term._core.renderer.dimensions.actualCellWidth;
+        var cols = Math.floor((width - padding) / realWidth);
+        this.state.term.resize(cols, 24);
+        cockpit.spawn(["sh", "-c", "echo '1 24 " + cols.toString() + "'>" + this.state.control_channel], { superuser: true });
+        this.setState({ cols: cols });
+    }
+
+    connectChannel() {
+        let self = this;
+        if (self.state.channel)
+            return;
+
+        if (self.props.containerStatus !== "running") {
+            let message = _("Container is not running");
+            this.setState({ errorMessage: message });
+            return;
+        }
+
+        varlink.call(utils.PODMAN_ADDRESS, "io.podman.GetAttachSockets", { name: this.state.container })
+                .then(out => {
+                    let opts = {
+                        payload: "packet",
+                        unix: out.sockets.io_socket,
+                        superuser: "require",
+                        binary: false
+                    };
+
+                    let channel = cockpit.channel(opts);
+                    channel.wait()
+                            .then(() => {
+                                // Show the terminal. Once it was shown, do not show it again but reuse the previous one
+                                if (!this.state.opened) {
+                                    this.state.term.open(this.refs.terminal);
+                                    this.setState({ opened: true });
+
+                                    self.state.term.on('data', function(data) {
+                                        if (self.state.channel)
+                                            self.state.channel.send(data);
+                                    });
+                                }
+
+                                channel.addEventListener("message", this.onChannelMessage);
+                                channel.addEventListener('close', this.onChannelClose);
+
+                                channel.send(String.fromCharCode(12)); // Send SIGWINCH to show prompt on attaching
+                                this.setState({ channel: channel, control_channel: out.sockets.control_socket, errorMessage: "" });
+                                this.resize(this.props.width);
+                            })
+                            .catch(e => {
+                                let message = cockpit.format(_("Could not open channel: $0"), e.problem);
+                                if (e.problem === "not-supported")
+                                    message = _("This version of the Web Console does not support a terminal.");
+                                this.setState({ errorMessage: message });
+                            });
+                })
+                .catch(e => this.setState({ errorMessage: cockpit.format(_("Could not attach to this container: $0"), e.problem) }));
+    }
+
+    componentWillUnmount() {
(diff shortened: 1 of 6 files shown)


marusak:
@larskarlitski reminded me one point I forgot to mention here:
We need the new `packet` payload from cockpit, without it it won't work. I should be possible to recognize that it is not supported. What should be done in such case? Shouldn't we show terminal at all (as we don't show it when container is not running?) Should we show terminal and write into it message like 'This version of cockpit does not support terminal...', Should we show this message just as line on the same place where terminal should be? Other ideas? @andreasn @garrett @larskarlitski

martinpitt:
My gut feeling is that this should not render the terminal at all in this case. This will avoid wasting a large amount of screen space for something useless. This is for sure better than bumping the cockpit-bridge dependency.

garrett:
Having a terminal available for a container is really cool, especially for debugging purposes.

However, I think most of the time people won't want a terminal.

We could move the terminal to its own tab instead of being in the "overview" — especially since there's just one tab right now. This is what the Machines page does with the graphical console for VMs (have it on another tab).

marusak:
Follow up with some questions and comments:
1) So let's have it in another tab, that is good idea, thanks @garrett . The first question, should it span through it fully? Should there be some padding? Or just make some smaller window of static size?
Spanning without any margin:
spanning

2) There has been discussion on IRC if the `Console` tab should be visible when we cannot provide console (such as when newer version of Cockpit is needed). Although majority thinks that this tab shouldn't be visible, @larskarlitski brought up great comment that this can confuse users who will not see the console. Also it is kinda tricky to implement since we know if we can connect only when we open the terminal tab - so until we open it we don't know if we need it. Some checks would be necessary to perform beforehand to see if we can show console.

3) Similarly to 2) - should the tab be shown when container is not running? This is easy to implement, but still the same question - wouldn't that confuse users why there is no terminal when they open container details?

Lets have some design discussion and possibly some simple mockup how to show that we cannot show terminal (if we decide to show the tab).

garrett:
We could use something like an empty state pattern with a message that the container needs to be running to use the terminal. (And then a button to start the container?)

andreasn:
I like the idea with the empty state.
So something like this?

mockup

andreasn:
Garrett felt it looked too much like an error, so make it look more like a regular page

mockup

marusak:
I pushed a new version, but now it is WIP, moving it into tab changed quiet a few things (mostly related to not being fixed size anymore, but we need to react to the changing size of the window).
I will put TODO list into first comment. Here are some screenshot how it looks like now.

When container is not running:
notrunning
When newer version is needed:
newerversionrequierd
And terminal:
withcontainer

src/ContainerTerminal.jsx
```diff
@@ -0,0 +1,187 @@
...
+                                window.addEventListener('resize', this.onWindowResize);
+
+                                channel.send(String.fromCharCode(12)); // Send SIGWINCH to show prompt on attaching
+                                this.setState({ channel: channel, control_channel: out.sockets.control_socket, errorMessage: "" });
+                                this.onWindowResize();
+                            })
+                            .catch(e => {
+                                let message = cockpit.format(_("Could not open channel: $0"), e.problem);
+                                if (e.problem === "not-supported")
+                                    message = _("Newer version of cockpit is required in order to show terminal");
```
andreasn: Maybe it's better to say _web console_ here?

garrett: Agreed. We never do (and never should) refer to "Cockpit" within Cockpit.

"Upgrade to newer version of the Web Console for an interactive terminal." is probably better text. It communicates the intent of the tab, states the problem, and includes a solution.

Backing up a bit: In what cases is this message necessary? Can we remove the tab completely for unsupported versions? Otherwise, it's basically UI leftovers that don't serve a purpose. None of the containers would have a console, yet they would all have a vestigial tab that teases one.

It's probably exceptionally rare that someone would run a version of Cockpit that's too old to have support for the terminal in the Podman page and also run the Podman page within Cockpit. (Of course, we should handle this exception so there's no outright _error_ if this happens, but I'm not sure we need a message here. Do we?)

marusak: > Backing up a bit: In what cases is this message necessary? Can we remove the tab completely for unsupported versions? Otherwise, it's basically UI leftovers that don't serve a purpose. None of the containers would have a console, yet they would all have a vestigial tab that teases one.

This is necessary, because we need to be able to connect to a different type of socket - for which support has been implemented just recently. As you said, this is very corner and rare case and it is going to be just more rare as time goes by. However, we still need to be able to tell users why it is not working. 
There has been discussion if we should  not show the tab and two major reasons for showing the tab with the message were:
1) It is less confusing for users - when they expect to have console, they are going to ask why it is not there - such as they read release notes but they don't see it there. (for example see this https://github.com/cockpit-project/cockpit/issues/11086 )
2) It is very tricky to implement nicely - there is no nice way how to find out if cockpit supports this feature already - other than trying to connect to some socket. That would mean the only reasonable way how to determine if we can do it would be to try to connect to some socket and see if we were successful.

marusak:
@martinpitt Could you please review? There is still the one where I use `echo` to write into FIFO, but I am afraid we cannot use cockpit stuff for writing into files.

marusak:
@KKoukiou I've fixed the found problem.
I also get one run of tests and F29 fails with needing a newer version of Cockpit. (https://209.132.184.41:8493/logs/pull-106-20190629-054545-d4ee4176-cockpit-fedora-29/TestApplication-testRunImage-fedora-29-127.0.0.2-2201-FAIL.png) I'll check this out why F29 image does not have newer version of Cockpit.

marusak:
Ech, so we don't build cockpit for F29 anymore? In bodhi I see the last version `cockpit-194-2.fc29`.

KKoukiou:
> Ech, so we don't build cockpit for F29 anymore? In bodhi I see the last version `cockpit-194-2.fc29`.

Right, which version and above you need?

src/ContainerTerminal.jsx
```diff
@@ -0,0 +1,187 @@
...
+    resize(width) {
+        var padding = 11 + 5 + 50;
+        var realWidth = this.state.term._core.renderer.dimensions.actualCellWidth;
+        var cols = Math.floor((width - padding) / realWidth);
+        this.state.term.resize(cols, 24);
+        cockpit.spawn(["sh", "-c", "echo '1 24 " + cols.toString() + "'>" + this.state.control_channel], { superuser: true });
+        this.setState({ cols: cols });
+    }
+
+    componentWillUpdate(nextProps, nextState) {
```
KKoukiou: See https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

This method will stop working with React 17.

marusak: Nice catch :) It was just for resizing, I moved it to `componentDidUpdate`.

marusak:
Some screenshot with the new style
consolenrewscreen

martinpitt:
.. and of course now we need a console with a dark-on-light theme! :grinning:  (SCNR) Great work! There are some failures on F30, though?

marusak:
> There are some failures on F30, though?

Hmm, on retry it did not happen. I'll try to reproduce locally but haven't seen it before.

marusak:
@KKoukiou @martinpitt I've rearranged tests a bit (to get rid of the `time.sleep(1)`) and also added some additional checks. Now it seems to behave nicely.

src/ContainerTerminal.jsx
```diff
@@ -0,0 +1,184 @@
...
+                                channel.addEventListener('close', this.onChannelClose);
+
+                                channel.send(String.fromCharCode(12)); // Send SIGWINCH to show prompt on attaching
+                                this.setState({ channel: channel, control_channel: out.sockets.control_socket, errorMessage: "" });
+                                this.resize(this.props.width);
+                            })
+                            .catch(e => {
+                                let message = cockpit.format(_("Could not open channel: $0"), e.problem);
+                                if (e.problem === "not-supported")
+                                    message = _("Upgrade to newer version of the Web Console for an interactive terminal");
```
martinpitt: Yesterday we discussed that a bit, and this may lead to confusion. E. g. in RHEL 8.1 or Fedora 29 there isn't really a newer version to upgrade to. How about "This version of the Web Console does not support a terminal."?

marusak: Right, this should not cause confusion when users try to update cockpit and won't find any newer version. Updated.

test/check-application
```diff
@@ -474,21 +472,54 @@ class TestApplication(testlib.MachineCase):
         self.assertIn('rw', rwmnt)
         self.assertIn(rwdir[4:], rwmnt)
 
+        if m.image != "fedora-29":
+            b.wait(lambda: "Hello World" in m.execute("podman logs busybox-with-tty"))
```
martinpitt: This previously worked on Fedora 29, too. Why would it stop working now?

test/check-application
```diff
@@ -474,21 +472,54 @@ class TestApplication(testlib.MachineCase):
...
+            b.click("a:contains('Console')")
```
martinpitt: Could this also be done on F-29, and check that there it shows the "not supported" message properly?

test/check-application
marusak: right, it wouldn't I just moved it and later on added condition. I'll move it back

test/check-application
marusak: Done

marusak:
Updated the tests a bit. And they went green on the first try (of course it is just luck :) )

PR #112, opened 2019-06-03
Don't include empty arguments when creating container

Gundersanne:


martinpitt:
What's the rationale for this? Does podman misbehave with empty arguments?

Gundersanne:
Not really, but the check itself is meaningless as it is now, as it'll always be true.

PR #118, opened 2019-06-11
navigateToItem shoud be a callback function calling navigateToImage

KKoukiou:


martinpitt:
@KKoukiou : FYI, fedora-30 tests are broken due to what looks like a podman regression in event handling -- it sends "channel closed" messages and disrupts event notifications.

KKoukiou:
> @KKoukiou : FYI, fedora-30 tests are broken due to what looks like a podman regression in event handling -- it sends "channel closed" messages and disrupts event notifications.

Thanks for heads up, I will take a look and open a podman bug if it's indeed regression.

PR #119, opened 2019-06-12
cockpituous: Fix srpm Release:

martinpitt:
Use release-srpm's `-V` option [1] so that the generated srpm will get a
proper changelog and Release "1" instead of "2".

[1] https://github.com/cockpit-project/cockpituous/commit/abb2bdb5


martinpitt:
Similar to https://github.com/weldr/cockpit-composer/pull/699

PR #122, opened 2019-06-13
Stop updating the whole state greedily with each container event

KKoukiou:


KKoukiou:
Locally with this PR all tests where green. Retrying to see if the [podman crash](https://209.132.184.41:8493/logs/pull-122-20190613-075252-4fb435c7-cockpit-fedora-30/TestApplication-testRunImage-fedora-30-127.0.0.2-2201-FAIL.log) is persistent here at least

PR #124, opened 2019-06-13
Sync form-layout.less with Cockpit master

martinpitt:
However, revert the change of reducing the number of supported splits
from 3 to 2, as the port and volumes lines do use 3 columns.

### DIFF
--- a/lib/form-layout.less
+++ b/lib/form-layout.less
@@ -20,7 +20,8 @@
 //
 // `ct-form-layout-split`: The grid can be split on a `form-control`
 // level by adding a this class. If you want two elements next to each
-// other, both should have this class.
+// other, both should have this class. Widths are equal by default.
+// See ct-form-layout-minmax & ct-form-layout-maxmin for alternate sizing.
 //
 // `ct-form-layout-relax`: Form elements normally stretch to take up the
 // full space. You can relax their width by adding this class to the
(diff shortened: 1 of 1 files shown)


martinpitt:
This breaks the layout right now:

podman-latest-formlayout

With current master:

podman-cur

[HTML/CSS/other files dump](https://github.com/cockpit-project/cockpit-podman/files/3285304/podman-create.tar.gz)

garrett:
A guess: https://github.com/cockpit-project/cockpit-podman/pull/124/files#diff-6729bfa75425a0393061b46592c7bddcL166:

```diff
-    // Only support 3 splits for now (can change to 4 later, if needed)
-    grid-template-columns: repeat(3, auto);
+    // Only support 2 splits for now (can change to 3 later, if needed)
+    grid-template-columns: repeat(2, auto);
```

I don't remember why I bumped it down (perhaps there was extra space on the side due to a gap?). I guess I assumed we were only using 2. See what happens when you change it back to 3.

martinpitt:
@garrett: Thanks! :heart: , this works perfectly! Want me to send a PR against cockpit with that?

garrett:
I probably changed it for a reason. I would suggest using repeat(auto-fit, …) in this case, but I think we'd have to use minmax() in the repeat and couldn't use auto for the width. (I wonder if that would work.)

Alternative approach is to support 2 by default and have a class for 3. But that's a bit of kludge.

garrett:
Another solution is to group the ports + dropdown elements together.

martinpitt:
@KKoukiou: [test failure](https://209.132.184.41:8493/logs/pull-124-20190613-095959-f38e6d18-cockpit-fedora-29/log), looks like fallout from #122 perhaps ? retrying

garrett:
Perhaps we should change the # into a variable here too, so we could either have a class that does it in ct-forms, or it can be adjusted in local CSS (and adapt accordingly).

Something like:
```
grid-template-columns: repeat(--ct-forms-layout-split-number, auto);
```

...but in the future. Having a slight fork is OK for now.

PR #127, opened 2019-06-14
Drop patternfly npm module

martinpitt:
Our pages include cockpit's shipped patternfly.min.css, and our build
system does not do anything with this NPM module, so save a few seconds
during `npm install`.


martinpitt:
Nevermind, bogus. lib/variables.less does include it.

PR #130, opened 2019-06-17
Fix Commit operation and allow commiting running containers

KKoukiou:


KKoukiou:
@larskarlitski according to baude we need to pass the  `more` parameter now in the commit' container API call. See issue in commit message for the issue link. Do you think this makes sense from API design POV? I don't really get the concept of calling some APIs with `more` parameter but without this change the commit operation from the UI is broken.

src/ContainerCommitModal.jsx
```diff
@@ -91,7 +91,8 @@ class ContainerCommitModal extends React.Component {
         }
         commitData.changes.push(...onbuildsArr);
 
-        varlink.call(utils.PODMAN_ADDRESS, "io.podman.Commit", commitData)
+        const more = true;
```
martinpitt: Why does this need to be a constant? Just to avoid a bare "true" argument and make it more legible? Or is there some hidden magic that I don't understand?

KKoukiou: Just to avoid bare 'true' that noone has any idea what is represents.

martinpitt:
Please ignore the failed rhel-8-1 test for now. This is fixed in https://github.com/cockpit-project/cockpit-podman/pull/131

PR #131, opened 2019-06-17
spec: Drop epoch from dependency

martinpitt:
This doesn't work with our current approach of building the srpm on the
container host, and it's also unnecessary now: All our supported target
OSes have a recent enough podman.

 - [x] Get fixed podman on rhel-8-1 image: https://github.com/cockpit-project/cockpit/issues/12078

### DIFF
--- a/cockpit-podman.spec.in
+++ b/cockpit-podman.spec.in
@@ -1,8 +1,3 @@
-# we generally want CentOS packages to be like RHEL; special cases need to check %{centos} explicitly
-%if 0%{?centos}
-%define rhel %{centos}
-%endif
-
 Name:           cockpit-podman
 Version:        @VERSION@
 Release:        1%{?dist}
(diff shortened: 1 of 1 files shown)


martinpitt:
@KKoukiou : Right, we need to rework this to build the rpm in the VM instead of the host. I'll do that, but honestly I'm a hurry today to get CI working on 8.1 at all (and then get a release out).

martinpitt:
The failing `podman.service` is fixed in 1.4.1, I triggered a rhel-8-1 image refresh to pick this up.

martinpitt:
The rebuilt rhel-8-1 image landed now, retriggering the test.

martinpitt:
This time the [failure](https://209.132.184.41:8493/logs/pull-131-20190618-124949-be710619-cockpit-rhel-8-1/log) looks like a flake, retrying.

martinpitt:
Included in https://github.com/cockpit-project/cockpit-podman/pull/132

PR #148, opened 2019-07-17
Pf4ish fix

garrett:
This addresses _most_ of the PF4ish related issues with Cockpit-Podman.

Known outstanding issues:

- [x] List selection colors are wrong (in the search dialog).
- [ ] + buttons are not uniformly sized

Please test this out and add more issues you might find.

### DIFF
--- a/src/ContainerHeader.jsx
+++ b/src/ContainerHeader.jsx
@@ -32,7 +32,7 @@ class ContainerHeader extends React.Component {
 
     render() {
         return (
-            <div className="content-filter" >
+            <React.Fragment>
                 <Select.Select id="containers-containers-filter" initial={this.state.filter} onChange={this.handleFilterChange}>
                     <Select.SelectEntry data='all'>{_("Everything")}</Select.SelectEntry>
                     <Select.SelectEntry data='running'>{_("Images and running containers")}</Select.SelectEntry>
@@ -43,7 +43,7 @@ class ContainerHeader extends React.Component {
                        className="form-control"
                        placeholder={_("Type to filter…")}
                        onChange={this.handleFilterTextChange} />
-            </div>
+            </React.Fragment>
         );
     }
 }
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -1,3 +1,9 @@
+/* Collapse top padding with additional container fluids */
+/* FIXME: This should go into PF overrides in Cockpit. */
+/* Remove it here when it lands there. */
+.container-fluid + .container-fluid {
+    padding-top: 0;
+}
 
 .vulnerability-row-ct-docker {
     line-height: 2.5em;
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -610,3 +610,12 @@ table.drive-list td:nth-child(2) img {
 .form-list-control input {
     width: 125px !important;
 }
+
+/* We need to be a little specific here to override without !important */
+.ct-form-layout .ct-form-layout .run-image-dialog-actions {
+    grid-template-columns: repeat(2, min-content);
+}
+
+.run-image-dialog-actions .btn {
+    padding: 0.25rem 0.75rem;
+}
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -308,15 +308,15 @@
     min-height: 4rem;
 }
 
-.content-filter input[type="text"] {
-    display: inline-block;
-    width: 200px;
-    margin-left: 6px;
-    margin-top: 2px;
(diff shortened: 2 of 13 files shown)


garrett:
It would be nice to get this ready to land by sometime tomorrow.

After this lands, we should do a release of Cockpit-Podman so that people using it can use it properly again. (The state Cockpit-Podman is in is mostly a cosmetic mess that looks bad, but it _also_ currently makes the list selection impossible to read. Oops.)

src/ContainerHeader.jsx
```diff
@@ -32,7 +32,7 @@ class ContainerHeader extends React.Component {
 
     render() {
         return (
-            <div className="content-filter" >
+            <React.Fragment>
```
martinpitt: This drops the content-filter class, but the next commit changes the CSS for this class. One of these two is inconsistent?

src/podman.scss
```diff
@@ -308,15 +308,15 @@
     min-height: 4rem;
 }
 
-.content-filter input[type="text"] {
-    display: inline-block;
-    width: 200px;
-    margin-left: 6px;
-    margin-top: 2px;
+.content-filter .ct-select {
```
martinpitt: This is the CSS place that updates the .content-filter class, but the previous commit removes it.

src/podman.scss
```diff
@@ -1,3 +1,9 @@
+/* Collapse top padding with additional container fluids */
+/* FIXME: This should go into PF overrides in Cockpit. */
+/* Remove it here when it lands there. */
+.container-fluid + .container-fluid {
```
martinpitt: I don't understand the + here. After the change above, there are no nested container-fluid elements any more, right? Just the two from app.jsx:
```
src/app.jsx:                <div key={"containerslists"} className="container-fluid">
src/app.jsx:                <div key={"imageslists"} className="container-fluid">
```

src/ContainerHeader.jsx
garrett: There were two content-filter divs. I removed the inner one.

src/podman.scss
garrett: Yep, this styles the remaining content-filter, instead of _both_. :wink:

src/podman.scss
garrett: This means style immediate sibling elements.

So if there are two `container-fluid`s in a row, remove the padding at the top of the second one, making the spacing between them the same as before and after (instead of doubled-up with both the bottom of the first and the top of the second).

garrett:
I should probably add another commit or two tomorrow for the button sizing... and then we have to decide how much of this gets squashed down. (We can add a closes # at that time too. I didn't want to rebase + squash + force-push while the PR is being reviewed if I don't have to.)

src/podman.scss
garrett: There were some other duplicate-nested issues elsewhere, and I removed those as well.

garrett: Specifically, https://github.com/cockpit-project/cockpit-podman/pull/148/commits/3a520f0a0dc33f41639b3f7ff5303e5f9e5a5e43 and https://github.com/cockpit-project/cockpit-podman/pull/148/commits/343c917fa06baf3b0bdf4dc5b7910784bb0ff8e4 (I kept container-fluid in the parents of each).

garrett:
I fine-tuned the layout for the dialog to search for a new image and fixed the button size in the run container.

I think this is now ready for review.

martinpitt:
With a PF4 cockpit this looks great now, thank you!

Unfortunately that breaks quite a bit when running with an older cockpit that still has PF3, such as in current RHEL 8.1:

podman-pf3

There are no table frames any more, everything is too cramped, and the "images and running containers" button looks just like  plain text.

So either we need to find a way to keep this a bit more backwards compatible, or we must avoid packaging the next release into RHEL 8.1 (and branch instead if necessary).

martinpitt:
I force-pushed to adjust the tests (changed three commits) and add a Closes #.

martinpitt:
[testRunImage fails on f-29](https://209.132.184.41:8493/logs/pull-148-20190718-151717-1d266659-cockpit-fedora-29/log) only. Downloading the image to check..

garrett:
> need to find a way to keep this a bit more backwards compatible

If it's just that the lines are gone, it sounds like it isn't picking up a CSS variable. We can redefine the missing ones locally. I _think_ we could even have it fall back if they're already defined (although it's probably not necessary).

PR #157, opened 2019-07-30
test: Adjust to changed Cockpit Services page

martinpitt:
https://github.com/cockpit-project/cockpit/pull/12265 changed the design
a lot, so that `#service-unit` does not exist any more.

The tests need to get along with the old and new design, so just relax
the check.


martinpitt:
Argh, need to support the old page layout as well.

PR #162, opened 2019-08-05
Show loading state while getting containers/images

marusak:
Getting containers takes a bit (even a few seconds for me). While it gets loaded it shows `No containers` which is confusing since they are not yet loaded. With this PR it shows firstly `Loading containers...`.

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -202,13 +202,17 @@ class Containers extends React.Component {
         const columnTitles = [_("Name"), _("Image"), _("Command"), _("CPU"), _("Memory"), _("State")];
 
         let emptyCaption = _("No containers");
-        if (this.props.textFilter.length > 0)
+        if (this.props.containers === null)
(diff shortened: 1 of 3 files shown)


src/Containers.jsx
```diff
@@ -202,13 +202,17 @@ class Containers extends React.Component {
         const columnTitles = [_("Name"), _("Image"), _("Command"), _("CPU"), _("Memory"), _("State")];
 
         let emptyCaption = _("No containers");
-        if (this.props.textFilter.length > 0)
+        if (this.props.containers === null)
+            emptyCaption = _("Loading containers...");
```
martinpitt: As the tiniest nitpick: Could we just have "Loading..." to make translation easier, and be more flexible in the future about other spinners without breaking i18n?

marusak:
> Could we just have "Loading..."

Sure, updated. I also would add another loadings (Loading user containers, etc.), so this makes sense :)

PR #171, opened 2019-08-06
test: Allow tests to run on existing testbed

martinpitt:
Can be tested with
```
make vm
bots/vm-run fedora-30
test/check-application -tv --machine 127.0.0.2:2201  --browser 127.0.0.2:9091
```


martinpitt:
I have a much more generic and robust approach now, which also works with the "all tests on a single testbed" scenario.

martinpitt:
Added a hack to get along with podman leaking mounts.

PR #173, opened 2019-08-08
Show user owned containers and images

### LINKED ISSUE #83: show and manage rootless containers
When I logging in using my regular user account with super priviledges (wheel group) I cannot see or manage container images/instances of that user account.

I can see and manage images/instances of containers owned by root, but not my own.

### LINKED ISSUE #84: Display owner
When logging in as a regular user with privileges (wheel group), I can see the root user's images and running instances, but it does not show who the owner of said images/instances are.

With rootless containers being a popular topic, it seems logical that there could be many different images and instances on the system owned by different users.  And thus, seems logical that the display should show the owner of each to avoid confusion.

marusak:
This implements showing of user owned containers. Can be tried with [this](https://mmarusak.fedorapeople.org/podman-1.4.5-1564481722.gita4928fff.fc30.x86_64.rpm) podman build.
Fixes #84
Fixes #83

Blocked on:
- [x] https://github.com/containers/libpod/pull/3662
- [x] https://github.com/containers/libpod/issues/3749

podmanuserstuff

TODOs:
- [x] Tests
- [x] Verify tests on F31
- [x] Add Fixes and closes lines

Follow-ups:
- [ ] Switching (filtering) between user, root and all containers
- [x] Show selection of root/user in image download only when privileged.
- [ ] Sort images and containers - first user(root?) and then root(user?). Now the order is rather random (depending which service replies quicker) and when some changes happen while cockpit is opened new or updated containers/images would end up on the bottom of the list (so mixing it up).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,10 @@ rpm: dist-gzip $(RPM_NAME).spec
 # build a VM with locally built rpm installed
 $(VM_IMAGE): rpm bots
 	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
+	# HACK: https://bugzilla.redhat.com/show_bug.cgi?id=1762663
+	if [ $(TEST_OS) = "fedora-31" ]; then bots/image-customize -v -r 'grub2-editenv - set "$$(grub2-editenv list | grep ^kernelopts) debug"' $(TEST_OS); fi
 	bots/image-customize -v -i cockpit-ws -i `pwd`/$(RPM_NAME)-*.noarch.rpm -s $(CURDIR)/test/vm.install $(TEST_OS)
+	if [ $(TEST_OS) = "fedora-31" ]; then bots/image-customize -v -r 'grub2-editenv - set "$$(grub2-editenv list | grep ^kernelopts | sed "s/debug//")"' $(TEST_OS); fi
 
 # convenience target for the above
 vm: $(VM_IMAGE)
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -65,7 +65,7 @@ class Containers extends React.Component {
 
         if (force)
             args.timeout = 0;
-        utils.podmanCall("StopContainer", args)
+        utils.podmanCall("StopContainer", args, container.isSystem)
                 .catch(ex => this.setState({
                     actionError: cockpit.format(_("Failed to stop container $0"), container.names),
                     actionErrorDetail: ex.parameters && ex.parameters.reason
@@ -73,7 +73,7 @@ class Containers extends React.Component {
     }
 
     startContainer(container) {
-        utils.podmanCall("StartContainer", { name: container.names })
+        utils.podmanCall("StartContainer", { name: container.names }, container.isSystem)
                 .catch(ex => this.setState({
                     actionError: cockpit.format(_("Failed to start container $0"), container.names),
                     actionErrorDetail: ex.parameters && ex.parameters.reason
@@ -85,7 +85,7 @@ class Containers extends React.Component {
 
         if (force)
             args.timeout = 0;
-        utils.podmanCall("RestartContainer", args)
+        utils.podmanCall("RestartContainer", args, container.isSystem)
                 .catch(ex => this.setState({
                     actionError: cockpit.format(_("Failed to restart container $0"), container.names),
                     actionErrorDetail: ex.parameters && ex.parameters.reason
@@ -93,16 +93,23 @@ class Containers extends React.Component {
     }
 
     renderRow(containersStats, container) {
-        const containerStats = containersStats[container.id];
+        const containerStats = containersStats[container.id + container.isSystem.toString()];
         const isRunning = container.status == "running";
         const image = container.image;
 
+        let proc = "";
+        let mem = "";
+        if (containerStats) {
+            proc = containerStats.cpu ? utils.format_cpu_percent(containerStats.cpu * 100) : <abbr title={_("not available")}>{_("n/a")}</abbr>;
+            mem = containerStats.mem_usage ? utils.format_memory_and_limit(containerStats.mem_usage, containerStats.mem_limit) : <abbr title={_("not available")}>{_("n/a")}</abbr>;
+        }
         let columns = [
             { name: container.names, header: true },
             image,
             utils.quote_cmdline(container.command),
-            isRunning ? utils.format_cpu_percent(containerStats.cpu * 100) : "",
-            containerStats ? utils.format_memory_and_limit(containerStats.mem_usage, containerStats.mem_limit) : "",
+            proc,
+            mem,
+            container.isSystem ? _("system") : this.props.user,
             container.status /* TODO: i18n */,
         ];
         let tabs = [{
@@ -112,7 +119,7 @@ class Containers extends React.Component {
         }, {
             name: _("Console"),
             renderer: ContainerTerminal,
-            data: { containerId: container.id, containerStatus: container.status, width:this.state.width }
+            data: { containerId: container.id, containerStatus: container.status, width:this.state.width, system:container.isSystem }
         }];
 
         var actions = [
@@ -152,7 +159,8 @@ class Containers extends React.Component {
         return (
             <ScrollableAnchor id={container.id} key={container.id}>
                 <Listing.ListingRow
-                        rowId={container.id}
+                        key={container.id + container.isSystem.toString()}
+                        rowId={container.id + container.isSystem.toString()}
                         columns={columns}
                         tabRenderers={tabs}
                         listingActions={actions}
@@ -172,7 +180,7 @@ class Containers extends React.Component {
         this.setState({
             selectContainerDeleteModal: false
         });
-        utils.podmanCall("RemoveContainer", { name: id })
+        utils.podmanCall("RemoveContainer", { name: id }, this.state.containerWillDelete.isSystem)
                 .catch(ex => console.error("Failed to do RemoveContainer call:", JSON.stringify(ex)));
     }
 
@@ -185,7 +193,7 @@ class Containers extends React.Component {
     // TODO: force
     handleForceRemoveContainer() {
         const id = this.state.containerWillDelete ? this.state.containerWillDelete.id : "";
-        utils.podmanCall("RemoveContainer", { name: id, force: true })
+        utils.podmanCall("RemoveContainer", { name: id, force: true }, this.state.containerWillDelete.isSystem)
                 .then(reply => {
                     this.setState({
                         setContainerRemoveErrorModal: false
@@ -201,7 +209,7 @@ class Containers extends React.Component {
     }
 
     render() {
-        const columnTitles = [_("Name"), _("Image"), _("Command"), _("CPU"), _("Memory"), _("State")];
+        const columnTitles = [_("Name"), _("Image"), _("Command"), _("CPU"), _("Memory"), _("Owner"), _("State")];
 
         let emptyCaption = _("No containers");
         if (this.props.containers === null)
--- a/src/ImageSearchModal.jsx
+++ b/src/ImageSearchModal.jsx
@@ -19,12 +19,14 @@ export class ImageSearchModal extends React.Component {
             imageList: [],
             searchInProgress: false,
             searchFinished: false,
+            isSystem: props.systemServiceAvailable,
         };
         this.onDownloadClicked = this.onDownloadClicked.bind(this);
         this.onItemSelected = this.onItemSelected.bind(this);
         this.onSearchTriggered = this.onSearchTriggered.bind(this);
         this.onValueChanged = this.onValueChanged.bind(this);
         this.onKeyPress = this.onKeyPress.bind(this);
+        this.onToggleUser = this.onToggleUser.bind(this);
     }
 
     componentDidMount() {
@@ -38,11 +40,15 @@ export class ImageSearchModal extends React.Component {
             this.activeConnection.close();
     }
 
+    onToggleUser(ev) {
+        this.setState({ isSystem: ev.target.id === "system" });
+    }
+
     onDownloadClicked() {
         let selectedImageName = this.state.imageList[this.state.selected].name;
 
         this.props.close();
-        this.props.downloadImage(selectedImageName, this.state.imageTag);
+        this.props.downloadImage(selectedImageName, this.state.imageTag, this.state.isSystem);
     }
 
     onItemSelected(key) {
@@ -63,7 +69,7 @@ export class ImageSearchModal extends React.Component {
 
         this.setState({ searchInProgress: true });
 
-        varlink.connect(utils.PODMAN_SYSTEM_ADDRESS)
+        varlink.connect(utils.getAddress(this.state.isSystem), this.state.isSystem)
                 .then(connection => {
                     this.activeConnection = connection;
 
@@ -110,6 +116,17 @@ export class ImageSearchModal extends React.Component {
     render() {
         let defaultBody = (
             <React.Fragment>
+                { this.props.userServiceAvailable && this.props.systemServiceAvailable &&
+                    <form className="ct-form">
+                        <label className="control-label" htmlFor="as-user">{_("Download as:")}</label>
+                        <fieldset id="as-user">
+                            <input type="radio" value="system" id="system" onChange={this.onToggleUser} checked={this.state.isSystem} />
+                            <label className="radio" htmlFor="system">{_("System")}</label>
+                            <input type="radio" value="user" id="user" onChange={this.onToggleUser} checked={!this.state.isSystem} />
+                            <label className="radio" htmlFor="user">{this.props.user.name}</label>
+                        </fieldset>
+                    </form>
+                }
                 <div className="input-group">
                     <span className="input-group-addon">
                         <span className="fa fa-search" />
--- a/src/ImageUsedBy.jsx
+++ b/src/ImageUsedBy.jsx
@@ -7,12 +7,18 @@ const _ = cockpit.gettext;
 
 const renderRow = (containerStats, container, showAll) => {
     const isRunning = container.status == "running";
+    let proc = "";
+    let mem = "";
+    if (containerStats) {
+        proc = containerStats.cpu ? utils.format_cpu_percent(containerStats.cpu * 100) : <abbr title={_("not available")}>{_("n/a")}</abbr>;
+        mem = containerStats.mem_usage ? utils.format_memory_and_limit(containerStats.mem_usage, containerStats.mem_limit) : <abbr title={_("not available")}>{_("n/a")}</abbr>;
+    }
 
     const columns = [
         { name: container.names, header: true },
         utils.quote_cmdline(container.command),
-        isRunning ? utils.format_cpu_percent(containerStats.cpu * 100) : "",
-        containerStats ? utils.format_memory_and_limit(containerStats.mem_usage, containerStats.mem_limit) : "",
+        proc,
+        mem,
         container.status /* TODO: i18n */,
 
     ];
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -27,32 +27,42 @@ import Images from './Images.jsx';
 import * as utils from './util.js';
 
 const _ = cockpit.gettext;
+const permission = cockpit.permission({ admin: true });
 
 class Application extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
-            serviceAvailable: null,
+            systemServiceAvailable: null,
+            userServiceAvailable: null,
             enableService: true,
-            images: null, /* images[Id]: detail info of image with Id from InspectImage */
-            containers: null, /* containers[Id] detail info of container with Id from InspectContainer */
-            containersStats:{}, /* containersStats[Id] memory usage of running container with Id */
+            images: null,
+            userImagesLoaded: false,
+            systemImagesLoaded: false,
+            containers: null,
+            containersStats: {},
+            userContainersLoaded: null,
+            systemContainersLoaded: null,
+            userServiceExists: false,
             onlyShowRunning: true,
             textFilter: "",
             dropDownValue: 'Everything',
             notifications: [],
         };
         this.onAddNotification = this.onAddNotification.bind(this);
+        this.updateState = this.updateState.bind(this);
         this.onDismissNotification = this.onDismissNotification.bind(this);
         this.onChange = this.onChange.bind(this);
         this.onFilterChanged = this.onFilterChanged.bind(this);
         this.updateImagesAfterEvent = this.updateImagesAfterEvent.bind(this);
         this.updateContainerAfterEvent = this.updateContainerAfterEvent.bind(this);
+        this.updateContainerStats = this.updateContainerStats.bind(this);
         this.startService = this.startService.bind(this);
         this.showAll = this.showAll.bind(this);
         this.goToServicePage = this.goToServicePage.bind(this);
         this.handleImageEvent = this.handleImageEvent.bind(this);
         this.handleContainerEvent = this.handleContainerEvent.bind(this);
+        this.checkUserService = this.checkUserService.bind(this);
     }
 
     onAddNotification(notification) {
@@ -88,24 +98,79 @@ class Application extends React.Component {
         });
     }
 
-    updateContainersAfterEvent() {
-        utils.updateContainers()
-                .then((reply) => {
-                    this.setState({
-                        containers: reply.newContainers,
-                        containersStats: reply.newContainersStats
-                    });
+    updateState(state, id, newValue) {
+        this.setState(prevState => {
+            let copyState = Object.assign({}, prevState[state]);
+
+            copyState[id] = newValue;
+
+            return {
+                [state]: copyState,
+            };
+        });
+    }
+
+    updateContainerStats(id, system) {
+        utils.podmanCall("GetContainerStats", { name: id }, system)
+                .then(reply => {
+                    this.updateState("containersStats", reply.container.id + system.toString(), reply.container);
                 })
                 .catch(ex => {
-                    console.warn("Failed to do Update Container:", JSON.stringify(ex));
+                    if (ex.error === "io.podman.ErrRequiresCgroupsV2ForRootless") {
+                        console.log("This OS does not support CgroupsV2. Some information may be missing.");
+                        this.updateState("containersStats", id + system.toString(), -1);
+                    } else
+                        console.warn("Failed to update container stats:", JSON.stringify(ex));
                 });
     }
 
-    updateImagesAfterEvent() {
-        utils.updateImages()
+    updateContainersAfterEvent(system) {
+        utils.podmanCall("ListContainers", {}, system)
                 .then(reply => {
-                    this.setState({
-                        images: reply
+                    this.setState(prevState => {
+                        // Copy only containers that could not be deleted with this event
+                        // So when event from system come, only copy user containers and vice versa
+                        let copyContainers = {};
+                        Object.entries(prevState.containers || {}).forEach(([id, container]) => {
+                            if (container.isSystem !== system)
+                                copyContainers[id] = container;
+                        });
+                        for (let container of reply.containers || []) {
+                            container.isSystem = system;
+                            copyContainers[container.id + system.toString()] = container;
+                            if (container.status === "running")
+                                this.updateContainerStats(container.id, system);
+                        }
+
+                        return {
+                            containers: copyContainers,
+                            [system ? "systemContainersLoaded" : "userContainersLoaded"]: true,
+                        };
+                    });
+                })
+                .catch(e => console.log(e));
+    }
+
+    updateImagesAfterEvent(system) {
+        utils.updateImages(system)
+                .then(reply => {
+                    this.setState(prevState => {
+                        // Copy only images that could not be deleted with this event
+                        // So when event from system come, only copy user images and vice versa
+                        let copyImages = {};
+                        Object.entries(prevState.images || {}).forEach(([id, image]) => {
+                            if (image.isSystem !== system)
+                                copyImages[id] = image;
+                        });
+                        Object.entries(reply).forEach(([id, image]) => {
+                            image.isSystem = system;
+                            copyImages[id + system.toString()] = image;
+                        });
+
+                        return {
+                            images: copyImages,
+                            [system ? "systemImagesLoaded" : "userImagesLoaded"]: true
+                        };
                     });
                 })
                 .catch(ex => {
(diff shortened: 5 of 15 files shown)


marusak:
The user service landed. We can continue with this (while we wait for release).
The problem is that we cannot show statistics (CPU and Memory of running containers). This needs CGroupsV2 (which are planed for Fedora 31, I have kinda doubts about rhel-8). Are we going to wait or do we want to work around this somehow? For example we can just show there some placeholder (to show that it is running but we don't have this information).
What do you think @martinpitt @garrett ?

martinpitt:
@marusak: I wouldn't block on that. We need to be able to run on Fedora 30, RHEL 8.1, or even custom kernels without cgroupsv2. So some placeholder, or simply not showing the graphs at all, seems prudent.

garrett:
I agree with @martinpitt. The cpu/memory graphs shouldn't be mixed into this PR. Also, don't show placeholders.

marusak:
Now I am reading comment from @martinpitt and I realized I misread it the first time. So I guess I should finish it up.

> The cpu/memory graphs shouldn't be mixed into this PR.

What do you mean by that? I am afraid I do not follow.

>  Also, don't show placeholders.

I just thought that we should show some information and not just leave those fields blank, as blank are for containers that are not running.

garrett:
The conversation is basically as follows:

>>> The problem is that we cannot show statistics (CPU and Memory of running containers). This needs CGroupsV2 (which are planed for Fedora 31, I have kinda doubts about rhel-8). Are we going to wait or do we want to work around this somehow?
>>
>> @martinpitt: I wouldn't block on that. 
>
> @garrett: I agree. The cpu/memory graphs shouldn't be mixed into this PR. Also, don't show placeholders.

Basically: Don't worry about the graphs. (Don't add in a placeholder, either.) Just ignore it for now, please.

marusak:
> Just ignore it for now, please.

Just to make sure, I meant these stats:
62718373-005a5180-ba06-11e9-8ebf-8df7f4876d27

Running image would have there some values, but we cannot get those for user containers, only for root containers. It may seem a bit empty and confusing for user.

garrett:
@marusak Oh, I thought you meant this (from the Docker page):

Screenshot_2019-09-03 Docker Containers - Rain

Whoops!

I guess if there are only user containers and no system containers, don't show those rows? 

If it makes the code too spaghetti-ish; we could have something like `has-system-containers` `has-user-containers` as classes on the list, then do something like the following CSS:

```css
/* Don't show CPU and Memory columns when there are only user containers */
.has-user-containers:not(.has-system-containers) .container-cpu,
.has-user-containers:not(.has-system-containers) .container-memory {
  display: none;
}

**Comment by garrett** [comment]
I get a bunch of blank named containers that I don't see when using podman from the command line...

Screenshot_2019-09-03 Podman Containers - Rain

**Comment by martinpitt** [comment]
The tests failed due to the internal mirror thing, that should be fixed now. Retrying.

**Comment by martinpitt** [comment]
Nice work! Initial testing:

 * I do see my system and user container images.
 * I also see my user container image (I don't have any system one). The nitpick here is that the "State" column is too narrow, as the command gets too much width. That might be unrelated to this PR, though:

podman-user
 * Trying to get a console of my toolbox doesn't work: It immediately says "disconnected". This could be because I already had a `toolbox enter` (aka `podman exec shell`), and possibly there can only be one of these? I tried to exit my podman shell.
 * After that, there is no "Reconnect" button, so I tried to reload the page. But that somehow upset the podman varlink process, it spun at 100% CPU and the container list was stuck at "Loading" forever. (This is not cockpit-podman's fault, but we should report it)
 * After some killing and reloading, I finally get the container view again. But console of my toolbox container still doesn't work. Possibly that's because that container has no tty enabled? When starting a new one, there's a "With terminal" checkbox, so whatever that controls might not be on for toolbox. Can we detect that and not show the Console tab in that case? (This may also not be related to this PR though, and affect system containers as well)
 * I started a new container from cockpit/unit-tests, with terminal. For that the Console works.
 * I force-deleted the container from the previous step. It went away from `podman ps`, but not the c-podman "Containers" list. Seems we are missing an event there? Reloading the page once again makes the podman varlink process freak out, and the page stuck at "Loading..".

So, ignoring all the podman varlink process bugs, and the unrelated Console unavailability, I think the only issue that's really related to *this* PR is the State column width.

**Comment by martinpitt** [comment]
It may be worth re-testing this with a current podman (the F30/F31release is a month old). Not sure how well that can be tested from a git build.

File: src/util.js [hunk 1]
```diff
@@ -58,14 +58,18 @@ function handleVarlinkCallError(ex) {
     console.warn("Failed to do varlinkcall:", JSON.stringify(ex));
 }
 
+export function podmanAction(name, args) {
```
martinpitt: Tiny nitpick: `podmanCall()`?

src/util.js
```diff
@@ -62,6 +62,11 @@ export function podmanAction(name, args) {
     return varlink.call(PODMAN_ADDRESS, "io.podman." + name, args);
 }
 
+export function monitor(name, args, callback) {
+    return varlink.connect(PODMAN_ADDRESS)
+            .then(connection => { return connection.monitor("io.podman." + name, args, callback) });
```
martinpitt: Nitpick: Drop the `{ return }` wrapper.

src/util.js
```diff
@@ -3,7 +3,7 @@ import varlink from './varlink.js';
 
 const _ = cockpit.gettext;
 
-export const PODMAN_ADDRESS = "unix:/run/podman/io.podman";
+export const PODMAN_ROOT_ADDRESS = "unix:/run/podman/io.podman";
```
martinpitt: Bikeshedding: This is a bit confusing, as "root" is also a user. How about `PODMAN_SYSTEM_ADDRESS`? system and user is the dichtonomy that systemd uses; libvirt has "system" and "session", so at least vaguely the same.

src/util.js
```diff
@@ -69,12 +69,18 @@ export function podmanAction(name, args, root) {
     return varlink.call(address, "io.podman." + name, args, root);
 }
 
-export function monitor(name, args, callback, root) {
+export function monitor(name, args, callback, error_callback, root) {
```
martinpitt: Let's please call this `on_close`, as that's what it does -- it does not receive general errors. Please also add a comment what argument it receives (`root`), and also what `root` actually means.

src/Containers.jsx
```diff
@@ -85,24 +85,31 @@ class Containers extends React.Component {
...
     renderRow(containersStats, container) {
-        const containerStats = containersStats[container.id];
+        const containerStats = containersStats[container.id + container.isRoot.toString()];
         const isRunning = container.status == "running";
         const image = container.image;
 
+        let proc = "";
+        let mem = "";
+        if (containerStats) {
+            proc = containerStats.cpu ? utils.format_cpu_percent(containerStats.cpu * 100) : <abbr title="not available">n/a</abbr>;
```
martinpitt: Title and n/a need i18n.

src/ImageSearchModal.jsx
```diff
@@ -38,11 +40,15 @@ export class ImageSearchModal extends React.Component {
             this.activeConnection.close();
     }
 
+    onToggleUser() {
+        this.setState({ isRoot: !this.state.isRoot });
```
martinpitt: While this seems to work, it feels kind of brittle -- wouldn't it be better to get the `event` and check its target.checked (or so) as to whether it's actually on?

src/ImageSearchModal.jsx
```diff
@@ -110,6 +116,15 @@ export class ImageSearchModal extends React.Component {
     render() {
         let defaultBody = (
             <React.Fragment>
+                <form className="ct-form">
+                    <label className="control-label" htmlFor="as-user">{_("Download as:")}</label>
+                    <fieldset id="as-user">
+                        <input type="radio" value="user" id="user" onChange={this.onToggleUser} checked={!this.state.isRoot} />
+                        <label className="radio" htmlFor="user">{_("user")}</label>
+                        <input type="radio" value="root" id="root" onChange={this.onToggleUser} checked={this.state.isRoot} />
+                        <label className="radio" htmlFor="root">{_("root")}</label>
```
martinpitt: Same "root" vs. "System" thing here. That may need @garrett 's opinion. Also, it feels like these words  should be capitalized.

src/ImageSearchModal.jsx
```diff
@@ -110,6 +116,15 @@ export class ImageSearchModal extends React.Component {
     render() {
         let defaultBody = (
             <React.Fragment>
+                <form className="ct-form">
```
martinpitt: This needs to check whether root is actually available. If I log in without superuser, I still see these radio buttons. The search then spins forever

src/ImageUsedBy.jsx
```diff
@@ -7,12 +7,18 @@ const _ = cockpit.gettext;
 
 const renderRow = (containerStats, container, showAll) => {
     const isRunning = container.status == "running";
+    let proc = "";
+    let mem = "";
+    if (containerStats) {
+        proc = containerStats.cpu ? utils.format_cpu_percent(containerStats.cpu * 100) : <abbr title="not available">n/a</abbr>;
```
martinpitt: i18n again

marusak:
> The nitpick here is that the "State" column is too narrow, as the command gets too much width.

Good point. This is most likely not related, it is now just more obvious. I'll try to style it.

> Trying to get a console of my toolbox doesn't work: It immediately says "disconnected"

Can you `podman attach <container>`? That is exactly what it is doing.

> Possibly that's because that container has no tty enabled? 

This works for me. I think the problem might be the executable of your toolbox container.

> It went away from podman ps, but not the c-podman "Containers" list. Seems we are missing an event there?

I think that podman was just acting up. Normally it works.

> Reloading the page once again makes the podman varlink process freak out

What version of podman do you use? I think I haven't seen such problems so often.

Thank you Martin for the review and comments, I'll look into those:)

src/ImageSearchModal.jsx
marusak: I am aware of this, it is in the `followups list`. But I may implement it here as well

martinpitt:
> > Trying to get a console of my toolbox doesn't work: It immediately says "disconnected"
> 
> Can you `podman attach <container>`? That is exactly what it is doing.

`podman attach devel` just hangs without showing any error. I tried to type `ls` but it doesn't do anything. So this is broken at the podman level then.

> What version of podman do you use? I think I haven't seen such problems so often.

podman-1.5.1-3.fc30.x86_64

src/ImageSearchModal.jsx
marusak: Fixed.

src/ImageSearchModal.jsx
marusak: I went for 'System' and real user name. So in my case `mmarusak` as it is the same as what is shown later on as `owner`.

src/util.js
marusak: Changed to `PODMAN_SYSTEM_ADDRESS`

src/app.jsx
```diff
@@ -196,59 +253,78 @@ class Application extends React.Component {
...
     }
 
     componentDidMount() {
-        this.init();
+        this.init(true);
+        this.init(false);
+    }
+
+    checkUserService() {
+        let argv = ["systemctl", "--user", "status", "io.podman.socket"];
```
martinpitt: This now does a lot of systemctl CLI calls, which are relatively expensive and also don't provide monitoring (i. e. send a signal when a service becomes (un)available. Time to copy cockpit's pkg/lib/service.js?

I'm ok with doing a conversion to D-Bus calls in a follow-up, though.

However, `status` is  really a no-go here. That will query the journal, which can take several seconds, and is entirely unnecessary here. Please use `is-active --quiet` instead.

src/app.jsx
```diff
@@ -196,59 +253,78 @@ class Application extends React.Component {
...
+
+        cockpit.spawn(argv)
+                .then(() => this.setState({ userServiceExists: true }))
+                .catch(err => {
+                    if (err.exit_status === 4) // systemctl status returns 3 when service is not running
+                        this.setState({ userServiceExists: false });
+                    else
+                        this.setState({ userServiceExists: true });
```
martinpitt: It seems a little bold to assume that any code except 4 means success. E. g. in the "not running" case, setting it to true still means that you will try and talk to it. It seems a bit more robust to check if the command succeeds, and assume "not available" otherwise.

test/check-application
```diff
@@ -12,7 +12,10 @@ import unittest
 TEST_DIR = os.path.dirname(__file__)
 sys.path.append(os.path.join(TEST_DIR, "common"))
 sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
+
 import testlib
+from testlib import skipImage
```
martinpitt: Why not simply write `testlib.skipImage()`? That's already imported, and it's confusing to mix both styles.

test/check-application
```diff
@@ -22,46 +25,132 @@ registries = ['localhost:5000']
...
         # HACK: sometimes podman leaks mounts
         super().setUp()
         self.machine.execute("cp -a /var/lib/containers /var/lib/containers.orig")
+
+        # Create admin session
+        self.machine.execute("mkdir /home/admin/.ssh")
+        self.machine.execute("cp /root/.ssh/* /home/admin/.ssh")
+        self.machine.execute("chown -R admin:admin /home/admin/.ssh")
+        self.machine.execute("chmod 700 /home/admin/.ssh")
+        self.machine.execute("chmod 600 /home/admin/.ssh/*")
```
martinpitt: you can also use a multi-line string and run through one .execute() command

How about `chmod -R go-wx /home/admin/.ssh` (but that's bikeshedding)

test/check-application
```diff
@@ -22,46 +25,132 @@ registries = ['localhost:5000']
...
+        self.admin_s = ssh_connection.SSHConnection(user="admin",
+                                               address=self.machine.ssh_address,
```
martinpitt: wrong indentation

test/check-application
```diff
@@ -22,46 +25,132 @@ registries = ['localhost:5000']
...
+            self.allow_journal_messages("Unit io.podman.socket could not be found.")
+
+        self.allow_journal_messages("/run.*/podman/io.podman: couldn't connect.*")
         self.addCleanup(self.machine.execute, "podman rm --force --all && "
             "findmnt --list -otarget | grep /var/lib/containers | xargs -r umount && "
             "rm -r /var/lib/containers && "
             "mv /var/lib/containers.orig /var/lib/containers")
 
-    def testBasic(self):
+    def execute(self, auth, cmd):
```
martinpitt: s/auth/system/?

test/check-application
```diff
@@ -22,46 +25,132 @@ registries = ['localhost:5000']
...
         b.wait_present(".content-filter input")
-        b.wait_in_text("#containers-images", "busybox:latest")
+
+        # Check all containers
+        if auth:
+            checkImage(b, "docker.io/library/busybox:latest", "system")
+            checkImage(b, "docker.io/library/alpine:latest", "system")
+            checkImage(b, "docker.io/library/registry:2", "system")
+
+        if m.image != "rhel-8-1":
```
martinpitt: This test case has a lot of these special cases, which makes it rather hard to read and understand. How about you split this into testSystemContainers and testUserContainers, and `skipImage` the latter on rhel-8-1?

src/Containers.jsx
```diff
@@ -85,24 +85,31 @@ class Containers extends React.Component {
...
-        const containerStats = containersStats[container.id];
+        const containerStats = containersStats[container.id + container.isRoot.toString()];
         const isRunning = container.status == "running";
         const image = container.image;
 
+        let proc = "";
+        let mem = "";
+        if (containerStats) {
+            proc = containerStats.cpu ? utils.format_cpu_percent(containerStats.cpu * 100) : <abbr title={_("not available")}>{_("n/a")}</abbr>;
+            mem = containerStats.mem_usage ? utils.format_memory_and_limit(containerStats.mem_usage, containerStats.mem_limit) : <abbr title={_("not available")}>{_("n/a")}</abbr>;
```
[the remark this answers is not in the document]
martinpitt: Actually, these are columns in a table. IMHO it wouln't be too bad to just keep them empty, but I don't have a strong opinion.

martinpitt:
Feel free to land the first few commits in a separate PR, to get them out of the way.

When running the tests locally (Fedora 30 toolbox), I get this failure:
```
testDownloadImage (__main__.TestApplication)
[..]
> error: Warning: A component is changing an uncontrolled input of type %s to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components%s radio 
    in input (created by ImageSearchModal)
    in fieldset (created by ImageSearchModal)
    in form (created by ImageSearchModal)
    in div (created by ModalBody)
    in ModalBody (created by ImageSearchModal)
    in div (created by CustomModalDialog)
    in div (created by CustomModalDialog)
    in div (created by CustomModalDialog)
    in CustomModalDialog (created by Modal)
    in Transition (created by Fade)
    in Fade (created by DialogTransition)
    in DialogTransition (created by Modal)
    in RefHolder (created by Modal)
    in div (created by Modal)
    in Portal (created by Modal)
    in Modal (created by Modal)
    in Modal (created by ImageSearchModal)
    in ImageSearchModal (created by Images)
    in div (created by Images)
    in Images (created by Application)
    in div (created by Application)
    in div (created by Application)
    in Application
```
This is an actual bug in c-podman (possibly not from this PR, though). This is in addition to the [testRunImageRoot failure](https://209.132.184.41:8493/logs/pull-173-20190918-130909-8e288bfb-cockpit-project-cockpit-podman--fedora-30/log) that happens on CI and also locally.

I'm still not that happy about the table layout. Even on a relatively wide screen there are lots of intra-word breaks, like "sys-tem" or "run-ing", or the container name:
podman

And on a slightly more narrow screen these become really messy:

podman-narrow

And that's with only two  containers. I think it would be better to reorganize the table so that each container gets two lines, the first for the current table columns minus the command, and the second line with the command (spanning all columns). Or, if that's impractical, drop the command -- it's already in the expander details anyway (which is currently redundant).

Thanks!

garrett:
> I'm still not that happy about the table layout. Even on a relatively wide screen there are lots of intra-word breaks, like "sys-tem" or "run-ing", or the container name

This is related to @KKoukiou's port of our list tables to PF4. It's a PF4 bug/"feature".

Previously, I optimized our old lists for narrow widths, including mobile. PF4 just forces wrapping in every case.

It's now bad across all of Cockpit, not specific to this PR. :disappointed:

KKoukiou:
> > I'm still not that happy about the table layout. Even on a relatively wide screen there are lots of intra-word breaks, like "sys-tem" or "run-ing", or the container name
> 
> This is related to @KKoukiou's port of our list tables to PF4. It's a PF4 bug/"feature".
> 
> Previously, I optimized our old lists for narrow widths, including mobile. PF4 just forces wrapping in every case.
> 
> It's now bad across all of Cockpit, not specific to this PR. disappointed

As we dicsussed on IRC this can't be related to the PF4-tables PR since that PR changed only 2-3 tables in machines page.
(Just to avoid confusion)

garrett:
> this can't be related to the PF4-tables PR since that PR changed only 2-3 tables in machines page

Ah, I saw the same problem with the PF4 tables. I guess we also have it here. Oops. :slightly_frowning_face: Regardless, we'll need to change it in both places.

---

We might want to rework the lists so that Name, Image, and Command are part of a "Container" cell, where each of these are on their own line inside the cell:


  
- **devel**
- registry:fedoraproject.org/f31/fedora-toolbox:31
- toolbox verbose init-container --home /var/home/martin --home-link --monitor-host shell /bin/bash --uuid 1000 --user martin

And then it's OK if the last line wraps (or is truncated, really). Combining all 3 into one cell would allow the items share the width and would allow for the other items to not wrap.

garrett:
Here's a mockup based on your screenshot:

podman-list

It prioritizes the critical information, especially the information for finding the unique container. I have the command line truncated. It's still enough to see what type of container it is (in this case, it's running via toolbox) and you'd want to show this information in full when the list item is expanded.

I've rounded the memory (perhaps GiB isn't the right unit here?) and made sure the values do not wrap (which is only possible if the memory isn't super-long, unless that's allowed to wrap, however).

The items under the main one use the small text (as referenced in [the PF4 typography guidelines](https://www.patternfly.org/v4/design-guidelines/styles/typography)). I think a `<small>` tag should do it?

src/Containers.jsx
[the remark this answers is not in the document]
marusak: > Are stats not available for user containers in general? Or when would that happen?

When you run it on system without CGroupsV2

>  IMHO it wouln't be too bad to just keep them empty, but I don't have a strong opinion.

I also don't have a strong option about it, but I think it is better to show there something. Mostly because when these values are not there it indicates that that container is not running and this could confuse users and I can see those reports coming in that sometimes we don't show usage.

src/app.jsx
```diff
@@ -88,87 +98,134 @@ class Application extends React.Component {
...
+    updateState(state, id, newValue) {
+        this.setState(prevState => {
+            let copyState = Object.assign({}, prevState[state]);
+
+            copyState[id] = newValue;
+
+            return {
+                [state]: copyState,
+            };
+        });
```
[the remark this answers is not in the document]
marusak: tbh, I find your version more complicated to understand what it does :) and I think `id` should be `[id]` to work correctly.

src/app.jsx
```diff
@@ -88,87 +98,134 @@ class Application extends React.Component {
...
+
+    updateContainerStats(id, root) {
+        utils.podmanCall("GetContainerStats", { name: id }, root)
+                .then(reply => {
+                    this.updateState("containersStats", reply.container.id + root.toString(), reply.container);
                 })
                 .catch(ex => {
-                    console.warn("Failed to do Update Container:", JSON.stringify(ex));
+                    if (ex.error === "io.podman.ErrRequiresCgroupsV2ForRootless") {
+                        console.log("This OS does not support CgroupsV2. Some information may be missing.");
```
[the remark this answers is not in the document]
marusak: No, because `podmanCall` does not wrap this method.
We really need to process errors better, but that is topic for another PR.

src/app.jsx
```diff
@@ -88,87 +98,134 @@ class Application extends React.Component {
...
-                        containersCopyStats[reply.container.id] = reply.containerStats;
+                        // Copy only images that could not be deleted with this event
+                        // So when event from root come, only copy user images and vice versa
+                        let copyImages = {};
+                        Object.entries(prevState.images || {}).forEach(([id, image]) => {
+                            if (image.isRoot !== root)
+                                copyImages[id] = image;
+                        });
+                        Object.entries(reply).forEach(([id, image]) => {
+                            image.isRoot = root;
```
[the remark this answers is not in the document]
marusak: Yes, it is.
The thing is that we get response from `ListImages` and we store it in the state as we got it. But we need to also store there if it is user or root(renamed to system) image. So the easiest way is to just add another property into the  reply and then store it.

src/app.jsx
marusak: > This now does a lot of systemctl CLI calls, which are relatively expensive and also don't provide monitoring

This is called only once when the page is first rendered. Also we don't need monitoring in this step, this just checks if user service exists as well (so at least podman 1.5 or higher) and if it does but is not running (discovered somewhere else) then we suggest starting it.

I can use `is-active` for sure.

marusak: Or probably `is-enabled`, since `is-active` on non-existing service replies with `inactive`.

marusak: Hmm, so return code for disabled is the same as for non existing (1). SO I guess the best way is to run it with C locale and check if it contains some string.

test/check-application
marusak: I want to check two scenarios:
1. Only user containers (there is `testBasicUser` which is skipped on rhel-8-1)
2. Test that user and system containers can both be used at once (this is this scenario. Problem is that rhel-8-1 does not have user containers so I need to skip half of the test).

I could copy the tests once over and have only root containers, which would be tested on rhel-8-1. Do you think I should do that?

marusak:
> When running the tests locally (Fedora 30 toolbox), I get this failure:

I cannot reproduce it. Can you tell me when this happens?

> I'm still not that happy about the table layout. Even on a relatively wide screen there are lots of intra-word breaks, like "sys-tem" or "run-ing", or the container name:

Yeah, but this also happens on master. Lets please do that in follow up where we restyle it as @garrett suggested.
This is master:
masteraswell

marusak:
@martinpitt thank you for the review. Addressed most of your comments and commented on the rest.

garrett:
Wrapping fix is @ https://github.com/cockpit-project/cockpit/pull/12909 (diff @ https://github.com/cockpit-project/cockpit/pull/12909/files). There's an upstream bug in PF4 to fix this properly, but meanwhile, we should use this hotfix.

src/app.jsx
```diff
@@ -196,59 +253,83 @@ class Application extends React.Component {
...
+                .then(() => this.setState({ userServiceExists: true }))
+                .catch((_, response) => {
+                    if (response.trim() !== "disabled")
+                        this.setState({ userServiceExists: false });
+                    else
+                        this.setState({ userServiceExists: true });
+                });
     }
 
     startService(e) {
```
martinpitt: It's rather odd that this function enables both the system and the user service, unconditionally. Shouldn't this also get a boolean flag like the others?

martinpitt:
Functionality test:

 * When I stopped my user's io.podman.{service,socket}, I got the header bar "User Podman service is also available". When I click on "Start", the console logs

```
Failed to enable unit: No such file or directory
cockpit.js:784:21
Failed to start user io.podman.socket: {"problem":null,"exit_status":1,"exit_signal":null,"message":"Failed to enable unit: No such file or directory"}
```
and nothing happens. `systemctl --user status io.podman.socket` is fine, though  (loaded/enabled/inactive). The same thing works fine for the system instance.

 * CPU/memory columns are empty for user containers (not even "n/a"), is that a known limitation? It works for system containers (I'm on current F31 with cgroupsv2 and standard podman install)

 * This is a bit hard to test due to podman varlink being so terribly unstable and locking up all the time. Not the fault of this PR, but it seriously impedes the user experience. Are there still open upstream bugs for these?

src/app.jsx
```diff
@@ -317,40 +417,62 @@ class Application extends React.Component {
...
+                    <span>{_("System Podman service is also available")}</span>
+                </div>
+                <button onClick={this.startService}>{_("Start")}</button>
+            </div>;
+        }
+        if (!this.state.userServiceAvailable && this.state.userServiceExists) {
+            startService = <div className="alert alert-info dialog-info">
+                <div className="info-message">
+                    <span className="fa fa-exclamation-triangle" />
+                    <span>{_("User Podman service is also available")}</span>
```
martinpitt: Related to this: When I click the "start" button, I briefly see the empty state pattern with the check box that selects whether to permanently enable the service or not. But this goes away after a blink of an eye, back to the overview page. If this is meant to completely replace the empty state pattern, then the latter should not be triggered any more?

marusak:
> When I stopped my user's io.podman.{service,socket}, I got the header bar "User Podman service is also available". When I click on "Start", the console logs

As discussed on IRC, this is bug in podman - resolved in https://github.com/containers/libpod/commit/59353b4dce692df0fb22a15427abbaafad58050e

> CPU/memory columns are empty for user containers (not even "n/a"), is that a known limitation? It works for system containers (I'm on current F31 with cgroupsv2 and standard podman install)

This is wrong. I tested on my machine and indeed does not work. I reported it https://github.com/containers/libpod/issues/4268

> This is a bit hard to test due to podman varlink being so terribly unstable and locking up all the time. Not the fault of this PR, but it seriously impedes the user experience. Are there still open upstream bugs for these?

I think they should be fixed in podman-1.6

src/app.jsx
marusak: This is also used when neither is running (and you get the screen with `Start podman` which should start both units). But it could at least check if one service is already not recognized as running.

src/app.jsx
marusak: It is not meant to replace the empty state pattern. There are 3 cases:
1. Neither service is running (show the empty pattern)
2. Only one service is running and other is present (load the one and show message about the other one)
3. Only one service is present and one is running (just load it normally and don't show anything)

> When I click the "start" button, I briefly see the empty state pattern with the check box that selects whether to permanently enable the service or not.

Hmm, I am not able to reproduce this, but I'll try a bit harder

src/app.jsx
marusak: But this function is called in following cases:
1. Only one unit is present and that unit is not active. We need to check both
2. There are two units, and exactly one is running. In that case the one is no-op and the other starts the second unit.

marusak:
And all F31 tests fail with `'systemctl --now --user enable io.podman.socket'` - which is fixed in https://github.com/containers/libpod/commit/59353b4dce692df0fb22a15427abbaafad58050e

marusak:
F31 failed with `> warning: Failed to update container stats: {"error":"io.podman.ErrorOccurred","parameters":{"reason":"Link not found"}}`, this is reported in https://github.com/containers/libpod/issues/4268 - I guess naughty?

marusak:
Needs https://github.com/cockpit-project/bots/pull/138

martinpitt:
The [fedora-30 failure](https://209.132.184.41:8493/logs/pull-173-20191015-123131-366c5d23-cockpit-project-cockpit-podman--fedora-30/log.html) looks like it's going to need a special case for not having cgroupsv2 yet?

marusak:
Fedora-31 very often does not shut down in image-customize. This needs to be looked at. https://logs-https-cockpit.apps.ci.centos.org/logs/pull-173-20191016-054343-ee67d486-cockpit-project-cockpit-podman--fedora-31/log.html (outside of this PR)

martinpitt:
Debugging the shutdown hang a bit. I can't reproduce any hang with the standard unmodified image:

 * A simple `bots/vm-run fedora-31` (in a clean tree, no overlay) doesn't hang, it reboots cleanly.
 * `time bots/image-customize -v -r true fedora-31` is 17s on my system, and works in a loop.
 * `bots/image-customize -v -i cockpit-ws -s test/vm.install fedora-31` doesn't hang (everything except installing the built cockpit-podman, which should be fairly unintrusive)
 * The full `TEST_OS=fedora-31 make vm` also works fine.

However, it *does* hang with the vm.install change from this PR. The smallest reproducer is

    bots/image-customize -v -r 'sudo -i -u admin podman ps' fedora-31

while running another command in the admin session doesn't hang:

    bots/image-customize -v -r 'sudo -i -u admin whoami' fedora-31

So the problem isn't with the `sudo -i -u admin` part, i. e. it's not the user session itself that hangs. When you follow along with `virsh console fedora-31-127.0.0.2-2201`, you see that it doesn't actually starts to shut down, there's just some kernel messages (just audit, nothing useful) and the entire thing freezes -- no ssh any more and no VT login.

Doing this more interactively with `bots/vm-run fedora-31`, VT-logging in directly as admin, and running `podman ps` works, though -- no lockup, one can `reboot` without problems. Same with logging in as root and doing it through sudo. Same with `ssh c 'sudo -i -u admin podman ps'`.

But it's also not the command itself that hangs:

    bots/vm-reset; bots/image-customize -v -r 'echo BOOT; sleep 5; echo CMD; sudo -i -u admin podman ps; sleep 5; echo DONE; sleep 5; echo REBOOT' fedora-31

During all phases I can login with ssh and the machine works, and it gets insta-broken at `REBOOT`.

With a slightly longer sleep, I can console-attach and log in after `DONE`, run `reboot`, and get the same.

martinpitt:
@marusak : I filed the kernel bug (see above); ideas for workaround:

 * enable debug, do vm.install, disable debug again (three invocations of image-customize)
 * append "sync; poweroff -f" to the command and ignore the failure

Makefile
```diff
@@ -123,7 +123,9 @@ rpm: dist-gzip $(RPM_NAME).spec
 # build a VM with locally built rpm installed
 $(VM_IMAGE): rpm bots
 	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
+	bots/image-customize -v -r 'grub2-editenv - set "$$(grub2-editenv list | grep ^kernelopts) debug"' $(TEST_OS)
```
martinpitt: Please restrict that to fedora-31 only.

martinpitt: .. and add a "# HACK:" link to bz

Makefile
```diff
@@ -123,7 +123,10 @@ rpm: dist-gzip $(RPM_NAME).spec
 # build a VM with locally built rpm installed
 $(VM_IMAGE): rpm bots
 	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
+	# HACK: https://bugzilla.redhat.com/show_bug.cgi?id=1762663
+	if [ $(TEST_OS) == "fedora-31" ]; then bots/image-customize -v -r 'grub2-editenv - set "$$(grub2-editenv list | grep ^kernelopts) debug"' $(TEST_OS); fi
```
martinpitt: Bashism, please use a single `=`

PR #178, opened 2019-08-13
Enable setting up memory

marusak:
This was never finished due to bug in podman.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1732713


marusak:
[TypeError: Cannot read property 'length' of undefine](https://209.132.184.41:8493/logs/pull-178-20190813-104040-94b188e2-cockpit-project-cockpit-podman-rhel-8-1/log) two times in row on rhel-8-1. I don't think it is related to this PR anyway, but should be fixed.

PR #197, opened 2019-09-11
CSS updates

martinpitt:
CSS minimization shamelessly stolen from https://github.com/cockpit-project/cockpit/pull/12723 . Thanks @KKoukiou for figuring that out!


marusak:
Thanks, I checked and it does not cause any regression with loading of css files.
However I am not quiet sure, how much this helps (size of `dist/` does not change when build with this or without this - Am I comparing something  wrongly or am I running it wrongly [I just run `make`])

martinpitt:
@marusak : Minimization is only done with `NODE_ENV=production make`. Then it loses about 20% size.

PR #202, opened 2019-09-13
Pin down version of webpack

marusak:
Version 4.40.0 has problem with production mode.

Related https://github.com/vuejs/vue-cli/issues/4572


marusak:
So webpack was fixed, no need for this

PR #212, opened 2019-10-01
Changes for user containers

marusak:
Picked from #173 so there are less changes and easier to review.

### DIFF
--- a/src/ContainerCommitModal.jsx
+++ b/src/ContainerCommitModal.jsx
@@ -3,7 +3,6 @@ import { Modal, Button, FormGroup, FormControl } from 'patternfly-react';
 import cockpit from 'cockpit';
 
 import * as utils from './util.js';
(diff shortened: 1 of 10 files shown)


martinpitt:
testNotRunning failed everywhere -- that's a bit more systematic than the podman flakes, as that's testing the systemd service integration, no?

marusak:
> testNotRunning failed everywhere -- that's a bit more systematic than the podman flakes, as that's testing the systemd service integration, no?

Kinda, it tests that we can react when podman disconnects. I think this is real issue. Locally when I do the same tests do it works, but since this looks suspicious I am currently  downloading image to debug it.

marusak:
Fixed. The problem was that I was checking wrong property, as you correctly [pointed out](https://github.com/cockpit-project/cockpit-podman/pull/173/files#r328928636) in the other PR :)

PR #225, opened 2019-10-18
Update eslint

marusak:


lib/cockpit-components-file-autocomplete.jsx
```diff
@@ -55,7 +55,7 @@ export class FileAutoComplete extends React.Component {
         this.debouncedChange = debounce(300, (value) => {
             this.updateIfDirectoryChanged(value, () => {
                 this.pendingCallback = false;
-                let stateUpdate = this.filterFiles(value);
+                const stateUpdate = this.filterFiles(value);
```
martinpitt: It seems that this file has diverged quite a bit from cockpit master. Can you please apply the changes there, and sync it over instead? e. g. this code does not exist at all any more since https://github.com/cockpit-project/cockpit/pull/12331

marusak:
>  But please sync lib/ from current Cockpit master.

Will do. Merging this and will send follow up

PR #226, opened 2019-10-18
Sync lib/ from cockpit

marusak:
Copy over from cockpit. Excluding some different imports and one css rule that we need to be different in podman (this was decided on the last copyover). Also I added one small fix in podman css.


marusak:
Oh yeah, I need to adjust tests

PR #232, opened 2019-10-30
Update translations from Fedora Zanata

cockpituous:
Update translations from Fedora Zanata
 * [x] po-refresh

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -1,0 +1,455 @@
+# Martin Pitt <mpitt@redhat.com>, 2019. #zanata
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE_VERSION\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2019-10-30 08:05+0000\n"
+"PO-Revision-Date: 2019-10-29 08:13+0000\n"
+"Last-Translator: Martin Pitt <mpitt@redhat.com>\n"
+"Language-Team: German\n"
+"Language: de\n"
+"X-Generator: Zanata 4.6.2\n"
+"Plural-Forms: nplurals=2; plural=(n != 1)\n"
+
+#: src/ImageDeleteModal.jsx:16
+msgid "Are you sure you want to delete this image?"
+msgstr "Sind Sie sicher, dass Sie dieses Image löschen wollen?"
+
+#: src/ContainerCommitModal.jsx:171 src/ImageDetails.jsx:28
+msgid "Author"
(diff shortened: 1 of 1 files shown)


po/de.po
```diff
@@ -0,0 +1,455 @@
...
+msgid "Are you sure you want to delete this image?"
+msgstr "Sind Sie sicher, dass Sie dieses Image löschen wollen?"
+
+#: src/ContainerCommitModal.jsx:171 src/ImageDetails.jsx:28
+msgid "Author"
+msgstr "Autor"
+
+#: src/app.jsx:401
+msgid "Automatically start podman on boot"
+msgstr "podman automatisch beim Hochfahren starten"
```
marusak: Maybe if it would start with capital letter would be better. Is this correct grammar-wise?

po/de.po
```diff
@@ -0,0 +1,455 @@
...
+msgstr "Befehl"
+
+#: src/ContainerCommitModal.jsx:215 src/Containers.jsx:137
+msgid "Commit"
+msgstr ""
+
+#: src/ContainerCommitModal.jsx:206
+msgid "Commit Image"
+msgstr ""
+
```
marusak: heh, I also had no idea what would be correct translation of `commit`

po/de.po
```diff
@@ -0,0 +1,455 @@
...
+#: src/Containers.jsx:120
+msgid "Console"
+msgstr "Konsole"
+
+#: src/ContainerCommitModal.jsx:133 src/ImageRunModal.jsx:287
+msgid "Container Name"
+msgstr "Container-Name"
```
marusak: This seems weird. But I guess it is correct that way in German? (Also it is with path and port, so I guess this was not typo).

marusak:
Every build fail with `Error: postcss-svgo: TypeError: Cannot set property 'multipassCount' of undefined`. Seems like some npm changed.

marusak:
https://github.com/svg/svgo/issues/1174

marusak:
I was not able to push here the fix, so I opened #233 (It will close this PR once landed)

po/de.po
martinpitt: Yes, German likes to build compound words, and as "Container" is a foreign word, we don't write "Containername".

po/de.po
martinpitt: Right, this may be a bit inconsistent. I guess I can't quite decide yet whether "podman" is a kind of trade mark/proper noun, or whether it's legit to capitalize it. I updated it in Zanata now.

po/de.po
martinpitt: In non-computer language it would be something like "verpflichten", but this sounds absurd in the context of containers. When German geeks talk to each other about database transactions, git revisions, etc, they literally say "etwas committen", because there just isn't any good German word for it that doesn't sound ridiculous.

PR #234, opened 2019-10-31
Update to new zanata cli

marusak:
* [x] po-refresh

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -2,17 +2,17 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE_VERSION\n"
-"MIME-Version: 1.0\n"
-"Content-Type: text/plain; charset=UTF-8\n"
-"Content-Transfer-Encoding: 8bit\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-10-30 08:05+0000\n"
-"PO-Revision-Date: 2019-10-29 08:13+0000\n"
+"POT-Creation-Date: 2019-10-31 10:10+0000\n"
+"PO-Revision-Date: 2019-10-30 09:23+0000\n"
 "Last-Translator: Martin Pitt <mpitt@redhat.com>\n"
 "Language-Team: German\n"
 "Language: de\n"
-"X-Generator: Zanata 4.6.2\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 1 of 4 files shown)


cockpituous:
po-refresh in progress on 1-cockpit-11.
Log: https://209.132.184.41:8493/logs/po-refresh-234-20191031-101010/

cockpituous:
po-refresh  done: https://github.com/cockpituous/cockpit-podman/commits/po-refresh-20191031-101111

marusak:
Ok, refresh worked. It needed one hack, since it was failing when manifest did not declare line number. I just put there 0 since the manifest is not that big it would make any practical difference. Also I have no idea why this was not crashing in cockpit, where manifest also does not have any line number.

po/de.po
```diff
@@ -24,7 +24,7 @@ msgstr "Autor"
 
 #: src/app.jsx:401
 msgid "Automatically start podman on boot"
-msgstr "podman automatisch beim Hochfahren starten"
+msgstr "Podman automatisch beim Hochfahren starten"
```
martinpitt: Hah, it picked up the capitalization question we were discussing before.

po/manifest2po
```diff
@@ -92,7 +92,7 @@ function process_menu(menu) {
         if (menu[m].label) {
             push({
                 msgid: menu[m].label,
-                locations: [ filename ]
+                locations: [ filename + ":0" ]
```
martinpitt: This should get a comment, as it's not obvious what that :0 does.

marusak: Done

po/manifest2po
```diff
@@ -92,7 +92,7 @@ function process_menu(menu) {
         if (menu[m].label) {
             push({
                 msgid: menu[m].label,
-                locations: [ filename ]
+                locations: [ filename + ":0" ] // Fake line number to zero
```
martinpitt: Well, that's the "what", which is quite obvious -- I meant document the "why", like "zanata crashes with BlahException when there is no line number"

po/manifest2po
```diff
@@ -92,7 +92,8 @@ function process_menu(menu) {
         if (menu[m].label) {
             push({
                 msgid: menu[m].label,
-                locations: [ filename ]
+                locations: [ filename + ":0" ] // Fake line number because when no line number
+                // present `zanata pull` fails with `ValueError: not enough values to unpack`
```
marusak: ...and that's best I can do :D

PR #240, opened 2019-11-07
tests: Drop unnecessary hack

marusak:
This is supposed to be fixed in podman-1.6.

This version is in Fedora-30 and Fedora-31. I was lazy to check RHEL-8-1, but tests let us know.


marusak:
Ha, not fixed even in F30 and F31 as they promote. I'll follow up back to them

PR #244, opened 2019-11-12
Enhance listing of containers on narrow screens

### LINKED ISSUE #223: Change listing of containers to better deal with narrow screens
See https://github.com/cockpit-project/cockpit-podman/pull/173#issuecomment-536487090

marusak:
Fixes #223

newwidescreen

One issue I have with css is that it does not want to wrap with ellipses... @garrett could you please take a quick look and give me some hint why is that so? Thank you!
doesnotwantotelipssis


marusak:
I'll need help with this. The table overflows the page, and I haven't figured out how to force the first column wrap more.. @garrett could I ask for some advice what would be a good approach?

garrett:
Sooooo... cockpit-podman is using Incorrect fonts.

While RedHatDisplay is used for the headings, Overpass is still being used for the text in the listings (instead of RedHatText). The font sizes are incorrect as well. This _will_ impact things, including breakpoints and the sizes of the UI.

Table headings (`<th>`) are also incorrectly styled.

Is cockpit-podman using a recent version of PF3 overrides? :thinking:

garrett:
I'm not able to get this branch to work. (It might not be specific to the branch. In fact, it probably isn't this branch.)

I just see "loading..." for containers and it won't let me run any of the images.

Screenshot_2020-02-03 Podman Containers - Rain(1)


That said, the Images list also has issues in mobile. Are we going to fix that in this PR too, or in another one?

marusak:
> I just see "loading..." for containers and it won't let me run any of the images.

eh, podman... It may take a bit longer to load. But I also saw broken podman, where I couldn't even do 'podman ps -a'. Can you do that?

> Are we going to fix that in this PR too, or in another one?

I was not planing on doing that. Do we also want to put more items in the same column?

garrett:
eh, podman... It may take a bit longer to load. 

It finally loads in >2.5 minutes.

> But I also saw broken podman, where I couldn't even do 'podman ps -a'. Can you do that?

Yep, that's _instant_.

garrett:
I've pushed some fixups, however:

- It takes _forever_ to iterate on this, as there's:
   - the page load bug which makes the page take a while to load still
   - the podman loading bug which causes the list to take around 2.5 minutes to load here ()
   - make clean is necessary before a make because the make system doesn't detect changes in CSS
- mobile & small widths are still broken
- the lists are lacking classes, which sucks for adding CSS
- the a11y fix for `<th>` scope should be merged into Cockpit upstream
- I think Cockpit itself is switching to PF4 for these lists?

marusak:
> I've pushed some fixups

Thanks!

> the page load bug which makes the page take a while to load still

What page load bug? I am not aware of any.

> the podman loading bug which causes the list to take around 2.5 minutes to load here

This was happening with your user service, right? The system one was much faster. Or is that slow as well?

garrett:
> What page load bug? I am not aware of any.

It's a Cockpit bug where the first page loaded takes a while. At first, people thought it was related to the Overview redesign, but it seems to be any first page you hit. (It's not cockpit-podman related, but does throw a wrench in the edit-compile-reload workflow.)

> This was happening with your user service, right? The system one was much faster. Or is that slow as well?

Oh right, I didn't turn off user services today. :man_facepalming:

It's still slow for the system containers too, but is not nearly as slow.

garrett:
For reference, here's an included version of the mockup:

podman container mockup

martinpitt:
> It's a Cockpit bug where the first page loaded takes a while. 

At least for Sanne and me this happens when running out of git with a `--enable-debug` build. Then the webpacks are ginormous and just take a few seconds to load.

garrett:
> this happens when running out of git with a --enable-debug build

Ah, I have `--enable-debug` on too. (I wonder what I'd miss by not using that flag.)

martinpitt:
@garrett: You'd miss useful stack traces on crashes, extra React warnings, and presumably introspectability of React state through browser tools. I. e. nothing that should concern your work in most cases. :-)

marusak:
I just tested it and was looking good. @garrett wanna take another look? Now with podman 2.0.2 it should be fast and stable (at least I hope so, we now switched to a new API).

garrett:
Looking again...

garrett:
Ah, this is a standard table and it doesn't deconstruct on smaller screen sizes, including mobile. 

I thought it would be ported to a PF4 list of some sort, either https://www.patternfly.org/v4/documentation/react/components/table or our ct-table (as used in the Machines page). _Note: Machines page is probably moving from an expandable to a details page in draft PR https://github.com/cockpit-project/cockpit/issues/14196. But it's using `ct-table` either way, at the moment._

That said, it is already an improvement on semi-wide to wide desktop screens...

marusak:
Tests are going to be red, but I feel like my chrome is acting up, so want to double check with CI if it is a real issue or just in my local dev.

webpack.config.js
```diff
@@ -202,6 +203,17 @@ module.exports = {
                             }
                         }
                     },
+                    {
+                        loader: 'sass-resources-loader',
```
KKoukiou, line 207: The reason you import this here is to redefine the breakpoint variables which are problematic in our case because they don't count in the side nav.
However when you use React Components they will import the CSS styles, in this case table.css and table-grid.css files which already have bundled in the old breakpoints.
So this has not effect here, your table still uses the old breakpoints. You can verify by checking that the table collapses in 768px screen sizes, instead of 428px which is the redfined md-breakpoint.

If you want to  use different breakpoints, you have to use null-loader to disable importing the table CSS stylesheets, so that the new imported ones take effect. See cockpit null-loader in cockpit webpack config for the approach.

(or ignore it and use the PF breakpoints but in this case you don't really need the global variables)

marusak: I had to import it, as the build was failing otherwise.

KKoukiou: Ack thought you needed it for different reason :)

garrett:
> The contain state is wrongly aligned to the right

That's a PF4 bug we reported last year...

https://github.com/patternfly/patternfly-react/issues/3301

marusak:
> That's a PF4 bug we reported last year...
> patternfly/patternfly-react#3301

right, so what should we do? Should we never align it to the right not even on big screens?

PR #265, opened 2019-12-17
test: RHEL-8-2 does not have CGroupsV2

marusak:


marusak:
Once rhel-8-2 passes here, we can then land https://github.com/cockpit-project/bots/pull/341

marusak:
Needs one more skip:
```
  File "test/check-application", line 737, in _testRunImage
    self.assertEqual(memory, '536870912')
```
but every other test failed on `virt-what-cpuid` dumping core. I thought this was fixed.

PR #266, opened 2019-12-17
Enable to limit cpu

### LINKED ISSUE #167: Enable to limit cpu
We can use the same approach as with docker
cpupriotiy

marusak:
Fixes #167

I am not sure what the default value should be. In docker we had 1024, but when I don't specify it, podman starts with 0. So should I use 0 as well?

cpushares

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -205,7 +205,9 @@ export class ImageRunModal extends React.Component {
             publish: [],
             image: props.image,
             memory: 512,
+            cpuShares: 0,
             memoryConfigure: false,
+            cpuSharesConfigure: false,
             memoryUnit: 'MiB',
             validationFailed: {},
             volumes: [],
@@ -227,6 +229,9 @@ export class ImageRunModal extends React.Component {
         if (this.state.memoryConfigure && this.state.memory) {
             const memorySize = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
             createConfig.memory = memorySize.toString();
+        }
+        if (this.state.cpuSharesConfigure) {
+            createConfig.cpuShares = this.state.cpuShares;
         }
         if (this.state.hasTTY)
             createConfig.tty = true;
@@ -331,6 +336,28 @@ export class ImageRunModal extends React.Component {
                         </Select.SelectEntry>
                     </Select.Select>
                 </div>
+
+                { this.state.image.isSystem &&
+                    <>
+                        <label className='control-label' htmlFor='run-image-cpu-priority'>
+                            {_("CPU Shares")}
+                        </label>
+                        <div role='group' className='form-inline' id="run-image-dialog-cpu-priority">
+                            <div className="checkbox">
+                                <input id="run-image-dialog-cpu-priority-checkbox" type="checkbox"
+                                       checked={this.state.cpuSharesConfigure}
+                                       onChange={e => this.onValueChanged('cpuSharesConfigure', e.target.checked)} />
+                            </div>
+                            <input className='form-control'
+                                   type='number'
+                                   value={dialogValues.cpuShares}
+                                   step={1}
+                                   min={0}
+                                   disabled={!this.state.cpuSharesConfigure}
+                                   onChange={e => this.onValueChanged('cpuShares', parseInt(e.target.value))} />
+                        </div>
+                    </>
+                }
 
                 <label className='control-label'> {_("With terminal")} </label>
                 <label className="checkbox-inline">
--- a/test/check-application
+++ b/test/check-application
@@ -673,6 +673,15 @@ class TestApplication(testlib.MachineCase):
             b.set_input_text("#run-image-dialog-memory-limit input.form-control", "0.5")
             b.set_val('#memory-unit-select', "GiB")
 
+        # CPU shares work only with system containers
+        if auth:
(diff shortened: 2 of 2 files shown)


martinpitt:
@marusak: IMHO the default should be whatever podman does if you don't specify something explicitly. That's the least surprise.

src/ImageRunModal.jsx
```diff
@@ -228,6 +230,9 @@ export class ImageRunModal extends React.Component {
             const memorySize = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
             createConfig.memory = memorySize.toString();
         }
+        if (this.state.cpuSharesConfigure && this.state.cpuShares) {
```
martinpitt: This means that a value of 0 would be ignored. Is that intended? 0 may be the internal default, but not sure if we should hardcode that knowledge. Why not just actually set 0 if the user specified it?

src/ImageRunModal.jsx
```diff
@@ -332,6 +337,27 @@ export class ImageRunModal extends React.Component {
                     </Select.Select>
                 </div>
 
+                <label className='control-label' htmlFor='run-image-cpu-priority'>
+                    {_("CPU Priority")}
```
martinpitt: I'm in favor of calling this "CPU Shares", to match the podman CLI and manpage. It also better points out that these are a relative number, not an absolute priority.

test/check-application
```diff
@@ -673,6 +673,11 @@ class TestApplication(testlib.MachineCase):
             b.set_input_text("#run-image-dialog-memory-limit input.form-control", "0.5")
             b.set_val('#memory-unit-select', "GiB")
 
+            b.set_checked("#run-image-dialog-cpu-priority-checkbox", True)
+            b.wait_present("#run-image-dialog-cpu-priority-checkbox:checked")
+            b.wait_present('div.modal-body label:contains("CPU Priority") + div.form-inline > input[value="1024"]')
+            b.set_input_text("#run-image-dialog-cpu-priority input.form-control", "512")
+
```
martinpitt: Could you please make this a separate test (not necessarily new function, just a new container), and also cover the default setting? We should test both.

martinpitt:
Changes LGTM, thanks!

The [fedora-31 failures](https://209.132.184.41:8493/logs/pull-266-20191219-115555-18a43340-cockpit-project-cockpit-podman--fedora-31/log.html#5-2) are likely flakes, but the [rhel-8-1 failure](https://209.132.184.41:8493/logs/pull-266-20191219-115555-18a43340-cockpit-project-cockpit-podman--rhel-8-1/log.html#7-2) looks relevant.

marusak:
> The fedora-31 failures are likely flakes, but the rhel-8-1 failure looks relevant.

Actually, it is not possible to limit cpu-shares for rootless containers. just try `podman run --cpu-shares 10 fedora bash` as normal user. I need to skip that test.
But I am also wondering, if we should hide this option when rootless image? Not sure what is more confusing - the fact that the item is missing, or the fact that it fails. Maybe we could disable this option and have there some reason why it is not possible.

Downloading rhel-8-1 to investigate.

marusak:
Lets just disable the checkbox and put a message there.
disableandmessage
I'll align it to the bottom, but was wondering what icon would be the most suitable. @andreasn WDYT?

andreasn:
What's a rootless container exactly?
Does it work to change the memory?

marusak:
> What's a rootless container exactly?

Assuming you are not root, then running `podman run fedora bash` is rootless/user and running `sudo podman run fedora bash` is root/system container. In the UI there is owner of image/container.

> Does it work to set the memory limit?

Yes

andreasn:
"Only for system containers" makes it sound like I need to select another container.
What about something like "You do not have permissions to set CPU shares" as a popup if you hover the checkbox? Similar to how we do everywhere else when you have insufficient permissions.

martinpitt:
@andreasn : It's not a permissions issue -- user ("rootless") containers generally can't do this. I still think it'd be better to completely hide the CPU share input, but if you want to show it, there should be a clearly visible distinction between the working Memory and the non-working CPU. On your screenshot they look exactly the same? Also, maybe this should be a notification bubble instead of a persistent text in the dialog?

andreasn:
I think hiding it alltogether makes most sense in that case. There is nothing you can do in the dialog to fix it, so makes little sense to show it.

andreasn:
And then we can document that behavior in the documentation for cockpit, or point to the podman documentation itself.

marusak:
Ok, so I am not showing setting of CPU on rootless containers.
Also fixed tests for rhel-8-1. Problem was that it used to be called in inspect `CPUShares` but now it is `CpuShares`. Well, when I say `used to be called`. `podman inspect` on rhel-8-1 shows `CpuShares` but only filters with `CPUShares` otherwise fails with unknown tag. Also this version has default `<nil>` and not 0. Since it is rhel-8-1 I skip checking of default value on this image.

martinpitt:
UI looks good to me now. I also did a full real-life test with this. I started one system container with 10 CPU shares, and ran

    while true; do true; done &

four times in parallel (to match my 4 cores). Every bash process then got 100% CPU.

Then I started a second system container with default shares, did the same 4 loops. Then the 4 bashes from the latter got 99% CPU, and the four bashes from the former got limited to around 1% CPU.

So this seems to work as expected. Thanks!

Issue #267, opened 2019-12-18
please migrate to the new Fedora translation platform


Jibec:
Hello, the Fedora project migrates its translation platform to Weblate [1].

This tool directly interact with your git repository, and requires us to know:
* [mandatory] which branch is your development branch?
* [mandatory] have you merged latest translation from Zanata and locked the project?
* [info] Weblate will handle updates when pot file changes, don't edit po files for this [2]
* [optional] what is the license of translation? (please use a code from https://spdx.org/licenses/)
* [optional] do you have any announcement/warning you would like to display to the translators? (it will be displayed in Weblate)
* [optional] do you need us to activate any specific checks? (this is a setting per component [3])
* [optional] do you need us to automatically detect new translation files? (typical usecase: website translation with one translation file per page)

Please note:
* For github hosted projects, Weblate open pull request. For other project, you'll have to add a key to allow commits.
* In Weblate's vocable, one project is a group of component. Each component is a translation file. You can have many projects or many components or both.
* You can change your mind over time, just reach trans@lists.fedoraproject.org

[1] https://communityblog.fedoraproject.org/fedora-localization-platform-migrates-to-weblate/ 
[2] https://docs.weblate.org/en/latest/admin/continuous.html#avoiding-merge-conflicts
[3] https://docs.weblate.org/en/latest/user/checks.html#translation-checks

marusak:
> - [mandatory] which branch is your development branch?

master

> - [mandatory] have you merged latest translation from Zanata and locked the project?

We have checked 3 days ago and we had the latests updates. I haven't yet locked the project, as first I would like to try interact with weblate and see how it goes and once we know we can do the migration, then I'll lock the project and manually push the newest translations into weblate.

> - [info] Weblate will handle updates when pot file changes, don't edit po files for this [2]

We want to do manual updates. We don't want any automatic PRs with `.po` files updates, nor we will have `.pot` file upstream.

> - [optional] what is the license of translation? (please use a code from https://spdx.org/licenses/)

LGPL-2.1-or-later

> - [optional] do you have any announcement/warning you would like to display to the translators? (it will be displayed in Weblate)

Nothing I can think of.

> - [optional] do you need us to activate any specific checks? (this is a setting per component [3])

All the basic ones plus javascript. Also we use `$n` where `n` is number as format string, not sure which rule would match this. But I guess we can tweak these later on.

> - [optional] do you need us to automatically detect new translation files? (typical usecase: website translation with one translation file per page)

[...]

PR #280, opened 2020-01-04
po: Switch from Zanata to Weblate

marusak:
This needs to do successful po-refresh. But before I trigger it, I want to get review.
I did one refresh:
1. It pushed [this commit](https://github.com/marusak/cockpit-podman-weblate/commit/864ab0cd34df1065e78bed564220d55549831dfe) updating `.pot` file
2. Created [refresh](https://github.com/cockpit-project/cockpit-podman/pull/279) (it contains two more commits, but the point is that it picked new language (Turkish), ignored French (as it has only 11% translated strings) and there was no update in already existing languages. (I'll do some change in Slovak so we have also normal update)

 * [x] po-refresh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ NODE_MODULES_TEST=package-lock.json
 # one example file in dist/ from webpack to check if that already ran
 WEBPACK_TEST=dist/index.html
 
+WEBLATE_REPO=tmp/weblate-repo
+WEBLATE_REPO_URL=https://github.com/cockpit-project/cockpit-podman-weblate.git
+WEBLATE_REPO_BRANCH=master
+
 all: $(WEBPACK_TEST)
 
 #
@@ -57,16 +61,19 @@ dist/po.%.js: po/%.po $(NODE_MODULES_TEST)
 	po/po2json -m po/po.empty.js -o $@.js.tmp $<
 	mv $@.js.tmp $@
 
-ZANATA_CLI = LANG=C.UTF-8 LC_ALL=C.UTF-8 zanata
+$(WEBLATE_REPO):
+	git clone --depth=1 -b $(WEBLATE_REPO_BRANCH) $(WEBLATE_REPO_URL) $(WEBLATE_REPO)
 
-upload-pot: po/$(PACKAGE_NAME).pot
-	$(ZANATA_CLI) push -f --srcdir=./po --push-type=source --project-config=./po/zanata.xml $(PACKAGE_NAME).pot
+upload-pot: po/$(PACKAGE_NAME).pot $(WEBLATE_REPO)
+	cp ./po/$(PACKAGE_NAME).pot $(WEBLATE_REPO)
+	git -C $(WEBLATE_REPO) commit -m "Update source file" -- $(PACKAGE_NAME).pot
+	git -C $(WEBLATE_REPO) push
 
 clean-po:
 	rm ./po/*.po
 
-download-po:
-	$(ZANATA_CLI) pull --min-doc-percent 50 --transdir=./po --project-config=./po/zanata.xml
+download-po: $(WEBLATE_REPO)
+	cp $(WEBLATE_REPO)/*.po ./po/
 
 #
 # Build/Install/dist
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ test/common/
 test/images/
 *.pot
 POTFILES*
+tmp/weblate-repo
--- a/po/cs.po
+++ b/po/cs.po
@@ -1,18 +1,20 @@
 # Pavel Borecki <pavel.borecki@gmail.com>, 2019. #zanata
+# Matej Marusak <mmarusak@redhat.com>, 2020.
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE_VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-11-29 05:51+0000\n"
-"PO-Revision-Date: 2019-11-22 11:56+0000\n"
-"Last-Translator: Pavel Borecki <pavel.borecki@gmail.com>\n"
-"Language-Team: Czech\n"
+"POT-Creation-Date: 2019-12-27 13:41+0000\n"
+"PO-Revision-Date: 2020-01-07 10:25+0000\n"
+"Last-Translator: Matej Marusak <mmarusak@redhat.com>\n"
+"Language-Team: Czech <https://translate.stg.fedoraproject.org/projects/"
+"cockpit-podman/master/cs/>\n"
 "Language: cs\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=3; plural=(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2\n"
-"X-Generator: Zanata 4.6.2\n"
+"Plural-Forms: nplurals=3; plural=(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2;\n"
+"X-Generator: Weblate 3.9.1\n"
 
 #: src/ImageDeleteModal.jsx:16
 msgid "Are you sure you want to delete this image?"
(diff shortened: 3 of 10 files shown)


marusak:
Fixes https://github.com/cockpit-project/cockpit-podman/issues/267

Makefile
```diff
@@ -12,6 +12,10 @@ NODE_MODULES_TEST=package-lock.json
 # one example file in dist/ from webpack to check if that already ran
 WEBPACK_TEST=dist/index.html
 
+WEBLATE_REPO=cockpit-podman-weblate
```
martinpitt: Just in case we want to port that to starter-kit, could we generalize this a bit? This is just a temporary directory name, so it doesn't have to contain the project name. Also, would be nice to put this into ./tmp/ or so, and add that to .gitignore, to further point out that it can be cleaned up. It's not currently removed automatically, as it's being used by two different rules. But it should at least be removed in `make clean`.

Makefile
```diff
@@ -12,6 +12,10 @@ NODE_MODULES_TEST=package-lock.json
...
+WEBLATE_REPO_URL=git@github.com:cockpit-project/cockpit-podman-weblate.git
```
martinpitt: You could use `$(PACKAGE_NAME)-weblate` as project name? (Again, for generalizing to starter-kit)

Makefile
```diff
@@ -12,6 +12,10 @@ NODE_MODULES_TEST=package-lock.json
...
+WEBLATE_REPO_BRANCH=master
```
martinpitt: This is a bit inconvenient to  hardcode, as you have to remember to change it when branching (this bit me several times from zanata.xml as well). Can we get away with just using `$(shell git rev-parse --abbrev-ref HEAD)` instead of using a variable? Or at least set the variable to that value?

Makefile
```diff
@@ -57,16 +61,20 @@ dist/po.%.js: po/%.po $(NODE_MODULES_TEST)
 	po/po2json -m po/po.empty.js -o $@.js.tmp $<
 	mv $@.js.tmp $@
 
-ZANATA_CLI = LANG=C.UTF-8 LC_ALL=C.UTF-8 zanata
+$(WEBLATE_REPO):
+	git clone --depth=1 $(WEBLATE_REPO_URL) $(WEBLATE_REPO)
```
martinpitt: Please use `-b $(WEBLATE_REPO_BRANCH)` here, instead of the second command.

Makefile
```diff
@@ -57,16 +61,20 @@ dist/po.%.js: po/%.po $(NODE_MODULES_TEST)
...
+	git -C $(WEBLATE_REPO) checkout $(WEBLATE_REPO_BRANCH)
 
-upload-pot: po/$(PACKAGE_NAME).pot
-	$(ZANATA_CLI) push -f --srcdir=./po --push-type=source --project-config=./po/zanata.xml $(PACKAGE_NAME).pot
+upload-pot: po/$(PACKAGE_NAME).pot $(WEBLATE_REPO)
+	cp ./po/$(PACKAGE_NAME).pot $(WEBLATE_REPO)
+	git -C $(WEBLATE_REPO) commit -m "Update source file" -- $(PACKAGE_NAME).pot
```
martinpitt: Oh wow, I expected a `git add` here, I've never seen that syntax. Nice, I learned something new about git :-)

Makefile
```diff
@@ -57,16 +61,20 @@ dist/po.%.js: po/%.po $(NODE_MODULES_TEST)
...
+	git -C $(WEBLATE_REPO) push
 
 clean-po:
 	rm ./po/*.po
 
-download-po:
-	$(ZANATA_CLI) pull --min-doc-percent 50 --transdir=./po --project-config=./po/zanata.xml
+download-po: $(WEBLATE_REPO)
+	cp $(WEBLATE_REPO)/*.po ./po/
```
martinpitt: Hm, where does it ignore < 50% translations now?

marusak: https://github.com/cockpit-project/bots/commit/fb34b6fe698a6edd04456312ddea6bc654c3b86f here :)

Makefile
marusak: Yeah, I know we were switching this on branching time in zanata.xml, so I thought it is fine to keep it here and we can switch it.
I like the idea that we can parse it. I'll try it with the command you provided.

Makefile
marusak: I would need to use `RPM_NAME`, since `PACKAGE_NAME` here is `podman`. But repo is `cockpit-podman`. In starter-kit, the `PACKAGE_NAME` is `starter-kit` and repo is also `starter-kit` and not `cockpit-starter-kit`. In `cockpit-composer` the `PACKAGE_NAME` is `cockpit-composer`, but there is different organization.
So either we would need to fix here what the `PACKAGE_NAME` is, but that would change other things as well (like name of `.pot` file, which would mean we would have to update weblate now...)
So it is all just mess and it is simpler to just specify the correct repo URL than trying to fit it in.

Makefile
marusak: Hmm, does it mean that I won't be able to do the refresh from this branch? Because bots prepare this branch and then do `po-refresh`, but this branch does not exist in weblate repo...

Makefile
martinpitt: Ack.

Makefile
martinpitt: As discussed on IRC, this doesn't work well with directly checking out refs from PRs, so let's keep the hardcoding for now indeed.

Makefile
martinpitt: Ah right, it removes them! It's a bit icky indeed, would be cleaner to do this sorting right here -- but that would mean more code duplication. OK for now, thanks!

cockpituous:
po-refresh in progress on 2-ci-srv-05.
Log: https://209.132.184.41:8493/logs/po-refresh-280-20200107-102727/

cockpituous:
Command '['make', 'upload-pot', 'clean-po', 'download-po']' returned non-zero exit status 2.: https://209.132.184.41:8493/logs/po-refresh-280-20200107-102727/

cockpituous:
po-refresh in progress on 3-ci-srv-02.
Log: https://209.132.184.41:8493/logs/po-refresh-280-20200107-105757/

cockpituous:
Command '['make', 'upload-pot', 'clean-po', 'download-po']' returned non-zero exit status 2.: https://209.132.184.41:8493/logs/po-refresh-280-20200107-105757/

martinpitt:
Summary from IRC. Right now, bots use https:// GitHub URLs via the API with an extra `Authorization: token 123ABC...` header. This can't be specified directly with `git push`.

We have two main options to fix the authentication:
  * Let the Makefile use an URL like https://cockpituous:$(shell cat ~/.config/github-token)@github.com... (not sure if that works, and hardcoding the user name is a bit sad, as that doesn't work for local development)
  * Use ssh URLs and add a github.com stanza to our secrets that uses /secrets/id_rsa for GitHub (easy to deploy); This is a bit icky as so far we don't use ssh in the tasks container with GitHub, so it may introduce new problems and is generally a duplication. But it may be the simplest solution to push to branches from shell scripts.

We could also adjust the container to set up ~/.ssh/id_rsa* symlinks to /secrets/ and simplify the ssh config, but that requires a container rebuild and I'm not sure if it's worth it.

cockpituous:
po-refresh in progress on 4-ci-srv-03.
Log: https://209.132.184.41:8493/logs/po-refresh-280-20200108-124949/

cockpituous:
Command '['make', 'upload-pot', 'clean-po', 'download-po']' returned non-zero exit status 2.: https://209.132.184.41:8493/logs/po-refresh-280-20200108-124949/

cockpituous:
po-refresh in progress on 2-cockpit-11.
Log: https://209.132.184.41:8493/logs/po-refresh-280-20200108-133636/

cockpituous:
po-refresh  done: https://github.com/cockpituous/cockpit-podman/commits/po-refresh-20200108-133838

marusak:
Refresh was successful. Some notes:
1. Update of the `.pot` file was done: https://github.com/cockpit-project/cockpit-podman-weblate/commit/243e687e0777bb825f35b680cac45262e04d0297
2. It did not pick up French and Portuguese(Brazil) as they are are less than 50% translated (https://translate.stg.fedoraproject.org/projects/cockpit-podman/master/)
3. It picked a few new languages, which were translated newly in the Weblate

PR #282, opened 2020-01-06
Wait until both services are detected

marusak:
We showed the page when one service was detected, which meant we always
shown alert mentioning the other service is available as well, even when
the service was running. We closed this warning after a few seconds.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -383,7 +383,7 @@ class Application extends React.Component {
     }
 
     render() {
-        if (this.state.systemServiceAvailable === null && this.state.userServiceAvailable === null) // not detected yet
+        if (this.state.systemServiceAvailable === null || this.state.userServiceAvailable === null) // not detected yet
             return null;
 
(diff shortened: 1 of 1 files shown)


src/app.jsx
```diff
@@ -383,7 +383,7 @@ class Application extends React.Component {
     }
 
     render() {
-        if (this.state.systemServiceAvailable === null && this.state.userServiceAvailable === null) // not detected yet
+        if (this.state.systemServiceAvailable === null || this.state.userServiceAvailable === null) // not detected yet
```
martinpitt: What happens if the session or user service are not enabled, or not available? Wouldn't this block page loading forever?

marusak: This checks for `null`, which means we did not yet checked. We check both services independently, and if they are active, we set it to 'true'. If they are disabled, we set it to 'false'. If the commands fails (like the unit does not exist), we also set it to 'false'.
We have exception, which checks if user service is disabled or missing. We do not do this with system service, as we expect it to always exist.

PR #284, opened 2020-01-07
Fixes and testing of commit 

marusak:
Issue https://github.com/containers/libpod/issues/3897 has been fixed a
few months ago, it reached Fedora 30 and 31 as well as RHEL 8.2.

^^ I haven't tested it, this is what I hope is true. Maybe it needs to be skipped on rhel-8-1, and maybe not (if the broken version never made it into 8.1).  Let's see what tests say.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -82,7 +82,7 @@ export function monitor(name, args, callback, on_close, system) {
                 if (e.name === "ConnectionClosed")
                     on_close(system);
                 else
-                    console.log(e);
+                    throw e;
             });
 }
 
--- a/test/check-application
+++ b/test/check-application
@@ -267,16 +267,12 @@ class TestApplication(testlib.MachineCase):
         b.click(".modal-dialog div .pf-c-alert .pf-c-button")
         b.wait_not_present(".modal-dialog div .alert")
 
-        # HACK: Disable checking for varlink error since it makes podman to coredump
-        # When hack removed, don't forget to also do the check for user containers
-        # See https://github.com/containers/libpod/issues/3897
         # check varlink error
-        # b.set_input_text("#commit-dialog-image-name", "TEST")
-        # b.click(".modal-dialog div .btn-ctr-commit")
-        # b.wait_present(".modal-dialog div")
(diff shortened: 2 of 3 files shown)


marusak:
eeeh, so it seems does not coredump anymore, but also does not show any error? I'll debug it locally.

marusak:
It reports error, does not coredump, but we close the dialog. This needs to be fixed.

marusak:
Ok, so it was a real issue. I also noticed another problem, so sent fix for it here as well.

test/check-application
```diff
@@ -267,16 +267,13 @@ class TestApplication(testlib.MachineCase):
         b.click(".modal-dialog div .pf-c-alert .pf-c-button")
         b.wait_not_present(".modal-dialog div .alert")
 
-        # HACK: Disable checking for varlink error since it makes podman to coredump
-        # When hack removed, don't forget to also do the check for user containers
```
marusak: before someone points this out, we run this part of test on both user container and system one as well.

src/util.js
```diff
@@ -82,7 +82,7 @@ export function monitor(name, args, callback, on_close, system) {
                 if (e.name === "ConnectionClosed")
                     on_close(system);
                 else
-                    console.log(e);
+                    return Promise.reject(e);
```
martinpitt: Instead of converting one exception (IOW, rejected promise) into another, `throw e;` would be a bit more straightforward and also keep the original stack trace.

marusak: I saw this commit https://github.com/cockpit-project/cockpit-podman/commit/2818fa6aae171701edd6ac684809ee184c71f758 and thought I would do the same. But if `throw` is better, I have no problem converting it.

martinpitt: Ah, I was not aware of that commit. @KKoukiou may know the details then. It just looks strange at first sight.

KKoukiou: You can use throw if you want. I already checked and we handle with catch any possible exceptions from these two functions so there is no danger of OOps here.
Also the is not code after this line, so stopping the control flow, which is what `throw` will do will not have any impact here.

PR #285, opened 2020-01-07
Keep order of containers and images

### LINKED ISSUE #222: Sort images and containers by name and by user
We want images and containers to be ordered by name and also grouped by user.
Now it is almost like that, until some container updates, in which case may no longer be grouped with other containers of the same user.

marusak:
Both images and containers need to be grouped by the user. And each
group needs to be ordered by names.
Before this commit, if you would create new container or download a new
image, it would be appended at the bottom of the list. Also there was no
given order of which group (if system or user) would be first.

I decided to show user images above system ones and the same with
containers.

Fixes #222

### DIFF
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -225,6 +225,20 @@ class Images extends React.Component {
                     return false;
                 });
         }
+
+        filtered.sort((a, b) => {
+            // User images are in front of system ones
+            if (this.props.images[a].isSystem !== this.props.images[b].isSystem)
+                return this.props.images[a].isSystem ? 1 : -1;
+            const name_a = this.props.images[a].repoTags ? this.props.images[a].repoTags[0] : "";
+            const name_b = this.props.images[b].repoTags ? this.props.images[b].repoTags[0] : "";
+            if (name_a === "")
+                return 1;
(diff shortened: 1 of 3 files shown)


src/Images.jsx
```diff
@@ -225,6 +225,16 @@ class Images extends React.Component {
...
                 });
         }
+
+        filtered.sort((a, b) => {
+            // User images are in front of system ones
+            if (this.props.images[a].isSystem !== this.props.images[b].isSystem)
+                return this.props.images[a].isSystem ? 1 : -1;
+            const name_a = this.props.images[a].repoTags ? this.props.images[a].repoTags[0] : "";
+            const name_b = this.props.images[b].repoTags ? this.props.images[b].repoTags[0] : "";
+            return name_a > name_b ? 1 : -1;
```
martinpitt: Nitpick: This will sort images without tags before images with a tag, which feels kind of backwards -- the tagged ones are the important ones, the untagged ones are usually just some intermediate layer noise.

marusak: Great point! I've fixed this. Also I plan to hide these images by default, but need to talk to @andreasn about some usable design.

PR #289, opened 2020-01-10
Fix default and minimum value for CPU Shares

### LINKED ISSUE #287: Minimum cpu-shares is 2
We enable to go down to 0, which is invalid. The minimum value is 2.
I am opening issue, since the default is also 1024 and not 0 as we have not set up, but podman behaves a bit differently - see https://github.com/containers/libpod/issues/4822 for discussion.
Once this discussion is resolved, I'll fix both of these issues in our code.

marusak:
Minimum value is 2.
Default value is 1024. We still test that if we don't set it up, it is
0, but for that see https://github.com/containers/libpod/issues/4822

Fixes #287

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -205,7 +205,7 @@ export class ImageRunModal extends React.Component {
             publish: [],
             image: props.image,
             memory: 512,
-            cpuShares: 0,
+            cpuShares: "",
             memoryConfigure: false,
             cpuSharesConfigure: false,
             memoryUnit: 'MiB',
@@ -230,7 +230,7 @@ export class ImageRunModal extends React.Component {
             const memorySize = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
             createConfig.memory = memorySize.toString();
         }
-        if (this.state.cpuSharesConfigure) {
+        if (this.state.cpuSharesConfigure && this.state.cpuShares !== "") {
             createConfig.cpuShares = this.state.cpuShares;
         }
         if (this.state.hasTTY)
@@ -352,9 +352,9 @@ export class ImageRunModal extends React.Component {
                                    type='number'
                                    value={dialogValues.cpuShares}
                                    step={1}
-                                   min={0}
+                                   min={2}
                                    disabled={!this.state.cpuSharesConfigure}
-                                   onChange={e => this.onValueChanged('cpuShares', parseInt(e.target.value))} />
+                                   onChange={e => this.onValueChanged('cpuShares', e.target.value === "" ? "" : parseInt(e.target.value))} />
                         </div>
                     </>
                 }
--- a/test/check-application
+++ b/test/check-application
@@ -710,7 +710,7 @@ class TestApplication(testlib.MachineCase):
         if auth:
             b.set_checked("#run-image-dialog-cpu-priority-checkbox", True)
             b.wait_present("#run-image-dialog-cpu-priority-checkbox:checked")
-            b.wait_present('div.modal-body label:contains("CPU Shares") + div.form-inline > input[value="0"]')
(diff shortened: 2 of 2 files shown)


src/ImageRunModal.jsx
```diff
@@ -205,7 +205,7 @@ export class ImageRunModal extends React.Component {
             publish: [],
             image: props.image,
             memory: 512,
-            cpuShares: 0,
+            cpuShares: 1024,
```
martinpitt: Admitteldy I don't like hardcoding this default very much, but of course the UI needs to know what to set the value to once you enable CPU shares. But given that "0" means "use the internal default", not literally "zero shares", could we just continue to interpret it that way? I. e. with "0" the setting would just appear disabled?

src/ImageRunModal.jsx
```diff
@@ -352,7 +352,7 @@ export class ImageRunModal extends React.Component {
                                    type='number'
                                    value={dialogValues.cpuShares}
                                    step={1}
-                                   min={0}
+                                   min={2}
```
martinpitt: 2 seems so arbitrary.. If we allow a container to have that little CPU share (0.2% of the CPU), then we may just as well cap it to 1. Unless we keep 0 as "default value", aka. "don't explicitly set it"?

src/ImageRunModal.jsx
marusak: I don't think that 0 means that. 0 is not defined anywhere what it mean, and seems that `podman inspects` shows 0 wrongly.
If we don't want 1024, then I would rather suggest not having any value.

src/ImageRunModal.jsx
marusak: `cpu-shares` is relative weight. So I you set up one with `2` and the other one with `4`, it is the same as `512` and `1024`.
```To modify the proportion from the default of 1024, use the --cpu-shares flag to set the weighting to 2 or higher.```

martinpitt: I  see -- I thought it was not just between containers, but between *all* processes in the system. If it's the former, how does the collective set of containers get prioritized to non-container processes?

marusak: from `man podman-run`:

> By  default,  all  containers  get the same proportion of CPU cycles. This proportion can be modified by changing the container's CPU share weighting **relative to the weighting of all other running containers**.



> how does the collective set of containers get prioritized to non-container processes?

I am not the right person to respond to that

martinpitt: With 

    cat /dev/urandom >/dev/null
    sudo podman run --cpu-shares=1024 -it --rm busybox sh -ec 'cat /dev/urandom >/dev/null'

the container cat process uses 25% total CPU (i. e. one of the four cores), in other words, it gets the maximum CPU amount possible. It's excactly the same with `--cpu-shares=2`, so it seems this is indeed only relative to other containers. But two podman run instances with shares 2 and 10 respectively, still both cats get 100% on their individual CPUs, and use up 50% of the CPU in total, though.

So I'm not sure how to even demonstrate the effect of that..

martinpitt: Anyway, closing this thread as the "2" is documented in the podman docs.

src/ImageRunModal.jsx
martinpitt: Right, if it works with `null` and then just not gets passed, that'd be even better of course -- we leave it as internal podman default then.

marusak: Ok, so I dropped the default value.
Forgot to change the commit message, but I'll update that on marging.

marusak:
> Thanks! "2" instead of "null" is so that it gets along with the input element, I suppose?

Yeah, we could either have no check for minimum value or check for specific value. And it is always the best approach to have native elements doing as much work as possible - rather than we would manually check if the value is `>= 2`.

PR #292, opened 2020-01-14
Adjust to new commit API

marusak:


src/app.jsx
```diff
@@ -50,6 +50,7 @@ class Application extends React.Component {
             dropDownValue: 'Everything',
             notifications: [],
             showStartService: true,
+            version: '1.3.0',
```
martinpitt: This looks very arbitrary. Could it default to `null,` or at least 0.0.0?

src/util.js
```diff
@@ -244,3 +244,20 @@ export function unquote_cmdline(text) {
 
     return words;
 }
+
+/*
+ * Return true if first argument is newer or equal version than the second argument.
+ * Both arguments are required to be strings, semver, wher delimiter is '.'.
```
martinpitt: it's not fully semver; let's just say "x.y.z"?

src/util.js
```diff
@@ -244,3 +244,20 @@ export function unquote_cmdline(text) {
...
+ */
+export function isVersionNewerOrEqual(v1, v2) {
```
martinpitt: FTR, a standard comparison function returns -1, 0, or 1. Like cockpit's `compare_versions()` in pkg/storaged/utils.js.

It's ok like that, but just in case you want to keep the code in sync.

src/ContainerCommitModal.jsx
```diff
@@ -79,8 +81,17 @@ class ContainerCommitModal extends React.Component {
...
 
         commitData.changes = [];
-        const cmdData = "CMD=" + cmdStr;
-        commitData.changes.push(cmdData);
+        if (this.state.command.trim() !== "") {
+            let cmdData = "";
+            if (utils.isVersionNewerOrEqual(this.props.version, "1.7.0")) {
+                const words = utils.unquote_cmdline(this.state.command.trim());
+                const cmdStr = words.map(quote).join(", ");
+                cmdData = "CMD [" + cmdStr + "]";
```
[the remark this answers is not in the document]
rhatdan: @martinpitt We need to add CI to libpod from cockpit, to prevent this from happening,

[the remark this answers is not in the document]
marusak: `JSON.stringify()` did not output what exactly was needed, so I went with this.

src/app.jsx
marusak: it is the version we require from spec file, so I think it is a safe and reasonable default.

src/util.js
marusak: nice, had no idea we have this implemented. I'll  take the code from there.

marusak:
Whoo, green.

PR #303, opened 2020-01-27
Webpack improvemts

marusak:


marusak:
I would like to hear options about this. Then I would also apply this in Cockpit.
The second commit (suppressing logs) is something we talked that is annoying (with @allisonkarlitskaya ) as when you have some error in your code, you need to stroll a lot to see this error. With this commit it shows right up.

package.json
```diff
@@ -32,7 +32,7 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-react": "^7.16.0",
     "eslint-plugin-standard": "^4.0.1",
-    "extract-text-webpack-plugin": "^4.0.0-beta.0",
+    "mini-css-extract-plugin": "^0.9.0",
```
martinpitt: :+1: We already did that in https://github.com/cockpit-project/cockpit/commit/dabebcab33d7 

Let's please do that in starter-kit, too.

marusak: Ah nice, was not aware that this is already in Cockpit. I'll do that in starter kit and send the logging suppression into cockpit.

PR #308, opened 2020-02-03
lib: Sync up with current Cockpit master

marusak:
The obvious difference is that items in listing are taller (have more padding):
taller


garrett:
Looks good! This fixes the fonts too.

PR #322, opened 2020-02-24
test: Clean up unexpected system podman images

martinpitt:
fedora-* ≥ 31 images start to have a cockpit/base and its parent fedora
container image in <https://github.com/cockpit-project/bots/pull/562>.
Our tests don't expect that. Keep the test strict (as we want to check
precise ordering) and remove the unexpected images instead.


martinpitt:
testBasicUser now succeeded against the new image. The other failures are unrelated to this PR.

PR #323, opened 2020-02-24
test: Fix testRunImageSystem for podman 1.8

martinpitt:
With https://github.com/containers/libpod/commit/c1e269 the default CPU
share now gets displayed as 1024 instead of 0.


martinpitt:
On the refreshed image it [now fails differently](https://logs.cockpit-project.org/logs/pull-323-20200224-214720-1c232e42-fedora-31-bots-562/log.html#7-2). I can't reproduce this yet, but it clearly uncovers an Oops. This is unrelated to this PR, though.

PR #329, opened 2020-03-03
Makefile: Drop hack for fixed kernel bug

martinpitt:
We have never observed https://bugzilla.redhat.com/show_bug.cgi?id=1762663
in Fedora 32, and it does not seem to happen with current Fedora 31 any
more either.


martinpitt:
This works for me locally, let's confirm on our CI.

PR #356, opened 2020-04-15
test: Make tests non-destructive

martinpitt:
Use recent Cockpit standard test API to handle the directory
restoration.

Clean up/restore user state as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,7 +162,7 @@ bots:
 # checkout Cockpit's test API; this has no API stability guarantee, so check out a stable tag
 # when you start a new project, use the latest release, and update it from time to time
 test/common:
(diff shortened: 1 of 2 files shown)


marusak:
What happened to the html log? The version 216 had some changes about this, but I though it was self contained in test/common.
 https://logs.cockpit-project.org/logs/pull-356-20200415-144703-15769ef2-fedora-31/log.html
https://logs.cockpit-project.org/logs/pull-356-20200415-144703-15769ef2-fedora-32/log.html

martinpitt:
@marusak: Indeed TAP formatting moved to run-tests. I'm not sure if we can easily get that functionality from cockpit-podman and other projects now, or whether we should adjust log.html to parse the `# Result` bits instead, and get along without knowing a total count in advance. @Gundersanne, do you have an idea?

PR #377, opened 2020-05-05
Only use "System service" when user is root

bastilian:
When the user is `root` it will show images and containers twice. In such a case the "system service" can be skipped and avoid lists with duplicate items.

**Before:**
Screenshot 2020-05-05 at 13 58 18

**After:**
Screenshot 2020-05-05 at 13 57 28

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -168,7 +168,6 @@ class Application extends React.Component {
                         // Copy only images that could not be deleted with this event
                         // So when event from system come, only copy user images and vice versa
                         const copyImages = {};
-                        console.log(prevState, reply);
                         Object.entries(prevState.images || {}).forEach(([id, image]) => {
                             if (image.isSystem !== system)
                                 copyImages[id] = image;
@@ -325,12 +324,24 @@ class Application extends React.Component {
         cockpit.script("echo $XDG_RUNTIME_DIR")
                 .done(xrd => {
                     sessionStorage.setItem('XDG_RUNTIME_DIR', xrd.trim());
-                    this.init(false);
+                    if (!utils.isRootUser()) {
+                        this.init(false);
+                    } else {
+                        this.setState({
+                            userImagesLoaded: true,
+                            userContainersLoaded: true,
+                            userServiceAvailable: false
+                        });
+                    }
                 })
                 .fail(e => console.log("Could not read $XDG_RUNTIME_DIR: ", e.message));
     }
 
     checkUserService() {
+        if (utils.isRootUser()) {
+            this.setState({ userServiceExists: false });
+            return;
+        }
         const argv = ["systemctl", "--user", "is-enabled", "io.podman.socket"];
 
         cockpit.spawn(argv, { environ: ["LC_ALL=C"], err: "out" })
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -120,16 +120,6 @@ class Application extends React.Component {
         });
     }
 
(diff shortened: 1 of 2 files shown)


src/app.jsx
```diff
@@ -324,12 +324,24 @@ class Application extends React.Component {
         cockpit.script("echo $XDG_RUNTIME_DIR")
                 .done(xrd => {
                     sessionStorage.setItem('XDG_RUNTIME_DIR', xrd.trim());
-                    this.init(false);
+                    if (!utils.isRootUser()) {
```
martinpitt, line 327: You can't (reliably) call isRootUser() that way, as it returns a promise. Also, I don't quite like the setState() call below, this feels like duplication.

How about simplifying this and checking root-ness in the shell command above? Like

    [ `id -u` -eq 0 ] || echo $XDG_RUNTIME_DIR

and handling the empty case in the `.done` section. This will also nicely cover the case when the runtime dir is not set due to other reasons, in which case podman won't work anyway.

src/app.jsx
```diff
@@ -324,12 +324,24 @@ class Application extends React.Component {
...
+                            userContainersLoaded: true,
+                            userServiceAvailable: false
+                        });
+                    }
                 })
                 .fail(e => console.log("Could not read $XDG_RUNTIME_DIR: ", e.message));
     }
 
     checkUserService() {
+        if (utils.isRootUser()) {
```
martinpitt, line 341: Please let's not repeat that check. Let's rather move the call of this inside the `XDG_RUNTIME_DIR` result processing above, when we already know the answer.

src/util.js
```diff
@@ -268,3 +268,9 @@ export function compare_versions(a, b) {
 
     return a_ints.length - b_ints.length;
 }
+
+export function isRootUser() {
+    return window.cockpit.user().then((user) => {
+        return user.name === 'root';
+    });
```
martinpitt, line 275: This function isn't needed, but just FTR: This should be written as

    return window.cockpit.user().then(user => user.name === 'root');

bastilian:
@martinpitt Thanks for taking reviewing this! I'll work in the changes you suggested and add a test.

I skimmed through `test/check-application`, which I assume are the integration tests and it looks straigthforward for the most part, but I am unclear about what "This requires adding an image to the root podman user instance..." is referring to.

martinpitt:
@bastilian : Maybe I am just confused about what happens. When I wrote the above, I thought that we would need to deal with *three* podman "namespaces" for images/containers: the system one, the "admin user" one, and the "root user" one. But on second thought, there is probably no such thing as "root user" images/containers, but interacting with podman as root would just control the system instances? So then you don't need any image copying, and just need to add a test case that logs in as root and make sure that stuff doesn't get duplicated.

Sorry for the confusion!

marusak:
>  just need to add a test case that logs in as root and make sure that stuff doesn't get duplicated.

Indeed, this is exactly what needs to be done. Should be very straightforward.

marusak:
This got unfortunately stalled. Can you please rebase? And also, do you need any help with this? We should address this ASAP.

KKoukiou:
@bastilian I will shortly post a fixup/squash of your commits + a test. I hope you are ok with me pushing to your branch.

KKoukiou:
@bastilian I force pushed to your branch but somehow the PR does not get updated - GitHub is acting up. 
I will close this one and just create another PR, keeping your as the commit author. Thanks for the contribution.

PR #378, opened 2020-05-05
test: RHEL-8.2 has sufficent podman version now

marusak:
Should fix https://github.com/cockpit-project/bots/pull/826

It is gonna be a bit tricky to get tests green, so I will do:
1. trigger all tests + `rhel-8-2@bots#826`
2. If passed merge the bots PR
3. re-test this


marusak:
Also triggered 8.3 to see if we can drop this special case or not (but we did refresh today, so it should be fine?)

martinpitt:
Current 8.3 nightly still has podman 1.6.4. 8.2.1 nightly indeed has podman 1.9.1.

PR #379, opened 2020-05-05
Update vagrant box

bastilian:
Fedora 28 is not available anymore as box this bumps it to the latest version 32.
It also adds `podman` as package to install for convenience.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,5 +1,5 @@
 Vagrant.configure(2) do |config|
-    config.vm.box = "fedora/28-cloud-base"
+    config.vm.box = "fedora/32-cloud-base"
     config.vm.network "forwarded_port", guest: 9090, host: 9090
 
     if Dir.glob("dist/*").length == 0
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -27,7 +27,7 @@ Vagrant.configure(2) do |config|
         systemctl enable cockpit.socket
         systemctl start cockpit.socket
 
-        systemctl enable io.podman.{socket,service}
-        systemctl start io.podman.{socket,service}
+        systemctl enable io.podman.socket
+        systemctl start io.podman.socket
     EOF
 end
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -20,11 +20,14 @@ Vagrant.configure(2) do |config|
...
-        sudo dnf install -y cockpit
+        sudo dnf install -y cockpit podman
 
         printf "[WebService]\nAllowUnencrypted=true\n" > /etc/cockpit/cockpit.conf
 
         systemctl enable cockpit.socket
         systemctl start cockpit.socket
+
+        systemctl enable podman
+        systemctl start podman
```
martinpitt, line 31: Shouldn't that be `io.podman.socket`? The .service unit shouldn't have an `[Install]` section in the first place, that's a bug in podman. Do you know what the difference is between the two sockets? Is podman.socket the new-style REST API (which current master doesn't support yet, @marusak is working on it), and io.podman.socket the varlink API?

marusak: Indeed, `podman.{socket,service}` is the new REST API that is not yet supported and `io.podman.{socket,service}` is the varlink one, that we currently support.

Vagrantfile
```diff
@@ -20,11 +20,14 @@ Vagrant.configure(2) do |config|
...
-        sudo dnf install -y cockpit
+        sudo dnf install -y cockpit podman
 
         printf "[WebService]\nAllowUnencrypted=true\n" > /etc/cockpit/cockpit.conf
 
         systemctl enable cockpit.socket
         systemctl start cockpit.socket
+
+        systemctl enable io.podman.{socket,service}
+        systemctl start io.podman.{socket,service}
```
martinpitt, line 31: Please only enable the .socket. The .service really shouldn't have an `[Install]` section in the first place (this is a podman bug).

bastilian:
@martinpitt Thank you for reviewing! It now enables only the socket unit.

PR #381, opened 2020-05-08
Update translations from Fedora Weblate

cockpituous:
Update translations from Fedora Weblate
 * [x] po-refresh


martinpitt:
@marusak: The failure on [fedora-31](https://logs.cockpit-project.org/logs/pull-381-20200510-110616-1e9014da-fedora-31/log.html) seems quite persistent, and it also happens in e. g. PR #382. It seems to me that this is simply fallout from the Services page rewrite, that the `#services` selector doesn't exist any more?

PR #390, opened 2020-05-26
Adjust to new cockpit + flake fix

marusak:
The flake was presented as
```
wait_js_cond(ph_text_is(".container-logs .xterm-accessibility-tree > div:nth-child(6)","6")): .container-logs .xterm-accessibility-tree > div:nth-child(6) not found
```
for example [here](https://logs.cockpit-project.org/logs/pull-385-20200523-055607-c3346b67-fedora-32/log.html)

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -86,7 +86,7 @@ class TestApplication(testlib.MachineCase):
             self.allow_authorize_journal_messages()
             self.allow_browser_errors("Failed to start system io.podman.socket.*")
 
-        self.login_and_go("/podman", authorized=auth)
+        self.login_and_go("/podman", authorized=auth, superuser=auth)
         b.wait_present("#app")
 
         b.wait_present(".content-filter input")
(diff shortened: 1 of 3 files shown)


test/check-application
```diff
@@ -704,7 +704,7 @@ class TestApplication(testlib.MachineCase):
         b.set_checked("#run-image-dialog-tty", True)
 
         # Set up command line
-        b.set_input_text('#run-image-dialog-command', "sh -c 'for i in $(seq 20); do sleep 1; echo $i; done'")
+        b.set_input_text('#run-image-dialog-command', "sh -c 'for i in $(seq 20); do sleep 1; echo $i; done; sleep infinity'")
```
martinpitt, line 707: That's a really curious command now :-) Do  we already  have a different one that just does `sleep infinity`, and we need that for disambiguation?

marusak: no really. I use this command to test logs, thats why I have there that for loop. It however cannot exit. I could just do the loop up to 1000 or so and that would do the trick, but thought why not just keeping it alive explicitly.

PR #392, opened 2020-05-26
Makefile: Call run-tests in make check target

Gundersanne:
- [x] https://github.com/cockpit-project/cockpit/pull/14157

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ vm: $(VM_IMAGE)
 
 # run the browser integration tests; skip check for SELinux denials
 check: $(NODE_MODULES_TEST) $(VM_IMAGE) test/common
-	TEST_AUDIT_NO_SELINUX=1 test/check-application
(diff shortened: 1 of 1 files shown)


martinpitt:
The cockpit PR to move run-tests landed now. This now needs to bump the test/common checkout tag, so for testing we can just check out a SHA -- and after tomorrow's release, we can use the 220 tag.

martinpitt:
This seems to work fine locally, I pushed a temporary hack to checkout the SHA to run tests on CI.

martinpitt:
Yay, green and proper logs! I don't see any naughties, just some retried flakes.

martinpitt:
Adding "blocked" again, let's finish/land this after cockpit 220 release tomorrow?

marusak:
:heart: thanks!
Lets indeed wait until 220 is tagged (later today) and do this properly.

PR #393, opened 2020-05-28
Stop importing cockpit's base1/patternfly.css

KKoukiou:
This is deprecated API and will be dropped at some point, in favor
of projects shipping their own CSS.

Follow https://github.com/cockpit-project/starter-kit/blob/master/webpack.config.js
approach on how to use PF CSS.


lib/patternfly/_opensans-fonts.scss
```diff
@@ -0,0 +1,31 @@
+/*
+ * Keep in sync with https://github.com/cockpit-project/cockpit/tree/master/src/base1/_opensans-fonts.scss
```
martinpitt, line 2: I'm a bit surprised -- I thought we'd use RedHat fonts for everything in Cockpit now. Why still OpenSans? I thought that was only a thing still for connecting to RHEL 7 version of Cockpit? (195, before the PF4 move)

package.json
```diff
@@ -32,17 +32,18 @@
...
+    "mini-css-extract-plugin": "^0.9.0",
     "optimize-css-assets-webpack-plugin": "^5.0.3",
-    "svgo": "1.3.0",
     "po2json": "^1.0.0-alpha",
-    "sass-loader": "^7.0.3",
+    "sass-loader": "^8.0.2",
     "sizzle": "^2.3.3",
     "stdio": "^0.2.7",
+    "string-replace-loader": "^2.3.0",
+    "svgo": "1.3.0",
```
martinpitt, line 46: This should use ^

webpack.config.js
```diff
@@ -123,28 +123,90 @@ module.exports = {
                 test: /\.(jsx|js)$/
             },
             {
-                test: /\.(scss|css)$/,
+                test: /\.less$/,
```
martinpitt, line 126: Are these just for our remaining three lib/*.less files (imported from cockpit)? (listing, variables, form-layout), or do we also need less for some node_modules/ imports? IOW, can we drop this once we convert these three lib files to scss in cockpit and sync them over?

martinpitt:
Tested and works okay. One thing that I noticed is that we neither minify nor compress index.css, so it grew quite a bit from 119 to 568 kB. It's fine to fix that in a follow-up, of course (beginning in starter-kit)

webpack.config.js
KKoukiou: We can drop this once we sync the following:
```
[kkoukiou@sequioa cockpit-podman]$ find lib/ -name *.less
lib/variables.less
lib/form-layout.less
lib/listing.less
[kkoukiou@sequioa cockpit-podman]$ find src/ -name *.less
src/css-overrides/toast.less

```
I am happy to do that in seperate PR.

lib/patternfly/_opensans-fonts.scss
KKoukiou: We do use RedHat fonts in everything AFAIK. tbh I just included it because of misunderstanding  - when I asked recently if we need to keep shipping the OpenSans fonts with patternfly.css in dist/base1

Anyway - I will remove it from here.

PR #408, opened 2020-06-22
eslint: Introduce eslint-a11y plugin

KKoukiou:


marusak:
Thanks! But seems that two tests are not happy. Can you please take a look?

KKoukiou:
> Thanks! But seems that two tests are not happy. Can you please take a look?

Ups, that was the Listing component sync - tests needed adjustment of course :)

PR #415, opened 2020-07-02
test: Adjust to the new format of `podman inspect`

marusak:


marusak:
I used image from https://github.com/cockpit-project/bots/pull/1037 locally and got completely different behaviour that the CI got. Lets see what it thinks here.

martinpitt:
I'd just accept either format and use `assertRegex()`, as otherwise we'll have to adjust the test a lot. But I don't veto this approach,  of course.

marusak:
Ok, tests are as expected (with one retry, it it not greatly stable, but hooopefully the new API could do things better...)

marusak:
> I'd just accept either format and use assertRegex(), as otherwise we'll have to adjust the test a lot. But I don't veto this approach, of course.

Right, it is that the regex will stay there forever, this workaround can be soon-ish dropped. I think this new version won't reach  F31, so we might want to just turn it into `if m.image == 'fedora-31'` once it is in 8.3.

martinpitt:
Unblocking. https://github.com/cockpit-project/bots/pull/1037 is ready to land, so let's land this now.

PR #419, opened 2020-07-09
test: Apply podman 2.0 inspect format to Fedora 31 as well

martinpitt:
Invert the test, so that we have a shrinking instead of growing list.


martinpitt:
This fixes https://github.com/cockpit-project/bots/pull/1056 and needs to land in lockstep.

PR #422, opened 2020-07-13
Ui fix - align of trash bin, border of container list

### LINKED ISSUE #410: Some CSS issues on Epiphany
Screenshot 2020-06-25 at 13 10 04

The Tabs underline is wrong and the Trash Bin icon is wrongly aligned.

EricJeffrey:
fix #410


EricJeffrey:
@marusak If they are sync from cockpit, will changing them affect other components in cockpit?

EricJeffrey:
Also, if the render part is written according to styles in `lib/`, maybe it's better to make changes in `src/` ?

marusak:
Would you mind if I would block this for a bit (like today)?
I am currently rewriting the listing to use tables. (As suggested in https://github.com/cockpit-project/cockpit-podman/pull/244#issuecomment-657593710) That may or may not resolve this issue. If it won't, then this is indeed bug in cockpit itself and we need to adjust it there.

EricJeffrey:
Yeah surely I'm ok with it, and honestly, changing file in `lib/` dir may not be a elegant way.

marusak:
It is still issue and there is one more issue in the library (see #244) so we need to address these problems in Cockpit itself.

EricJeffrey:
Actually, I'm not very clear about these stuff😂, should I just close this PR?

marusak:
> Actually, I'm not very clear about these stuffjoy, should I just close this PR?

I'll investigate on this, so I think we can close this. Thank you anyway!

PR #424, opened 2020-07-13
Arrange container and image details in a grid

garrett:
When reviewing #244, I noticed the details for the containers and images are in a simple definition list. I've added a bit of CSS that uses PF4 variables to arrange the same content in a simple grid, consistent with the look of PF4's forms. (This isn't a form, so it shouldn't use `<label>`s and various form widgets. A `dl` is fine.)

### DIFF
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -635,3 +635,23 @@ table.drive-list td:nth-child(2) img {
 .pf-c-alert__description {
     overflow-wrap: anywhere;
 }
+
+/* Place details definition list into a grid, with PF4-like style */
+.container-details,
+.image-details {
+    display: grid;
+    grid-template-columns: [label] auto [value] 1fr;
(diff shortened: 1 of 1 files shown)


garrett:
## Before
Screenshot_2020-07-13 Podman Containers - garrett Rain(1)

## After
Screenshot_2020-07-13 Podman Containers - garrett Rain

garrett:
Yeah, this is more proper.

That said; you're right — the VM's overview does use ct-form. However, ct-form does support various elements under it. Perhaps I should make sure that ct-form works when applied to a dl with dt below (and therefore make this solution more generic)? (We'd have to add classes to the dt at the moment to make them treated like labels... but supporting `> dt` as labels probably has merit.)

marusak:
Thank you!

> and therefore make this solution more generic

if you feel like doing it in this PR, then please go ahead, otherwise we can land this.

PR #426, opened 2020-07-14
[rhel8] metainfo: Fix launchable and update description

martinpitt:
`<launchable>` must coincide with the actual URL path defined by the
manifest.

Also remove the redundant "Cockpit" from `<name>`, as all of these
extensions are about Cockpit. This makes the "Applications" page more
consistent.

Remove the period from `<summary>`, as the spec [1] suggests.

[1] https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent

https://bugzilla.redhat.com/show_bug.cgi?id=1854673

Cherry-picked from master commit 8dc58334d

### DIFF
--- a/org.cockpit-project.podman.metainfo.xml
+++ b/org.cockpit-project.podman.metainfo.xml
@@ -2,9 +2,9 @@
 <component type="addon">
   <id>org.cockpit_project.podman</id>
   <metadata_license>CC0-1.0</metadata_license>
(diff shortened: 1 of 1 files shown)


martinpitt:
We haven't run centos-8-stream anywhere yet, but let's give it a try. If it fails, please ignore :wink:

Issue #363, opened 2020-04-23
Clone or Copy feature is missing


S-B-V:
the possibility to clone or copy an existing container is really annoying. It is not funny to configure the same stuff again and again in case of update or SELinux destroying the container...

Please no comments like do it in CLI it is quick, I do not care... Otherwise we can delete Cockpit ;-)

marusak:
Interesting idea, haven't seen it before :)
Does podman (or docker for that matter) provide such functionality? If so, can you please provide cli command that does that? (I am not suggesting you should do it on command line, I just need to see how this behaves and what it actually does to be able to model it for cockpit-podman as well).
If it is not a thing in command line clients, what should it do?

S-B-V:
Hi marusak,
example for update of a container:
You stop the existing one and you issue the same command again, you just change the Name.
But as Cockpit is UI based it would be great to just click on clone and, as it has all information,  fills the needed fields out so that I just have to change the name. Everything is filled automatically by using the known information from the existing container, 

Here an example how it is handled by RancherOS which I currently use as main system for container
I just started to test FCOS and so I am testing cockpit as well :-).

![Bildschirmfoto von 2020-04-24 09-32-09](https://user-images.githubusercontent.com/11391862/80190798-21ef7d80-8615-11ea-8696-58da899987d6.png)
![Bildschirmfoto von 2020-04-24 09-32-16](https://user-images.githubusercontent.com/11391862/80190866-36337a80-8615-11ea-84f8-6c958a54f540.png)
![Bildschirmfoto von 2020-04-24 09-32-19](https://user-images.githubusercontent.com/11391862/80190868-37fd3e00-8615-11ea-9f0f-4159549cc70c.png)
![Bildschirmfoto von 2020-04-24 09-32-21](https://user-images.githubusercontent.com/11391862/80190876-3895d480-8615-11ea-9a42-6b72d6423f8d.png)
![Bildschirmfoto von 2020-04-24 09-32-24](https://user-images.githubusercontent.com/11391862/80190880-39c70180-8615-11ea-81db-c6b3dda55a24.png)
![Bildschirmfoto von 2020-04-24 09-32-32](https://user-images.githubusercontent.com/11391862/80190882-3b90c500-8615-11ea-933b-4df0681c6c1c.png)
[...]

PR #430, opened 2020-07-15
Support local checkpoint and restore

lenticularis39:
Adds support of local container checkpoint and restore (without export and import) using the CRIU integration with Podman.

Tests for this feature are not included, because Podman checkpoint and restore doesn't work on Fedora 31 yet.

checkpoint-1
checkpoint-2
checkpoint-3
checkpoint-4

### DIFF
--- a/src/ContainerCheckpointModal.jsx
+++ b/src/ContainerCheckpointModal.jsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import { Modal } from 'patternfly-react';
+import { Button, Checkbox } from '@patternfly/react-core';
+import cockpit from 'cockpit';
+import * as utils from './util.js';
+
+const _ = cockpit.gettext;
+
+class ContainerCheckpointModal extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            keep: false,
+            leaveRunning: false,
+            tcpEstablished: false,
+            ignoreRootFS: false
+        };
+        this.handleChange = this.handleChange.bind(this);
+    }
+
+    handleChange(checked, event) {
+        if (event.target.type === "checkbox")
+            this.setState({ [event.target.name]: event.target.checked });
+    }
+
+    render() {
+        return (
+            <Modal show={this.props.selectContainerCheckpointModal}>
+                <Modal.Header>
+                    <Modal.Title>
+                        {cockpit.format(_("Checkpoint container $0"),
+                                        utils.truncate_id(this.props.containerWillCheckpoint.Id))}
+                    </Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    <div className="ct-form">
+                        <Checkbox label={_("Keep all temporary checkpoint files")} id="checkpoint-dialog-keep"
+                                  name="keep" isChecked={this.state.keep} onChange={this.handleChange} />
+                        <Checkbox label={_("Leave running after writing checkpoint to disk")}
+                                  id="checkpoint-dialog-leaveRunning" name="leaveRunning"
+                                  isChecked={this.state.leaveRunning} onChange={this.handleChange} />
+                        <Checkbox label={_("Support preserving established TCP connections")}
+                                  id="checkpoint-dialog-tcpEstablished" name="tcpEstablished"
+                                  isChecked={this.state.tcpEstablished} onChange={this.handleChange} />
+                        <Checkbox label={_("Do not include root file-system changes when exporting")}
+                                  id="checkpoint-dialog-ignoreRootFS" name="ignoreRootFS"
+                                  isChecked={this.state.ignoreRootFS} onChange={this.handleChange} />
+                    </div>
+                </Modal.Body>
+                <Modal.Footer>
+                    <Button variant="primary" isDisabled={this.props.checkpointInProgress}
+                            onClick={() => this.props.handleCheckpointContainer(this.state)}>
+                        {_("Checkpoint")}
+                    </Button>
+                    <Button variant="link" onClick={this.props.handleCheckpointContainerDeleteModal}>
+                        {_("Cancel")}
+                    </Button>
+                    {this.props.checkpointInProgress && <div className="spinner spinner-sm pull-right" />}
+                </Modal.Footer>
+            </Modal>
+        );
+    }
+}
+
+export default ContainerCheckpointModal;
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -10,6 +10,8 @@ import ContainerTerminal from './ContainerTerminal.jsx';
 import ContainerLogs from './ContainerLogs.jsx';
 import { DropDown } from './Dropdown.jsx';
 import ContainerDeleteModal from './ContainerDeleteModal.jsx';
+import ContainerCheckpointModal from './ContainerCheckpointModal.jsx';
+import ContainerRestoreModal from './ContainerRestoreModal.jsx';
 import ContainerRemoveErrorModal from './ContainerRemoveErrorModal.jsx';
 import * as utils from './util.js';
 import * as client from './client.js';
@@ -23,8 +25,14 @@ class Containers extends React.Component {
         super(props);
         this.state = {
             selectContainerDeleteModal: false,
+            selectContainerCheckpointModal: false,
+            selectContainerRestoreModal: false,
             setContainerRemoveErrorModal: false,
             containerWillDelete: {},
+            containerWillCheckpoint: {},
+            containerWillRestore: {},
+            checkpointInProgress: false,
+            restoreInProgress: false,
             width: 0,
         };
         this.renderRow = this.renderRow.bind(this);
@@ -34,7 +42,11 @@ class Containers extends React.Component {
         this.stopContainer = this.stopContainer.bind(this);
         this.deleteContainer = this.deleteContainer.bind(this);
         this.handleCancelContainerDeleteModal = this.handleCancelContainerDeleteModal.bind(this);
+        this.handleCheckpointContainerDeleteModal = this.handleCheckpointContainerDeleteModal.bind(this);
+        this.handleRestoreContainerDeleteModal = this.handleRestoreContainerDeleteModal.bind(this);
         this.handleRemoveContainer = this.handleRemoveContainer.bind(this);
+        this.handleCheckpointContainer = this.handleCheckpointContainer.bind(this);
+        this.handleRestoreContainer = this.handleRestoreContainer.bind(this);
         this.handleCancelRemoveError = this.handleCancelRemoveError.bind(this);
         this.handleForceRemoveContainer = this.handleForceRemoveContainer.bind(this);
 
@@ -56,11 +68,18 @@ class Containers extends React.Component {
                 setContainerRemoveErrorModal: true,
             }));
         } else {
-            this.setState((prevState) => ({
+            this.setState({
                 containerWillDelete: container,
                 selectContainerDeleteModal: true,
-            }));
+            });
         }
+    }
+
+    checkpointContainer(container) {
+        this.setState({
+            containerWillCheckpoint: container,
+            selectContainerCheckpointModal: true,
+        });
     }
 
     stopContainer(container, force) {
@@ -93,6 +112,13 @@ class Containers extends React.Component {
                     const error = cockpit.format(_("Failed to restart container $0"), container.Names);
                     this.props.onAddNotification({ type: 'danger', error, errorDetail: ex.message });
                 });
+    }
+
+    restoreContainer(container) {
+        this.setState({
+            containerWillRestore: container,
+            selectContainerRestoreModal: true,
+        });
     }
 
     renderRow(containersStats, container, containerDetail) {
@@ -156,11 +182,18 @@ class Containers extends React.Component {
             </Button>,
         ];
         if (!isRunning) {
-            actions.push(
-                <Button key={container.Id + "start"} variant="secondary" onClick={() => this.startContainer(container)}>
-                    {_("Start")}
-                </Button>
-            );
+            if (container.isSystem && container.hasCheckpoint) {
+                const runActions = [];
+                runActions.push({ label: _("Start"), onActivate: () => this.startContainer(container) });
+                runActions.push({ label: _("Restore"), onActivate: () => this.restoreContainer(container) });
+                actions.push(<DropDown key={_(container.Id) + "stop"} actions={runActions} />);
+            } else {
+                actions.push(
+                    <Button key={container.Id + "start"} variant="secondary" onClick={() => this.startContainer(container)}>
+                        {_("Start")}
+                    </Button>
+                );
+            }
         } else {
             const restartActions = [];
             const stopActions = [];
@@ -171,6 +204,8 @@ class Containers extends React.Component {
 
             stopActions.push({ label: _("Stop"), onActivate: () => this.stopContainer(container) });
             stopActions.push({ label: _("Force Stop"), onActivate: () => this.stopContainer(container, true) });
+            if (container.isSystem)
+                stopActions.push({ label: _("Checkpoint"), onActivate: () => this.checkpointContainer(container) });
             actions.push(<DropDown key={_(container.Id) + "stop"} actions={stopActions} />);
         }
 
@@ -204,6 +239,50 @@ class Containers extends React.Component {
                     const error = cockpit.format(_("Failed to remove container $0"), this.state.containerWillDelete.names);
                     this.props.onAddNotification({ type: 'danger', error, errorDetail: ex.message });
                 });
+    }
+
+    handleCheckpointContainer(args) {
+        const container = this.state.containerWillCheckpoint;
+        this.setState({ checkpointInProgress: true });
+        client.postContainer(container.isSystem, "checkpoint", container.Id, args)
+                .catch(ex => {
+                    const error = cockpit.format(_("Failed to checkpoint container $0"), container.Names);
+                    this.props.onAddNotification({ type: 'danger', error, errorDetail: ex.message });
+                })
+                .finally(() => {
+                    this.setState({
+                        checkpointInProgress: false,
+                        selectContainerCheckpointModal: false
+                    });
+                });
+    }
+
+    handleRestoreContainer(args) {
+        const container = this.state.containerWillRestore;
+        this.setState({ restoreInProgress: true });
+        client.postContainer(container.isSystem, "restore", container.Id, args)
+                .catch(ex => {
+                    const error = cockpit.format(_("Failed to restore container $0"), container.Names);
+                    this.props.onAddNotification({ type: 'danger', error, errorDetail: ex.message });
+                })
+                .finally(() => {
+                    this.setState({
+                        restoreInProgress: false,
+                        selectContainerRestoreModal: false
+                    });
+                });
+    }
+
+    handleCheckpointContainerDeleteModal() {
+        this.setState((prevState) => ({
+            selectContainerCheckpointModal: !prevState.selectContainerCheckpointModal,
+        }));
+    }
+
+    handleRestoreContainerDeleteModal() {
+        this.setState((prevState) => ({
+            selectContainerRestoreModal: !prevState.selectContainerRestoreModal,
+        }));
     }
 
     handleCancelRemoveError() {
@@ -268,6 +347,22 @@ class Containers extends React.Component {
                 handleCancelContainerDeleteModal={this.handleCancelContainerDeleteModal}
                 handleRemoveContainer={this.handleRemoveContainer}
             />;
+        const containerCheckpointModal =
+            <ContainerCheckpointModal
+                selectContainerCheckpointModal={this.state.selectContainerCheckpointModal}
+                handleCheckpointContainer={this.handleCheckpointContainer}
+                handleCheckpointContainerDeleteModal={this.handleCheckpointContainerDeleteModal}
+                containerWillCheckpoint={this.state.containerWillCheckpoint}
+                checkpointInProgress={this.state.checkpointInProgress}
+            />;
+        const containerRestoreModal =
+            <ContainerRestoreModal
+                selectContainerRestoreModal={this.state.selectContainerRestoreModal}
+                handleRestoreContainer={this.handleRestoreContainer}
+                handleRestoreContainerDeleteModal={this.handleRestoreContainerDeleteModal}
+                containerWillCheckpoint={this.state.containerWillRestore}
+                restoreInProgress={this.state.restoreInProgress}
+            />;
         const containerRemoveErrorModal =
             <ContainerRemoveErrorModal
                 setContainerRemoveErrorModal={this.state.setContainerRemoveErrorModal}
@@ -293,6 +388,8 @@ class Containers extends React.Component {
                     rows={rows}
                 />
                 {containerDeleteModal}
+                {containerCheckpointModal}
+                {containerRestoreModal}
                 {containerRemoveErrorModal}
                 {this.state.showCommitModal && containerCommitModal}
             </div>
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -143,8 +143,28 @@ class Application extends React.Component {
                 .catch(e => console.log(e));
     }
 
+    isContainerCheckpointPresent(id, system) {
+        return client.inspectContainer(system, id)
+                .then(inspectResult => {
+                    const checkpointPath = inspectResult.StaticDir + "/checkpoint";
+                    return cockpit.script(`test -d ${checkpointPath}; echo $?`, [],
+                                          system ? { superuser: "require" } : {});
+                })
+                .then(scriptResult => scriptResult === "0\n");
+    }
+
     updateContainersAfterEvent(system, init) {
         client.getContainers(system)
+                .then(reply => Promise.all(
+                    (reply || []).map(container =>
+                        this.isContainerCheckpointPresent(container.Id, system)
+                                .then(checkpointPresent => {
+                                    const newContainer = Object.assign({}, container);
+                                    newContainer.hasCheckpoint = checkpointPresent;
+                                    return newContainer;
+                                })
+                    )
+                ))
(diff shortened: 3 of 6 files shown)


adrianreber:
> Tests for this feature are not included, because Podman checkpoint and restore doesn't work on Fedora 31 yet.

With cgroup2 Podman uses crun which does not yet support checkpoint/restore (https://github.com/containers/crun/issues/71). The checkpoint/restore support in crun is almost finished it just needs one more CRIU PR to be merged (https://github.com/checkpoint-restore/criu/pull/1063). Once the CRIU PR is merged, the necessary changes to crun need be merged and then cgroup2 based systems can use checkpoint/restore with Podman. All code is ready it just waits on being merged and make its way to the distributions.

marusak:
Oh wow, this is cool! Thank you for working on this :)
Firstly I would like to hear from our designers (@andreasn can you please take a look?) as I am not convinced about some of those design decisions (like 'checkpoint'/'restore' being under stop/start actions).

And if I understand this correctly, there is currently no enviroment in which I could try it? (without building my own crun/criu)

adrianreber:
You can switch to cgroup1 using

```
grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
```

on Fedora. Or on centos/rhel 8

src/Containers.jsx
```diff
@@ -94,6 +111,13 @@ class Containers extends React.Component {
                 });
     }
 
+    restoreContainer(container) {
+        this.setState((prevState) => ({
```
KKoukiou, line 115: Same here!

src/Containers.jsx
```diff
@@ -198,6 +230,42 @@ class Containers extends React.Component {
                 });
     }
 
+    handleCheckpointContainer(args) {
+        const container = this.state.containerWillCheckpoint;
+        this.setState((prevState) => ({
+            selectContainerCheckpointModal: false
+        }));
+        client.postContainer(container.isSystem, "checkpoint", container.Id, args)
```
KKoukiou, line 238: In trivial cases both checkpoint and restore operations are very fast. But if the container in interest is running some memory heavy operation when we want to checkpoint that it might take some time to store all these into files right?

So I think we should keep the modal open and only close that when the operation is really finished.
We should show a spinner, disable the Checkpoint button etc.
Something you can achieve with the diff below:
``` diff
diff --git src/ContainerCheckpointModal.jsx src/ContainerCheckpointModal.jsx
index d44bdee8..e5bf3d83 100644
--- src/ContainerCheckpointModal.jsx
+++ src/ContainerCheckpointModal.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Modal } from 'patternfly-react';
-import { Button, Checkbox } from '@patternfly/react-core';
+import { Button, Checkbox, Spinner } from '@patternfly/react-core';
 import cockpit from 'cockpit';
 import * as utils from './util.js';
 
@@ -48,8 +48,8 @@ class ContainerCheckpointModal extends React.Component {
                     </div>
                 </Modal.Body>
                 <Modal.Footer>
-                    <Button variant="primary" onClick={() => this.props.handleCheckpointContainer(this.state)}>
-                        {_("Checkpoint")}
+                    <Button variant="primary" isDisabled={this.props.checkpointInProgress} onClick={() => this.props.handleCheckpointContainer(this.state)}>
+                        {!this.props.checkpointInProgress ? _("Checkpoint") : <Spinner size="sm" className="pull-right" />}
                     </Button>
                     <Button variant="link" onClick={this.props.handleCheckpointContainerDeleteModal}>
                         {_("Cancel")}
diff --git src/Containers.jsx src/Containers.jsx
index dbdcded0..7a70c564 100644
--- src/Containers.jsx
+++ src/Containers.jsx
@@ -232,13 +232,14 @@ class Containers extends React.Component {
 
     handleCheckpointContainer(args) {
         const container = this.state.containerWillCheckpoint;
-        this.setState((prevState) => ({
-            selectContainerCheckpointModal: false
-        }));
+        this.setState({ checkPointInProgress: true });
         client.postContainer(container.isSystem, "checkpoint", container.Id, args)
                 .catch(ex => {
                     const error = cockpit.format(_("Failed to checkpoint container $0"), container.Names);
                     this.props.onAddNotification({ type: 'danger', error, errorDetail: ex.message });
+                })
+                .finally(() => {
+                    this.setState({ selectContainerCheckpointModal: false, checkPointInProgress: false });
                 });
     }
 
@@ -333,6 +334,7 @@ class Containers extends React.Component {
                 selectContainerCheckpointModal={this.state.selectContainerCheckpointModal}
                 handleCheckpointContainer={this.handleCheckpointContainer}
                 handleCheckpointContainerDeleteModal={this.handleCheckpointContainerDeleteModal}
+                checkPointInProgress={this.state.checkPointInProgress}
                 containerWillCheckpoint={this.state.containerWillCheckpoint}
             />;
         const containerRestoreModal =

```

Let me know if you have some objections here or you prefer that I implement it as a followup.

KKoukiou:
So @lenticularis39 let's add at least a test that checks the error message from the API.
Should be 'Configured runtime does not support checkpoint/restore'

@marusak I would be actually ok to have it in with a test checking the error. I tested it locally with runc runtime and cgroupsv1 of course and it works as expected.

@lenticularis39 let me know if you are interested to add the test or I should do it.

marusak:
> I would be actually ok to have it in with a test checking the error. I tested it locally with runc runtime and cgroupsv1 of course and it works as expected.

Hmm, I am not so convinced. Surely we should test that if fails in some reasonable manner when on cgroupsv2 but I think we should also test it with cgroupsv1. We can edit VM to run on V1.

src/Containers.jsx
lenticularis39: That's a good point, checkpoint/restore can take longer on heavier containers (I only tested it on a simple Node.js counter) - I'll implement the spinner.

lenticularis39:
As of the test - I've already written it and didn't include it only because it doesn't work with cgroup2 (therefore on Fedora 31/32). I tried adding the switch to cgroup1 suggested by @adrianreber to `test/vm.install` and it works, so I can include it into the PR.

KKoukiou:
> As of the test - I've already written it and didn't include it only because it doesn't work with cgroup2 (therefore on Fedora 31/32). I tried adding the switch to cgroup1 suggested by @adrianreber to `test/vm.install` and it works, so I can include it into the PR.

@marusak what do you think about this? switching to cgroup v1 in vm.install means that all test which run with cgroupv1 and as a result with the non default container runtime, aka runc. I guess instead we should just check that the expected error is shown by the API, 'container runtime does not support checkpoint/restore', instead of forcing all tests run on cgroupv1 setup. What do you think?

marusak:
>  what do you think about this? switching to cgroup v1 in vm.install

No ideal. But I guess we can create destructive test, that would set the grubby option and reboot the VM and then do the testing.

KKoukiou:
> > what do you think about this? switching to cgroup v1 in vm.install
> 
> No ideal. But I guess we can create destructive test, that would set the grubby option and reboot the VM and then do the testing.

Good point - that would work. And it would not affect the rest of the tests.

lenticularis39:
OK, I'll add the grubby command to the test instead of vm.install.

andreasn:
> Firstly I would like to hear from our designers (@andreasn can you please take a look?) as I am not convinced about some of those design decisions (like 'checkpoint'/'restore' being under stop/start actions).

Maybe I got this backwards, but as I understand from [an article on podman checkpoints](https://podman.io/blogs/2018/10/10/checkpoint-restore.html) you run **Restore** to start the container from the checkpoint rather than **Start**. So in that case I can see how the items fit in that menu.

src/ContainerCheckpointModal.jsx
```diff
@@ -0,0 +1,63 @@
...
+                    </Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    <div className="ct-form">
+                        <Checkbox label={_("Keep all temporary checkpoint files")} id="keep" name="keep"
+                                  isChecked={this.state.keep} onChange={this.handleChange} />
+                        <Checkbox label={_("Leave the container running after writing checkpoint to disk")}
+                                  id="leaveRunning" name="leaveRunning" isChecked={this.state.leaveRunning}
+                                  onChange={this.handleChange} />
+                        <Checkbox label={_("Checkpoint a container with established TCP connections")}
```
andreasn, line 42: Is this the container you're working with or another container? For me it reads as the latter.
Would need to rephrase for clarity. Would `remember established TCP connections` work? (if that is what it does, I'm not super-familiar with exactly what this option does)

src/ContainerCheckpointModal.jsx
```diff
@@ -0,0 +1,63 @@
...
+                    <Modal.Title>
+                        {cockpit.format(_("Checkpoint container $0"),
+                                        utils.truncate_id(this.props.containerWillCheckpoint.Id))}
+                    </Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    <div className="ct-form">
+                        <Checkbox label={_("Keep all temporary checkpoint files")} id="keep" name="keep"
+                                  isChecked={this.state.keep} onChange={this.handleChange} />
+                        <Checkbox label={_("Leave the container running after writing checkpoint to disk")}
```
andreasn, line 39: I would make **the** into **this** for slightly more clarity. Was also considering if we should drop the **container** part, but even though that makes the text shorter, it might make it more unclear.

src/ContainerCheckpointModal.jsx
lenticularis39: Good point - I copied the description of the options from the API documentation and from Podman command line help, but the context is clearly different here. I don't think that "remember" is the right word for what the option does, `support preserving established TCP connections` sounds better to me.

lenticularis39: A similar problem is in the restore option description, where simply dropping the `a container` (making it `restore with established TCP connections`) should suffice.

src/ContainerCheckpointModal.jsx
lenticularis39: "Leave" (in the meaning used here) is a transitive verb, I think it should have an object. "This container" seems to me like implying that there are another containers in the context, which isn't the case (the dialog mentions a specific container in the heading). In my opinion using "the container" is better than "this container" or dropping the object altogether.

lenticularis39:
> Maybe I got this backwards, but as I understand from [an article on podman checkpoints](https://podman.io/blogs/2018/10/10/checkpoint-restore.html) you run **Restore** to start the container from the checkpoint rather than **Start**. So in that case I can see how the items fit in that menu.

That's correct, restore is similar to starting the container (a stopped container becomes running), as is checkpoint to stopping the container (although you don't have to stop the container when making a checkpoint). Checkpoint/restore is in a way like suspend/restore in virtual machines.

src/ContainerCheckpointModal.jsx
andreasn: Sounds good to me!

src/ContainerCheckpointModal.jsx
andreasn: Could also make it `Leave container running after writing checkpoint to disk`

andreasn: Missed the last part apparently. Dropping the object all together sounds like a good idea. As it's implied by the context.

lenticularis39: I actually didn't mean that by the last part, but I've changed my mind about it - considering I dropped the "container" object in another case (above) I now don't see a problem with doing it here.

lenticularis39:
I've implemented the requested changes and put the testing of checkpoint/restore into a separate destructive test. For the spinner I've decided to use the `spinner` class instead of the PatternFly `spinner` component to match the force delete dialog (not sure whether this is a better option, but I think it shouldn't be implemented differently at different places).

KKoukiou:
> I've implemented the requested changes and put the testing of checkpoint/restore into a separate destructive test. For the spinner I've decided to use the `spinner` class instead of the PatternFly `Spinner` component to match the force delete dialog (not sure whether this is a better option, but I think it shouldn't be implemented differently at different places).

Thanks for the changes, regarding the spinner the visual result is the same, <Spinner /> or <div className='spinner' />, consistency is nice of course.

test/check-application
```diff
@@ -555,6 +559,75 @@ class TestApplication(testlib.MachineCase):
...
+        b = self.browser
+        m = self.machine
+
+        # Set machine to cgroup1 and reboot
+        self.execute(True, 'grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"; reboot')
+        m.wait_reboot()
+
+        self.login_and_go("/podman", authorized=True, superuser=True)
+        b.wait_present("#app")
+        self.filter_containers('all')
```
marusak, line 572: These two lines should not be needed.

test/check-application
```diff
@@ -555,6 +559,75 @@ class TestApplication(testlib.MachineCase):
...
+
+        # Run a container
+        self.execute(True, "podman run -dit --name swamped-crate busybox sh; podman stop swamped-crate")
```
marusak, line 575: Does checkpoint/restore work only for root containers and not for rootless?
If so, then we should check that these actions don't show up for rootless containers. If it works, then this test should have two modes System/User. (See how for example testBasic{System,User} is done)

test/check-application
```diff
@@ -555,6 +559,75 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait(lambda: self.execute(True, "podman ps --all | grep -e swamped-crate -e Exited"))
+
+        b.wait_present("#containers-containers")
```
marusak, line 578: This line is not needed, since the line underneath it is superset of this selector.

test/check-application
```diff
@@ -555,6 +559,75 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_present('#containers-containers tr:contains("swamped-crate")')
+        b.click('#containers-containers tbody tr:contains("busybox:latest") td.listing-ct-toggle')
+
+        # Check that the restore option is not present (i.e. start is a regular button)
+        b.wait_present('#containers-containers tbody tr:contains("busybox:latest") + tr button.pf-c-button:contains(Start)')
+
+        # Start the container
+        b.click('#containers-containers tbody tr:contains("busybox:latest") + tr button:contains(Start)')
+        b.wait(lambda: b.text('#containers-containers tr:contains(swamped-crate) td:nth-of-type(7)') == 'running')
```
marusak, line 587: This can be written simpler as `b.wait_in_text('#containers-containers tr:contains(swamped-crate)', 'running') but surely fine as is :)

test/check-application
```diff
@@ -555,6 +559,75 @@ class TestApplication(testlib.MachineCase):
...
+
+        # Checkpoint the container
+        b.wait_present('#containers-containers tr:contains("busybox:latest")')
```
marusak, line 590: No need to wait for something, when you do `click` as the next command. `wait` (as majority of other actions) firstly do `wait_visible` which on its own does `wait_present`.

test/check-application
```diff
@@ -555,6 +559,75 @@ class TestApplication(testlib.MachineCase):
...
+        b.click('#containers-containers tr:contains("busybox:latest") + tr button:contains(Stop) + button')
+        b.click('#containers-containers tr:contains("busybox:latest") + tr #Stop-dropdown ul li:has(a:contains(Checkpoint))')
+        b.click('.modal-dialog input#checkpoint-dialog-keep')
+        b.click('.modal-dialog input#checkpoint-dialog-tcpEstablished')
+        b.click('.modal-dialog input#checkpoint-dialog-ignoreRootFS')
+        b.click('.modal-dialog button:contains(Checkpoint)')
```
marusak, line 596: If it is checkbox, use `b.set_checked`, that also makes sure it has the correct value and not just blindly click on checkboxes.

lenticularis39:
> Just random idea while I was looking at #363 - we will need kebab menu ([this](https://upload.wikimedia.org/wikipedia/commons/thumb/8/8e/Symbol_rain_64.svg/768px-Symbol_rain_64.svg.png)) next to this start/stop select. So if it does not fit perfectly into this select, that would be a good place for it. (but if it fits, which seems it kind of does, then please ignore, just wanted to share :) )

Checkpoint and restore fits into stop and start selects fine. The kebab menu could however be useful later - I'm going to implement container migration after checkpoint/restore is finished and that could nicely fit into the menu, since it's probably too complex to be included in the Stop select.

> Also unrelated, but unfortunately this PR has conflicts now, so needs rebase.

OK. (Unfortunately the changing of the test class to destructive seems to cause a conflict with any change in tests, therefore probably even more rebasing will be needed.)

marusak:
> Checkpoint and restore fits into stop and start selects fine. The kebab menu could however be useful later - I'm going to implement container migration after checkpoint/restore is finished and that could nicely fit into the menu, since it's probably too complex to be included in the Stop select.

makes sense, thanks!

test/check-application
lenticularis39: Checkpoint/restore in Podman is currently supported only on root containers. The action actually shows up for user containers and causes an error - I'll implement a check that will hide the buttons for user containers and a test for it.

test/check-application
lenticularis39: I'll be doing changes in the tests anyway, so I'll simplify it - it will also help me remember that the method exists when writing Cockpit tests in the future.

test/check-application
lenticularis39: They are needed, because by default only running containers are shown, and the test operates on stopped containers several times.

marusak: oh, right, sorry!

lenticularis39:
> Can you please also check in some other test (testBasic most likely) how it behaves on cgroupsV2? Thanks!

I don't see how this is related to the PR. All tests except for `testCheckpointRestore` operate on cgroupsV2, because it is default since Fedora 31 (https://fedoraproject.org/wiki/Releases/31/ChangeSet#Modify_Fedora_31_to_use_CgroupsV2_by_default) - nothing changed about that with the implementing of checkpoint/restore.

marusak:
@lenticularis39 would it be useful for you if we would have info if currently cgroups V1 or V2 are in use? (I am doing some different work and can get this info while initially loading this plugin). If so, I'll send PR that will include it right now.

marusak:
> I don't see how this is related to the PR. All tests except for testCheckpointRestore operate on cgroupsV2, because it is default since Fedora 31 (https://fedoraproject.org/wiki/Releases/31/ChangeSet#Modify_Fedora_31_to_use_CgroupsV2_by_default) - nothing changed about that with the implementing of checkpoint/restore.

So on cgroupsV2 there is no checkpoint/restore button? (then it should verify that this button is not present) Or is this button there but the action fails in some expected manner? (then it should verify that the error is expected).

lenticularis39:
> @lenticularis39 would it be useful for you if we would have info if currently cgroups V1 or V2 are in use? (I am doing some different work and can get this info while initially loading this plugin). If so, I'll send PR that will include it right now.

Most likely not (at least not right now), because soon checkpoint/restore will work with cgroupsV2/crun, therefore the feature won't be dependent on the cgroups version.

src/app.jsx
```diff
@@ -134,8 +134,27 @@ class Application extends React.Component {
                 });
     }
 
+    isContainerCheckpointPresent(id, system) {
```
marusak, line 137: Oh I see, so you actually verify that it is present for each and every container. So no need to bother with checking of cgroupsV2. (But still there should be test that the button is not there, when on cgroupsV2 and once it will be available, we will notice in tests :) )

lenticularis39:
> So on cgroupsV2 there is no checkpoint/restore button? (then it should verify that this button is not present) Or is this button there but the action fails in some expected manner? (then it should verify that the error is expected).

On cgroupsV2 the buttons are shown and a correct error message (Configured runtime does not support checkpoint/restore) is shown (I can add checking this to the tests). I don't like the idea of hiding them in this case, because the reason why they aren't visible wouldn't be clear (the error message makes the reason why it doesn't work clear, unlike in the user container case).

src/app.jsx
lenticularis39: This checks whether a checkpoint is present for the container (i.e. the container can be restored), not whether the container can be checkpointed.

marusak:
> On cgroupsV2 the buttons are shown and a correct error message (Configured runtime does not support checkpoint/restore) is shown (I can add checking this to the tests). I don't like the idea of hiding them in this case, because the reason why they aren't visible wouldn't be clear (the error message makes the reason why it doesn't work clear, unlike in the user container case).

Agree. Please add test that verifies that the error is shown.

lenticularis39:
Checkpoint and restore buttons are now hidden for user containers. Tests have been added for checkpoint/restore of a user container (checks the absence of the Checkpoint button) and for checkpoint/restore with crun (currently fails) - the latter will be removed in the future when checkpoint/restore support becomes available in crun.

I've also fixed a bug that broke cockpit-podman with limited access (the same bug caused the integration tests to fail) - `isContainerCheckpointPresent` now doesn't request superuser privileges on user containers.

lenticularis39:
> Just one nitpick - you check that you can click checkpoint/restore but you don't verify that it has actually some effect on those containers. I think it would be nicer, if you would do some changes in those containers (like touch file -> checkpoint -> touch other file -> restore -> just the first file is there) so we know it actually does something. Would that make sense?

Touching a file wouldn't check anything, because the container filesystem is also preserved when the container is stopped (and I already check that) - the state of a running process would have to be changed and checked to see whether the checkpoint works properly, however I think this belongs to the tests for podman, not cockpit-podman. I think that here it's enough to verify that a checkpoint has been created - this is done by restoring, because if the checkpoint was missing (e.g. clicking the Checkpoint button stopped the container instead), the button wouldn't appear.

KKoukiou:
The new test seems little bit flaky https://logs.cockpit-project.org/logs/pull-430-20200722-092233-1e36b46e-fedora-32/TestApplication-testCheckpointRestoreSystem-fedora-32-127.0.0.2-2301-FAIL.png

Retrying for comparison.

marusak:
> The new test seems little bit flaky 

Oh, maybe we just need to wait for proper state?

lenticularis39:
> Oh, maybe we just need to wait for proper state?

I don't think that is the problem. The error message says that a restore of a running or paused container was attempted, but I can't see how this could happen - the Restore button shouldn't be shown for a running container and the container shouldn't be paused anywhere in the test.

KKoukiou:
> > Oh, maybe we just need to wait for proper state?
> 
> I don't think that is the problem. The error message says that a restore of a running or paused container was attempted, but I can't see how this could happen - the Restore button shouldn't be shown for a running container and the container shouldn't be paused anywhere in the test.

The second run is passing on the new tests. I think we should keep an eye on it but otherwise that's good to go.
What do you think @marusak?

KKoukiou:
For the release notes:

```
### Podman: Support checkpoint and restore operations for containers

Checkpointing a container stops the container and writes the state of all container processes to disk.
Such a checkpoint can be restored later to resume the container in the exact state as it was at checkpoint time.
This capability requires [CRIU 3.11](https://www.criu.org/Main_Page) or later installed. Currently
only the `runc` container runtime is supported; this is only available with `cgroupv1` systems.

Thanks to Tomáš Glozar for this contribution!

```

marusak:
> For the release notes:

Can we also have some screenshot? :)

We also won't get this into today release as the tests are flaky. It is most likely related to those two new nondetrucive ones and how we treat nond tests. I'll debug this later on.

marusak:
Hmm, trying this locally and cannot reproduce this problem. Lets retry both tests twice as hard.

marusak:
Now the [same coredump](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-430-20200723-072140-1e36b46e-fedora-31/log.html) on F31, [F-32](https://logs.cockpit-project.org/logs/pull-430-20200723-072143-1e36b46e-fedora-32/log.html) shows that the new test is indeed not the most stable.

lenticularis39:
>  [F-32](https://logs.cockpit-project.org/logs/pull-430-20200723-072143-1e36b46e-fedora-32/log.html) shows that the new test is indeed not the most stable.

I tried testCheckpointRestoreSystem several times separately on F-31 and it always succeeded - I'll try to switch the test VM to F-32.

marusak:
> I tried testCheckpointRestoreSystem several times separately on F-31 and it always succeeded - I'll try to switch the test VM to F-32.

Yeah, this seem like a strange flake... Gonna be tricky to find the cause

I'll push here some changes to get some results from CI as I cannot get anything locally. It will be separate commits for easy reverting.

marusak:
The [coredump](https://logs.cockpit-project.org/logs/pull-430-20200723-084240-3c9d7824-fedora-31/log.html) happened now in random test, unrelated to this PR :/ I have corefile, backtrace, so I just send it to podman guys and will hope they can get something out of it, since I am out of ideas how to actually reproduce it.

lenticularis39:
Thank you for your help with the tests.

> The [coredump](https://logs.cockpit-project.org/logs/pull-430-20200723-084240-3c9d7824-fedora-31/log.html) happened now in random test, unrelated to this PR :/ I have corefile, backtrace, so I just send it to podman guys and will hope they can get something out of it, since I am out of ideas how to actually reproduce it.

That makes sense to me, a coredump usually means a bug in the software where it happens. 

I wasn't able to reproduce the container improper state error in testCheckpointRestoreUser - strangely it seems to happen a lot in the CI, but never locally (or on F-31), even when running it after the two new non-destructible tests.

marusak:
It is green but I don't think we should land this. I have one more idea what could be the problem, lets see if I am right.

marusak:
First run both [f31](https://logs.cockpit-project.org/logs/pull-430-20200723-104145-c927298a-fedora-31/log.html) and [f32](https://logs.cockpit-project.org/logs/pull-430-20200723-104148-c927298a-fedora-32/log.html) were successful.

Maybe I even got the instability of that one test, see https://github.com/cockpit-project/cockpit-podman/pull/430/commits/5070f02464abc096359f9029196b9b27da3e8bab

Now I'll squash things down, run tests again.

PR #431, opened 2020-07-15
Dynamically update cpu and memory info

### LINKED ISSUE #376: CPU and memory usage is static
`podman top` shows cpu usage 98.309%,  but cockpit-podman show cpu usage 0%, and seems not changing anyway.
image

image

EricJeffrey:
Fix #376

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -121,17 +121,17 @@ class Application extends React.Component {
     }
 
     updateContainerStats(id, system) {
-        client.getContainerStats(system, id)
-                .then(reply => {
-                    this.updateState("containersStats", reply.Id + system.toString(), reply);
-                })
-                .catch(ex => {
-                    if (ex.error === "io.podman.ErrRequiresCgroupsV2ForRootless") { // TODO - Check if this message still applies
-                        console.log("This OS does not support CgroupsV2. Some information may be missing.");
-                        this.updateState("containersStats", id + system.toString(), -1);
-                    } else
-                        console.warn("Failed to update container stats:", JSON.stringify(ex));
-                });
+        client.getContainerStats(system, id, reply => {
+            if (reply.response) // executed when container stop, with reply: {cause, message, response}
+                console.warn("Failed to update container stats:", JSON.stringify(reply.message));
+            else
+                this.updateState("containersStats", reply.Id + system.toString(), reply);
+        }).catch(ex => {
+            if (ex.cause == "no support for CGroups V1 in rootless environments") {
+                console.log("This OS does not support CgroupsV2. Some information may be missing.");
+            } else
+                console.warn("Failed to update container stats:", JSON.stringify(ex.message));
+        });
     }
 
     updateContainersAfterEvent(system) {
--- a/src/client.js
+++ b/src/client.js
@@ -63,14 +63,13 @@ export function getContainers(system, id) {
     });
 }
 
-export function getContainerStats(system, id) {
+export function getContainerStats(system, id, callback) {
     return new Promise((resolve, reject) => {
         const options = {
-            stream: false, // FIXME that it actually streams
+            stream: true,
         };
-        podmanCall("libpod/containers/" + id + "/stats", "GET", options, system)
-                .then(reply => resolve(JSON.parse(reply)))
-                .catch(reject);
+        podmanMonitor("libpod/containers/" + id + "/stats", "GET", options, callback, system)
+                .then(resolve, reject);
     });
 }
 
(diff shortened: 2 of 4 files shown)


EricJeffrey:
well, something wrong with the 'cgroup v2 not support' part, checking now

marusak:
> well, something wrong with the 'cgroup v2 not support' part, checking now

How do you mean? I haven't noticed any problem with cgroupsV2.

EricJeffrey:
@marusak `app.jsx > updateContainerStats > catch`, the if statement `if (ex.cause == "no support for CGroups V1 in rootless environments") ` won't work actually. the `ex` argument is not as I expected. It is:
```
{status: 500, reason: "Internal Server Error", message: "Failed to obtain Container e937b97031d60bf70ad8bef…o support for CGroups V1 in rootless environments", problem: null, valueOf: ƒ, …}
```
But if I do `curl` in cli, it shows this:
```json
{"cause":"no support for CGroups V1 in rootless environments","message":"Failed to obtain Container uhalo stats: unable to load cgroup at /libpod_parent/libpod-e937b97031d60bf70ad8bef21373059956ff6f401f3d577c950c5db77ea3e07c: no support for CGroups V1 in rootless environments","response":500}
```

marusak:
right, can you please just check that `message` contains this substring?

EricJeffrey:
Well I've located the problem and seems `manage_error` need a change:
```js
function manage_error(reject, error, content) {
    let message = "";
    if (content) {
        try {
            message = JSON.parse(content).message;
        } catch (e) {
            message = content;
        }
    }
    const c = { ...error };
    if (message)
        c.message = message;
    reject(c);
}
```
The `cause` filed is in `content`, should we extract it out like this?:
```js
function manage_error(reject, error, content) {
    let message = "";
    let cause = "";
    if (content) {
        try {
            const content_parsed = JSON.parse(content);
            message = content_parsed.message;
            cause = content_parsed.cause;
        } catch (e) {
            message = content;
        }
    }
    const c = { ...error };
    if (message)
        c.message = message;
    if (cause)
        c.cause = cause;
    reject(c);
}
```
😂 Not very cool, any better idea?

EricJeffrey:
trying to clean up redundant commits, failed 😂

EricJeffrey:
Looks like work now. Anyway, how should I do the integration test? Just run command and see whether it goes well as you said all by myself?

marusak:
> The cause filed is in content, should we extract it out like this?:

We could just take everything from `content` and put it into `c`. So something like:
```js
 if (content) {
        let content_o = {};
        try {
            content_o = JSON.parse(content);
        } catch {
            content_o.message = content;
        }
    }
    const c = { ...error, ...content_o };
```

> Anyway, how should I do the integration test? Just run command and see whether it goes well as you said all by myself?

If you don't feel like writing those tests, I'll gladly help with that.
To start, [here](https://github.com/cockpit-project/starter-kit#automated-testing)  is how to prepare VM and how to run tests.
It should go into `test/check-application`, not sure where exactly, maybe it could be a separate test.
Some pointers:
1. search for `podman run` in the file to see, how to run new container
2. search for `cpu =` to see, how we read the value from the UI

EricJeffrey:
Ok got it, I'll take a try about the test.

EricJeffrey:
Well, you take the test part. 😂 `make check` doesn't work on my Ubuntu20 Server, and got no idea about the error message.

marusak:
Taking over this. I'll write tests and make commits fit our style.

EricJeffrey:
Also noticed sometimes the `client.streamEvents` will catch a `truncated response` exception, causing the `systemServiceAvailable` or `userServiceAvailable` changed. I tried to reproduce, but failed sometime. Don't know if it is related to my `podman.socket` (I wrote the systemd unit file by myself).

marusak:
Pushed version with test and polished commits. Also triggering tests.

marusak:
> Also noticed sometimes the client.streamEvents will catch a truncated response exception, causing the systemServiceAvailable or userServiceAvailable changed. I tried to reproduce, but failed sometime. Don't know if it is related to my podman.socket (I wrote the systemd unit file by myself).

Yeah, this smells like podman coredumping. When that happens, can you check journal if there is some info?

marusak:
@martinpitt or @KKoukiou can you please review since I touched this? Thanks!

EricJeffrey:
@marusak Mostly it happen when I try to stop/start more than one containers in a short time. I look up the log by `journalctl -xe` but no very useful message found.

Issue #261, opened 2019-12-12
Duplicate containers showing in the UI


bastilian:
When logged in with `root` into Cockpit, podman shows containers twice.

![Screenshot 2019-12-12 at 19 33 40](https://user-images.githubusercontent.com/7757/70739285-c633f000-1d16-11ea-9ae9-a00887c5ada6.png)

`podman` ouput:
```
[bastilian@titan ~]$ podman ps
CONTAINER ID  IMAGE  COMMAND  CREATED  STATUS  PORTS  NAMES
[bastilian@titan ~]$ su root
Password:
[root@titan bastilian]# podman ps
CONTAINER ID  IMAGE                              COMMAND          CREATED       STATUS           PORTS                                                              NAMES
ac08b3b6675e  docker.io/gitlab/gitlab-ce:latest  /assets/wrapper  19 hours ago  Up 19 hours ago  0.0.0.0:2424->22/tcp, 0.0.0.0:8585->80/tcp, 0.0.0.0:9443->443/tcp  gitlab
[root@titan bastilian]# exit
[bastilian@titan ~]$ sudo podman ps
CONTAINER ID  IMAGE                              COMMAND          CREATED       STATUS           PORTS                                                              NAMES
ac08b3b6675e  docker.io/gitlab/gitlab-ce:latest  /assets/wrapper  19 hours ago  Up 19 hours ago  0.0.0.0:2424->22/tcp, 0.0.0.0:8585->80/tcp, 0.0.0.0:9443->443/tcp  gitlab
[bastilian@titan ~]$
```

I'm unsure if this is expected. I believe that this might happen specifically only with `root` logged in.

marusak:
@bastilian are you still interested in fixing this and/or do you need some help?
Or you want me to take over?

bastilian:
@bastilian Sorry, I was on vacation over the past weeks. Yes, I am still interested in fixing this. I'll find some time soon to open a PR.

bmansvk:
I have the same issue on RHEL 8.1.

ReinforcedPanda:
Is this being worked on? Can confirm this is still a bug with the latest version of cockpit / podman containers

![image](https://user-images.githubusercontent.com/4663874/80899510-832ee500-8cd6-11ea-9f6b-ad1ae20203fa.png)

bastilian:
I finally started working on this today in https://github.com/cockpit-project/cockpit-podman/pull/377. The plan is to only user one "service" in the case of the root user.

PR #433, opened 2020-07-15
Check if root user and skip user service if true

KKoukiou:
Carried over from https://github.com/cockpit-project/cockpit-podman/pull/377


KKoukiou:
Needs Fixes #261 on merging.

PR #435, opened 2020-07-16
Enable setting up selinux label on mounted volumes

### LINKED ISSUE #362: Ability to set selinux contexts for volumes
SELinux will deny it and it is not possible to use rw:z or ro:z and there for SELinux needs to be deactivated permanently.
If I create a container with permissive SELinux it will work until reboot that SELinux is back and will destroy the container. Even after deactivating SELinux it is not possible to the container again.
I am using FCOS for testing FCOS with cockpit and podman but I may stay longer with RancherOS until it is really usable (cockpit-podman).

marusak:
Fixes #362
Resolves rhbz#1811773

Locally `TestApplication.testRunImageUser` is failing, but having hard time to find an actual reproducer. Here is what I am trying: [reportfull.txt](https://github.com/cockpit-project/cockpit-podman/files/4931590/reportfull.txt)

Screenshot from 2020-07-16 14-58-32

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -119,6 +119,7 @@ const Volume = ({ id, item, onChange, idx, removeitem, additem }) =>
                        placeholder={_("Container path")}
                        value={item.containerPath || ''}
                        onChange={e => onChange(idx, 'containerPath', e.target.value)} />
+                <div />
                 <Select.Select className='form-control'
                                initial={item.mode}
                                onChange={value => onChange(idx, 'mode', value)}>
@@ -127,6 +128,19 @@ const Volume = ({ id, item, onChange, idx, removeitem, additem }) =>
                     </Select.SelectEntry>
                     <Select.SelectEntry data='rw' key='rw'>
                         {_("ReadWrite")}
+                    </Select.SelectEntry>
+                </Select.Select>
+                <Select.Select className='form-control'
+                               initial={item.mode}
+                               onChange={value => onChange(idx, 'selinux', value)}>
+                    <Select.SelectEntry data='' key=''>
+                        {_("No SELinux label")}
+                    </Select.SelectEntry>
+                    <Select.SelectEntry data='z' key='z'>
+                        {_("Shared")}
+                    </Select.SelectEntry>
+                    <Select.SelectEntry data='Z' key='Z'>
+                        {_("Private")}
                     </Select.SelectEntry>
                 </Select.Select>
             </div>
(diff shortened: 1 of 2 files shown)


src/ImageRunModal.jsx
```diff
@@ -129,6 +130,19 @@ const Volume = ({ id, item, onChange, idx, removeitem, additem }) =>
                         {_("ReadWrite")}
                     </Select.SelectEntry>
                 </Select.Select>
+                <Select.Select className='form-control'
+                               initial={item.mode}
+                               onChange={value => onChange(idx, 'selinux', value)}>
+                    <Select.SelectEntry data='' key=''>
+                        {_("None SELinux label")}
```
martinpitt, line 137: "No SELinux label". "None" is indefinite, and ungrammatical in this context.

src/ImageRunModal.jsx
```diff
@@ -119,6 +119,7 @@ const Volume = ({ id, item, onChange, idx, removeitem, additem }) =>
                        placeholder={_("Container path")}
                        value={item.containerPath || ''}
                        onChange={e => onChange(idx, 'containerPath', e.target.value)} />
+                <div />
```
martinpitt, line 122: Was that on purpose, or some kind of leftover noise?

KKoukiou, line 122: This div either should not be here or has wrong indentation

marusak: It is on the same level as other components. I am afraid I do not follow.
It is a small hack, as I needed some empty element to make things wrap into the next row.

KKoukiou: This </div> line that you added just closes the line  <div role='group' className='ct-form-split'> on 112.
If that's done on purpose you should remove 4 spaces because the line that it closes is 4 spaces behind the one you added :)

marusak: On purpose. There are 4 elements we want to show. I want to have two in one row and two in the next. But the form puts 3 items on one line and then wraps to another, so putting one empty div makes things wrap as expected.

src/ImageRunModal.jsx
marusak: Fixed.

src/ImageRunModal.jsx
KKoukiou: Ups I just saw it's an empty div. Overlooked that as an closing element.

marusak: No, this is `<div />` which is short form for `<div></div>` and not `</div>` :)

marusak:
Release not draft:

**Enable setting up SELinux label when mounting volumes**

When creating new container and mounting volumes `:z` or `:Z` labels can be specified. These options tell Podman to relabel file objects on the shared volumes.

selinuxlabels

PR #436, opened 2020-07-16
Cleanups

marusak:
Can someone be my second pair of eyes and double check that I did not break some UI? :) Thanks!


martinpitt:
Both tests failed, though. RunImageSystem failed with

    Error: remount /home/admin/.local/share/containers/storage/overlay, flags: 0x40000: invalid argument

which again sounds like "the user has no user systemd instance, and no XDG_RUNTIME_DIR" again?

test/check-application
```diff
@@ -66,6 +66,10 @@ class TestApplication(testlib.MachineCase):
 
         self.allow_journal_messages("/run.*/podman/podman: couldn't connect.*")
 
+    def cgroupsV2(self):
+        # We will introduce images that are still cgroupsV1 later on
```
martinpitt, line 70: This will be rhel-8-4?

marusak: Indeed

PR #438, opened 2020-07-17
test: Include testRootUser into testBasicSystem

marusak:
This simple tests was too quick and it left VM in destructed state.
I could not figure out the exact problem, it seems that we cleaned up
podman directories while it was still init-ing them.

Thus moved it into testBasicSystem. However if I logged out too quickly
and there was still some event coming in, it coredumped podman so I just
wait until the last container is exited. This needs report, but rather
tricky to reproduce.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -72,18 +72,6 @@ class TestApplication(testlib.MachineCase):
         else:
             return self.admin_s.execute(cmd)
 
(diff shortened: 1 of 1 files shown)


marusak:
Ok, that worked out, can I get review, it block 4 or so PRs.

PR #444, opened 2020-07-20
Better manage intermediate images

### LINKED ISSUE #214: Better manage intermediate images
Should we show them by default? (`podman images` does not do that, you need to use `--all`).
How should we show them? Now we have empty name, `podman images -a` lists them as `<none>`.

Maybe it would be great to rethink the top filtering and have there options to show all images or not to show those intermediate ones. Also when we show user containers and user images, we want to filter and show only user or only system containers and images.

marusak:
If there is at least one such image, show 'Show intermediate images'
button, but don't show these images in the list. When this button is
clicked, these images are shown and button label changes to 'Hide
intermediate images'.

Fixes #214

opened
hidden

### DIFF
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -26,6 +26,7 @@ class Images extends React.Component {
             selectImageDeleteModal: false,
             setImageRemoveErrorModal: false,
             imageWillDelete: {},
+            intermediateOpened: false,
         };
 
         this.deleteImage = this.deleteImage.bind(this);
@@ -115,7 +116,7 @@ class Images extends React.Component {
             </div>
         );
         const columns = [
-            { title: image.RepoTags ? image.RepoTags[0] : "", header: true },
+            { title: image.RepoTags ? image.RepoTags[0] : "<none>:<none>", header: true },
             moment(image.Created, "YYYY-MM-DDTHH:mm:ss.SZ").calendar(),
             cockpit.format_bytes(image.Size),
             image.isSystem ? _("system") : this.props.user.name,
@@ -178,15 +179,19 @@ class Images extends React.Component {
                 {_("Get new image")}
             </Button>
         ];
+
+        const intermediateOpened = this.state.intermediateOpened;
+
         let filtered = [];
         if (this.props.images !== null) {
-            filtered = Object.keys(this.props.images);
-            if (this.props.textFilter.length > 0) {
-                filtered = filtered.filter(id =>
-                    (this.props.images[id].RepoTags || []).some(tag =>
-                        tag.toLowerCase().indexOf(this.props.textFilter.toLowerCase()) >= 0)
-                );
-            }
+            filtered = Object.keys(this.props.images).filter(id => {
+                const tags = this.props.images[id].RepoTags || [];
+                if (!intermediateOpened && tags.length < 1)
+                    return false;
+                if (this.props.textFilter.length > 0)
+                    return tags.some(tag => tag.toLowerCase().indexOf(this.props.textFilter.toLowerCase()) >= 0);
+                return true;
+            });
         }
 
         filtered.sort((a, b) => {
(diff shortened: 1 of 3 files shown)


src/Images.jsx
```diff
@@ -115,7 +116,7 @@ class Images extends React.Component {
             </div>
         );
         const columns = [
-            { title: image.RepoTags ? image.RepoTags[0] : "", header: true },
+            { title: image.RepoTags ? image.RepoTags[0] : "<none>:<none>", header: true },
```
martinpitt, line 119: Is "<none>" some kind of standard convention? It's not i18n'ed, and also looks a bit weird. I wondered how the CLI shows them, but I don't see any:
```
❱❱❱ podman images --all
REPOSITORY                                     TAG     IMAGE ID      CREATED      SIZE
docker.io/library/busybox                      latest  c7c37e472d31  2 weeks ago  1.45 MB
registry.fedoraproject.org/f32/fedora-toolbox  32      7e4d1b69a358  3 weeks ago  394 MB
docker.io/library/debian                       sid     e1d741d6634c  5 weeks ago  123 MB
```

Could this perhaps be `_("Unnamed intermediate image")` or `_("Intermediate image")` or so, with italic formatting?

src/Images.jsx
```diff
@@ -178,15 +179,19 @@ class Images extends React.Component {
                 {_("Get new image")}
             </Button>
         ];
+
+        const openedInterm = this.state.intermediateOpened;
```
martinpitt, line 183: Let's please be consistent in the alias names, or avoid the alias at all.

test/check-application
```diff
@@ -331,6 +331,23 @@ class TestApplication(testlib.MachineCase):
         else:
             b.wait_collected_text("#containers-containers .container-name", "abc")
 
+        # Test intermediate images
+        b.wait_not_present(".listing-action")
+        self.execute(auth, "echo 'FROM docker.io/library/registry:2\nRUN ls' > Dockerfile")
```
martinpitt, line 336: Could this happen in `self.vm_tmpdir`, to avoid possible collisions in non-destructive tests?

src/Images.jsx
marusak: Yes, that is how CLI shows that, thats why I went with that:
```
REPOSITORY                         TAG     IMAGE ID      CREATED            SIZE
localhost/foobar                   latest  66fe3398edde  About an hour ago  17.5 MB
<none>                             <none>  afdffa38735b  About an hour ago  11.7 MB
docker.io/library/fedora           latest  a368cbcfa678  9 days ago         189 MB
docker.io/library/ubuntu           latest  adafef2e596e  13 days ago        76.3 MB
localhost/tnuro                    5       27c9a4e9a337  2 weeks ago        1.05 GB
localhost/tnuro                    latest  27c9a4e9a337  2 weeks ago        1.05 GB
<none>                             <none>  5f359173d668  2 weeks ago        1.05 GB
<none>                             <none>  beafbcb1a253  2 weeks ago        1.05 GB
<none>                             <none>  a30842c5884e  2 weeks ago        1.05 GB
<none>                             <none>  2784fd2faf41  2 weeks ago        1.05 GB
<none>                             <none>  8b2bf089d928  2 weeks ago        1.05 GB
<none>                             <none>  e29e1eb8a53b  2 weeks ago        1.05 GB
<none>                             <none>  d0a4984abc21  2 weeks ago        1.05 GB
<none>                             <none>  c1b358886b1a  2 weeks ago        1e+03 MB
<none>                             <none>  744ca788c9e0  2 weeks ago        991 MB
<none>                             <none>  b1a5b6945399  2 weeks ago        991 MB
<none>                             <none>  3dc83a0616a6  2 weeks ago        593 MB
<none>                             <none>  46a0609ac99f  2 weeks ago        199 MB
docker.io/library/mongo            latest  2b2cc1f48aed  4 weeks ago        392 MB
docker.io/library/alpine           latest  a24bb4013296  7 weeks ago        5.85 MB
registry.fedoraproject.org/fedora  31      a7a37f74ff86  2 months ago       199 MB
k8s.gcr.io/pause                   3.1     da86e6ba6ca1  2 years ago        749 kB

```

marusak: I would prefer not to go with some arbitrary name/description, as we really want to indicate that there is no name.

martinpitt: The CLI is not i18n'ed, though. Well, it's not a very exposed feature, so if you want to keep `<none>` for now, that's ok with me.

test/check-application
marusak: Well, kind of - We run this a root and as admin. For root it is fine, but admin cannot write into `self.vm_tmpdir`.

src/Images.jsx
marusak: Changed

test/check-application
martinpitt: The test has to `mkdir` it anyway, it could use `--mode 777`

martinpitt: .. or use `mktemp -d` of course, if  you like that better.

marusak: Ok, using `mktemp -d`

marusak:
Release note draft:

**Don't show intermediate images by default**

Intermediate images are not shown by default but are hidden and user can opt-in to see them.

intermdiate

PR #446, opened 2020-07-21
Filter page by user/system

### LINKED ISSUE #221: Filter page by user/system
In https://github.com/cockpit-project/cockpit-podman/pull/173 support for user containers was introduced. It would be great, if we could have 3 modes - user/system/both. User mode would show only user containers and user images and would not show user/system selector in downloading of new image. The same for system.

Would simple `<select>` on the top of the page work, or does anyone have better idea? (I was thinking toggle switch, but we need 3 states). @andreasn

marusak:
Fixes #221

Screenshot from 2020-07-21 08-09-43

### DIFF
--- a/src/ContainerHeader.jsx
+++ b/src/ContainerHeader.jsx
@@ -7,11 +7,13 @@ class ContainerHeader extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
+            owner: 'all',
             filter: 'running',
             filterText: ''
         };
         this.handleFilterChange = this.handleFilterChange.bind(this);
         this.handleFilterTextChange = this.handleFilterTextChange.bind(this);
+        this.handleOwnerChange = this.handleOwnerChange.bind(this);
     }
 
     filterChanged() {
@@ -26,6 +28,13 @@ class ContainerHeader extends React.Component {
         }
     }
 
+    handleOwnerChange (value) {
+        this.setState({ owner: value });
+        if (this.props.onOwnerChanged) {
+            this.props.onOwnerChanged(value);
+        }
+    }
+
     handleFilterTextChange() {
         this.setState({ filterText: this.refs.filterTextInput.value }, this.filterChanged);
     }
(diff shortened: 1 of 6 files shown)


src/ContainerHeader.jsx
```diff
@@ -26,17 +28,35 @@ class ContainerHeader extends React.Component {
...
 
     render() {
         return (
             <>
+                <label className="heading-label" htmlFor="containers-containers-filter">{_("Containers")}</label>
                 <Select.Select id="containers-containers-filter" initial={this.state.filter} onChange={this.handleFilterChange}>
-                    <Select.SelectEntry data='all'>{_("Everything")}</Select.SelectEntry>
-                    <Select.SelectEntry data='running'>{_("Images and running containers")}</Select.SelectEntry>
+                    <Select.SelectEntry data='running'>{_("Only running")}</Select.SelectEntry>
+                    <Select.SelectEntry data='all'>{_("All")}</Select.SelectEntry>
```
martinpitt, line 48: Unrelated, but good change.

test/check-application
```diff
@@ -144,6 +144,27 @@ class TestApplication(testlib.MachineCase):
...
+            b.wait_js_func("ph_count_check", "#containers-containers tbody", 1)
+            b.wait_in_text("#containers-containers", "system")
+
+            b.set_val("#containers-containers-owner", "user")
+            b.wait_js_func("ph_count_check", "#containers-images tbody", 3)
+            b.wait_in_text("#containers-images", "admin")
+            b.wait_js_func("ph_count_check", "#containers-containers tbody", 1)
+            b.wait_in_text("#containers-containers", "admin")
+
+            b.set_val("#containers-containers-owner", "all")
```
martinpitt, line 164: This could check that this gets back to 8 images and 2 containers.

marusak: Done

src/ContainerHeader.jsx
marusak: Yeah, I also gave this select label that was not there before to be consistent with the new select. I can split it into separate commit, but I feel like not really needed. But if you think that would be better, then I'll happily do that.

martinpitt: No worries, this was just thinking aloud.

marusak:
Release note draft:

**Enable filtering images and containers by owner**

If user wants to work only with system or rootless containers now can switch between these two modes.

owner

Issue #85, opened 2019-04-18
RFE: Configure /etc/containers/registries.conf


TomSweeneyRedHat:
Podman relies on the the /etc/containers/registries.conf file to determine where container registries exist.  It's most important use is to add a private registry like 'acme.com' to the list so it will search for images there first.  The category to update for this in the file is `[registries.search]`

The file can also be used to  block registries in the category `[registries.block]]`  or to allow for insecure registries `['insecure.io']`, being able to toggle those would be handy too, but of lesser importance.

marusak:
Blocked on https://github.com/containers/libpod/issues/4793

marusak:
Selecting of registries when downloading image has been introduced in #447
Managing registries is gonna be a rather difficult one - There is no API for that, so we would need to parse `/etc/containers/registries.conf`.

We would just add a new one or dropped one from `unqualified-search-registries`, which is fine. We could also add/remove `[[registry]]` entry which still would be fine. But the thing that concerns me is that there is `v1` and `v2` version on this config file and combining these two will break stuff badly. What I described is `v2`, `v1` is described in the first comment.

If I understand it correctly, if you always only updated podman, you can still be using `v1` version of this file. We of course could have simple logic, that if `unqualified-search-registries` is present, it is `v2` and thus we need to do it that way otherwise different way. Or we could be lazy and when we would recognize that `v1` is present we would say that this feature is supported only with version 2.

Lykos94:
Any update on this issue? I was wondering if it is possible to add the authentication to a private registry, like the Gitlab one for example.

While i can add the registry to the registry list in `/etc/containers/registries.conf`,  I'm actually unable to browse it correctly or authenticate to it

marusak:
> Any update on this issue? 

Unfortunately not. Patches are welcomed though ;)

>  I was wondering if it is possible to add the authentication to a private registry, like the Gitlab one for example.

I would think so. Never tried it though. Quick google search shows some results but for specific queries it is best if you contact podman team directly: https://github.com/containers/podman

shonpaz123:
> > Any update on this issue?
> 
> Unfortunately not. Patches are welcomed though ;)
> 
> > I was wondering if it is possible to add the authentication to a private registry, like the Gitlab one for example.
> 
> I would think so. Never tried it though. Quick google search shows some results but for specific queries it is best if you contact podman team directly: https://github.com/containers/podman

[...]

PR #447, opened 2020-07-21
Enable selecting registries

marusak:
Enables users to select which registry should be used.

Part of #85 (the other part is managing these registries).

No tests yet, I'll write them now, but wanted to ask @garrett what he thinks about this design.

Screenshot from 2020-07-21 11-00-45
registryselection

### DIFF
--- a/src/ImageSearchModal.jsx
+++ b/src/ImageSearchModal.jsx
@@ -125,12 +125,12 @@ export class ImageSearchModal extends React.Component {
             <>
                 { this.props.userServiceAvailable && this.props.systemServiceAvailable &&
                     <form className="ct-form">
-                        <label className="control-label" htmlFor="as-user">{_("Download as:")}</label>
+                        <label className="control-label" htmlFor="as-user">{_("Owner")}</label>
                         <fieldset id="as-user">
                             <div className="radio">
                                 <label>
                                     <input type="radio" value="system" id="system" onChange={this.onToggleUser} checked={this.state.isSystem} />
-                                    {_("System")}
+                                    {_("system")}
                                 </label>
                                 <label>
                                     <input type="radio" value="user" id="user" onChange={this.onToggleUser} checked={!this.state.isSystem} />
@@ -177,17 +177,16 @@ export class ImageSearchModal extends React.Component {
                 {this.state.searchInProgress && <div id='search-image-dialog-waiting' className='spinner' />}
 
                 {this.state.searchFinished && !this.state.imageIdentifier == '' && <>
-                    <h5> {_("Images")} </h5>
-                    {this.state.imageList.length == 0 && <div> {cockpit.format(_("No results for $0. Please retry another term."), this.state.imageIdentifier)} </div>}
+                    {this.state.imageList.length == 0 && <div className="no-results"> {cockpit.format(_("No results for $0. Please retry another term."), this.state.imageIdentifier)} </div>}
                     {this.state.imageList.length > 0 && <ListGroup>
                         {this.state.imageList.map((image, iter) => {
                             return (
                                 <ListGroupItem active={this.state.selected == iter} onClick={() => this.onItemSelected(iter)} key={iter}>
                                     <span className='image-list-item'>
-                                        <label className='control-label'>
+                                        <label className='image-name control-label'>
                                             { image.Name }
                                         </label>
-                                        <span className='pull-right'> { image.Description } </span>
+                                        <span className='image-description'> { image.Description } </span>
                                     </span>
                                 </ListGroupItem>
                             );
@@ -201,7 +200,7 @@ export class ImageSearchModal extends React.Component {
             <Modal show onHide={this.props.close} className="podman-search">
                 <Modal.Header>
                     <Modal.CloseButton onClick={this.props.close} />
-                    <Modal.Title> {_("Search Image")} </Modal.Title>
+                    <Modal.Title> {_("Search for an image")} </Modal.Title>
                 </Modal.Header>
                 <Modal.Body>
                     {defaultBody}
@@ -213,7 +212,7 @@ export class ImageSearchModal extends React.Component {
                         <input className="form-control image-tag-entry"
                                id="image-search-tag"
                                type='text'
-                               placeholder={_("Tag")}
+                               placeholder="latest"
                                value={this.state.imageTag || ''}
                                onChange={e => this.onValueChanged('imageTag', e.target.value)} />
                     </div>
--- a/src/ImageSearchModal.jsx
+++ b/src/ImageSearchModal.jsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { ListGroup, ListGroupItem, Modal } from 'patternfly-react';
-import { Button, InputGroup, InputGroupText } from '@patternfly/react-core';
-import { SearchIcon } from '@patternfly/react-icons';
+import { Button, InputGroup } from '@patternfly/react-core';
 
+import * as Select from '../lib/cockpit-components-select.jsx';
 import { ErrorNotification } from './Notification.jsx';
 import cockpit from 'cockpit';
 import rest from './rest.js';
@@ -22,6 +22,7 @@ export class ImageSearchModal extends React.Component {
             searchInProgress: false,
             searchFinished: false,
             isSystem: props.systemServiceAvailable,
+            registry: "",
         };
         this.onDownloadClicked = this.onDownloadClicked.bind(this);
         this.onItemSelected = this.onItemSelected.bind(this);
@@ -72,12 +73,16 @@ export class ImageSearchModal extends React.Component {
         this.setState({ searchInProgress: true });
 
         this.activeConnection = rest.connect(client.getAddress(this.state.isSystem), this.state.isSystem);
+
+        const rr = this.state.registry;
+        const registry = rr.length < 1 || rr[rr.length - 1] === "/" ? rr : rr + "/";
+
         const options = {
             method: "GET",
             path: "/v1.12/libpod/images/search",
             body: "",
             params: {
-                term: this.state.imageIdentifier,
+                term: registry + this.state.imageIdentifier,
             },
         };
         this.activeConnection.call(options)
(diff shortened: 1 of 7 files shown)


garrett:
Overall: The feature is great at a first glance.

However: There are issues with the dialog; some are especially exacerbated with this addition:
- "Download as" should probably be "Download to", I suppose? It's more about a system versus user container, not who you're downloading it as.
- Why is there no space between the radio button and the text labels for each radio button at the top? (This should happen by default, so I really don't know what's going on here... Very odd.)
- Similarly, there's no space between the repo and the entry
- Additionally, adding the repo dropdown obfuscates the fact that the search is a search entry
- Tag feels like an afterthought. Is there any way to retrieve the available tags?

marusak:
> "Download as" should probably be "Download to", I suppose? It's more about a system versus user container, not who you're downloading it as.

Unrelated to this PR, but I think `D. as` makes more sense, as you want to choose if you are downloading as root or yourself. But this might be english grammar thing, where it makes more sense with `to`?

> Why is there no space between the radio button and the text labels for each radio button at the top? (This should happen by default, so I really don't know what's going on here... Very odd.)

Unrelated to this PR, this was issue even before. We should address this in a different PR for sure.

> Similarly, there's no space between the repo and the entry

This is on purpose. As if you would want to download from `docker.io` repository, you would do `podman pull docker.io/image` and I thought we may want to imitate that behaviour if we put `docker.io` right next to `image`.

> Additionally, adding the repo dropdown obfuscates the fact that the search is a search entry

Yeah, probably, it may not be clear it is search, but may indiacte, that you have to type the whole name. But tbh. I think you usually know what you are looking for anyway, so you usually type it fully.

> Tag feels like an afterthought. Is there any way to retrieve the available tags?

Maybe? I would need to properly check response from the API. You suggest then replacing this input with select with all possible tags? (but out of scope for this PR, but we may want to think where we want to place it then)

garrett:
> Unrelated to this PR

We should go through an do a review/cleanup PR together. But after this PR. :wink:

> As if you would want to download from docker.io repository, you would do podman pull docker.io/image and I thought we may want to imitate that behaviour if we put docker.io right next to image.

If we're recreating this, we should have a / between the two parts. I'm not sure how valid that would be. And also, if we're recreating this, perhaps we should have labels for the fields... but above the entries?

But I'm still not sold on `docker.io/image`… as it's is more for pulling down an image, not for searching, which is what this dialog does. (Yes, it pulls, but only after you select the image. And add a tag, if you don't go with the default one.)

marusak:
> We should go through an do a review/cleanup PR together. But after this PR.

Yes, lets do that :)

> But I'm still not sold on docker.io/image… as it's is more for pulling down an image, not for searching

Yeah, but `podman pull` == `search` + `download 1.st`, we do the search part, but instead of pulling the first image, we let user to select. And even now in this PR I just concat these two fields and search with that.

> we should have a / between the two parts

That was my other idea, a bit trickier to get nicely aligned. Thus I am just lazily adding `/` to the registries name. But I can take a look at that.

> we should have labels for the fields... but above the entries?

That would work. A bit scared it is going to make the UI a bit messy, but surely worth trying it out.

garrett:
Does it search across all repos by default? Is the search just a limiter?

marusak:
> Does it search across all repos by default? Is the search just a limiter?

Exactly. By default is searches all registries. And we still have 'All registries' option, which will keep the same behaviour. It is to limit response time but also users often know, where they want theirs images from. But also, I've seen lately that podman requires sometimes exact registry (like https://github.com/cockpit-project/cockpit/commit/2256ae091ee731f80a69ec2e3f01b756f8620b07) so this might become required sooner or later.

marusak:
@garrett could you please draft/describe some design I can work towards? Thanks!

marusak:
> Why is there no space between the radio button and the text labels for each radio button at the top? (This should happen by default, so I really don't know what's going on here... Very odd.)

Fixed in #459 

> Tag feels like an afterthought. Is there any way to retrieve the available tags?

Checked the API and it does not return such information.

marusak:
@garrett updated to have labels, WDYT?
Screenshot from 2020-07-27 11-55-26

marusak:
Wrote also tests and they pass. Ready for code and design review.

garrett:
Looks nice in the screenshot. Thanks!

I'll check locally now.

garrett:
I've pushed up a ton of changes. Here's what it looks like at the moment:

Screenshot_2020-07-28 Podman Containers - garrett Rain(1)

And mobile, while still a little wonky, is a lot better:
Screenshot from 2020-07-28 16-02-50

(You have to slide up the dialog while not touching the list)

garrett:
I was wondering if we should move download under tag... or perhaps even split download into two action buttons:

[ Download to system ] [ Download to $user ] Cancel

src/podman.scss
```diff
@@ -236,3 +257,6 @@
     font-weight: 400;
 }
 
+#registry-select option:first-child {
```
KKoukiou, line 260: This is specific to  src/ImageSearchModal.css  and should be moved there.

src/podman.scss
```diff
@@ -145,6 +145,27 @@
     padding-left: 30px;
 }
 
+.vertical-group {
```
KKoukiou, line 148: Same here - these are only be used on imageSearchModal.jsx file. I really prefer per-component CSS overrides and only in the case where we have global fixes above PF we should bring it up to files that are imported from the base of the package.

src/podman.scss
garrett: Great point.

I added CSS where it already existed, or when I made a mistake. :grin:  But it should be moved.

src/podman.scss
garrett: Yep. Fully agreed.

garrett: Also @marusak. But I should've removed this CSS when I removed vertical-group from the layout (and just used a ct-form-split class instead). Doing so now.

src/podman.scss
garrett: Ah, that's from @marusak, a week ago. But it doesn't work in anything but Blink. Non-Chromium-based browsers just ignore styles on options.

garrett: I cannot find any good reference which browsers support what CSS for `<select>`'s `<option>`s, but I do know from experience that browsers are generally unhappy to mess with default platform styling for the popup. I _think_ it's a security-related problem, as the select's popup is rendered at OS level (above the page) and modifying it (which _used to be possible_, years ago) meant that people _could_ hijack it with CSS to make impostor widgets which could appear outside of the page. This is why even Chrome only supports a strict subset of CSS on `<options>`... and other browsers basically disable it outright.

As I cannot find a reference to this, I could be misremembering or making it up. :shrug: But I'm pretty sure this is the case.

Anyway:

Summary: Not supported in anything but Chrome. Should be in ` src/ImageSearchModal.css` as @KKoukiou states, or it should be dropped (as it's Chrome-only).

src/ImageSearchModal.jsx
```diff
@@ -212,7 +212,7 @@ export class ImageSearchModal extends React.Component {
                         <input className="form-control image-tag-entry"
                                id="image-search-tag"
                                type='text'
-                               placeholder={_("Tag")}
+                               placeholder={_("latest, 3.1, ...")}
```
marusak, line 215: Wonder if it makes sense to translate this. Maybe for languages that do not use the same alphabet?

src/podman.scss
marusak: I removed it. I am squashing those commits + breaks one test so fixing that. I'll push force once I am done.

src/ImageSearchModal.jsx
```diff
@@ -125,7 +125,7 @@ export class ImageSearchModal extends React.Component {
             <>
                 { this.props.userServiceAvailable && this.props.systemServiceAvailable &&
                     <form className="ct-form">
-                        <label className="control-label" htmlFor="as-user">{_("Download as:")}</label>
+                        <label className="control-label" htmlFor="as-user">{_("Download to")}</label>
```
KKoukiou, line 128: Why 'Download to'?
Download to I would expect to be a label for a location where the images are stored.  This is for the owner of the image.

marusak: @garrett

garrett: It downloads to the system store or the user's store?

src/ImageSearchModal.jsx
garrett: The tags themselves aren't translated, right? Is "latest" the default tag?

Also: Should as have another example or two?

src/ImageSearchModal.jsx
garrett: Although, I guess it's really more about running as a system or rootless user container, than downloading at all? (But this dialog is about searching and downloading an image, of course.)

KKoukiou: There is the concept of stores as in separate storage locations for different users. But still 'Download to: garrett' does not make sense to me, it could be 'Download to: /storage' but 'Download to: garett' seems wrong seems there is no garrett storage/store call it how you want. But a default image location configured for garrett user.

src/ImageSearchModal.jsx
marusak: > The tags themselves aren't translated, right?

Indeed, they would always be what someone tagged them. So if someone tags their container "najnovsi" (slovak for "latest") you would need to write "najnovsi". So I guess we don't want to translate this. I'll remove the translation.

> Is "latest" the default tag?

Yes, if you don't specify tag, it tries to pull `latest`. Should we then "placehold" just that, as that is what is going to happen if you don't change it.

> Also: Should as have another example or two?

I don't think that is necessary. If users know what tag is, they know which specific one they want. Otherwise latest is implied.

garrett: > Should we then "placehold" just that, as that is what is going to happen if you don't change it.

Yeah, that could work. Having multiple shows that you can have numerical tags, but it does also sort of imply that commas and spaces are supported (for pulling down multiple tags?!)... and I'm pretty sure that's not the case.

Showing a value that is equivalent to the default state is a pattern that's commonly used.

Let's just go with an untranslated "latest", like you suggest.

src/ImageSearchModal.jsx
garrett: Downloading _as_ system or a user doesn't necessarily imply where it's downloaded or that you can only run it as system or user after downloading. That's the problem I have with "Downloading as". Additionally, Cockpit is doing the downloading (one way or the other).

I guess "Downloading as" or "Downloading to" are both a bit problematic here. I figured "to" at least indicates where it's going.

marusak:
Release note draft:
When searching for new images a registry can be selected in which user want to look for this image.

Screenshot from 2020-07-29 12-56-10

PR #452, opened 2020-07-23
Debian packaging and testing

martinpitt:
podman 2.0 is [now in Debian unstable](https://tracker.debian.org/pkg/libpod), so I'd like to start testing it there. It would be nice to upload cockpit-podman to Debian once we have something working decently.

After hammering on it a lot, I now have `TEST_OS=debian-testing make vm` working. Let's see how it fares in actual testing!


martinpitt:
Bah -- no `dpkg-dev` in our test containers. I suppose we need to build the dsc inside the VM. [log](https://logs.cockpit-project.org/logs/pull-452-20200723-103156-585b463f-debian-testing/log.html)

martinpitt:
On debian-testing, system podman.service fails with

    [graphdriver] prior storage driver overlay failed: remount /var/lib/containers/storage/overlay, flags: 0x40000: invalid argument

user podman seems to work better, I can at least do stuff. So while most tests will still be broken, I think this is good enough progress to land this -- debian-testing isn't yet triggered automatically.

martinpitt:
https://bugs.debian.org/966118 got fixed in the meantime, so dropping the hack. Rebased to current master.

marusak:
Been playing with this for a bit and here are two things I noticed:
1) The `[graphdriver] prior storage driver overlay failed` seem to be related to us doing `self.restore_dir("/var/lib/containers")` in `SetUp`. We sometimes see this in fedora as well. It would be worth figuring this one out. In fedora it happens only sometimes, so it might be that we bind-mount while podman is doing some actions that cause this problem.
2) The new build is affected by https://github.com/containers/podman/issues/7078 so unusable

martinpitt:
Yay, 2.0.2 is in testing now, so I dropped the "get from unstable" hack and put back the "install correct units" hack. I filed https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=966501 to block 2.0.3 from entering testing. Once 2.0.4 hits testing, we can clean all that up.

packaging/debian/control
```diff
@@ -3,7 +3,6 @@ Section: admin
 Priority: optional
 Maintainer: Cockpit <cockpit-devel@lists.fedorahosted.org>
 Build-Depends: debhelper-compat (= 12),
-               appstream-util,
```
marusak: This commit `debian: Drop appstream-util check ` could be merged into the previous one as it directly fixes it.

martinpitt: That was by intent -- it's easier to revert once we get appstream-util into the Debian image's pbuilder.

martinpitt:
I think I fixed the /overlay thing now. Tests still fail but "much less bad" :grin:

martinpitt:
Yay, there's (at least) one successful test now :tada:

g3blv:
Does this mean that `cockpit-podman` will be available in the Ubuntu repos?

marusak:
> Does this mean that cockpit-podman will be available in the Ubuntu repos?

Please see https://github.com/cockpit-project/cockpit-podman/issues/481

PR #456, opened 2020-07-24
Use libpod events endpoint

marusak:


martinpitt:
Again [these weird flakes due to storage overlay mounting](https://logs.cockpit-project.org/logs/pull-456-20200724-183331-f6fdd8e8-fedora-32/log.html#7-2), retrying. But looks good if that works.

PR #461, opened 2020-07-28
If container does not have tty, 'exec' and then connect

### LINKED ISSUE #297: Terminal should 'exec' shell and connect to it
When command is not interactive shell, then terminal only logs instead of being usable as terminal.

marusak:
Fixes #297

### DIFF
--- a/src/ContainerTerminal.jsx
+++ b/src/ContainerTerminal.jsx
@@ -56,6 +56,9 @@ class ContainerTerminal extends React.Component {
         this.disconnectChannel = this.disconnectChannel.bind(this);
         this.connectChannel = this.connectChannel.bind(this);
         this.resize = this.resize.bind(this);
+        this.connectToTty = this.connectToTty.bind(this);
+        this.execAndConnect = this.execAndConnect.bind(this);
+        this.setUpBuffer = this.setUpBuffer.bind(this);
 
         const term = new Terminal({
             cols: 80,
@@ -70,6 +73,7 @@ class ContainerTerminal extends React.Component {
         this.state = {
             term: term,
             container: props.containerId,
+            sessionId: props.containerId,
             channel: null,
             buffer: null,
             opened: false,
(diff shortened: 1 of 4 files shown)


src/ContainerTerminal.jsx
```diff
@@ -95,7 +99,7 @@ class ContainerTerminal extends React.Component {
         var realWidth = this.state.term._core._renderService.dimensions.actualCellWidth;
         var cols = Math.floor((width - padding) / realWidth);
         this.state.term.resize(cols, 24);
-        client.resizeContainersTTY(this.props.system, this.state.container, cols, 24)
+        client.resizeContainersTTY(this.props.system, this.state.sessionId, this.props.tty, cols, 24)
```
KKoukiou, line 102: Why don't you keep using the existing state 'container' here? It has the same value. (Or does it change somewhere I don't see?)

marusak: It does, when we connect through exec, it is not container ID, but exec ID. It is set up [here](https://github.com/cockpit-project/cockpit-podman/pull/461/files#diff-ce6eac7021a57afa8c484caf011f3e15R195)

KKoukiou: Ack thanks, thought that's still the containerId but aparently it's different value of that (r.Id)

src/Containers.jsx
```diff
@@ -132,7 +134,7 @@ class Containers extends React.Component {
         }, {
             name: _("Console"),
             renderer: ContainerTerminal,
-            data: { containerId: container.Id, containerStatus: container.State, width:this.state.width, system:container.isSystem }
+            data: { containerId: container.Id, containerStatus: container.State, width:this.state.width, system:container.isSystem, tty: tty }
```
KKoukiou, line 137: not big of a deal - but this is a new property here, maybe you want to add it in PropTypes. (which is missing  half of the others already so we can fix this in the future)

marusak: ah right, I added the new one and also added the forgotten ones. Thanks for spotting!

PR #466, opened 2020-07-29
test: Drop existing podman images

martinpitt:
Our Fedora/RHEL CI images have the cockpit/base and the underlying
fedora container image, but the Debian image does not. Avoid the
ambiguity by cleaning up all existing images in vm.install.

 - [x] Work around podman API breakage after idle timeout: PR #473

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -155,11 +155,11 @@ class TestApplication(testlib.MachineCase):
 
         # Test owner filtering
         if auth:
-            b.wait_js_func("ph_count_check", "#containers-images tbody", 8)
+            b.wait_js_func("ph_count_check", "#containers-images tbody", 6)
             b.wait_js_func("ph_count_check", "#containers-containers tbody", 2)
 
             b.set_val("#containers-containers-owner", "system")
-            b.wait_js_func("ph_count_check", "#containers-images tbody", 5)
+            b.wait_js_func("ph_count_check", "#containers-images tbody", 3)
             b.wait_in_text("#containers-images", "system")
             b.wait_js_func("ph_count_check", "#containers-containers tbody", 1)
             b.wait_in_text("#containers-containers", "system")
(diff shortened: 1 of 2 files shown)


martinpitt:
Should fix [this failure](https://logs.cockpit-project.org/logs/pull-452-20200729-135554-dce96016-debian-testing/log.html#1-2)

martinpitt:
Somehow I managed to [break the Fedora tests](ohttps://logs.cockpit-project.org/logs/pull-466-20200729-162257-09d3e58d-fedora-32/log.html). Certainly not due to the changes here, as they don't touch these parts. I have a feeling that commit 3f78931902b1c88e3ed3b5180302d062c232901c was related somehow.

martinpitt:
That [was clearly not it](https://logs.cockpit-project.org/logs/pull-466-20200729-164250-bea46157-fedora-32/log.html), reverting the revert.

martinpitt:
I sent the second commit to PR #470, as that's independent and should really work.

martinpitt:
Adding the commits from #473 to test everything together.

martinpitt:
Ah, debian-testing is now broken, as 2.0.3 went into unstable (the release-critical bug was downgraded, with my consent). I'll add a hack to downgrade to 2.0.2, but in a separate PR.

martinpitt:
This works -- the stabilization in #473 now keeps Fedora tests intact, and in Debian testBasicSystem() succeed. (Plus testDownloadImage() as well now, which was an unexpected bonus).

martinpitt:
Rebased to master.

martinpitt:
Whee! podman 2.0.4 is out, and [got packaged in Debian](https://tracker.debian.org/news/1165013/accepted-libpod-204dfsg1-1-source-into-unstable/). Locally it breaks a few tests again, but it's still better to test the current version rather than an old snapshot.

martinpitt:
The [testNotRunning regression](https://logs.cockpit-project.org/logs/pull-466-20200803-045534-71f5291d-debian-testing/log.html#8-2) is actually just because the Debian package now properly ships the unit. The unit itself [has been wrong for a long time](https://github.com/containers/podman/issues/7190), I added a hack around this.

PR #467, opened 2020-07-30
Enable untagging images

### LINKED ISSUE #172: Enable to remove image with more tags
Image can have more tags. Deleting such image asks if we want to remove the image itself with all of its tags.
CLI users usually remove tags and when the last tag is removed the image itself is removed.
Maybe we should show images by their tags the same as `podman images` would show it and then this deleting would not be confusing?

@andreasn @garrett, WDYT?

marusak:
Fixes #172

untag

### DIFF
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -9,7 +9,7 @@ import ImageDetails from './ImageDetails.jsx';
 import ImageUsedBy from './ImageUsedBy.jsx';
 import { ImageRunModal } from './ImageRunModal.jsx';
 import { ImageSearchModal } from './ImageSearchModal.jsx';
-import ModalExample from './ImageDeleteModal.jsx';
+import { ImageDeleteModal } from './ImageDeleteModal.jsx';
 import ImageRemoveErrorModal from './ImageRemoveErrorModal.jsx';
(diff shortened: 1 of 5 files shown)


src/Images.jsx
```diff
@@ -73,18 +73,30 @@ class Images extends React.Component {
...
+        else {
+            // Call another untag once previous one resolved. Calling all at once can result in undefined behavior
+            const tag = tags.shift();
+            const i = tag.lastIndexOf(":");
+            client.untagImage(this.state.imageWillDelete.isSystem, image, tag.substring(0, i), tag.substring(i + 1, tag.length))
+                    .then(() => {
+                        if (tags.length > 0)
+                            this.handleRemoveImage(tags, all);
+                    })
+                    .catch(console.log);
```
KKoukiou, line 98: I know we do that everywhere in podman but if untagImage fails this won't show an error in the UI.

marusak: Yeah, I want to address all of those in some other PR. We do this rather often and it is not good. Thought some global alert handler would be great, thus just `console.log` for now.

KKoukiou: Sure - I am fine with that.

PR #468, opened 2020-07-30
Add support for pods (and some Ui refactoring)

KKoukiou:
* [x] fix tests

### DIFF
--- a/src/ContainerHeader.jsx
+++ b/src/ContainerHeader.jsx
@@ -8,10 +8,8 @@ class ContainerHeader extends React.Component {
         super(props);
         this.state = {
             owner: 'all',
-            filter: 'running',
             filterText: ''
         };
-        this.handleFilterChange = this.handleFilterChange.bind(this);
         this.handleFilterTextChange = this.handleFilterTextChange.bind(this);
         this.handleOwnerChange = this.handleOwnerChange.bind(this);
     }
@@ -19,13 +17,6 @@ class ContainerHeader extends React.Component {
     filterChanged() {
         if (this.props.onFilterChanged)
             this.props.onFilterChanged(this.state.filterText);
-    }
-
-    handleFilterChange (value) {
-        this.setState({ filter: value });
-        if (this.props.onChange) {
-            this.props.onChange(value);
-        }
     }
 
     handleOwnerChange (value) {
@@ -42,11 +33,6 @@ class ContainerHeader extends React.Component {
     render() {
         return (
             <>
-                <label className="heading-label" htmlFor="containers-containers-filter">{_("Containers")}</label>
-                <Select.Select id="containers-containers-filter" initial={this.state.filter} onChange={this.handleFilterChange}>
-                    <Select.SelectEntry data='running'>{_("Only running")}</Select.SelectEntry>
-                    <Select.SelectEntry data='all'>{_("All")}</Select.SelectEntry>
-                </Select.Select>
                 { this.props.twoOwners &&
                     <>
                         <label className="heading-label" htmlFor="containers-containers-owner">{_("Owner")}</label>
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -18,6 +18,8 @@ import * as client from './client.js';
 import ContainerCommitModal from './ContainerCommitModal.jsx';
 import ScrollableAnchor from 'react-scrollable-anchor';
 
+import './Containers.scss';
+
 const _ = cockpit.gettext;
 
 class Containers extends React.Component {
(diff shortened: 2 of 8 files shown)


KKoukiou:
@marusak @garrett I did not show yet the open ports and created time of the pod since the current mock up does not incoorporate them and I did not want to innovate too much.
I think it's fine that the pods attributes will me implemented in a followup.

marusak:
List of TODOs: (I'll convert them into issues when this PR lands)
- Create new pod
- Pod actions (Delete, start/stop, pause/unpause)
- Add container to pod
- Add podless container
- pod details (ID, name, status, created, owner, ports) (owner not needed for containers then)

src/ContainerHeader.jsx
```diff
@@ -8,10 +8,8 @@ class ContainerHeader extends React.Component {
         super(props);
```
marusak, line 8: Why the second commit has `Closes #475`?
And I think the `Fixes` line would be better for the commit that actually touches pods.

test/check-application
```diff
@@ -77,6 +77,47 @@ class TestApplication(testlib.MachineCase):
         else:
             return self.admin_s.execute(cmd)
 
+    def waitPodRow(self, podName):
+        self.browser.wait_present("#table-" + podName)
+
+    def waitPodContainer(self, podName, containerList):
+        print(containerList)
```
marusak, line 84: Forgotten debugging

test/check-application
```diff
@@ -77,6 +77,47 @@ class TestApplication(testlib.MachineCase):
...
+        if len(containerList):
+            for container in containerList:
+                self.check_container(container["id"], True, [container["name"], container["image"], container["command"], container["state"]])
```
marusak, line 87: This does not check that the container is associated with the pod, just that it is displayed.

src/Containers.jsx
```diff
@@ -314,45 +316,58 @@ class Containers extends React.Component {
...
         }
 
-        if (this.props.textFilter.length > 0) {
-            const lcf = this.props.textFilter.toLowerCase();
-            filtered = filtered.filter(id => this.props.containers[id].Names[0].toLowerCase().indexOf(lcf) >= 0 ||
-                    this.props.containers[id].Image.toLowerCase().indexOf(lcf) >= 0
-            );
-        }
+        if (this.props.containers !== null && this.props.pods !== null) {
+            if (this.props.containers !== null && this.props.pods !== null)
```
marusak: Checking twice the same condition. That seems pointless, right?

src/Containers.jsx
```diff
@@ -314,45 +316,58 @@ class Containers extends React.Component {
...
+                filtered = Object.keys(this.props.containers).filter(id => !this.props.onlyShowRunning || this.props.containers[id].State == "running");
```
marusak: I think here is some misunderstanding.
pod is owned by root or system and is running/stopped.

We want to show owner (fine as followup of course, we may also think about not showing owner for separate containers if they are in pod owner by root/user). That means we want to hide the whole user pod if we switch filter from `all` to `system` and the other way around and not just containers inside it. (Because it leaves "empty" pod).

Pod can be running or stopped. We want to show that, fine as followup ofc. That of course means that the filter for running should not filter running containers in pods, but pods as whole. If pod is running but has one died container inside it, we should show also this stopped container (imo).

src/Containers.jsx
```diff
@@ -314,45 +316,58 @@ class Containers extends React.Component {
...
+                });
+            }
 
-        filtered.sort((a, b) => {
-            // User containers are in front of system ones
-            if (this.props.containers[a].isSystem !== this.props.containers[b].isSystem)
-                return this.props.containers[a].isSystem ? 1 : -1;
-            return this.props.containers[a].Names > this.props.containers[b].Names ? 1 : -1;
-        });
+            if (this.props.textFilter.length > 0) {
```
marusak, line 353: This also does not seem to filter pod names. I think the behaviour we want;
1. If pod name fits the filter, show it with all containers (don't filter those down)
2. If container inside pod fits the filter, show the filter and just containers that fit the name.
3. If nor the pod name nor any container inside the pod fit the filter, hide the whole pod.

src/Containers.jsx
```diff
@@ -314,45 +316,58 @@ class Containers extends React.Component {
...
+                const lcf = this.props.textFilter.toLowerCase();
+                filtered = filtered.filter(id => this.props.containers[id].Names[0].toLowerCase().indexOf(lcf) >= 0 ||
+                        this.props.containers[id].Image.toLowerCase().indexOf(lcf) >= 0
+                );
+            }
+
+            // Remove infra containers
+            filtered = filtered.filter(id => !this.props.containers[id].IsInfra);
```
marusak, line 361: This seems like a weird place to do this. I think it should not be in the list of containers. Or would that overly-complicate things?

src/Containers.jsx
```diff
@@ -314,45 +316,58 @@ class Containers extends React.Component {
...
-        const rows = filtered.map(id => this.renderRow(containersStats, this.props.containers[id], this.props.containersDetails[id]));
+            filtered.sort((a, b) => {
+                // User containers are in front of system ones
+                if (this.props.containers[a].isSystem !== this.props.containers[b].isSystem)
+                    return this.props.containers[a].isSystem ? 1 : -1;
+                return this.props.containers[a].Names > this.props.containers[b].Names ? 1 : -1;
+            });
+
+            Object.keys(this.props.pods || {}).forEach(pod => { partitionedContainers[pod] = [] });
+            filtered.forEach(id => {
```
marusak, line 371: If I see correctly, you don't order pods. I think we want to have the same logic with pods as we have with containers - first goes user ones, then system ones and otherwise ordered by name.

src/Containers.jsx
KKoukiou: This was delibatate. Some pods details like port mappings are present only in the infra Container details, and not in the Pod object itself.
Thus it makes sense to keep the infra container until this component, where we can get the details we are interested in.

src/Containers.jsx
KKoukiou: Good point for the owner. I fixed it. I can in a follup remove the owner state from the container columns and move it up to the pod details.

Regarding the state I though it's not so straightforward.
You can have a start/stop a while pod, but you can also stop once container in a pod while all the others are running.

marusak: > Regarding the state I though it's not so straightforward.
You can have a start/stop a while pod, but you can also stop once container in a pod while all the others are running.

Yeah, but pod being the higher abstraction, I think most filters should apply to it. It might be confusing showing pod with 2 running containers, even though there are 5 containers in the pod. It feels misleading. Maybe if we could indicate somehow that there are additional containers that are hidden for some reason? (Or show number of containers in details).

@garrett WDYT?

garrett:
OK, so I've fixed the page's height (so it always stretches), fixed the stickiness of the filter bar, and adjusted the spacing of the pods in all the scenarios I've encountered.

@KKoukiou: I hope I didn't break anything in 628e995, when I removed a nested wrapper.

src/Containers.jsx
KKoukiou: I would rather have more fine grained filters like demoonstrated here https://www.patternfly.org/2020.06/documentation/react/demos/filtertabledemo and get rid of this logic.

marusak: That is one approach for sure. It just feels a bit too heavy. I think user just wants to quickly find some container to check if it is running or some image to run it. Going to specific panel, picking column, searching... instead of just typing on the top of the page. But of course, that is just my idea and I want to keep it simple, but maybe it is a bit heavier that I think of it :)

KKoukiou: @marusak I added the running filters for the pods in the end as you suggested.

marusak:
Thank you @KKoukiou for the changes, now it works much better!
There seems to be some problem though:
```
$ podman pod ps
POD ID        NAME      STATUS   CREATED        # OF CONTAINERS  INFRA ID
88fb67086715  user_one  Running  3 minutes ago  3                ebcfad8f79be
a25674684d34  user_pod  Created  21 hours ago   1                9378c45561e0
```
So I have one rootless pod that is running and there are 2 running containers in it, but when I pick `Show Only Running` it disappears. Are you able to reproduce, or should I do it locally?

```
$ podman pod create --name=user_one
$ podman run -dit --pod user_one fedora bash
$ podman run -dit --pod user_one fedora bash
In UI switch to see only running.
```

KKoukiou:
> $ podman pod create --name=user_one
> $ podman run -dit --pod user_one fedora bash

Added a HACK in our code and opened an issue to podman.

PR #470, opened 2020-07-30
test: Adjust checkpoint tests for debian-testing

martinpitt:
criu is not in Debian testing, and will not be for a while:
https://bugs.debian.org/796536

Ensure that the UI shows a proper error message there.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -66,6 +66,7 @@ class TestApplication(testlib.MachineCase):
         self.addCleanup(self.admin_s.execute, "podman rm --force --all")
 
         self.allow_journal_messages("/run.*/podman/podman: couldn't connect.*")
+        self.allow_journal_messages(".*/run.*/podman/podman.*Connection reset by peer")
 
         self.has_criu = self.machine.image not in ["debian-testing"]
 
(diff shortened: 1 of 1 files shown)


marusak:
This is of course fine, just wonder how it behaves when criu is not present. Just those buttons are missing? Would it be worth checking that they are not present on debian - then we will notice once they are present and we can verify it works as expected.

martinpitt:
@marusak : Good point. They actually are present, but then cause a dialog that says "needs criu of at least version XXX". criu is available in unstable. So the page could check if the `criu` command is available and otherwise not show the menu entry; or we leave it there for discoverability.

marusak:
> They actually are present, but then cause a dialog that says "needs criu of at least version XXX". criu is available in unstable. So the page could check if the criu command is available and otherwise not show the menu entry; or we leave it there for discoverability.

Error that is clear is fine. It will show users that they miss something and they may install it or rise it with upstream to get this delivered. I suggest just checking this error in tests for debian.

martinpitt:
Whoops, [ambiguous selector](https://logs.cockpit-project.org/logs/pull-470-20200731-092820-34312228-debian-testing/log.html#4-2), I didn't get that locally.

martinpitt:
Still [one unexpected message](https://logs.cockpit-project.org/logs/pull-470-20200731-114148-aa67fa3d-debian-testing/log.html#4) which I don't get locally, but in general the test is working.

martinpitt:
The two checkpoint tests now succeeded on Debian. @marusak , could you please review?

PR #473, opened 2020-07-31
test: Work around leaking podman processes and broken API after idle timeout

martinpitt:


martinpitt:
Much much better! Still breaks [testNotRunning](https://logs.cockpit-project.org/logs/pull-473-20200731-140430-92a0e633-fedora-32/log.html#8-2)

But also, I lied in that commit -- it's not really a race condition in podman.service, the socket is already handed over correctly. With curl it works reliably all the time. I also don't get that error interactively, it works. So there must be something more subtle going on.

martinpitt:
I finally have this working on my local system now, let's see what CI says to that. Now I will try to find CLI reproducers for these problems, and report them upstream.

martinpitt:
Yay, at last! [F32](https://logs.cockpit-project.org/logs/pull-473-20200801-073038-a606f875-fedora-32/log.html) is perfectly green now, and [F31](https://logs.cockpit-project.org/logs/pull-473-20200801-073038-a606f875-fedora-31/log.html) just has a single retry from an unrelated flake. I reported the "podman pause" leak upstream, I'm still working on a reproducer for the hanging API request.

martinpitt:
Hooray, I *finally* understand the madness that's going on here: https://github.com/containers/podman/issues/7180#issuecomment-667496141

martinpitt:
Alas, that was a hard one. In hindsight it's pretty obvious, but I got led astray in so many different wrong directions here.. @marusak, this is ready for review. :tada:

PR #477, opened 2020-08-03
Only offer selinux labels on systems that support selinux

marusak:


src/ImageRunModal.jsx
```diff
@@ -106,7 +106,7 @@ const EnvVar = ({ id, item, onChange, idx, removeitem, additem }) =>
         </>
     );
 
-const Volume = ({ id, item, onChange, idx, removeitem, additem }) =>
+const Volume = ({ id, item, onChange, idx, removeitem, additem, options }) =>
```
martinpitt, line 109: I'd rather leave out the `options` indirection and specify them as direct properties. More robust and easier to check with propTypes and such if they are flat.

That's not a veto though, if you prefer that it's fine.

src/app.jsx
```diff
@@ -406,6 +407,9 @@ class Application extends React.Component {
                     }
                 })
                 .fail(e => console.log("Could not read $XDG_RUNTIME_DIR: ", e.message));
+        cockpit.spawn("selinuxenabled")
```
martinpitt, line 410: I checked that `selinuxenabled` works as non-root. But I think this should use `{ error: "ignore" }` to not spam the journal with "no such command" etc. errors.

src/ImageRunModal.jsx
marusak: The thing is that this is wrapped with `DynamicListForm` which is used for different components. Adding random options for whichever component we would need them would make it rather messy, thus I decided  to go with general `options` so any component can specify what they need without making this wrapper one big mess.

src/app.jsx
marusak: updated

src/ImageRunModal.jsx
martinpitt: Makes sense, thanks!

marusak:
Fedora is green, debian got beyond selinux things. It is broken on exec session closing immediately, but did not figure that one out yet - but it is surely unrelated to this PR.

PR #478, opened 2020-08-03
test: Install podman 2.0.4 from updates-testing

martinpitt:
This avoids running into the API breakage of 2.0.3:
https://github.com/containers/podman/issues/7078

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -46,8 +46,10 @@ if [ -d /var/tmp/debian ]; then
     systemctl disable docker.service io.podman.service
 fi
 
-# HACK: avoid leaking processes with podman < 2.0.4; see https://github.com/containers/podman/commit/2b6dd3fb4384
-sed -i '/KillMode=process/d; s/Type=oneshot/Type=simple/' /lib/systemd/user/podman.service /lib/systemd/system/podman.service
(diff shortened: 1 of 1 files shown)


martinpitt:
This should unblock https://github.com/cockpit-project/bots/pull/1109

test/vm.install
```diff
@@ -46,8 +46,10 @@ if [ -d /var/tmp/debian ]; then
     systemctl disable docker.service io.podman.service
 fi
 
-# HACK: avoid leaking processes with podman < 2.0.4; see https://github.com/containers/podman/commit/2b6dd3fb4384
-sed -i '/KillMode=process/d; s/Type=oneshot/Type=simple/' /lib/systemd/user/podman.service /lib/systemd/system/podman.service
+if rpm -q podman >/dev/null 2>&1; then
```
marusak, line 49: Why not using 'else' branch? Seems a bit clearer but just nitpick of course :)

martinpitt: You mean for "Debian vs. not"? No particular reason, I suppose we don't support a third package format right now. The hack should disappear completely in a week anyway, though.

marusak: Yeah, as the condition is there just to avoid breaking debian. But fair enough since it is rather tmp thing :)

PR #479, opened 2020-08-03
Fix showing 'n/a' on stats when on CGroupsV1

marusak:
This got broken with commit b91d6906

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -75,8 +75,7 @@ class TestApplication(testlib.MachineCase):
         self.has_selinux = self.machine.image not in ["debian-testing"]
 
     def cgroupsV2(self):
-        # We will introduce images that are still cgroupsV1 later on
-        return True
+        return self.machine.image not in ["debian-testing"]
 
     def execute(self, system, cmd):
         if system:
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -129,8 +129,8 @@ class Containers extends React.Component {
         let proc = "";
         let mem = "";
         if (containerStats) {
-            proc = containerStats.cpu_stats ? containerStats.cpu_stats.cpu.toFixed(2) + "%" : <abbr title={_("not available")}>{_("n/a")}</abbr>;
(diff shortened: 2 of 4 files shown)


marusak:
Should fix `testLifecycleOperationsUser` for debian-testing. `testLifecycleOperationsSystem` has different issue, debugging now.

marusak:
All 3 remaining debian failures seem to be caused by the same problem:
```
$ podman exec...
timed out waiting for file /var/lib/containers/storage/overlay-containers/d2a7d4f207d0c50bc473c8e2ad159b9a1fc5a553f332255b1c2cea959d2f4212/userdata/db2cb73cd85f55226c37b3323a9dfa98e7e3bf1f55077ed4638e7b083d636776/exit/d2a7d4f207d0c50bc473c8e2ad159b9a1fc5a553f332255b1c2cea959d2f4212: internal libpod error
```
Doing exec twice usually works...

marusak:
Bummer - https://github.com/containers/podman/issues/5339 which seems to be fixed in `conmon 2.0.13`. 

```
$ apt list conmon
Listing... Done
conmon/testing,unstable,now 2.0.9-1 amd64 [installed,automatic]
```

Trying to update conmon did not yield any results. @martinpitt you know these stuff for debian, can you check somewhere if there is plan to update conmon?

test/check-application
```diff
@@ -75,8 +75,7 @@ class TestApplication(testlib.MachineCase):
         self.has_selinux = self.machine.image not in ["debian-testing"]
 
     def cgroupsV2(self):
-        # We will introduce images that are still cgroupsV1 later on
-        return True
+        return self.machine.image not in ["debian-testing"]
```
martinpitt, line 78: Hm, so "hybrid" is not "v2" enough? Both APIs should be available with that. Or is that just due to the outdated conmon?

marusak: Does not seems so. At least not for providing stats for rootless containers.

martinpitt:
@marusak : Right, it's already requested in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=964858

marusak:
> The extra <div> wrappers seem unrelated? But either way, the core change looks fine, thanks!

It is related, it is due to how listing components works, without it, it was oopsing.

martinpitt:
@marusak : I prodded the Debian bug, and in #480 I want to see how it behaves with conmon 2.0.18.

PR #484, opened 2020-08-04
Bump podman dependency to 2.0.4

martinpitt:
This makes sure that we don't use this with the broken 2.0.3 version.


martinpitt:
I also sent https://salsa.debian.org/debian/libpod/-/merge_requests/1 to bump podman's dependency to conman, like in Fedora.

PR #491, opened 2020-08-04
Add actions for Pods

KKoukiou:
Screen Shot 2020-08-05 at 09 35 16

### DIFF
--- a/lib/table.css
+++ b/lib/table.css
@@ -1,0 +1,267 @@
+.panel .table {
+    font-size: var(--font-size-s);
+}
+
+/* Panels don't draw borders between them */
+.panel > .table > tbody:first-child td {
+    border-top: 1px solid rgb(221, 221, 221);
+}
+
+/* Table headers should not generate a double border */
+.panel .table thead tr th {
+    border-bottom: none;
+}
+
+/* Fix panel heading alignment & mobile layout */
+
+.panel-heading {
+    align-items: center;
+    background: #f5f5f5;
+    display: flex;
+    flex-wrap: wrap;
+    /* (28px small size widget height) + (0.5rem * 2) */
+    min-height: calc(28px + 1rem);
+    padding: 0.5rem 1rem 0.5rem 1rem;
+    position: relative;
+    z-index: 100;
+}
+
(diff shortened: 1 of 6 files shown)


src/Containers.jsx
```diff
@@ -485,6 +488,10 @@ class Containers extends React.Component {
                                                 <div className='pod-name'>{caption}</div>
                                                 <div>{_("pod group")}</div>
                                             </h4>
+                                            <div className='panel-actions'>
```
marusak, line 491: Badge is cool, of course, just maybe it should be the same for container statu as well then?

lib/table.css
```diff
@@ -0,0 +1,271 @@
+/*
+ * Keep in sync with https://github.com/cockpit-project/cockpit/tree/master/pkg/lib/table.css
```
marusak, line 2: I prefer to not to write this header - as it makes sync a that bit more difficult :)

KKoukiou: ack I plan anyway to make sync automatic through a makefile rule

marusak: That would be cool :) I just do `cp ../cockpit/pkg/lib/* lib` and then `git add -u`, but rule that would copy just the existing files would be great :)

src/Containers.jsx
KKoukiou: Indeed it looks nicer to be consistent. Fixed.

KKoukiou:
@marusak fixed your comments and localized the pod status strings as well.

KKoukiou:
> Thank you, looks great!

oh wait - I just noticed the extra { t: 0 } I pass by mistake in all API calls, I need to remove this - sorry.

marusak:
This is release-noty for sure, but since we are just introducing pods, I think we can just do one entry like "Show pods - Show pods with containers in them and enable stopping/starting and deleting them"

KKoukiou:
> This is release-noty for sure, but since we are just introducing pods, I think we can just do one entry like "Show pods - Show pods with containers in them and enable stopping/starting and deleting them"

Yes, it should be one entry for pods definitely.

PR #492, opened 2020-08-05
Implement Pod Deletion

KKoukiou:
I had to replace the panels with PF4 Cards because the first mess up with the dropdown, causing them to hide behind the panels of the next rows.
Anyway, panel class is from bootstrap, I should not have used it in the first place.


lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -0,0 +1,53 @@
...
+        justify-content: center;
+    }
+}
+
+ul.pf-c-select__menu {
+    max-height: 20rem;
+    overflow-y: scroll;
+}
+
+/* Drop-down menu should overlay other elements */
```
KKoukiou, line 50: This was added by me now, without it, when one has more than two pods and tries to select the dropdown in the first pod, it gets hidden by the next pod action buttons.

KKoukiou:
@marusak feel free to review this PR, it might get some UI improvements from @garrett today but I think you can do a code review already.

garrett:
I adjusted the spacing in 12ef52e.

(I also have some changes locally that adjust the badges/labels, but that's probably out of scope for this PR.)

KKoukiou:
> I adjusted the spacing in [12ef52e](https://github.com/cockpit-project/cockpit-podman/commit/12ef52eb190bb12de7938b62ad5ca3e311df06d0).
> 
> (I also have some changes locally that adjust the badges/labels, but that's probably out of scope for this PR.)

Nice thanks a lot! Let's merge the badges separately.

marusak:
Release note draft:

**Podman: Support for pod deletion**

The cockpit-podman plugin gained also ability to delete pod groups.

delete

PR #494, opened 2020-08-06
test: Get input and output for 'dd' from container

marusak:
This command sometimes fails in CI, I cannot reproduce locally but thought that maybe this could help?

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -723,7 +723,7 @@ class TestApplication(testlib.MachineCase):
             self.assertIn(full[1], ["GiB", "MiB"])
 
             # Test that the value is updated dynamically
-            self.execute(auth, "podman exec -i swamped-crate dd bs=1024 count=1000000 < /dev/urandom > outfile")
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -723,7 +723,7 @@ class TestApplication(testlib.MachineCase):
             self.assertIn(full[1], ["GiB", "MiB"])
 
             # Test that the value is updated dynamically
-            self.execute(auth, "podman exec -i swamped-crate dd bs=1024 count=1000000 < /dev/urandom > outfile")
+            self.execute(auth, "podman exec -i swamped-crate sh -c 'dd bs=1024 count=1000000 < /dev/urandom > outfile'")
```
martinpitt, line 726: Ah, you figure that piping all that data into and out of podman exec somehow makes it crash? That would be an interesting upstream bug report, as that is an use case for getting lots of data into or out of the container. But it's worth trying if that stabilizes the test, of course.

Do we even need `outfile`? I don't see it being used anywhere. Could just be /dev/null ? Maybe the container just runs into ENOSPC because the 1 GB file doesn't fit?

PR #495, opened 2020-08-06
test: Add Fedora rawhide scenario

martinpitt:
Upgrade podman/container related packages and the kernel to rawhide, so
that we can test it on demand.

Note: The `--releasever` option is necessary to force dnf to download
and install the correct GPG keys. This assumes that this runs on the
most recent stable Fedora release (i.e. fedora-32 at the time of
writing) that precedes rawhide.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,10 +155,16 @@ VM_DEP=$(RPMFILE)
 VM_PACKAGE=--install cockpit-ws --install `pwd`/$(RPMFILE)
 endif
 
+ifeq ($(TEST_SCENARIO),rawhide)
+# needs to run as a separate command, as --run-command is executed before --script
+RAWHIDE=bots/image-customize -v --run-command 'dnf update -y --releasever=$$(. /etc/os-release; echo $$(( VERSION_ID + 1)) ) podman conmon crun containernetworking-plugins containers-common kernel' $(TEST_OS)
+endif
+
 # build a VM with locally built rpm/dsc installed
 $(VM_IMAGE): $(VM_DEP) bots
 	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
 	bots/image-customize -v $(VM_PACKAGE) -s $(CURDIR)/test/vm.install $(TEST_OS)
+	$(RAWHIDE)
 
 # convenience target for the above
 vm: $(VM_IMAGE)


martinpitt:
Triggered rawhide test run now, and I'm also running it locally. I hope that this reproduces at least one of the two failures in [yesterday's Fedora gating run](https://jenkins-continuous-infra.apps.ci.centos.org/blue/organizations/jenkins/fedora-rawhide-build-pipeline/detail/fedora-rawhide-build-pipeline/14944/pipeline/).

Makefile
```diff
@@ -155,10 +155,16 @@ VM_DEP=$(RPMFILE)
 VM_PACKAGE=--install cockpit-ws --install `pwd`/$(RPMFILE)
 endif
 
+ifeq ($(TEST_SCENARIO),rawhide)
+# needs to run as a separate command, as --run-command is executed before --script
+RAWHIDE=bots/image-customize -v --run-command 'dnf install -y fedora-repos-rawhide && dnf update -y --enablerepo=rawhide --releasever=$$(. /etc/os-release; echo $$(( VERSION_ID + 1)) ) podman conmon crun containernetworking-plugins containers-common' $(TEST_OS)
```
marusak, line 160: I thought that just `dnf -y --releasever=33 update podman_and_friends` would be enough. Why do we need `fedora-repo-rawhide`?

martinpitt: Because [StackOverflow](https://stackoverflow.com/questions/24968410/install-single-package-from-rawhide) said so :grin: Indeed it works fine without, and that's much cleaner. I'll update it.

martinpitt:
Both in my local test run and [on CI](https://logs.cockpit-project.org/logs/pull-495-20200806-073024-f3a74b21-fedora-32-rawhide/log.html) all tests succeeded.

But today we got new versions of conmon and podman (they update Rawhide almost every day), so maybe we just had bad luck. However, let me include the kernel in the upgraded packages, as that's rather relevant.

martinpitt:
Now [rawhide did fail](https://logs.cockpit-project.org/logs/pull-495-20200806-074952-39577223-fedora-32-rawhide/log.html), but that's an unrelated common flake.

martinpitt:
Indeed, the rawhide iso used podman and conmon versions which are a week old already. I *can reproduce* these errors in a local run with the same image. After upgrading podman and conmon to their current versions, they still fail in the same way, so there's still something funky going on. But I think this PR is already helpful/useful?

martinpitt:
@marusak: Done in https://github.com/cockpit-project/bots/pull/1133

PR #497, opened 2020-08-06
Dowload test fixes

marusak:
I have no idea how it worked before...

### DIFF
--- a/src/ImageDeleteModal.jsx
+++ b/src/ImageDeleteModal.jsx
@@ -5,13 +5,21 @@ import cockpit from 'cockpit';
 
 const _ = cockpit.gettext;
 
+function sortTags(a, b) {
+    if (a.endsWith(":latest"))
+        return -1;
+    if (b.endsWith(":latest"))
+        return 1;
+    return a.localeCompare(b);
+}
+
 export class ImageDeleteModal extends React.Component {
     constructor(props) {
         super(props);
 
         const tags = {};
         const repoTags = this.props.imageWillDelete.RepoTags ? this.props.imageWillDelete.RepoTags : [];
-        repoTags.forEach((x, i) => {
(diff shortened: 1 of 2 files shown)


src/ImageDeleteModal.jsx
```diff
@@ -5,13 +5,21 @@ import cockpit from 'cockpit';
 
 const _ = cockpit.gettext;
 
+function sortTags(a, b) {
+    if (a.endsWith(":latest"))
+        return -1;
+    if (b.endsWith(":latest"))
+        return 1;
+    return a > b;
```
martinpitt, line 13: This looks wrong. The compare function is supposed to return -1, 0, or 1, not false/true -- turning the latter into a number in Python just  yields NaN. How about `a.localeCompare(b)`?

src/ImageDeleteModal.jsx
```diff
@@ -5,13 +5,21 @@ import cockpit from 'cockpit';
 
 const _ = cockpit.gettext;
 
+function sortTags(a, b) {
+    if (a.endsWith(":latest"))
```
martinpitt, line 9: so `RepoTags` really contains the `:` in the names? (That sounds strange)

martinpitt:
I applied the first two commits (test fixes) to my local Fedora gating VM, and they do fix the test. :balloon: thanks!

src/ImageDeleteModal.jsx
marusak: repo tags are the full strings with registry, name and tag.
`[ "docker.io/library/busybox:latest", "docker.io/library/busybox:1", "docker.io/library/busybox:2", "docker.io/library/busybox:3" ]` for example

src/ImageDeleteModal.jsx
marusak: fixed using `localCompare`

PR #498, opened 2020-08-06
Recolor badges

garrett:
Recolor badges based on status

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import ReactDOM from "react-dom";
 import {
-    Button, Badge,
+    Button,
     Card, CardBody, CardHeader, CardTitle, CardActions,
+    Label,
     Text, TextVariants, FormSelect, FormSelectOption,
 } from '@patternfly/react-core';
 import { TrashIcon } from '@patternfly/react-icons';
@@ -158,7 +159,14 @@ class Containers extends React.Component {
             proc,
             mem,
             container.isSystem ? _("system") : this.props.user,
-            { title: <Badge isRead>{_(container.State)}</Badge> }, // States are defined in util.js
+            {
+                title: (
+                    <Label color={container.State == 'running' ? 'green' : undefined}
+                           className={"container-state container-state-" + container.State}>
+                        {_(container.State)}
+                    </Label>
+                )
+            }, // States are defined in util.js
         ];
 
         const tty = containerDetail ? !!containerDetail.Config.Tty : undefined;
@@ -490,7 +498,8 @@ class Containers extends React.Component {
                                                     <span>{_("pod group")}</span>
                                                 </CardTitle>
                                                 <CardActions className='panel-actions'>
-                                                    <Badge isRead>{_(this.props.pods[section].Status)}</Badge>
+                                                    <Label color={this.props.pods[section].Status == 'Running' ? 'green' : undefined}
+                                                       className={"pod-state-" + this.props.pods[section].Status.toLowerCase()}>{_(this.props.pods[section].Status)}</Label>
                                                     <PodActions onAddNotification={this.props.onAddNotification} pod={this.props.pods[section]} />
                                                 </CardActions>
                                             </CardHeader>}
--- a/src/Containers.scss
+++ b/src/Containers.scss
@@ -33,3 +33,18 @@
 .ct-table-empty td {
     padding: var(--pf-global--spacer--sm) var(--pf-global--spacer--md) var(--pf-global--spacer--lg);
 }
+
+// Fix label color and alignment in the table cell
+.container-state {
+  // Unless there is a modifier (ex: pf-m-green for running), drop the color
+  &:not([class*=pf-m-]) {
+    --pf-c-label--BackgroundColor: transparent;
+    padding-right: 0;
+    // Compensate for padding
+    position: relative;
+    // Using margin to remove the extra padded space,
+    // instead of keeping the space and moving contents w/ just left/top
+    margin-left: calc(-1 * var(--pf-c-label--PaddingLeft));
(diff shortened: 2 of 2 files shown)


garrett:
This builds on top of #492. We'll eventually drop the other commits and rebase on master at some point in time.

This branch isn't done yet, but I wanted to share what I had before the weekend hit. :wink:

garrett:
PF4 recently introduced "labels". It's an overloaded term, but it's basically a more generic badge, that is less for an read/unread status and more for adding keywords (aka: tags, and \*ahem\* labels) to something. In commit d7489, I switched them over in the React and CSS.

While I'm using CSS for styling right now, perhaps we want to assign color states in the `<Label>` attributes instead? (This would just be for notable ones, like running, error, and perhaps another one or two. The default state is good enough for all the rest, including the ones that aren't currently running.) We might want to consider an icon in some of the label widgets too.

(Note: These widgets called "label" are not to be confused with the form widgets called "label" or the HTML element the PF4 form "label" is based on _also_ called "label", despite all being called the same thing.)

marusak:
@garrett is this blocked on something or just too busy? :) Please let me know if I can help with it in any way.

garrett:
I've been busy. Please pick this up if you'd like. That'd be great.

KKoukiou (commented):
I will rebase this one, it's a pity to leave this here so long time. It's almost ready.

src/Containers.jsx
```diff
@@ -154,7 +154,14 @@ class Containers extends React.Component {
             proc,
             mem,
             container.isSystem ? _("system") : this.props.user,
-            { title: <Badge isRead>{_(container.State)}</Badge> }, // States are defined in util.js
+            {
+                title: (
+                    <Label color={container.State == 'running' ? 'green' : undefined}
```
garrett, line 159: I see "green" specified here, but I have a light blue on my page.

Screenshot_2020-11-17 Podman Containers - garrett Rain

garrett (commented):
I think this PR is fine, although I think the "green" isn't working for some reason. I have blue here. Weird, right?

src/Containers.jsx
garrett: Scratch the color thing. That was the result of being on an old checkout of the branch. After running git clean, doing a hard reset to the remote branch, and rebuilding, it's green (and up to date with where this branch is right now).

Everything looks fine now. :+1:

garrett:
LGTM :+1:

src/Containers.jsx
marusak: Seems green on my checkout:
green

I thought that other badges (like `exited` would have also that badge styling, but they dont)

KKoukiou:
> `exited` does not have that badge wrapping. I though it would have the gray one.
> 
> green

Not - this is overriden with CSS_ @garrett  wanted the exited/inactive states to not have background.

garrett:
Oh, the badges don't have the alignment adjustment. 😟

KKoukiou:
> Oh, the badges don't have the alignment adjustment. worried

The alignment adjustment is needed because for some of them we remove the background and make it transparent, right? And then we need to realign all labels. @marusak pointed out earlier that he finds weird the transparency in 'some' labels. Maybe we can reconsider having some grey background and drop the overrides in the CSS? Code vise it would be much more straightforward.

garrett:
The alignment adjustment is needed to make the text line up. That is, for label and non-label text to both align properly... this includes the headers.

Labels with transparent backgrounds and labels with colorful backgrounds should both line up without adjustment, but they won't align with the rest of the text.

The background color draws attention, and this is why I suggest dropping the background color for standard, non-error, non-action states that really shouldn't have attention.

src/Containers.scss
```diff
@@ -33,3 +33,19 @@
 .ct-table-empty td {
     padding: var(--pf-global--spacer--sm) var(--pf-global--spacer--md) var(--pf-global--spacer--lg);
 }
+
+// Fix label color and alignment in the table cell
+.container-state {
+  // Container states: configured created running stopped paused exited removing
+  &-configured, &-created, &-stopped, &-exited {
```
marusak, line 40: Since we add color to only 'running' can we just turn this around and remove it if it is not `running`? Seems easier to maintain and less likely to break with some state we don't count on.

marusak:
Needs rebase.

src/Containers.scss
garrett: We don't add any color. It uses the "green" PF property, specified in the JS, and PF handles the style. We want these specific states to be transparent.

We could add a transparent state in the JS, but I didn't want to specify style in the code; I wanted to keep it separate.

src/Containers.scss
```diff
@@ -33,3 +33,18 @@
...
+.container-state {
+  // Unless there is a modifier (ex: pf-m-green for running), drop the color
+  &:not([class*=pf-m-]) {
+    --pf-c-label--BackgroundColor: transparent;
+    padding-right: 0;
+    // Compensate for padding
+    position: relative;
+    // Using margin to remove the extra padded space,
+    // instead of keeping the space and moving contents w/ just left/top
+    margin-left: calc(-1 * var(--pf-c-label--PaddingLeft));
```
KKoukiou, line 47: With this the mobile mode looks nice but on regular monitor sizes it looks worse:
Without this line:

Screen Shot 2021-02-04 at 12 17 44

With this line:

Screen Shot 2021-02-04 at 12 17 16

KKoukiou (changes requested):
I feel we should have the desktop in priority against mobile so that margin-left is a problem

src/Containers.scss
garrett: I had something to offset the state for the running label too, but I don't see it here in your screenshots. Perhaps it got lost in a rebase.

PR #500, opened 2020-08-06
test: Wait for proper containers in the UI

marusak:
In this test we create and delete many containers named 'test-sh' from
'alpine' image. There exists race, where in the UI we still see the
previous 'test-sh' container, that we think we already deleted and
expect to see a new 'test-sh' container in the UI we just created. When
we are fast enough, we have time to toggle open the container details,
but then it gets deleted, a new container is rendered but it does not
have details expanded.

Thus let's name every container uniquely and when we target specific
container line, let's do it by its name and not by its image (which
makes so much more sense anyway).

Should fix flake like [this](https://logs.cockpit-project.org/logs/pull-499-20200806-160515-6e8157f5-fedora-32-rawhide/log.html)

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -352,35 +352,32 @@ class TestApplication(testlib.MachineCase):
 
         # run a container (will exit immediately) and test the display of commit modal
         if auth:
-            self.execute(True, "podman run -d --name test-sh alpine sh")
-            container_commit("test-sh")
+            self.execute(True, "podman run -d --name test-sh0 alpine sh")
+            container_commit("test-sh0")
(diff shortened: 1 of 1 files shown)


marusak:
PR #500 :partying_face: 
Let's hope it is the last flaky test fix and the next 500 PRs will be flake free :smile: :crossed_fingers:

marusak:
Oh, wait https://github.com/cockpit-project/cockpit-podman/pull/497 did not land yet, so there might be some flakes that PR fixes. Lets land that first, then spin this 3 times

PR #501, opened 2020-08-07
Refresh translations from weblate

cockpituous:
There has been issue with weblate not syncing pot files properly, so I want to test it out.

 * [x] po-refresh


marusak:
oooh, I cannot approve as I opened the issue in the first place.

PR #502, opened 2020-08-07
Adjust weblate branch to this rhel8 branch

martinpitt:


martinpitt:
Maybe we don't need this after all, as translations for 8.3 are pretty much done.

PR #505, opened 2020-08-11
Add import and export of checkpoints

lenticularis39:
Adds export (download of checkpoint tarball after making a checkpoint) and import (upload of checkpoint tarball for restore) support.

Upload is done by directly reading a file selected by the PatternFly file upload component; ~download is handled using a service worker (it's a little bit hacky, but seems to work fine in both Firefox and Chromium and doesn't store the tarball in memory; maybe there is a better way, but as far as I know Cockpit can download only files from the file system, not channels/HTTP streams)~ download uses channel redirection to save the checkpoint into a file, which is then downloaded directly using a read channel. (Direct download of the checkpoint stream is not possible, because the checkpoint process takes some time, and also errors have to be handled in the UI, not downloaded instead of the tarball).

There are currently no tests - I'm not sure how Selenium handles downloads (the download has to be done in the test browser because of the service worker); I can have a look at it.

checkpoint
restore

### DIFF
--- a/src/ContainerCheckpointModal.jsx
+++ b/src/ContainerCheckpointModal.jsx
@@ -11,7 +11,8 @@ class ContainerCheckpointModal extends React.Component {
             keep: false,
             leaveRunning: false,
             tcpEstablished: false,
-            ignoreRootFS: false
+            ignoreRootFS: false,
+            export: false
         };
         this.handleChange = this.handleChange.bind(this);
     }
@@ -51,6 +52,9 @@ class ContainerCheckpointModal extends React.Component {
                     <Checkbox label={_("Do not include root file-system changes when exporting")}
                                   id="checkpoint-dialog-ignoreRootFS" name="ignoreRootFS"
                                   isChecked={this.state.ignoreRootFS} onChange={this.handleChange} />
+                    <Checkbox label={_("Export the checkpoint into a tarball and download it")}
+                                  id="checkpoint-dialog-export" name="export" isChecked={this.state.export}
+                                  onChange={this.handleChange} />
                 </Form>
             </Modal>
         );
--- a/src/ContainerRestoreModal.jsx
+++ b/src/ContainerRestoreModal.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Button, Checkbox, Form, Modal } from '@patternfly/react-core';
+import { Button, Checkbox, FileUpload, Form, Modal } from '@patternfly/react-core';
 import cockpit from 'cockpit';
 import * as utils from './util.js';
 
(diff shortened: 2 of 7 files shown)


marusak:
Sorry for getting to you only now. It looks cool! I have a few questions first.

@andreasn can you please take a look design-wise? Thanks!

src/download-sw.js
```diff
@@ -0,0 +1,82 @@
+/*
```
marusak, line 1: Have you written this file on your own or is it taken from some other place?
In the first case please add here the same copyright notice as we have in `src/app.jsx` or so.
If not, then this might be a bit trickier.

marusak: Haven't actually read this yet, just thought if it is necessary. We download archives for example in sosreport package - https://github.com/cockpit-project/cockpit/tree/master/pkg/sosreport

lenticularis39: > Have you written this file on your own or is it taken from some other place?
> In the first case please add here the same copyright notice as we have in `src/app.jsx` or so.
> If not, then this might be a bit trickier.

I've written this myself - I'll add the copyright notice.

lenticularis39: > Haven't actually read this yet, just thought if it is necessary. We download archives for example in sosreport package - https://github.com/cockpit-project/cockpit/tree/master/pkg/sosreport

That looks like it could be used to download the response, but I think it would start the download immediately after sending the request, causing the download to be stuck at 0 B/s until the checkpoint is done on the server- I'll try it.

lenticularis39: > I think it would start the download immediately after sending the request, causing the download to be stuck at 0 B/s until the checkpoint is done on the server- I'll try it.

I checked it and this doesn't happen, but there are other problems:

- The session timeouts when the checkpoint takes slightly longer (around 15 seconds). (I don't know whether this is expected - it might be a bug in the handling of external channels in cockpit-ws.)
- The Cockpit page doesn't get notified when the download starts (i.e. the checkpoint is finished).

marusak:
I think I cannot make this work. Let see.
When I pick export on my local (cgroupsv2) system, I get my favourite error:
whataniceerrormessage
When I try this in fedora-31 VM, sitting in tests (so it switches to cgroupsv1) it does not produce any error,, but the dialog just gets closed after a bit of waiting without any download prompt (thats what I was expecting). Tried on chromium and firefox, same result. Am I missing here something?

lenticularis39:
I tried it again myself and I discovered a possible cause of your problems. 

There is a bug that causes the download not to trigger in Firefox when opening Cockpit for the first time in the browser session (it probably has something to do with the service worker not running) - reloading the page works for me. I thought this was fixed, but apparently that's not the case.

In Chromium, the service worker doesn't start when the SSL certificate couldn't be verified (it throws an error in my case, but it's possible that doesn't happen on every version of Chromium).

I'll try to fix the bug and push the changes.

lenticularis39:
I've fixed the issue in Firefox (which is apparently caused by the fetch event not being registered at the point of the fetch request in the iframe).

lenticularis39:
This is apparently still buggy in Firefox - marking as draft for now.

andreasn:
Since this is still a draft, I'll take this off my UI review queue for now.
Judging from the screenshots, it looks like a good approach though.

lenticularis39:
Dropped the idea of downloading through a service worker due to continuing problems in Firefox, the new solution uses channel redirection (only locally, not the controversial remote redirection that might be rejected in Cockpit). 

This means it's now also blocked on cockpit-project/cockpit#14641 (like #590).

lenticularis39:
Like #590, blocked on closed https://github.com/cockpit-project/cockpit/pull/14641. Closing.

PR #507, opened 2020-08-12
Listen on event for image build

marusak:
In https://github.com/containers/podman/pull/7202 a new event was
introduced. Listen on it as well.

This is not going to work out just now, but I am opening it now, so we won't forget. Once new podman (I guess 2.0.5 should get it) is out and in our VMs, we can nicely verify it with retrying tests in this PR.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -326,6 +326,7 @@ class Application extends React.Component {
         case 'untag':
         case 'remove':
         case 'prune':
+        case 'build':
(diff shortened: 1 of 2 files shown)


marusak:
Finally it landed in all images so is ready. @KKoukiou can you please take a look? Thanks!

Gundersanne (approved):
looks good! Add fixes# & closes#?

PR #511, opened 2020-08-18
Image create time moves to timestamp

marusak:
This was introduced somewhere between 2.0.4 and 2.1.0.
Our `fedora-32/rawhide` already sees that, but not our other images.

I have to add there some condition so it works also with 2.0.4 but I want to spin tests once now to see it really is needed for rawhide.

### DIFF
--- a/src/ContainerDetails.jsx
+++ b/src/ContainerDetails.jsx
@@ -2,12 +2,11 @@ import React from 'react';
 import cockpit from 'cockpit';
 import * as utils from './util.js';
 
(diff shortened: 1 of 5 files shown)


marusak:
As expected [f32](https://logs.cockpit-project.org/logs/pull-511-20200818-083042-3b9b063d-fedora-32/log.html) and [rawhide](https://logs.cockpit-project.org/logs/pull-511-20200818-083042-3b9b063d-fedora-32-rawhide/log.html)

PR #515, opened 2020-08-20
Adding support for PPC64LE

ezeeyahoo:
Hi,

I have tested these changes included to support PPC64LE and request you for a review.

Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,26 @@
 dist: trusty
-sudo: false
-language: node_js
-node_js:
+
+language: nodejs
+
+nodejs:
   - "8"
+
+arch:
+  - amd64
+  - ppc64le
+  
+matrix:
+  include:
+    - arch: ppc64le
+      dist: focal
+      
+  exclude:
+    - arch: ppc64le
+      dist: trusty
+      
 script:
   - npm install
   - npm run build
+
+before_install:
+  - curl https://raw.githubusercontent.com/creationix/nvm/v0.35.3/install.sh | bash


marusak:
May I ask why do you need this?
We dropped `.travis.yml` file from other cockpit projects and maybe it is time to do the same for c-podman. We surely won't do that if there are users that need it, but if it is not necessary to keep it in the repository, then I vote for removing it.

gerrith3:
@marusak this project happens to be a part of the Red Hat distributions, including their ppc64le builds.  We are enabling upstream CI for those to help reduce the cost of maintaining/debugging issues and to ensure the port stays current on ppc64le.  If you dropped .travis.yml from other projects, what CI are you using?  We have very few that support Power today (jenkins, Travis, and possible github actions).  We would like to stay in sync with what you are already doing as much as possible.

martinpitt:
@ezeeyahoo : That travis config won't do anything useful, I'm afraid. The build process is merely about calling `npm install` and `webpack`, none of which is particularly architecture specific or interesting -- and it will be completely irrelevant for actually *running* cockpit-podman.noarch.rpm on e.g. ppc64el, as the official RPM builds use the release tarballsl.

What *would* be useful is to run the integration tests on non-x86 hardware for sure! We don't have hardware for this right now. A while ago I was playing around with running integration tests in travis/semaphore/github actions, but these are usually hampered by not exposing /dev/kvm to the test environment. We can only meaningfully test cockpit-* in VMs, not some random ubuntu container on travis/semaphore.

gerrith3:
@martinpitt We do provide free open source access to Power VMs at http://osuosl.org/powerdev - let me know if you are interested and I'd be willing to be your IBM sponsor.  That would be a real Power8 or Power9 VM.

PR #526, opened 2020-09-02
cockpituous: Stop releasing to Fedora 31

martinpitt:
Fedora 33 is branched now, let's not support too many releases in
parallel.


martinpitt:
If you are ok with this, we should stop the f31 tests next.

marusak (approved):
sure, lets do that :) I'll remove the requested context and send PR for dropping it from the testmap.

PR #540, opened 2020-09-21
DBG - let the bots run

mvollmer:
I get a mysterious failure locally on debian-testing; let's see what the official bots say.


mvollmer:
Thank you, bots.

PR #541, opened 2020-09-21
test: Disable podman.socket globally for the tests

mvollmer:
Otherwise, "systemctl --user disable podman.socket" has no effect on
"systemctl --user is-enabled".

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -63,6 +63,8 @@ class TestApplication(testlib.MachineCase):
         # Enable user service as well
         self.restore_dir("/home/admin/.local/share/containers")
         self.admin_s.execute("systemctl --now --user enable podman.socket")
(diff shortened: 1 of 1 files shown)


mvollmer:
See #540 for an example of the failure that otherwise happens.  I have no idea why this hasn't bitten us before. We install a number of packages at the time of running the test, so some change in those might explain it.

On Fedora, podman.socket is not enabled by vendor preset, but on Debian it is. I think on Debian every user unit is enabled by vendor preset, actually, since it doesn't have the same /usr/lib/systemd/user-preset/99-default-disable.preset file as Fedora has.

This failure is blocking the debian-testing image refresh.

martinpitt (approved):
Danke!

PR #547, opened 2020-09-27
test: Don't require docker.service

martinpitt:
We are dropping docker from the debian image.

See https://github.com/cockpit-project/bots/pull/1262


martinpitt:
Oh, [io.podman.service](https://logs.cockpit-project.org/logs/pull-547-20200927-084608-97d920da-debian-testing/log.html) is gone as well, debian-testing has podman 2.0.6 now. A lot of stuff in vm.install is obsolete now, I dropped it.

marusak (approved):
Thanks! I like when we remove hacks :)

PR #549, opened 2020-09-28
test: Explicitly install runc on Debian testing

martinpitt:
The podman package's dependencies now prefer crun, so the previous hack
for https://bugs.debian.org/961016 is obsolete (and that bug is closed).

However, podman *really* defaults to runc, so explicitly install it
for now until the default configuration or dependencies get fixed:
https://bugs.debian.org/971253

This was previously hidden because the debian-testing image contained
docker.io, which pulls in runc. But
https://github.com/cockpit-project/bots/pull/1262 removes docker.io, and
runc with it.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -7,8 +7,8 @@ if [ -d /var/tmp/debian ]; then
     apt-get update
     eatmydata apt-get install -y cockpit-ws cockpit-system podman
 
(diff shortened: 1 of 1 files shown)


marusak (approved):
uh oh, nice catch! Thanks

Issue #550, opened 2020-09-29
Update translations from Fedora Weblate


cockpituous:
Update translations from Fedora Weblate
 * [ ] FAIL: po-refresh

cockpituous:
po-refresh in progress on 3-cockpit-7.
Log: https://logs.cockpit-project.org/logs/po-refresh-550-20200929-121414/

cockpituous:
Command '('bots/make-source', 'update-po')' returned non-zero exit status 2.: https://logs.cockpit-project.org/logs/po-refresh-550-20200929-121414/

marusak:
```
Option "--directory" requires 1 arguments, but 0 were provided
USAGE: node manifest2po [OPTION1] [OPTION2]... arg1 arg2...
```
what?

martinpitt:
@marusak: I just fixed that in starter-kit yesterday: https://github.com/cockpit-project/starter-kit/pull/375

This needs to be ported to podman. This was fallout from bumping the "stdio" npm dependency, and we failed to have `make pot` in our tests anywhere.

marusak:
Sent #551

PR #551, opened 2020-09-30
Fix building PO template

marusak:
The current version failed with

    Option "--directory" requires 1 arguments, but 0 were provided

In the current stdio npm module version, getopt() now requires a
`default:` attribute, otherwise it considers the options as required.

Also sync PO helpers with current cockpit.

Cherry-picked from cockpit-project/starter-kit commit cbedf06

Checked and .pot file is built correctly. When we land this we need to retry #550

### DIFF
--- a/po/html2po
+++ b/po/html2po
@@ -45,9 +45,9 @@ try {
 }
 
 var opts = stdio.getopt({
-    directory: { key: "d", args: 1, description: "Base directory for input files" },
+    directory: { key: "d", args: 1, description: "Base directory for input files", default: "." },
     output: { key: "o", args: 1, description: "Output file" },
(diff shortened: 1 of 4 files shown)


martinpitt (changes requested):
Can you please also add this to travis, so that we don't fall into the same trap again? The original starter-kit commit does that.

marusak:
> Can you please also add this to travis, so that we don't fall into the same trap again?

done

martinpitt:
Sorry, I didn't realize that we don't have travis set up for c-podman. How about adding it to test/run instead?

marusak:
@Gundersanne can you please take a look? Thanks!

Gundersanne (approved):
lgtm :)

PR #558, opened 2020-10-12
cockpituous: Run release in a GitHub workflow

### LINKED ISSUE #153: Check for NPM dependency updates
Check for any outstanding update of point version of any outstanding NPM package dependency
 * [ ] npm-update

martinpitt:
Enter the new world of GitHub actions [1]/GitLab pipelines [2]. This
simplifies our end of the infrastructure considerably:

 * No need any more to set up webhooks, all the relevant configuration
   is right in the workflow file.

 * Does not need any infrastructure on our side any more, and thus works
   for third-party projects. They just need to set up their own secrets.

 * Gets rid of our webhook's special case for releases, which spawns an
   OpenShift Job (which means we can replace that webhook code with
   something more generic). Structually, spawning a Job is exactly like
   running an action/pipeline.

 * GitHub automatically provides a temporary `GITHUB_TOKEN` with
   sufficient write access to the project to publish a release, so we
   don't need to carry around that secret.

[1] https://github.com/features/actions
[2] https://docs.gitlab.com/ee/ci/pipelines/

Closes #153

 - [x] Disable release webhook

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,32 @@
+name: release
+on:
+  push:
+    tags:
+      # this is a glob, not a regexp
+      - '[0-9]*'
+jobs:
+  cockpituous:
+    runs-on: ubuntu-latest
+    container:
+      image: docker.io/cockpit/release
+    steps:
+      - name: Set up configuration and secrets
+        run: |
+          # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
+          export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
+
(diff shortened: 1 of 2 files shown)


martinpitt:
This is exactly the same as https://github.com/cockpit-project/cockpit-ostree/pull/153 , and that just [ran successfully](https://github.com/cockpit-project/cockpit-ostree/runs/1241925956?check_suite_focus=true). c-podman has exactly the same release targets as -ostree.

marusak (approved):
:heart:

martinpitt:
OK, let's do this :grin: webhook disabled. @Gundersanne, FYI, this will affect the release on Wednesday (if you are going to release c-podman). Of course I am standing by to pick up the pieces should anything go wrong.

PR #560, opened 2020-10-16
Update README: cockpit-docker no longer exists

slokhorst:
cockpit-docker was dropped: cockpit-project/cockpit#14585

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,7 @@
 This is the [Cockpit](https://cockpit-project.org/) user interface for [podman
 containers](https://podman.io/).
 
-It is being actively developed and has not yet reached feature parity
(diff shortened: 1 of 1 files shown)


marusak (changes requested):
I think we reached future parity - at least in the sense where it made the most sense.
We also support `pod`, not just `image and container` tasks.
I would drop the whole line about it being actively developed.
Probably also the next line, or would write there a bit more what it can and cannot do. Are you up for that?

slokhorst:
@marusak Thanks for the explanation. I'm not that familiar with this project yet (I was just looking how to get my containers into Cockpit).. so I think I better leave this up to you.

PR #567, opened 2020-10-20
Run npm-update in a GitHub workflow

martinpitt:
This avoids a lot of moving parts in our infrastructure (webhook,
npm-trigger roundtrip, tasks container) and does not need *any* secret
other than the automatically provided GitHub token.

Adopted from starter-kit: https://github.com/cockpit-project/starter-kit/pull/384

PR #568, opened 2020-10-20
Run po-refresh in GitHub workflow

martinpitt:
Run them on Tuesday evening, to maximize the time for translators, and
still give us enough time to test/review/land the translations before
Wednesday releases.

This entirely replaces the cockpituous .tasks file.

 - [x] Builds on top of PR #567 (conflict on `.tasks` file)


martinpitt:
I [ran this on my repo](https://github.com/martinpitt/cockpit-podman/runs/1279337121?check_suite_focus=true) and it behaves as expected: It picks up new translations, but then fails to commit to the weblate repo as ~martinpitt does not have access there. Once this lands, we can trigger it manually or wait until this evening for the automatic run.

marusak (approved):
Thanks! Can we trigger it manually to see it works?

martinpitt:
@marusak: Yes, see the README -- on the "Actions" tab here you will see the "po-refresh" workflow, just like you already see the "npm-update" workflow. If you click on it, you have a button to run it manually. That's the `on.workflow_dispatch` trigger in the yml file.

martinpitt:
The main thing I'm not sure about is whether the automatic github token is good for committing to the weblate repo -- I have a gut feeling that it won't, but it's the same org. So I propose I'll land this, run this manually, see what breaks and how to fix it (at worst, we have to add our cockpituous token to secrets and install that), and if it gets too bad I'll revert this. Tomorrow is not release day, so it won't be super-urgent.

PR #574, opened 2020-10-23
 package.json: Add 'file-selector' dependency

marusak:
Locally I had to edit package.json to be actually able to build this project, so want to check with CI if there is the same problem or not.


marusak:
https://github.com/cockpit-project/cockpit/pull/14789

So it is an issue. But it is likely to get resolved very soon. Lets see if we also need it here.

allisonkarlitskaya (approved):

PR #576, opened 2020-10-25
Migrate modals to PF4

marusak:
Now only very few things from PF3 are still being used. I would like to migrate that in another PR as this one is rather big on its own.


KKoukiou (changes requested):
You need to sync PF4 overrides CSS with commit e6d4403e14  from master because otherwise alerts are displayed inline with the buttons in the footer.

Like this:
Screen Shot 2020-10-26 at 09 23 49

marusak:
Good catch! I've updated this PR, but we need to release these changes first. I could use sha, but lets just block this for a few days.

marusak:
Now looks correct.

worksnow

marusak:
Tests are red! (they are just now not showing it). Lets not merge this.

marusak:
Green, ready for review

KKoukiou (commented):
The spinner in the footer is going crazy, I will check what's wrong and comment here.

KKoukiou (commented):
So footer from PF4 has display: flex now.
spinner does not behave well as a child of a flex container.
So we need to choose one of the following:
* either change the spinner to have a wrappre
* move the spinner inside the button 
* override PF4 footer flex property and have a block a it was before.

KKoukiou (approved):

marusak:
Nice, thanks! Looks good to me :)

PR #577, opened 2020-10-25
test: Support generic Debian environments

martinpitt:
This makes the script reusable in environments like GitHub, GitLab,
Travis, or Debian/Ubuntu autopkgtest.

- Build debs directly on the machine if pbuilder is not available.

- Create admin user and set up SSH to root for test.

- Make test/vm.install executable, so that it can be called easily
  directly.

- Support passing additional options to `apt-get install`.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -5,7 +5,7 @@ set -eu
 # for Debian based images, build and install debs; for RPM based ones, the locally built rpm gets installed separately
 if [ -d /var/tmp/debian ]; then
     apt-get update
-    eatmydata apt-get install -y cockpit-ws cockpit-system podman
+    eatmydata apt-get install ${APT_INSTALL_OPTIONS:-} -y cockpit-ws cockpit-system podman
 
     # HACK: podman dependencies prefer crun, but config defaults to runc: https://bugs.debian.org/971253
     eatmydata apt-get install -y runc
@@ -21,11 +21,15 @@ if [ -d /var/tmp/debian ]; then
     cp -r ../debian .
     sed -i "s/(0-1)/(${VERSION}-1)/" debian/changelog
     dpkg-buildpackage -S -us -uc -nc
-    cd ..
 
-    # build and install binary package
-    pbuilder build --buildresult . *.dsc
-    dpkg -i *.deb
+    # build and install binary package; prefer pbuilder if available (on Cockpit test VMs)
+    if [ -e /var/cache/pbuilder/base.tgz ]; then
+        pbuilder build --buildresult .. ../*.dsc
+    else
+        eatmydata apt-get install ${APT_INSTALL_OPTIONS:-} -y build-essential debhelper
+        dpkg-buildpackage -us -uc -b
+    fi
+    dpkg -i ../*.deb
 
     # Debian does not enable user namespaces by default
     echo kernel.unprivileged_userns_clone = 1 > /etc/sysctl.d/00-local-userns.conf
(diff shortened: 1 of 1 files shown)


martinpitt:
[Tested on my own fork](https://github.com/martinpitt/cockpit-podman/actions/runs/327767566)

martinpitt:
The centosci OCP log links are wrong, see https://github.com/cockpit-project/cockpituous/pull/348 . The real links are:

 - [fedora-32/rawhide](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-577-20201025-200850-27e422f7-fedora-32-rawhide/log.html)
 - [fedora-32](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-577-20201025-200850-27e422f7-fedora-32/log.html)

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
...
+          cp -r packaging/debian/ /var/tmp
+          sudo -n env APT_INSTALL_OPTIONS="-t focal-backports" test/vm.install
+
+      - name: Install test dependencies
+        run: |
+          sudo -n apt-get install -y chromium-browser python3-libvirt
+          make test/common bots
+
+      # interactive debugging
+      # - uses: mxschmitt/action-tmate@v3
```
martinpitt, line 28: Not sure if we want to leave that in, but I thought it's convenient to have.

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Run test
+        # FIXME: debian-testing is not quite accurate, but our tests don't recognize ubuntu yet
+        run: TEST_AUDIT_NO_SELINUX=1 TEST_OS=debian-testing test/common/run-tests -tv --machine 127.0.0.1:22 --browser 127.0.0.1:9090
```
martinpitt, line 32: We may want to drop `-tv` for the final version to make the logs less massive.

martinpitt:
BTW, I'm completely fine to drop this if you don't like this -- it was mostly a study whether this works. I *would* like to do this to starter-kit though, as there it may be really useful for people who clone it to have a fully working CI without further setup (if only on Ubuntu 20.04 LTS).

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
+name: test
+on: [push, pull_request]
```
martinpitt, line 2: Also, if we want to go ahead with this, let's limit this to just `pull_request` to avoid running this twice.

.github/workflows/test.yml
marusak: No problem with leaving it in

.github/workflows/test.yml
marusak: Yeah, I think `-v` is enough. `-t` for normal runs is just too much.

.github/workflows/test.yml
marusak: push would be mostly just on merging, right? but  agree that `pull_request` is just enough.

marusak:
> BTW, I'm completely fine to drop this if you don't like this -- it was mostly a study whether this works. I would like to do this to starter-kit though, as there it may be really useful for people who clone it to have a fully working CI without further setup (if only on Ubuntu 20.04 LTS).
However, I think it is useful to test c-podman on Ubuntu as well -- whether we do that this way or the more classic way through our ubuntu-latest image is mostly a matter of taste, I suppose.

Like this! As alternative to our CI.  I still prefer our weekly refreshes where we have control over our images. Here we don't have such control and when the github image changes we may have unexpected broken tests. But I think it is nice to have and podman seems like a good candidate. If it would become problematic we can easily disable it. And agree that it should be in starter kit, very useful for new projects!

.github/workflows/test.yml
martinpitt: merging and also just pushing to own repos before sending a PR.

martinpitt:
I did the cleanups above. I'm also experimenting with running qemu on Travis -- it seems to generally support that now (there is a /dev/kvm), but I still didn't actually manage to get this  running.

So indeed, let's just let this run for a while, and re-disable it if it gets in the way. Or maybe I manage to get KVM working on travis (or semaphore), then we can again test with our own regular VM images.

martinpitt:
I figured out how to run tests with our normal machinery [in Travis](https://github.com/cockpit-project/starter-kit/pull/386). I'll try to apply the same thing to cockpit-podman, and send an alternate PR. However, I'd like to keep at least the first commit here, as that's useful for re-using in downstream tests (Fedora/RHEL/Debian/Ubuntu).

martinpitt:
This was mostly a PoC and exploring alternative infrastructure. I think we should either take #578 instead, or just keep what we have. Our CI works fine, after all.

But I'll re-use this to define a proper autopkgtest for Debian/Ubuntu. Hence I just kept the first commit here.

marusak (approved):
Thanks!

PR #578, opened 2020-10-27
Run integration tests in Travis

martinpitt:
Travis now offers /dev/kvm, and its machines are powerful enough to run
our browser integration tests, at least for small projects.

Building an RPM on the Ubuntu host environment is a bit tricky, as there
are no installed RPMs. Thus ignore the BuildRequires, and install
appstream-util explicitly. In the future, the rpm build should happen
inside the VM (like Cockpit does).

Add initial scenario for current Fedora.

Switch to the "minimal" environment to make the test easier to reproduce
locally and more explicit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,29 @@
-dist: bionic
-sudo: false
-language: node_js
-node_js:
-  - "lts/*"
+dist: focal
+sudo: true
+language: minimal
+addons:
+  apt:
+    packages:
+      - appstream-util
+      - chromium-browser
+      - curl
+      - git
+      - libvirt-daemon-system
+      - npm
+      - python3-libvirt
+      - qemu-kvm
+      - qemu-utils
+      - rpm
+env:
+  - TEST_OS=fedora-33
 script:
-  - npm install
-  - npm run build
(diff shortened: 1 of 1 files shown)


martinpitt:
Ugh, this works [on my repo](https://travis-ci.com/github/martinpitt/cockpit-podman/builds/193091166), but apparently we removed Travis from the official PRs a while ago? Presumably because it was failing all the time.

This is modelled after https://github.com/cockpit-project/starter-kit/pull/386 and an alternative to https://github.com/cockpit-project/cockpit-podman/pull/577 .

martinpitt:
I re-enabled Travis, this time through the modern GitHub Apps API instead of the old webhook one. So it now properly appears in "Checks" and such.

I also had an initial run for ubuntu-stable, but that requires a tad more work. I'll do this in a follow-up.

martinpitt:
[Success](https://github.com/cockpit-project/cockpit-podman/pull/578/checks?check_run_id=1313105529)!

I pushed another version to enable `$TEST_JOBS`, so that it runs faster.

marusak (approved):
cool, thanks!

Issue #481, opened 2020-04-27
Add cockpit-podman to Ubuntu


durzo:
The blog at https://cockpit-project.org/blog/cockpit-215.html says cockpit-docker has been deprecated in favor of cockpit-podman, but there are no packages available for download for either addon.

Furthermore, https://github.com/cockpit-project/cockpit-podman says that it "has not yet reached feature parity with cockpit-docker".

I can see the 217-2 packages in proposed available at https://launchpad.net/ubuntu/+source/cockpit/217-2 but there is also a lack of docker or podman.

Can you please advise which package we should use, and how to install it on Ubuntu 20.04 LTS ?

martinpitt:
I transfered this to the cockpit-podman project, as it's more appropriate to track here. podman is finally in [Debian](https://tracker.debian.org/pkg/libpod) and the current [Ubuntu development release](https://launchpad.net/ubuntu/+source/libpod), so the biggest blocker got resolved. There is now also Debian packaging files in our upstream, and I just sent an [intent to package](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=967030) to Debian.

Next steps:
 - [ ] Wait for an updated conmon in Debian: https://bugs.debian.org/cgi-bin/964858
 - [ ] Upload cockpit-podman to Debian, get it through NEW queue
 - [ ] Wait until it hits testing
 - [ ] Get it auto-imported into Ubuntu development release

I'm not sure how well libpod backports, it has quite a lot of build dependencies. So this will most probably not be a 20.04 LTS thing anytime soon, but at least be available in a near release -- it may still happen for 20.10 ("groovy") if all goes well.

tborychowski:
I'm currently on cockpit 215 and Ubuntu 20.04 says it's the latest version. There's also no podman.
Is there any way for me to get it sooner (both the latest version of Cockpit and podman)?

martinpitt:
@tborychowski : Newer cockpit versions are in the [official backports](https://help.ubuntu.com/community/UbuntuBackports), you can just install from there.

[...]

PR #586, opened 2020-11-02
Add testing on ubuntu-stable

martinpitt:
cockpit-podman is in Ubuntu now, so let's make sure it stays working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ addons:
       - rpm
 env:
   - TEST_OS=fedora-33
+  - TEST_OS=ubuntu-stable
 script:
   # HACK: /dev/kvm is root:kvm 0660 by default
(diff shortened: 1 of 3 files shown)


martinpitt:
I locally ran `TEST_OS=ubuntu-stable make check` and it succeeded. I'm happy to add that to our testmap as well, if you prefer that over running on Travis.

marusak:
Cool!!
Can we have `Fixes #481` ?
I'll wait for test results before approving.
I don't mind travis, but at least it should be in `__manual__`

martinpitt:
@marusak: Indeed, I closed #481 now. That happened independent from this PR. I sent https://github.com/cockpit-project/bots/pull/1353 to add it to testmap, _manual is a good idea.

marusak (approved):
nice, thanks!

PR #587, opened 2020-11-02
container.names does not exist - property should be capitalzed

KKoukiou:


marusak (changes requested):
Oh good catch! This is one of those annoying changes when we migrated to the REST API.

Can you please also fix it in the following places?
```
src/Containers.jsx:                    const error = cockpit.format(_("Failed to remove container $0"), this.state.containerWillDelete.names);
src/Containers.jsx:                    const error = cockpit.format(_("Failed to force remove container $0"), this.state.containerWillDelete.names);
```

marusak (approved):
Thanks!

PR #589, opened 2020-11-05
Makefile: Don't parse rpm name from `spec.in`

marusak:
This file is not shipped in tar file.

Oh wow, someone took picture of me while I was writing this patch:
i-have-no-idea-what-im-doing

seriously, WDYT @martinpitt, does it make sense? The point it that if someone tries to build it from tar file it reports:
```
/usr/bin/make install DESTDIR=/home/jnovy/rpmbuild/BUILDROOT/cockpit-podman-25-2.el8.x86_64 'INSTALL=/usr/bin/install -p'
error: Unable to open cockpit-podman.spec.in: No such file or directory
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ TEST_OS = fedora-32
 endif
 export TEST_OS
 TARFILE=cockpit-$(PACKAGE_NAME)-$(VERSION).tar.gz
-RPMFILE=$(shell rpmspec -D"VERSION $(VERSION)" -q cockpit-podman.spec.in).rpm
+RPMFILE=$(shell rpmspec -D"VERSION $(VERSION)" -q `ls cockpit-podman.spec.in cockpit-podman.spec 2>/dev/null | head -n1`).rpm
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 # stamp file to check if/when npm install ran
 NODE_MODULES_TEST=package-lock.json


Makefile
```diff
@@ -127,9 +126,7 @@ srpm: $(TARFILE) $(RPM_NAME).spec
 	  --define "_srcrpmdir `pwd`" \
 	  $(RPM_NAME).spec
 
-rpm: $(RPMFILE)
-
-$(RPMFILE): $(TARFILE) $(RPM_NAME).spec
+rpm: $(TARFILE) $(RPM_NAME).spec
```
martinpitt, line 129: I would really like to keep the current version where the phony `rpm` target is just convenience, and the actual file is used for the dependency. Otherwise we would be back in a situation where make can't check if dependencies are current, and thus `make check` would *always* rebuild the rpm and thus the VM even if nothing changed (i.e. iterating on tests just sucks).

Makefile
```diff
@@ -7,7 +7,6 @@ TEST_OS = fedora-32
 endif
 export TEST_OS
 TARFILE=cockpit-$(PACKAGE_NAME)-$(VERSION).tar.gz
-RPMFILE=$(shell rpmspec -D"VERSION $(VERSION)" -q cockpit-podman.spec.in).rpm
```
martinpitt, line 10: How about supporting both .spec.in and .spec here? something like

    ... -q `ls cockpit-podman.spec.in cockpit-podman.spec 2>/dev/null| head -n1`).rpm

?

martinpitt (changes requested):

Makefile
marusak: yeah, that was my other idea to just check which one is present. Found it a bit hacky to do that, but if we want to keep the current behaviour, then I can go with that. Will just have to check if rpmspec with specified version behaves the same for spec.in and spec.

marusak:
Updated

martinpitt (approved):
Dakujem!

Issue #504, opened 2020-08-10
Forced deletion modal can be closed while delete is in progress


lenticularis39:
The forced deletion modal (`ContainerRemoveErrorModal`) can be closed by the user using the Cancel button while the forced deletion is in progress. This creates counterintuitive behavior, because the operation is actually not cancelled and after clicking the container delete button again, the window reappears, showing the spinner the same as before it was closed.

Moreover when the user opens the modal for a different container while forced deletion is in progress (and the modal has been hidden using the Cancel button), it will display the name of the other container along with the spinner, although it's the first container, not the second one that is the one being deleted.

![2020-08-10-09-17-24](https://user-images.githubusercontent.com/1266171/89759589-ca551f00-daea-11ea-89c4-c97c7e329b85.gif)

lenticularis39:
This probably also affects checkpoint and restore modals, since they are implemented the same way.

marusak:
Right, so this is problem that if the action is still in progress.
I can think of 3 ways how we can solve this:
- Disable the 'Cancel' button if the operation is in progress
- 'Cancel' button would actually "cancel" this operation - which might be tricky, as I don't think you can cancel deletion in the middle.
- 'Cancel' button would not physically cancel the operation, it would just close the dialog and cleanup the state that there is something being removed. It would act more like 'Close'.

marusak:
Well, one more - keep list of containers that are being removed + if they are still in progress.

PR #590, opened 2020-11-09
Add container migration support

lenticularis39:
This requires a few changes in client.js, namely specifiying a host
where the operation is executed with each Podman command.

---

This patch introduces a graphical interface for [CRIU Podman container migration](https://criu.org/Podman#Container_Live_Migration), making use of cockpit-bridge sessions on the source and target hosts to transfer the data.

- Shows the state of the migration process using a label (see picture below)
- Supports transferring the image if not present on the target host
- Requires cockpit-project/cockpit#14641 to work and cockpit-project/cockpit#14849 to handle larger images/checkpoints (currently both open, first one likely close to merging)
- Uses raw channel API a lot, because the higher level cockpit.js API doesn't fit well with channel redirects
- Uses Cockpit session storage data to get the list of available host to migrate to
- I implemented the kebab menu using the PatternFly Dropdown component (as in the adjecent dropdown buttons) - the kebab icon's position looks a little bit off (see picture below)
- There are currently no tests - I'm working on them (a multi-machine setup is required for that, which is a little bit tricky)
- Export phases don't report errors correctly (error message is absent, because it gets saved into the checkpoint/image file) - this will probably be fixed with the tests, since it will be likely needed there

Screenshot_20201214_123611
Screenshot_20201214_123632
Screenshot_20201214_123653
Screenshot_20201214_123711
Screenshot_20201214_123738

### DIFF
--- a/src/ContainerMigrateModal.jsx
+++ b/src/ContainerMigrateModal.jsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import { Modal } from 'patternfly-react';
+import { Button, Checkbox, Select, SelectOption } from '@patternfly/react-core';
+import cockpit from 'cockpit';
+
+const _ = cockpit.gettext;
+const targetHostPlaceholder = _("Choose target host...");
+
+class ContainerMigrateModal extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            targetHost: targetHostPlaceholder,
+            targetHostOpen: false,
+            keep: false,
+            leaveRunning: false,
+            tcpEstablished: false,
+            ignoreRootFS: false
+        };
+        this.handleChange = this.handleChange.bind(this);
+    }
+
+    handleChange(checked, event) {
+        if (event.target.type === "checkbox")
+            this.setState({ [event.target.name]: event.target.checked });
+    }
+
+    render() {
+        return (
+            <Modal show={this.props.selectContainerMigrateModal}>
+                <Modal.Header>
+                    <Modal.Title>
+                        {_("Migrate running container to another host")}
+                    </Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    <div className="ct-form">
+                        <label className="control-label" htmlFor="migrate-dialog-container">
+                            {_("Container")}
+                        </label>
+                        <div id="migrate-dialog-container">{this.props.containerWillMigrate.Names}</div>
+
+                        <label className="control-label" htmlFor="migrate-dialog-target-host">
+                            {_("Target host")}
+                        </label>
+                        <Select id="migrate-dialog-target-host" isOpen={this.state.targetHostOpen}
+                                selections={this.state.targetHost}
+                                onToggle={targetHostOpen => this.setState({ targetHostOpen })}
+                                onSelect={(event, targetHost) => this.setState({ targetHost, targetHostOpen: false })}>
+                            {
+                                Object.keys(this.props.remoteHosts).map(host =>
+                                    <SelectOption value={this.props.remoteHosts[host]} key={host} />)
+                                        .concat([<SelectOption value={targetHostPlaceholder} key="_placeholder"
+                                                               isPlaceholder />])
+                            }
+                        </Select>
+
+                        <label className="control-label" htmlFor="migrate-dialog-keep">
+                            {_("Keep all temporary checkpoint files")}
+                        </label>
+                        <Checkbox id="migrate-dialog-keep" name="keep" isChecked={this.state.keep}
+                                  onChange={this.handleChange} />
+
+                        <label className="control-label" htmlFor="migrate-dialog-leaveRunning">
+                            {_("Leave running after migration")}
+                        </label>
+                        <Checkbox id="migrate-dialog-leaveRunning" name="leaveRunning"
+                                  isChecked={this.state.leaveRunning} onChange={this.handleChange} />
+
+                        <label className="control-label" htmlFor="migrate-dialog-tcpEstablished">
+                            {_("Preserve established TCP connections")}
+                        </label>
+                        <Checkbox id="migrate-dialog-tcpEstablished" name="tcpEstablished"
+                                  isChecked={this.state.tcpEstablished} onChange={this.handleChange} />
+
+                        <label className="control-label" htmlFor="migrate-dialog-ignoreRootFS">
+                            {_("Do not migrate root file-system changes")}
+                        </label>
+                        <Checkbox id="migrate-dialog-ignoreRootFS" name="ignoreRootFS"
+                                  isChecked={this.state.ignoreRootFS} onChange={this.handleChange} />
+
+                        <label className="control-label" htmlFor="migrate-dialog-ignoreStaticIP">
+                            {_("Ignore IP address if set statically")}
+                        </label>
+                        <Checkbox id="migrate-dialog-ignoreStaticIP" name="ignoreStaticIP"
+                                  isChecked={this.state.ignoreRootFS} onChange={this.handleChange} />
+
+                        <label className="control-label" htmlFor="migrate-dialog-ignoreStaticMAC">
+                            {_("Ignore MAC address if set statically")}
+                        </label>
+                        <Checkbox id="migrate-dialog-ignoreStaticMAC" name="ignoreStaticMAC"
+                                  isChecked={this.state.ignoreStaticMAC} onChange={this.handleChange} />
+                    </div>
+                </Modal.Body>
+                <Modal.Footer>
+                    <Button variant="primary" isDisabled={this.props.migrateInProgress ||
+                                                          this.state.targetHost === targetHostPlaceholder}
+                            onClick={() => this.props.handleMigrateContainer(this.state, this.state.targetHost)}>
+                        {_("Migrate")}
+                    </Button>
+                    <Button variant="link" onClick={this.props.handleMigrateContainerDeleteModal}>
+                        {_("Cancel")}
+                    </Button>
+                    <label className="migration-progress">
+                        {this.props.migrationStateLabel}
+                    </label>
+                    {this.props.migrateInProgress && <div className="spinner spinner-sm pull-right" />}
+                </Modal.Footer>
+            </Modal>
+        );
+    }
+}
+
+export default ContainerMigrateModal;
--- a/src/ContainerMigrateModal.jsx
+++ b/src/ContainerMigrateModal.jsx
@@ -1,23 +1,36 @@
 import React from 'react';
-import { Modal } from 'patternfly-react';
-import { Button, Checkbox, Select, SelectOption } from '@patternfly/react-core';
+import { Button, Checkbox, Modal, Select, SelectOption, TextInput } from '@patternfly/react-core';
 import cockpit from 'cockpit';
+import * as client from './client.js';
 
 const _ = cockpit.gettext;
-const targetHostPlaceholder = _("Choose target host...");
 
 class ContainerMigrateModal extends React.Component {
     constructor(props) {
         super(props);
+        this.targetHostPlaceholder = _("Choose target host...");
+        this.targetContainerPlaceholder = _("Create new container");
+        this.targetContainerLoading = _("Loading container list...");
+        this.targetContainerError = _("Error loading container list");
         this.state = {
-            targetHost: targetHostPlaceholder,
+            targetHost: this.targetHostPlaceholder,
             targetHostOpen: false,
+            targetContainer: this.targetContainerPlaceholder,
+            targetContainerList: [],
+            targetContainerOpen: false,
+            targetContainerLoading: false,
+            targetContainerError: false,
+            targetContainerName: "",
             keep: false,
             leaveRunning: false,
             tcpEstablished: false,
             ignoreRootFS: false
         };
         this.handleChange = this.handleChange.bind(this);
+        this.handleTargetHostSelect = this.handleTargetHostSelect.bind(this);
+        this.handleTargetContainerSelect = this.handleTargetContainerSelect.bind(this);
+        this.isTargetContainerDisabled = this.isTargetContainerDisabled.bind(this);
+        this.getContainerName = this.getContainerName.bind(this);
     }
 
     handleChange(checked, event) {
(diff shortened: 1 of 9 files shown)


src/client.js
```diff
@@ -13,133 +14,385 @@ export function getAddress(system) {
...
+                image: containerInfo.Image,
+                terminal: true,
+                command: containerInfo.Args
+            };
+            await createContainer(system, dummyConfig, targetHost);
+            containerCreated = true;
+        } else {
+            throw error;
+        }
+    }
```
marusak, line 266: So we need container with the same name to exist on the target host? What happens if it already exits? Can you please also add comment above this section what it does?

src/client.js
```diff
@@ -13,133 +14,385 @@ export function getAddress(system) {
...
+        ignoreRootFS: args.ignoreRootFS,
+        ignoreStaticIP: args.ignoreStaticIP,
+        ignoreStaticMAC: args.ignoreStaticMAC
+    };
+    await exportContainer(system, id, checkpointArgs, host, containerSaveChannel.id);
+    containerSaveChannel.control({ command: "done" });
+    await new Promise(resolve => containerSaveChannel.addEventListener("close", resolve));
+
+    // Restore it into the created dummy container
+    callbacks.containerImport();
```
marusak, line 291: Ok, I don't get this dummy container thing. According to https://criu.org/Podman#Container_Live_Migration you export into a file and then restore from a file. No need for any tmp containers.

src/util.js
```diff
@@ -175,3 +175,39 @@ export function compare_versions(a, b) {
 
     return a_ints.length - b_ints.length;
 }
+
+/*
+ * Gets a dictionary of remote hosts with hosts as keys and addresses as values.
+ */
+export function getRemoteHosts() {
```
marusak, line 182: this one is a bit meh. Don't take it wrongly, today I discussed this with @mvollmer and we realized there is no good way of doing this right now. We would love to have support for this in the future though. While this seems fine( I guess, haven't tested it and did not get in depth really) we have `pkg/lib/machines.js` in cockpit directly that we use in host switcher and dashboard to get list of hosts. It wraps this storage anyway. So maybe that would be a bit better approach to use that as when we change this lib we can easily sync it here as well rather than just try to rewrite this to work with whatever changes we do in Cockpit. We have some ideas and plans to change how we store and treat remote hosts, so there is a reason to be a bit worrier about it - especially since we don't call this "stable API".
(we copy things from pkg/lib here into lib/ so nothing new.If it would be impractical we can go with this for sure).

marusak (changes requested):
Super cool!
Setting to blocked on those cockpit issues and will try to move them forward.

> There are currently no tests - I'm working on them (a multi-machine setup is required for that, which is a little bit tricky)

If you want to know how we do multi host testing checkout cockpit tests. Just grep for `provision` in `test/verify`.

src/ContainerMigrateModal.jsx
```diff
@@ -0,0 +1,114 @@
+import React from 'react';
+import { Modal } from 'patternfly-react';
+import { Button, Checkbox, Select, SelectOption } from '@patternfly/react-core';
+import cockpit from 'cockpit';
+
+const _ = cockpit.gettext;
+const targetHostPlaceholder = _("Choose target host...");
```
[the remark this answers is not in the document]
lenticularis39: That makes sense, I'll move it to the class. (The primary purpose of the placeholder being a constant was to avoid a typo/mismatch breaking the comparison in the `Migrate` button.)

src/ContainerMigrateModal.jsx
```diff
@@ -0,0 +1,114 @@
+import React from 'react';
+import { Modal } from 'patternfly-react';
```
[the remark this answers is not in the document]
lenticularis39: Nice, I like the PF4 modals :)

src/client.js
```diff
@@ -13,133 +14,385 @@ export function getAddress(system) {
...
+        imageImport: () => {},
+        imageTagging: () => {},
+        containerInspect: () => {},
+        containerExport: () => {},
+        containerCreate: () => {},
+        containerImport: () => {}
+    }, callbacks);
+
+    // Make sure Podman system service is present and active on target host
+    const systemd = cockpit.dbus("org.freedesktop.systemd1", { bus: "system" });
```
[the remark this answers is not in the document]
lenticularis39: You're right - I somehow completely missed the mismatch between the comment and the options.

src/client.js
lenticularis39: That is true when using the Podman CLI, but with the REST API it's different - the container has to exist (see https://docs.podman.io/en/latest/_static/api.html#operation/libpodRestoreContainer and https://github.com/containers/podman/blob/master/pkg/api/handlers/libpod/containers.go).

src/client.js
lenticularis39: If the container exists the checkpoint is restored into it. If it doesn't exist, it is created first.

marusak: > If the container exists the checkpoint is restored into it. 

Does that change the container? I guess we should notify that such container exists on the host and we are fine with changing it to avoid unexpected results.

lenticularis39: > Does that change the container? I guess we should notify that such container exists on the host and we are fine with changing it to avoid unexpected results.

That depends on the migration settings. If "Do not migrate root file-system changes" is checked, it won't change the container's contents, only restore its state (if the container is running, it fails). I'm not sure about the behavior when the option is unchecked and there are conflicting filesystem changes between the containers.

As of the behavior of the migration UI - perhaps a better solution would be to let the user choose the target container using a select (only valid containers, i.e. ones that are not running and use the same image, would be shown), with an option for creating a new container.

src/util.js
lenticularis39: Just to make this clear: would that mean copying the entire `pkg/lib/machines.js` file into the cockpit-podman source, as with the cockpit-components-* files? That seems a little hacky to me, it has the advantage of being able to update it more easily when it changes in cockpit, though.

marusak: yes, thats exactly what I meant.

lenticularis39:
New changes:

- The UI was changed to let the user choose a container on the other host or a new container as the target (I'll also update the screenshots in the main post).
- Migration modal was migrated to PF4.
- Reporting of container export errors was fixed (the error message is now displayed).
- Constants using gettext in ContainerMigrateModal.jsx were moved inside the class.
- #504 was fixed for the migration modal.
- getRemoteHosts was rewritten to use the code in lib/machines.js, which was copied into cockpit-podman from cockpit
- max_read_size in the container/image import implementation was changed to fit the actual size of the file instead of being fixed.

I don't have the tests yet - I managed to provision multiple guests by replicating the multi-machine tests in cockpit, but unlike in cockpit I couldn't get guest-to-guest communication to work (IP addresses are set correctly, but neither ping nor the Cockpit remote host connections works), so I have to do some troubleshooting on this.

garrett:
Checkboxes shouldn't have empty labels.

Quick fix suggestion: Make an options group with all the checkboxes instead:


  |  
-- | --
Container | counter<br>_(ehhh? what is this?)_
  |  
Target container name | [___________________________]
Target host | [shadowman@cockpit-podman-2]
Target container | [Create new container ⏷]
  |  
Options | ☐ Keep temporary checkpoint files<br> ☐ Leave running after migration<br> ☐ Preserve established TCP connections<br> ☐ Do not migrate root file-system changes<br> ☐ Ignore statically-set IP address<br> ☐ Ignore statically-set MAC address 

(Ignore alignment; this is a side effect of GitHub tables.)

Notes:
- I don't know what container counter is or why it's here or why it's not editable
- I don't know why the container is an action to create a new container or why it's in a dropdown.
- Shouldn't the target container name default to the same as the original name?
- Are these option strings from some setting somewhere? Or are we writing ones fresh? I changed some with some simple suggestions above, but they'd benefit from more editing.
- I've grouped related items above.

We could also possibly redo the target fields with something like:


  |  
-- | --
Destination | Container name<br>[___________________]<br><br>Host<br>[shadowman@cockpit-podman-2]<br><br>Action<br>[Create new container ⏷]

lenticularis39:
> Checkboxes shouldn't have empty labels.
> 
> Quick fix suggestion: Make an options group with all the checkboxes instead:

I modeled this after the image run modal, which has two checkboxes without a label, but I agree this would be a better solution, especially considering there are multiple simple checkboxes here. I remember trying to do this when I was writing the migration modal code and having some problems with it - I'll have another look into this.

> Notes:
> 
> - I don't know what container counter is or why it's here or why it's not editable

It's not a container counter, it's the name of the selected container, which is called counter, similarly to how the image run modal displays the name of the selected image.

> 
> - I don't know why the container is an action to create a new container or why it's in a dropdown.
> 

As described above - the user either migrates the container state into an existing (stopped) container on the other host, or creates a new container. The dropdown is there to select the existing container on the other host or to select an option to create a new one (it loads the list of existing containers from the the other host).

> - Shouldn't the target container name default to the same as the original name?
> 

That would make sense, but it could lead to confusing behavior in case a container with such name already exists on the other host (the migration code doesn't differentiate between an existing container being selected in the dropdown and its name being entered via the "Create new container" option - in both cases the existing container is used), only if there is no container of the selected name a new one is created).

> - Are these option strings from some setting somewhere? Or are we writing ones fresh? I changed some with some simple suggestions above, but they'd benefit from more editing.
> 

They are based on the Podman REST API documentation for the checkpoint and restore operations, with slight changes. They also match the strings in the checkpoint and restore modals where possible. The expressions used in the Podman docs aren't the best, though, they would certainly benefit from editing.

> 
> 
> We could also possibly redo the target fields with something like:
>   	 
> Destination 	Container name
> [___________________]
> 
> Host
> [shadowman@cockpit-podman-2]
> 
> Action
> [Create new container ⏷]

That looks very nice. The "Container name" field could then alter between a text input and a select, based on the action selected (maybe a radio button could also be used since there are only two options).

garrett:
> the image run modal, which has two checkboxes without a label

Ah, then the run model also needs updating (outside of this PR). Thanks for pointing it out!

> It's not a container counter, it's the name of the selected container, which is called counter, similarly to how the image run modal displays the name of the selected image.

Aha! OK, that wasn't clear to me. I thought it was some "counter" value. (Perhaps I just didn't have enough coffee this Monday morning when looking over the PR. :coffee::wink:)

We might want to do source / destination to differentiate between the containers.

In this case, "Source container" and the name. Or, if source is problematic: perhaps "Original container" or "Origin container".

> That would make sense, but it could lead to confusing behavior in case a container with such name already exists on the other host

There should be a check to ensure the name isn't a duplicate. When opening the dialog, it could check the default and show an error that the name already exists, and that the user needs to select another. This is the same case if someone were to happen to type in an already-existing name, except that it's checking on what the computer suggests by default.

> That looks very nice. The "Container name" field could then alter between a text input and a select, based on the action selected (maybe a radio button could also be used since there are only two options).

Oh, then action should go first, before the other elements it affects.

And, yes, radio buttons are better when there are <=3 options. (Dropdowns should only be used when there are > 3 mutually exclusive options or the number of items are completely unpredictable (but generally still more than 3). That 3 number is fuzzy depending on context; it's generally between 3 to 5 items when one would consider a select/dropdown instead of a radio button list.)

We use a pattern like this in Cockpit already, where the topmost item has two inline radio buttons.

Something like:
**Action**  (x) Add  (_) Replace

garrett:
I've made some mockups, where I went back on the destination group, as I realized if we split the add/replace action into its own toplevel element, then that means there will only be a new container name on the destination host _or_ it will replace a container on the host. (At least if I'm understanding this correctly...)

### Standard migration, adding a new container on the destination

container-migrate

### Standard migration, adding a new container, but the name conflicts

container-migrate-conflict

### Replace another existing container

container-migrate-replace

---

How does this look?

garrett:
The combobox for the destination host would be a list of known hosts (perhaps currently used), with an option to add more by typing. It would use the PatternFly typeahead component with the isCreatable option and would override the `Create "..."` string using the `createText` property set to... "Select", I guess?

lenticularis39:
Thank you for doing the mockups, this is definitely a significant design improvement. There are only two things that should be addressed:

- The labels in the Action option do not correspond well to what the operation does. When an existing container is selected, its state is restored from a checkpoint generated from the source container's state rather than it being replaced as a whole ("replace" would imply e.g. that its configuration is discarded and replaced, which is not the case). Something like "Create new container" and "Restore into existing container" would better describe the actions.

-  Currently a non-typeahead select is used for the know host list - a combobox is certainly a good idea, because it allows searching, but creating a new entry doesn't make much sense. The list consists of hosts added via the remote host functionality in Cockpit. Because the container migration feature uses a Cockpit session to transfer the checkpoint, only hosts with an active Cockpit session can be used, therefore only those are shown and it doesn't make sense to add new ones (this should be done the usual way via the sidebar, handling authentication and so on).

I'll change the modal and once again update the screenshots once it is done.

lenticularis39:
UI has been updated. 

There are a few minor problems (the helper texts are black instead of red and for some reason the target container text input clears when closing the modal without the update event being called, leading to the validated attribute not being updated), but otherwise it should work (will be tested further in the Selenium tests once they are updated to the new UI).

garrett:
> Something like "Create new container" and "Restore into existing container" would better describe the actions.

OK, sure.

"Restore into existing container" does sound a little weird to me. Can you select an unrelated container to migrate into?

> The list consists of hosts added via the remote host functionality in Cockpit.

OH! I wasn't aware! Neat! I guess this is one of the first places in the UI where we're re-using the multi-host configuration, aside from the dashboard (which we've just deprecated). We should probably think how we can improve the flow of adding new hosts. 

...But for now:

We need to message how to adjust this list in the UI. I guess adding a new host from the autocomplete might be a little complex, as it'd force a dialog on top of the other dialog and it'd have to be done by calling an API that might not be cross-page capable at the moment?

Perhaps something like this:

> 🛈 Add destination hosts using [the host selector](#).

(With an actual icon, not UTF-8, of course)

It would be _really cool_ if you could click the link on that and have the host switcher show itself and somehow indicate the "add new host" button with something like a short pulse animation. However, I don't think we can do this at the moment. 

But we _can_, at least, have the text (without the link part). It's better than nothing. Anything more than just text here should be done in a follow-up (if at all).

Also, what happens when:
1. there's only the current host (nothing to migrate to)?
   - would we have an empty state that says something about adding additional hosts as destinations using the host selector?
2. when there's only one remote destination host to migrate to? (perhaps a dropdown, but it's disabled but has the host preselcted?)

lenticularis39:
> 
> "Restore into existing container" does sound a little weird to me. Can you select an unrelated container to migrate into?
> 

You can, but the migration would fail - the target container has to have the same image as the source container. The UI can of course filter the container list for that, but there might be some corner cases where the image is virtually the same, but has a different ID (e.g. it has been imported from a different image format). This was the reason why I didn't want to make the image IDs matching a hard requirement for the container to be possible to select. The actual best solution would depend on whether the migration succeeds in the aforementioned corner cases - I'll test this.

> We need to message how to adjust this list in the UI. I guess adding a new host from the autocomplete might be a little complex, as it'd force a dialog on top of the other dialog and it'd have to be done by calling an API that might not be cross-page capable at the moment?
> 
> Perhaps something like this:
> 
> > 🛈 Add destination hosts using [the host selector](#).
> 
> (With an actual icon, not UTF-8, of course)
> 
> It would be _really cool_ if you could click the link on that and have the host switcher show itself and somehow indicate the "add new host" button with something like a short pulse animation. However, I don't think we can do this at the moment.
> 
> But we _can_, at least, have the text (without the link part). It's better than nothing. Anything more than just text here should be done in a follow-up (if at all).
> 

I'll have a look on how the code works and what can be done with that.

> Also, what happens when:
> 
>     1. there's only the current host (nothing to migrate to)?
>        
>        * would we have an empty state that says something about adding additional hosts as destinations using the host selector?
> 
>     2. when there's only one remote destination host to migrate to? (perhaps a dropdown, but it's disabled but has the host preselcted?)

Nothing special happens. In the first case the target host select is empty (except for the placeholder selection):

image

In the second case the dropdown simply shows the placeholder and the one remote host:

image

I'm not sure if it's better to keep the placeholder option or to remove it and set the placeholder text (as it is with the target container:)

image

Perhaps the placeholder option could be removed, along with preselecting the one remote host in the case where there is only one (and also doing the same with the target container).

lenticularis39:
- The action checkbox was changed to inline.
- Validated error labels are now red.
- Problem with target container name text field input was fixed.
- Both the target host and target container selects pre-select the only available option when there's only one.
- Target host select placeholder was changed from a placeholder item to placeholder text to match the target container select.
- A bug in client.js migration code that caused a read channel done message not to be detected was fixed.
- The branch was rebased to the current master.
- The dialog was migrated to PatternFly React forms to match the other dialogs.

About the unrelated containers in the select question - the problematic case I encountered before doesn't seem to happen anymore, so I couldn't test it. However I don't like the idea of the user not being able to select the desired container in case an unexpected image hash mismatch happens, so I kept the select as is.

Also jquery was added to package.json, because it it used by machines.js (a slightly modfified source file from Cockpit).

lenticularis39:
Closing, since it depends on the closed cockpit-project/cockpit#14641, which I won't be working on anymore.

PR #591, opened 2020-11-10
Drop PF3 dependency

KKoukiou:


KKoukiou:
@garrett can you please review this one?

KKoukiou:
@marusak I am currently waiting for the tests - but this is review ready. It might need some test update, but I would like your review here as well.

marusak:
Codewise looks fine. Also happy to see that we are PF4!! :)

A few things:
1. Commits could be squashed a bit, like the test one or dropping PF3 dep (squash to commit which was the last user or it)

2. There is a lot of space between inputs: (like after "in" or after "Tag")
(this PR)
medzerysematam
(master)
lessspacy

3. Run image and the volume add button is unhappy: (should be on the same line)
(this PR)
breakybreaky
(master)
sameline

4. 'With terminal' label moved to the left. Feels a bit inconsistent. See screenshots from point 3.


5. Image searching misses any spinner: (Empty state with big spinner in the middle on the top would be better, but that can be separate PR, here at least we should still show it).
(this PR)
no-search
(master)
dasspiner

6. The play button looks weird:
(this PR)
playsisate
(master)
buttonsbetter

marusak (changes requested):
See my other comment

KKoukiou:
@marusak reposted with fixes.
The label - fields spacing is different in PF4 forms than what garrett had implemented for us using ct-form. For this reason, differences in this kind of spacing are expected. Exception in this rule is the Search - in combination, PF4 does not offer support for such things, as two labels on the same row, I wrote some CSS to support this scenario.

The play buttons (all icon buttons) indeed look again different, I would not say broken. I did not add any styling to them, so that's how PF4 icon buttons should look like I guess.

src/ImageSearchModal.jsx
```diff
@@ -170,48 +157,58 @@ export class ImageSearchModal extends React.Component {
...
                             </Select.Select>
-                        </div>
-                    </div>
-                </InputGroup>
+                        </FormGroup>
+                    </Flex>
+                </Form>
 
-                {this.state.searchInProgress && <div id='search-image-dialog-waiting' className='spinner' />}
+                {<Bullseye><Spinner id='search-image-dialog-waiting' /></Bullseye>}
```
marusak, line 164: There is spinner when I open "get new image". The spinner should be there only when it is searching for something. It is also there when  I get some results or even when there are no results.

It needs that `this.state.searchInProgress` check

marusak (changes requested):
Almost there :)

src/ImageSearchModal.jsx
KKoukiou: Done

marusak (approved):
Nice, thanks!

PR #592, opened 2020-11-10
Disable all ways of closing dialogs while an operation is in progress

### LINKED ISSUE #504: Forced deletion modal can be closed while delete is in progress

KKoukiou:
Fixes #504

Note: ideally we should be able to cancel the operations but podman does
not offer that, so let's just be safe with not allowing the users to
close the dialogs.

### DIFF
--- a/src/ContainerCheckpointModal.jsx
+++ b/src/ContainerCheckpointModal.jsx
@@ -24,8 +24,8 @@ class ContainerCheckpointModal extends React.Component {
     render() {
         return (
             <Modal isOpen={this.props.selectContainerCheckpointModal}
+                   showClose={false}
                    position="top" variant="medium"
-                   onClose={this.props.handleCheckpointContainerDeleteModal}
                    title={cockpit.format(_("Checkpoint container $0"), this.props.containerWillCheckpoint.Names)}
                    footer={<>
                        <Button variant="primary" isDisabled={this.props.checkpointInProgress}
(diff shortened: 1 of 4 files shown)


src/ContainerCheckpointModal.jsx
```diff
@@ -24,6 +24,8 @@ class ContainerCheckpointModal extends React.Component {
     render() {
         return (
             <Modal isOpen={this.props.selectContainerCheckpointModal}
+                   onEscapePress={() => undefined}
+                   showClose={false}
                    position="top" variant="medium"
                    onClose={this.props.handleCheckpointContainerDeleteModal}
```
marusak, line 30: We don't need to define `onClose` then anymore. And if you drop that, then there is no need to define `onEscapePress`.

I guess we cannot disable the X button as well, if we could, that would be the best approach. But otherwise not showing the X button works as well.

marusak (changes requested):

src/ContainerCheckpointModal.jsx
KKoukiou: We can't - but I will open an issue to PF4 and maybe I can implement it the following week.

marusak: That would be great, thanks! :)

KKoukiou: https://github.com/patternfly/patternfly-design/issues/952

marusak (approved):
Thank you :)

PR #593, opened 2020-11-11
Limited access dialog has been updated in cockpit in release 231

KKoukiou:


skobyda (approved):
Since it's just temporary, I'm okay with that. Let's just not forget to remove it when it's no longer needed.

KKoukiou:
> Since it's just temporary, I'm okay with that. Let's just not forget to remove it when it's no longer needed.

I will send already the revert PR.

PR #594, opened 2020-11-11
Remove test workaround for the limited access modals

KKoukiou:
…ase 231"

This reverts commit 75a9eb08dd3770782529a08f801609e2df907754.

It was meant to be a temporary workaround anyway.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -182,12 +182,8 @@ class TestApplication(testlib.MachineCase):
         # Drop privileges
         b.switch_to_top()
         b.click("#super-user-indicator button")
-        try:
-            b.click(".pf-c-modal-box:contains('Switch to limited access') button:contains('Limit access')")
-            b.wait_not_present(".pf-c-modal-box:contains('Switch to limited access')")
-        except testlib.Error:
-            b.click(".modal-dialog:contains('Switch to limited access') button:contains('Limit access')")
-            b.wait_not_present(".modal-dialog:contains('Switch to limited access')")
+        b.click(".pf-c-modal-box:contains('Switch to limited access') button:contains('Limit access')")
+        b.wait_not_present(".pf-c-modal-box:contains('Switch to limited access')")
         b.wait_text("#super-user-indicator", "Limited access")
         b.switch_to_frame("cockpit1:localhost/podman")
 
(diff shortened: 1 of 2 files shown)


marusak (dismissed):
Thanks! I am of course fine with this as long as all tests are green :)

marusak:
ubuntu-stable in travis is the one place where this one has not yet been fixed. I wonder when it will get there...

KKoukiou:
@marusak Do you know what's wrong with this one? All tests passed by travis failed by showing that tests failed. I even retried it but did not change anything.

marusak:
The image for ubuntu-stable that we use in travis does not yet have adequate cockpit version.

martinpitt (approved):
Looks good, and I'm fine with not blocking on ubuntu-stable for now. Now that the new development series is open, I can do backports again, I'll do that now. And then I'll send a PR to adjust travis to pull from backports.

martinpitt:
I [backported 231](https://launchpad.net/ubuntu/+source/cockpit) to 20.10 and 20.04. Will take a few hours to appear on the mirrors, then we can pull it in. After that it actually ought to be enough to bump the cockpit-{bridge,system} dependencies in packaging/debian/control, everything else should then Just Work™. I'll test this tomorrow or Monday.

martinpitt (changes requested):
Needs a packaging fix to bump cockpit dependency

martinpitt:
It's already on the mirrors (I forgot how fast this is in Ubuntu). It still did need a little vm.install tweak, as apt doesn't use backports automatically to satisfy versioned dependencies.

martinpitt:
Meh, tests on e2e fail due to the docker pull rate limit. But at least Travis is happy :-)

So it seems we need to fix/push PR #602 ASAP

src/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
     "version": "0.1",
     "requires": {
-        "cockpit": "137"
+        "cockpit": "231"
```
marusak, line 4: strictly speaking, it is not really required. The page would work just fine with 230 for example. Even though I don't really think 137 is adequate version, but this one seem rather strict for no obvious reason.

packaging/debian/control
```diff
@@ -10,8 +10,8 @@ Package: cockpit-podman
 Architecture: all
 Multi-Arch: foreign
 Depends: ${misc:Depends},
-         cockpit-bridge (>= 138),
-         cockpit-system (>= 138),
+         cockpit-bridge (>= 231),
```
marusak, line 13: The same reason as for manifest.

src/manifest.json
martinpitt: Hmm, good point, this really was for the tests only. That was a thinko then. Reverted.

martinpitt:
Argh, commit message was out of sync, updated.

marusak (approved):

PR #602, opened 2020-11-19
Replace docker.io with quay.io

marusak:
Docker now has strict limits for how much we can pull from it. Lets use
something different.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -172,8 +172,8 @@ class TestApplication(testlib.MachineCase):
         self.execute(False, "podman pod create --infra=false --name pod_user")
         self.execute(True, "podman pod create --infra=false --name pod_system")
 
-        checkImage(b, "docker.io/library/registry:2", "system")
-        checkImage(b, "docker.io/library/registry:2", "admin")
+        checkImage(b, "quay.io/cockpit/registry:2", "system")
+        checkImage(b, "quay.io/cockpit/registry:2", "admin")
         b.wait_present("#containers-containers .pod-name:contains('pod_user')")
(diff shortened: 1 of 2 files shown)


marusak:
oh common, I though those images were just the common one, but seems neither of them exists on `registry.fedoraproject.org` nor `registry.access.redhat.com`.

martinpitt:
Indeed registry.fedoraproject isn't very good/popular for general upstreams. How about these:

 - https://quay.io/repository/libpod/busybox?tab=info (`podman run -it --rm quay.io/libpod/busybox`)
 - https://quay.io/repository/libpod/alpine?tab=tags (`podman run -it --rm quay.io/libpod/alpine`

These seem quite reasonable to me. I didn't find registry:2 on quay, though. GitHub also has a registry now (ghcr.io), but there is no browser yet, or I didn't find it.

IMHO it would be better to pull these images during image-create, then we need to pull them less often, and at least we run into registry problems during image rebuild (where they hurt less than in project PRs).

martinpitt:
I created a cockpit org on quay and pushed the registry:2 image [to quay.io](https://quay.io/repository/cockpit/registry?tab=info). Let's just use it from there. @marusak , if you create an account on quay.io, I'm happy to add you to the org as admin.

martinpitt:
Lazy Saturday evening "let the bots test it for me" force-push.

martinpitt (approved):
That works. Formally approving, but @marusak please cross-check and land.

marusak:
Amazing, thanks @martinpitt !

PR #607, opened 2020-11-27
Empty states improvements

marusak:
(most likely breaks tests so I'll let it run, then fix tests and repush also with closes)

Before: (Searching)
Screenshot from 2020-11-27 11-07-55
After:
Screenshot from 2020-11-27 11-31-53

Before: (Not found)
notfoundbefore
After:
notfoundnew

Before: (when you open the dialog)
currentsearch
After:
emptynew

Before: (terminal/logs)
notrunning
After:
newnotrunning

### DIFF
--- a/src/ContainerTerminal.jsx
+++ b/src/ContainerTerminal.jsx
@@ -89,7 +89,11 @@ class ContainerTerminal extends React.Component {
     }
 
     componentDidUpdate(prevProps, prevState) {
-        if (!this.state.channel && this.props.containerStatus === "running" && prevProps.containerStatus !== "running")
+        // Connect channel when there is none and either container started or tty was resolved
+        // tty is being read as part of containerDetails and it is asynchronous so we need to wait for it
+        if (!this.state.channel && (
+            (this.props.containerStatus === "running" && prevProps.containerStatus !== "running") ||
+            (this.props.tty !== undefined && prevProps.tty === undefined)))
             this.connectChannel();
         if (prevProps.width !== this.props.width) {
             this.resize(this.props.width);
@@ -111,6 +115,9 @@ class ContainerTerminal extends React.Component {
             return;
 
         if (this.props.containerStatus !== "running")
+            return;
+
+        if (this.props.tty === undefined)
             return;
 
         if (this.props.tty)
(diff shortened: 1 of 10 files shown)


src/ImageSearchModal.jsx
```diff
@@ -161,10 +162,12 @@ export class ImageSearchModal extends React.Component {
...
 
-                {this.state.searchInProgress && <Bullseye><Spinner id='search-image-dialog-waiting' /></Bullseye>}
+                {this.state.searchInProgress && <EmptyStatePanel loading title={_("Searching...")} /> }
 
-                {this.state.searchFinished && !this.state.imageIdentifier == '' && <>
-                    {this.state.imageList.length == 0 && <div className="no-results"> {cockpit.format(_("No results for $0. Please retry another term."), this.state.imageIdentifier)} </div>}
+                {this.state.imageIdentifier == "" && <EmptyStatePanel title={_("No images found")} paragraph={_("Start typing to look for images.")} /> }
+
+                {this.state.searchFinished && this.state.imageIdentifier !== '' && <>
+                    {this.state.imageList.length == 0 && <EmptyStatePanel icon={ExclamationCircleIcon} title={cockpit.format(_("No results for $0. Please retry another term."), this.state.imageIdentifier)} />}
```
KKoukiou, line 170: Please add the Please ... part in the paragraph like you have for the hint above. Let's put the hints in the paragraph section always.

src/ContainerTerminal.jsx
```diff
@@ -246,18 +248,9 @@ class ContainerTerminal extends React.Component {
...
-                </span>
-            </div>
-            );
-
-        return (
-            <>
-                { element }
-            </>
-        );
+            element = <EmptyStatePanel icon={ExclamationCircleIcon} title={this.state.errorMessage} />;
```
KKoukiou, line 251: Also I would say that this does not need an ExclamationCircleIcon since it's not an error here, I would just say something like VM page does.
Screen Shot 2020-11-27 at 16 54 57

KKoukiou (commented):
Two minor issues - other wise good!

src/ContainerTerminal.jsx
marusak: noptrunning20

src/ImageSearchModal.jsx
marusak: noreultsplease
pleaaseprettyplewase

marusak:
Two more commits :)

KKoukiou (approved):

PR #608, opened 2020-11-27
run: Make hostPort optional

### LINKED ISSUE #365: Implement omiting hostPort when defining ports
If I do not configure the needed ports for the container, there is no random generation of the ports like K8s or docker will do. If I configure no port the container will be isolated which is a bit annoying. If I test something I do not care about the ports and I am not willing to check what is already in use or free and to enter something. It should be done for me, I can later check what port the system  assigned to the container.

marusak:
Follow what CLI allows:
```
--publish, -p=ip:hostPort:containerPort | ip::containerPort | hostPort:containerPort | containerPort
```

Does not yet supports IP addresses. (so just the last two cases)
Also changing order as it is like that everywhere including in our UI
where we show `IP:hostPort->ContainerPort/protocol`

Fixes #365

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -32,6 +32,13 @@ const PublishPort = ({ id, item, onChange, idx, removeitem, additem }) =>
     (
         <>
             <InputGroup className='ct-input-group-spacer-sm'>
+                <TextInput type='number'
+                           step={1}
+                           min={1}
+                           max={65535}
+                           placeholder={_("Host port (optional)")}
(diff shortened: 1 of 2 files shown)


KKoukiou (approved):
The PR is correct. Reading the docs I realized that this support range as well, I guess we could extend the functionality in the future.

In addition I would like to see a more clear info message, what would happen it no host port is speficied.
>>> If it is not, the container port will be randomly assigned a port on the host.

but this PR can go in nicely as is.

PR #610, opened 2020-11-30
run: Enable setting up IP address for exposed ports

### LINKED ISSUE #412: Unable to put IP:Port Mappings in the UI
UI doesn't let me `podman run -d -p 192.168.1.2:80:80 nginx` do this.

marusak:
Fixes #412

For release notes:

Enable setting up IP address when exposing container ports

If host IP is set to 0.0.0.0 or not set at all, the port will be bound on all IPs on the host.
Screen Shot 2020-11-30 at 11 12 10

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -32,6 +32,11 @@ const PublishPort = ({ id, item, onChange, idx, removeitem, additem }) =>
     (
         <>
             <InputGroup className='ct-input-group-spacer-sm' id={id}>
+                <TextInput aria-label={_("IP (optional)")}
+                           type='text'
+                           placeholder={_("IP (optional)")}
+                           value={item.IP || ''}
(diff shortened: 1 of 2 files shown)


src/ImageRunModal.jsx
```diff
@@ -32,6 +32,11 @@ const PublishPort = ({ id, item, onChange, idx, removeitem, additem }) =>
     (
         <>
             <InputGroup className='ct-input-group-spacer-sm' id={id}>
+                <TextInput aria-label={_("IP (optional)")}
+                           type='text'
+                           placeholder={_("IP (optional)")}
+                           value={item.IP || ''}
+                           onChange={value => onChange(idx, 'IP', value)} />
```
KKoukiou, line 39: We could also make IP and TCP/UDP is one line and the Port mapping in next, so that they don't squeeze so much. but design improvements come come later on of course.

KKoukiou (approved):

PR #612, opened 2020-12-02
workflows: Update release container registry

martinpitt:
With [1] the release container moved from dockerhub to GitHub's
container registry, as we both build and use it from GitHub. This avoids
running into docker.io pull limits.

https://github.com/cockpit-project/cockpituous/pull/353


martinpitt:
Ah, required-statuses.. pushing with Closes# and without no-test.

marusak (approved):

PR #616, opened 2020-12-02
Move #app into body

### LINKED ISSUE #615: webkit: Page does not extend to fill the content area
This seems to be a WebKit only issue; the page does not fully extend to fill the content area:

Screenshot from 2020-12-02 10-55-19

Firefox and Chromium are both doing the right thing, so this is lower priority, and my guess is that it's probably something subtle in the CSS.

garrett:
Fixes #615

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -29,7 +29,5 @@ along with this package; If not, see <http://www.gnu.org/licenses/>.
     <script type="text/javascript" src="index.js"></script>
 </head>
 
-<body class="pf-m-redhat-font">
-    <div id="app"></div>
-</body>
+<body class="pf-m-redhat-font" id="app"></body>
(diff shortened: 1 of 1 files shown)


marusak:
I know we are doing this in cockpit, but it is discouraged, even some browsers complain about it: https://github.com/cockpit-project/cockpit/issues/14871

KKoukiou:
> I know we are doing this in cockpit, but it is discouraged, even some browsers complain about it: [cockpit-project/cockpit#14871](https://github.com/cockpit-project/cockpit/issues/14871)

yes I agree, I did this in many places in cockpit but I would rather that we figure out if PF really wants us to do this before extending it to more projects. IOW Let's fix it properly here.

PR #617, opened 2020-12-03
tests: Make it pass on rhel-8-3 and rhel-8-4

mvollmer:
rhel-8-4:

- [x] Shell dialogs use old-style classes
- [x] synchronous reboot causes self.execute to fail.
- [x] cgroupv1 vs cgroupv2 stuff
- [x] Rootless podman doesn't work on rhel-8-4: https://bugzilla.redhat.com/show_bug.cgi?id=1903983

rhel-8-3:

- [x] Shell dialogs use old-style classes
- [x] synchronous reboot causes self.execute to fail.
- [x] cgroupv1 vs cgroupv2 stuff
- [x] TestApplication.testCheckpointRestoreSystemCrun can unexpectedly make snapshots

### DIFF
--- a/src/Notification.jsx
+++ b/src/Notification.jsx
@@ -23,7 +23,19 @@ import cockpit from 'cockpit';
 
 const _ = cockpit.gettext;
 
+let last_error = "";
+
+function log_error_if_changed(error) {
+    // Put the error in the browser log, for easier debugging and
+    // matching of known issues in the integration tests.
+    if (error != last_error) {
+        last_error = error;
+        console.error(error);
+    }
+}
+
 export const ErrorNotification = ({ errorMessage, errorDetail, onDismiss }) => {
+    log_error_if_changed(errorMessage + (errorDetail ? ": " + errorDetail : ""));
     return (
         <Alert isInline variant='danger' title={errorMessage}
             actionClose={onDismiss ? <AlertActionCloseButton onClose={onDismiss} /> : null}>
--- a/test/check-application
+++ b/test/check-application
@@ -69,7 +69,7 @@ class TestApplication(testlib.MachineCase):
         # But disable it globally so that "systemctl --user disable" does what we expect
         self.machine.execute("systemctl --global disable podman.socket")
 
-        self.addCleanup(self.machine.execute, "killall -9 podman")
+        self.addCleanup(self.machine.execute, "killall -9 podman || true")
 
         self.allow_journal_messages("/run.*/podman/podman: couldn't connect.*")
         self.allow_journal_messages(".*/run.*/podman/podman.*Connection reset by peer")
@@ -938,9 +938,13 @@ class TestApplication(testlib.MachineCase):
 
     @testlib.nondestructive
     def testCheckpointRestoreSystemCrun(self):
-        # With crun (CgroupsV2) checkpoint doesn't work yet
         b = self.browser
         m = self.machine
+
+        if not self.cgroupsV2() and self.has_criu:
+            return
+
+        # With crun (CgroupsV2) or without criu checkpoints don't work yet
 
         self.login_and_go("/podman", superuser=True)
         b.wait_present("#app")
--- a/test/check-application
+++ b/test/check-application
@@ -78,7 +78,7 @@ class TestApplication(testlib.MachineCase):
         self.has_selinux = self.machine.image not in ["debian-testing", "ubuntu-stable"]
 
     def cgroupsV2(self):
-        return self.machine.image not in ["debian-testing", "ubuntu-stable"]
+        return self.machine.image not in ["debian-testing", "ubuntu-stable", "rhel-8-3", "rhel-8-4"]
 
     def execute(self, system, cmd):
         if system:
--- a/test/check-application
+++ b/test/check-application
@@ -179,11 +179,16 @@ class TestApplication(testlib.MachineCase):
         b.wait_present("#containers-containers .container-name:contains('a')")
         b.wait_present("#containers-containers .container-name:contains('b')")
 
+        if self.system_before(231): # Changed in #14740
+            modal_sel = ".modal-dialog"
+        else:
+            modal_sel = ".pf-c-modal-box"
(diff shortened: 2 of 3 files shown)


mvollmer:
Rootless podman doesn't work on rhel-8-4: https://bugzilla.redhat.com/show_bug.cgi?id=1903983

test/check-application
```diff
@@ -179,11 +179,16 @@ class TestApplication(testlib.MachineCase):
         b.wait_present("#containers-containers .container-name:contains('a')")
         b.wait_present("#containers-containers .container-name:contains('b')")
 
+        if self.machine.image in [ "rhel-8-3" ]:
```
marusak, line 182: Would be so nice, if we could use this one: https://github.com/cockpit-project/cockpit/commit/58d1d00d3dcce89ddcdd0a1f5f585f43e7382d0e :) (needs a new `test/common`)

README.md
```diff
@@ -1,3 +1,5 @@
+Hi!
```
marusak, line 1: Hey :)

test/check-application
mvollmer: Yes, we should probably add some convenience functions for gaining/dropping privs to testlib and hide all the complexity there.

README.md
mvollmer: :-) we can drop this now...

marusak:
Almost :) https://logs.cockpit-project.org/logs/pull-617-20201203-200210-746ad9cf-rhel-8-4-bots-1437/TestApplication-testRunImageUser-rhel-8-4-127.0.0.2-2301-FAIL.png
We might need to do just `skipImage` or so also in tests directly. There does not seem to be anything in the log, that naughty could safely use, so I would say the naughty is good, it can land and we need to also skip this one test with reference to the naughty.

mvollmer:
> There does not seem to be anything in the log, that naughty could safely use

In cockpit itself, we usually log dialog errors also with console.log, see https://github.com/cockpit-project/cockpit/blob/master/pkg/lib/cockpit-components-dialog.jsx#L102 for example. That makes them matchable with a naughty pattern (I think).

We can do that in cockpit-podman as well, I'd say.

marusak:
> We can do that in cockpit-podman as well, I'd say.

Yeah. But honestly, I am not a huge fan of introducing code changes just to make tests to work. Since this test is just busted, we can skip it or add there check for version of podman on 8-4 and if less that where it should be fixed, then `return`.

martinpitt:
> I am not a huge fan of introducing code changes just to make tests to work.

No, but IMHO it *is* useful for debugging tests in general to always log errors to the console as well, not just putting it into the dialog.

mvollmer:
> But honestly, I am not a huge fan of introducing code changes just to make tests to work.

Well, testing is important and we should write testable code.

test/check-application
martinpitt: Note that 8.4 also has the old cockpit still; there is some trouble with the erratum creation.

mvollmer: The tests don't seem to agree: they pass without special-casing rhel-8-4.  Does that make sense?

mvollmer: Heh, did they maybe all get skipped before getting to this point?

mvollmer: So, yeah, let's just use the approach in https://github.com/cockpit-project/cockpit/pull/14981.  Unfortunately, it's not released yet... should we just get test/common from 58d1d00?

mvollmer: > should we just get test/common from 58d1d00?

That doesn't work because git doesn't allow fetching by SHA.. silly me.

mvollmer: Ok, done anyway!

test/check-application
```diff
@@ -179,11 +179,16 @@ class TestApplication(testlib.MachineCase):
         b.wait_present("#containers-containers .container-name:contains('a')")
         b.wait_present("#containers-containers .container-name:contains('b')")
 
+        if self.before_system(231): # Changed in #14740
```
mvollmer, line 182: What the flip?!?  I would have sworn I have tested this locally...

src/Notification.jsx
```diff
@@ -23,7 +23,19 @@ import cockpit from 'cockpit';
...
 const _ = cockpit.gettext;
 
+let last_error = "";
+
+function log_error_if_new(error) {
+    // Put the error in the browser log, for easier debugging and
+    // matching of known issues in the integration tests.
+    if (error != last_error) {
+        last_error = error;
+        console.log("ERROR: ", error);
```
marusak, line 33: what about `console.error`?

src/Notification.jsx
```diff
@@ -23,7 +23,19 @@ import cockpit from 'cockpit';
 
 const _ = cockpit.gettext;
 
+let last_error = "";
```
marusak, line 26: This could be array to really just log `if new`.

marusak (approved):
Ideas for improvement, but otherwise a nice job, thanks!

src/Notification.jsx
mvollmer: Good point... cockpit-components-dialog uses `console.warn`... let's use console.error.

src/Notification.jsx
mvollmer: Yes, but I think we really want `log_error_if_changed`, no?  We only want to filter out repeated output from re-rendering the currently open dialog.  I rename the function.

marusak (approved):
Thanks!
(Misses Closes line though but I recall your last week's suggestion ;) )

martinpitt (commented):
Argh, I reviewed this 5 days ago, and somehow it didn't actually get through. :sob: But not that important, we'll release today and thus can bump it back to the release soon.

mvollmer:
Let's wait for 234, it'll happen any minute now.

mvollmer:
> Let's wait for 234, it'll happen any minute now.

234 tag is available, let's use it.

martinpitt (approved):
Yay!

PR #618, opened 2020-12-06
Drop sass-resources-loader

martinpitt:
Explicitly include the required PF variables definitions into components
instead. This lessens the already gross duplication in the built
index.css (1200 → 950 kB in debug mode). It also makes separate
compilation and reuse easier, and paves the way for using sassc (which
doesn't support global resources).

Also drop the `~` prefix from cockpit-components-table.scss imports.
It's not necessary for sass-loader, and does not work with sassc.

 - [x] Do this in cockpit first, where these files got copied from: https://github.com/cockpit-project/cockpit/pull/15042

### DIFF
--- a/lib/cockpit-components-table.scss
+++ b/lib/cockpit-components-table.scss
@@ -1,5 +1,6 @@
-@import "~@patternfly/patternfly/components/Table/table.scss";
-@import "~@patternfly/patternfly/components/Table/table-grid.scss";
+@import "@patternfly/patternfly/base/patternfly-variables.scss";
+@import "@patternfly/patternfly/components/Table/table.scss";
+@import "@patternfly/patternfly/components/Table/table-grid.scss";
 
 .ct-table {
     // Reverse out the default padding for table bodies,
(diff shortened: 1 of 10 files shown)


martinpitt:
@KKoukiou : This is not ready to land yet as we shouldn't divert these files from cockpit. But I'd like your opinion whether this is the right way to go, or whether I am missing something about the global resources loading? Visually I see no difference between both builds.

lib/cockpit-components-table.scss
```diff
@@ -1,5 +1,6 @@
-@import "~@patternfly/patternfly/components/Table/table.scss";
-@import "~@patternfly/patternfly/components/Table/table-grid.scss";
+@import "@patternfly/patternfly/base/patternfly-variables.scss";
```
KKoukiou, line 1: Hmm, I really thought it's a good idea to have patternfly-variables everywhere accessible. (no in podman but in cockpit) almost all our CSS files use PF4 variables. I am not sure if it makes sense to include in every single files the PF variables. We currently import the variables with sass-resource-loader but maybe I can find another solution.

KKoukiou: For example in this project here, you missed src/podman.scss, it also uses PF4 variables.

martinpitt: Hmm, sass-loader (and sassc) complains loudly and bitterly if it does not know about a variable, so it seemed fairly complete to me (it takes quite a lot of effort in cockpit to get it to build, due to tracking down all the missing dependencies).

My gut feeling was that including the bits into page.scss is a good enough "global" place? Every page ought to import that, no?

KKoukiou: Right, we can do it like this. page.scss is global enough indeed.

martinpitt:
The cockpit rework landed, so I updated this PR to sync up the lib/ files with cockpit, and drop sass-resources-loader on top. Visually it all seems good to me.

For the lolz, we've had this line with a typo (.scs instead of  .scss) in our webpack for a while, apparently without any noticeable error:
```js
                             path.resolve(nodedir, './@patternfly/patternfly/base/patternfly-variables.scs'),
```
sass-resources-loader ought to fail on missing files, to avoid hard to debug and much more subtle CSS problems.

KKoukiou (approved):

PR #620, opened 2020-12-11
workflows: Fix apt installation

martinpitt:
GitHub's base VMs don't automatically refresh package indexes, so it can
happen that one of the package dependencies get out of date and are not
available on the mirrors any more.

Run `apt update` first to ensure that the workflow installs the latest
packages.


martinpitt:
Ported from https://github.com/cockpit-project/starter-kit/pull/400

mvollmer (approved):

PR #628, opened 2020-12-17
spec: Add BuildRequires: make

martinpitt:
Taken from Troy Dawson's downstream commit:
https://src.fedoraproject.org/rpms/cockpit-podman/c/04f4a72c7033056690
due to removing `make` from the buildroot:
https://fedoraproject.org/wiki/Changes/Remove_make_from_BuildRoot


marusak (approved):
Thanks!

PR #629, opened 2020-12-17
webpack cleanups

martinpitt:
Mostly ported from https://github.com/cockpit-project/starter-kit/pull/395 .

PR #630, opened 2020-12-17
Use sassc instead of node-sass

martinpitt:
node-sass is a compiled ELF module, which is problematic for
distributions that want to rebuild everything from source. The sassc CLI
program is packaged everywhere, and both use the same libsass library.

So drop node-sass and sass-loader, and replace it with a simple loader
wrapper around sassc.

This also saves 122 npm packages (16 MB in node_modules/).

Ported from starter-kit commits 233b5cfe04142, 8821db1daf1, and 50ef8a93623.

------

 - [x] Builds on top of PR #629


marusak:
#629 landed

martinpitt:
Unblocked, rebased, ready for review.

marusak (approved):
Thanks!

PR #634, opened 2020-12-18
Always fill the page with gray background

### LINKED ISSUE #615: webkit: Page does not extend to fill the content area

marusak:
This was not working in webkit browser. Now it works in webkit, firefox
and chrome.

Fixes #615
Closes #616

Webkit before:
before
Webkit after:
after

### DIFF
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -14,7 +14,7 @@ body {
     flex-direction: column;
 
     @at-root #app {
-        min-height: 100%;
+        height: 100%;
     }
 }
 


martinpitt (approved):
That simple? nice!

marusak:
> That simple? 

I am a bit skeptical as well. I guess I'll give @KKoukiou and @garrett chance to comment on this.

garrett:
Just check to make sure the content overflows. But if #app is in the right place, it still should. I've previously fixed this elsewhere (and I thought in podman too? at least in some commit, which might not've been merged? or even pushed?) by making #app the topmost element in the file, instead of one level down.

marusak:
> Just check to make sure the content overflows.

It does.

> I've previously fixed this elsewhere (and I thought in podman too? at least in some commit, which might not've been merged? or even pushed?) by making #app the topmost element in the file, instead of one level down.

You mean #616?

garrett:
Ah, OK.

Well, if this works cross-browser, then that's great.

(I'm on PTO, so I'm not on my work computer, so I can't really test. I just happen to be on GitHub at the moment. :wink:)

PR #641, opened 2020-12-31
Fedora 32 and 33 now can checkpoint containers on CGroupsV2

marusak:
As seen in https://github.com/cockpit-project/bots/pull/1497
I expect Fedora-32(/rawhide) tests to fail, but the `@bots#` versions to succeed.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -857,7 +857,45 @@ class TestApplication(testlib.MachineCase):
         b.wait_text("#containers-containers tbody tr:contains('busybox:latest') > td:nth-child(3)", "")
         b.wait_text("#containers-containers tbody tr:contains('busybox:latest') > td:nth-child(4)", "")
 
-    def testCheckpointRestoreSystem(self):
+    @testlib.nondestructive
+    def testCheckpointRestoreCGroupsV2(self):
+        b = self.browser
+        m = self.machine
+
+        if not self.cgroupsV2():
+            return;
+
+        # Only Fedora works with CGroupsV2
+        if m.image in ["fedora-32", "fedora-33"]:
+            return self._testCheckpointRestore()
+
+        # On other systems just check that we get expected error messages
+
+        self.login_and_go("/podman", superuser=True)
+        b.wait_visible("#app")
+        self.filter_containers('all')
+
+        # Run a container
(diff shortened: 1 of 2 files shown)


test/check-application
```diff
@@ -77,7 +77,7 @@ class TestApplication(testlib.MachineCase):
         self.allow_journal_messages("/run.*/podman/podman: couldn't connect.*")
         self.allow_journal_messages(".*/run.*/podman/podman.*Connection reset by peer")
 
-        self.has_criu = self.machine.image not in ["debian-testing", "ubuntu-stable"]
+        self.has_criu = self.machine.image not in ["debian-testing", "ubuntu-stable", "fedora-32"]
```
marusak, line 80: fedora-33 will get here on the next refresh as well, so then we should just flip the list.

marusak:
Oh, fedora-32 supports checkpoint, I'll look at the test how it behaves now vs. how it behaves with the new version.

marusak:
Both F32 and F33 with their respective refreshes worked. Unfortunately it fails with the current images so this needs to land in lockstep with those refreshes.

test/check-application
```diff
@@ -857,7 +857,45 @@ class TestApplication(testlib.MachineCase):
...
         b.wait_text("#containers-containers tbody tr:contains('busybox:latest') > td:nth-child(4)", "")
 
-    def testCheckpointRestoreSystem(self):
+    @testlib.nondestructive
+    def testCheckpointRestoreCGroupsV2(self):
+        b = self.browser
+        m = self.machine
+
+        if not self.cgroupsV2():
+            return;
```
martinpitt, line 866: Wrote a lot of JS lately? (semicolon) :grin:

test/check-application
```diff
@@ -857,7 +857,45 @@ class TestApplication(testlib.MachineCase):
...
+
+        # Only Fedora works with CGroupsV2
+        if m.image in ["fedora-32", "fedora-33"]:
```
martinpitt, line 869: FYI: Debian's systemd switched to cgroupsv2 a few days ago, so we'll soon get it there as well.

martinpitt (approved):
Nice, thank you!

PR #642, opened 2021-01-04
Fix error and success handling on image pull

### LINKED ISSUE #640: JSON Error on Arch
When attempting to install an image from the Cockpit UI, this error message is thrown instead of deploying the image:
```
Danger alert:Failed to download image docker.io/bytemark/smtp:latest
Error message:Unexpected token { in JSON at position 64
```

marusak:
We have shown error on every pull because we tried to parse the whole
    response as one JSON, but in fact it is a group of JSONs. That meant we
    failed to notice the fact that we were not showing any useful error when
    downloading actually failed. (We always just shown that we could not
    parse the response)
    
    Fixes #640

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -256,7 +256,18 @@ export function pullImage(system, reference) {
             reference: reference,
         };
         podmanCall("libpod/images/pull", "POST", options, system)
-                .then(reply => resolve(JSON.parse(reply)))
+                .then(r => {
+                    // Need to check the last response if it contains error
+                    const responses = r.trim().split("\n");
+                    const response = JSON.parse(responses[responses.length - 1]);
+                    if (response.error) {
+                        response.message = response.error;
+                        reject(response);
+                    } else if (response.cause) // present for 400 and 500 errors
+                        reject(response);
+                    else
+                        resolve();
+                })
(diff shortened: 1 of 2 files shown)


marusak:
I suspect that we will need naughty for https://github.com/containers/podman/issues/8870 but lets see first what our bots think.

KKoukiou (approved):
Hmm the API is weird here or is it me? libpod pull should fail properly when the pull fails and we should catch it in catch without searching for error in the successful response.

Your changes make the error show up, so we should merge them, but should we report a bug in podman for the API? 

In addition, when in non-existent tag is used when downaloading an image, the API fail with a not-so-good error message.

:note the screenshot is with your change included.
Screen Shot 2021-01-04 at 14 05 38

marusak:
> Hmm the API is weird here or is it me? libpod pull should fail properly when the pull fails and we should catch it in catch without searching for error in the successful response.

Yeah, kinda what I expected as well. But it is streaming so it does not have chance to response with different error code.

> In addition, when in non-existent tag is used when downaloading an image, the API fail with a not-so-good error message.

this is bug in podman, reported as https://github.com/containers/podman/issues/8870

PR #643, opened 2021-01-04
test: Don't assume $XDG_RUNTIME_DIR

marusak:
Fixes https://github.com/cockpit-project/bots/pull/1489 where `$XDG_RUNTIME_DIR` is `/run/user/1001`.


martinpitt (approved):
That's a really nice fix indeed! The commit message is confusing though, I first thought it means "don't assume that X-R_D is set" (which would be weird, as it is a non-negotiable API for these purposes), where in fact the fix does the opposite. :-) So maybe squash it as "Don't hardcode X_R_D"?

PR #644, opened 2021-01-07
Drop travis CI

martinpitt:
Since Travis changed its payment model, we have not had a free plan any
more for the cockpit-project organization. Drop it again, and let all
tests run on our own infra (see https://github.com/cockpit-project/bots/pull/1526)


marusak (approved):
Thanks!

PR #646, opened 2021-01-07
test: Drop forgotten sit()

marusak:
I wonder who introduced this :thinking:


martinpitt (approved):
That happens if you forget to sit back down after a standup!

PR #648, opened 2021-01-08
Packit fixes

marusak:
I just want to run `fedora-33/rawhide` but though why not fixing this? :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,7 @@ devel-install: $(WEBPACK_TEST)
 	ln -s `pwd`/dist ~/.local/share/cockpit/$(PACKAGE_NAME)
 
 dist-gzip: $(TARFILE)
+	echo `pwd`/$(TARFILE)
 
 # when building a distribution tarball, call webpack with a 'production' environment
 # we don't ship node_modules for license and compactness reasons; we ship a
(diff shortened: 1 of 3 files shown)


martinpitt (dismissed):
Thanks!

marusak:
Ok, two failed tests. I'll debug them locally and push fix here.

marusak:
There is some weird issue, that the first test always fails. Second test or retry works just fine. I saw it in CI as well. I am playing with it, it seems like race but I am having no luck solving it yet. Pushing to get one more run.

marusak:
Ok, so the very first test fails as expected and also testDownload is failing. It fails the same way, not sure though why. Looking.

martinpitt (approved):
Thanks!

martinpitt:
[testDownloadImage](https://logs.cockpit-project.org/logs/pull-648-20210108-142833-d09bd223-rhel-8-4/log.html#5-2) failed on rhel-8-4, retrying once to compare.

martinpitt:
FTR: the packit fixes don't seem to work, but I'm happy to address that in #649.

marusak:
F33 rawhide is green! I feel like there is going to be some flakiness to it, but it is not required context, so let's go with this.
Also I'll open issue about that first failing test.

PR #649, opened 2021-01-08
Add integration tests in FMF format

martinpitt:
This will allow us to use the same test execution scripts for downstream
gating and upstream PRs via packit. Copy browser.sh and run-test.sh from
Fedora's downstream gating tests, and adjust browser.sh's path variables
to work for both STI and FMF.

Enable GitHub PR integration.

See https://docs.fedoraproject.org/en-US/ci/tmt/ for details.

------

 - [x] rebase after #648
 - [x] wait for https://github.com/packit/sandcastle/pull/90 to get rolled out (asked in that PR)
 - [x] fix [incomplete logs](https://github.com/packit/packit-service/issues/917)
 - [x] increase sandcastle memory to fix webpack OOM: https://github.com/packit/sandcastle/pull/92 (not rolled out yet)
 - [x] rebase on top of #656 
 - [x] apply the browser.sh changes to downstream in a PR, and ensure that they get in sync: https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/17
 - [x] unbreak packit test runs: https://github.com/packit/packit/issues/1078
 - [x] enable all Fedora versions

I am currently testing this with

    bots/vm-run fedora-33
    tmt -v run --all provision --how connect --guest c

It would be much more elegant to test with just `tmt run`, but that [requires system libvirt](https://bugzilla.redhat.com/show_bug.cgi?id=1914194) which I don't have on my system. The tests generally work, but I see a bunch of failures still.

PR #650, opened 2021-01-11
Is sassc installed?

marusak:
I've tried the new container and it worked, lets see if it is deployed

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -5,4 +5,6 @@ upstream_package_name: cockpit-podman
 downstream_package_name: cockpit-podman
 actions:
   post-upstream-clone: make cockpit-podman.spec
-  create-archive: make dist-gzip
+  create-archive:
+    - "make dist-gzip"
+    - "sh -c 'find . -name \"*tar.gz\"'"


marusak:
:man_shrugging: locally it works, in sandbox it works, here it does not work

martinpitt:
I'm also trying this in PR #649. I am not sure whether we need a separate rule to output the name after all, as the error message says "the single line of the output". So in #649 I split it again.

marusak:
> I am not sure whether we need a separate rule to output the name after all, as the error message says "the single line of the output". So in #649 I split it again.

Ah right, you are trying it out as well. It it fine when it is at the of the log, does not need to be separate command (from my local tests). I am trying it now with separate command in packit.yaml with `find` as 1. If it fails to be created, this command should not find it 2. Maybe since not all logs are parsed from `make dist-gzip` it might not see it.

marusak:
Ok, `find` did not find it, that means `make dist-gzip` still fails.

TomasTomecek:
playing with this a little bit more and finally getting to something:
```
ERROR in ./src/Containers.scss                                                                                                                                                                
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):                                                                                                             
ModuleBuildError: Module build failed (from ./lib/sassc-loader.js):                                                                                                                           
Error: EACCES: permission denied, mkdtemp 'sassc-loader.XXXXXX'                                                                                                                               
    at Object.mkdtempSync (fs.js:1847:3)                                                                                                                                                      
    at Object.module.exports (/src/lib/sassc-loader.js:9:24)                                                                                                                                  
    at /src/node_modules/webpack/lib/NormalModule.js:316:20                                                                                                                                   
    at /src/node_modules/loader-runner/lib/LoaderRunner.js:367:11                                                                                                                             
    at /src/node_modules/loader-runner/lib/LoaderRunner.js:233:18                                                                                                                             
    at runSyncOrAsync (/src/node_modules/loader-runner/lib/LoaderRunner.js:143:3)                                                                                                             
    at iterateNormalLoaders (/src/node_modules/loader-runner/lib/LoaderRunner.js:232:2)                                                                                                       
    at Array.<anonymous> (/src/node_modules/loader-runner/lib/LoaderRunner.js:205:4)                                                                                                          
    at Storage.finished (/src/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:55:16)                                                                                               
    at /src/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:91:9                                                                                                                   
    at /src/node_modules/graceful-fs/graceful-fs.js:123:16                                                                                                                                    
    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:63:3)  
```
this looks like a problem with openshift's anyuid + the truncated output seems indeed to be caused by npm's black sorcery

martinpitt:
@TomasTomecek : OOh! So `mktemp` is broken in the container?

TomasTomecek:
> @TomasTomecek : OOh! So `mktemp` is broken in the container?

I thought that we are hitting openshift's anyuid but it wasn't the case - I tried to mimic how openshift runs containers on my own but failed, hence the error above, sorry for the noise.

The actual problem is, that node segfualts on being out of memory: https://github.com/packit/packit-service/issues/917#issuecomment-760948922

martinpitt:
I sent #656 to do some fixes which have proven to work on packit staging. I think we can close this now?

PR #651, opened 2021-01-12
Revert "test: Explicitly install runc on Debian testing"

skobyda:
This reverts commit 03944ea904ab41282922c2260a5e3ee828e6fc89.

Since https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=971253 has been
fixed, we can safely revert back to the 'crun'

Fixes a failure in https://github.com/cockpit-project/bots/pull/1531

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -10,8 +10,14 @@ if [ -d /var/tmp/debian ]; then
     fi
     eatmydata apt-get install ${BACKPORTS:-} ${APT_INSTALL_OPTIONS:-} -y cockpit-ws cockpit-system podman
 
-    # HACK: podman dependencies prefer crun, but config defaults to runc: https://bugs.debian.org/971253
-    eatmydata apt-get install -y runc
+    # HACK: starting podman.service complains about missing crun: https://bugs.debian.org/961016
+    # use crun for Debian, but Ubuntu so far still has runc
+    OS_RELEASE=$(grep '^NAME' /etc/os-release | awk -F'=' '{print $2}')
+    if [ $OS_RELEASE != "Ubuntu" ]; then
+        eatmydata apt-get install -y runc
+    else
+        eatmydata apt-get install -y crun
+    fi
 
     # build source package
     cd /var/tmp


martinpitt:
This is apparently not fixed in [ubuntu-stable](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-651-20210112-214843-3409cd89-ubuntu-stable/log.html) yet:
```
Error: default OCI runtime "runc" not found: invalid argument
Command '/var/tmp/vm.install' returned non-zero exit status 125.
```

Indeed [ubuntu groovy](https://launchpad.net/ubuntu/+source/libpod) still has podman 2.0.6, so I suggest to turn this into a condition on `if [ "${TEST_OS#ubuntu}" != "$TEST_OS" ]`.

martinpitt (changes requested):
see above

skobyda:
> This is apparently not fixed in [ubuntu-stable](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-651-20210112-214843-3409cd89-ubuntu-stable/log.html) yet:
> 
> ```
> Error: default OCI runtime "runc" not found: invalid argument
> Command '/var/tmp/vm.install' returned non-zero exit status 125.
> ```
> 
> Indeed [ubuntu groovy](https://launchpad.net/ubuntu/+source/libpod) still has podman 2.0.6, so I suggest to turn this into a condition on `if [ "${TEST_OS#ubuntu}" != "$TEST_OS" ]`.

I'm not so profound in shell, what does '#ubuntu' do? just concatenate string? Then it could be just ubuntu-stable#ubuntu != ubuntu-stable ?

martinpitt:
@skobyda: See man bash "Parameter expansion". With `${var..}` you can manipulate strings. For example with `X=foo123.bar`, you can cut off a prefix, with `${X#foo} = "123.bar"`, or cut off a suffix with `${X%.bar} = "foo". This even works with `*` patterns, for example` ${X%.*} = "foo123"` (cut off a `.` and everything after it). These are the most common string operations, but you can e.g. also use default values if a variable is undefined (`${X:-defaultvalue}`) and others.

So `[ "${TEST_OS#ubuntu}" != "$TEST_OS" ]` is the POSIX shell way to ask "does the string `$TEST_OS` start with 'ubuntu'". If the string is different after cutting of an "ubuntu" prefix, then that is true. If it starts with anything else, the string does not have such a prefix, and `${TEST_OS#garbage}` = `$TEST_OS` (unchanged). Does that make sense?

It's similar to `echo "$TEST_OS" | grep -q '^ubuntu'`, except that it does not need two forks, an exec of grep, and grep's rather expensive string matching.

skobyda:
I see SRPM build failures (What does that mean)
Apart from that, both debian and ubuntu are now passing

test/vm.install
```diff
@@ -10,8 +10,14 @@ if [ -d /var/tmp/debian ]; then
     fi
     eatmydata apt-get install ${BACKPORTS:-} ${APT_INSTALL_OPTIONS:-} -y cockpit-ws cockpit-system podman
 
-    # HACK: podman dependencies prefer crun, but config defaults to runc: https://bugs.debian.org/971253
-    eatmydata apt-get install -y runc
+    # HACK: starting podman.service complains about missing crun: https://bugs.debian.org/961016
+    # use crun for Debian, but Ubuntu so far still has runc
+    OS_RELEASE=$(grep '^NAME' /etc/os-release | awk -F'=' '{print $2}')
```
martinpitt, line 15: OK for now, just a hint for the future: The canonical way to use that is to source it as a shell script:
```sh
. /etc/os-release
if [ "$NAME" = Ubuntu ] ...
```

martinpitt (approved):
Thanks! Not quibbling too much over style here, as this unblocks image refreshes.

Ignore the packit failures for now -- we can't fix them yet, but they are also something entirely new.

PR #652, opened 2021-01-14
test: Allow debian-testing to cgroupsV2

skobyda:
Fixes a failure in debian-testing image refresh:
https://github.com/cockpit-project/bots/pull/1544

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -81,7 +81,7 @@ class TestApplication(testlib.MachineCase):
         self.has_selinux = self.machine.image not in ["debian-testing", "ubuntu-stable"]
 
     def cgroupsV2(self):
-        return self.machine.image not in ["debian-testing", "ubuntu-stable", "rhel-8-3", "rhel-8-4"]
+        return self.machine.image not in ["ubuntu-stable", "rhel-8-3", "rhel-8-4"]
 
     def execute(self, system, cmd):
         if system:


marusak:
@skobyda about contexts, you triggered some of these manually, right?

`debian-testing@cockpit-project/cockpit ` - is really pointless here, and it will do something weird
`debian-testing@cockpit-project/cockpit-podman` == `debian-testing` so no point in triggering that

You should trigger though `debian-testing@bots#1544`. Can be done simply by using `--bots-pr` option in `tests-trigger`.

skobyda:
@marusak Thanks. I also see some failure in https://logs.cockpit-project.org/logs/pull-652-20210114-114153-5aa6b5bb-rhel-8-4/log.html#5-2 . Have you seen it before?

skobyda:
Same thing happend in https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-652-20210114-121415-5aa6b5bb-debian-testing-bots-1544/log.html, but on second try it passed. So I will retry it here, and if it passes, it should be good to go

marusak (approved):
Nice, thanks!
The `imageDownload` failure is kinda known.

skobyda:
So there are failures in rhel-8-4 (known) and in an old debian-testing image (expected since it's using the old image, the new image used in test debian-testing@bots#1544 passes).

skobyda:
@marusak Since rhel-8-4 failure is known, is there a way to get around the Required status check? Seems like I don't have enough rights to merge it with that flag (administrators apparently can)

marusak:
This won't be possible to land because `debian-testing` will never succeed.
However, it is possible to turn of branch protection, land it and turn it back on. I'll do that and then we need to land https://github.com/cockpit-project/bots/pull/1544 ASAP as well.

PR #654, opened 2021-01-18
Stop hardcoding library paths

martinpitt:
Just tell webpack to resolve modules in lib/ instead.

This is preparation for moving the library to src/lib for consistency
with starter-kit.


martinpitt:
Argh, once more without git pilot error, sorry..

KKoukiou (approved):

PR #655, opened 2021-01-18
Use cockpit library components and webpack helpers from cockpit.git

martinpitt:
Drop the cockpit-components-*, cockpit-po-plugin.js and sassc-loader.js
code copies, and check them out from a stable cockpit tag, similarly to
how we already get the PatternFly CSS. Also move them from lib/ to
src/lib for consistency with starter-kit.

Ported from starter-kit commit XXXXX

 - [x] Builds on top of PR #654 
 - [x] Land corresponding starter-kit PR and adjust hash in commit message
 - [x] Wait for next cockpit release, checkout stable 236 tag instead of master
 - [x] Diff copy vs. imported files (jsx+css) and visually check for regressions


martinpitt:
I diffed all files in lib with

    for f in lib/*; do diff -u $f ../cockpit-podman/src/$f; done

and the [only diff](https://github.com/cockpit-project/cockpit-podman/files/5831457/lib.diff.txt) is in the PO plugin and sass loader to accomodate some extra cockpit bits (but I kept these backwards compatible). Turned out that the cockpit-components-* bits were already perfectly in sync. I did a visual comparison with master as well, just to be on the safe side. I did see a handful of little warts, but absolutely no difference between the two.

KKoukiou (approved):
woohoo

PR #658, opened 2021-01-23
Python style

marusak:


test/run
```diff
@@ -4,5 +4,12 @@
 
 set -eu
 
+# run static code checks like pyflakes and pep8
```
martinpitt, line 7: I think this would be a little more useful as `Makefile` rules, so that you can do `make codecheck` or something.

test/run
```diff
@@ -4,5 +4,12 @@
...
+PYEXEFILES="$(git grep -lI '^#!.*python') $(git ls-files "*.py")"
+python3 -m pyflakes $PYEXEFILES
+
+# TODO: Fix long lines
+python3 -m pycodestyle --ignore E501 $PYEXEFILES
```
martinpitt, line 12: Hint: Start with `--max-line-length=<longest line that we currently have`, and work your way down to ~ 130 or so. That seems to be our current practice (80 is really a bit low IMHO)

martinpitt (approved):
Dakujem!

test/run
marusak: Updated, our recordman is 194 though :)

test/run
marusak: Done

martinpitt (approved):
Cheers!

PR #663, opened 2021-01-29
spec: Drop cockpit-system dependency

martinpitt:
cockpit-system is not strictly a requirement. In "remote ssh host" mode,
the Podman UI works just fine with just cockpit-{podman,bridge}.

----

It looks a little unfamiliar, but there is not actually anything wrong with this:

cockpit-podman-only

### DIFF
--- a/cockpit-podman.spec.in
+++ b/cockpit-podman.spec.in
@@ -28,7 +28,6 @@ BuildRequires:  libappstream-glib
 BuildRequires:  make
 
 Requires:       cockpit-bridge >= 138
-Requires:       cockpit-shell >= 138
 Requires:       podman >= 2.0.4
 
 %description
--- a/packaging/debian/control
+++ b/packaging/debian/control
@@ -11,7 +11,6 @@ Architecture: all
 Multi-Arch: foreign
 Depends: ${misc:Depends},
          cockpit-bridge (>= 138),
-         cockpit-system (>= 138),
          podman (>= 2.0.4),
 Description: Cockpit component for Podman containers
  The Cockpit Web Console enables users to administer GNU/Linux servers using a
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -580,11 +580,13 @@ class Application extends React.Component {
                     <Button onClick={this.startService}>
                         {_("Start podman")}
                     </Button>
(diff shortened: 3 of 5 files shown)


martinpitt:
I want to drop the dep in debian as well, but let's wait for first round of tests.

martinpitt:
Ugh, [something bad](https://logs.cockpit-project.org/logs/pull-663-20210129-163836-55599dd6-fedora-33-rawhide/log.html) happened in rawhide..

marusak:
> Ugh, something bad happened in rawhide..

Yeah, I saw this yesterday already, I prepared environment locally but haven't got to it yet. Not related to this PR.

marusak (commented):
Thanks!
The only worry I see with this is, that we have 'Troubleshoot' button which redirects to services page. I think we should not show that button when services page is not present (we can check that easily with manifest).

martinpitt:
@marusak: Good point, thanks for spotting! I added a commit that handles that.

marusak (approved):
Nice, thanks!

martinpitt:
@marusak: [debian-testing also fails](https://logs.cockpit-project.org/logs/pull-663-20210201-073807-b9dbbbd7-debian-testing/log.html). Does that ring a bell?
```
> error: Failed to commit container test-sh0: CommitFailure: error resolving name "testimg:testtag": could not get either XDG_CONFIG_HOME or HOME: Internal Server Error
```

marusak:
> Does that ring a bell?

Not at all, I see this for the first time.

martinpitt:
I can reproduce the debian-testing failure locally without a problem; it also happens in other PRs like #665. It did not happen [4 days ago in the d-testing image refresh](https://github.com/cockpit-project/bots/pull/1594), and also not two days ago in #664.

I don't think `make vm` does any package installation. So, mystery..

martinpitt:
I investigated the debian-testing regression. Reported at https://bugs.debian.org/981813 , known issue https://github.com/cockpit-project/bots/issues/1635 , sent naughty pattern to https://github.com/cockpit-project/bots/pull/1636

martinpitt:
Re-running debian-testing after landing the naughty.

PR #667, opened 2021-02-04
cockpituous-release: Rename Fedora rawhide branch to "rawhide"

martinpitt:
Turns out "main" is just some kind of alias. Let's use "rawhide"
directly.


marusak (approved):
I feel like I am becoming master in deleting tags and retagging it back up :)

PR #669, opened 2021-02-06
Enough is enough

marusak:


marusak:
All green without single retry!! We havn't see that in a while.
rawhide has one test genuinely broken, I sent naughty in https://github.com/cockpit-project/bots/pull/1648

I'll cleanup a bit and push force and test ^ as well.

src/client.js
```diff
@@ -46,9 +46,11 @@ export function streamEvents(system, callback) {
 
 export function getInfo(system) {
     return new Promise((resolve, reject) => {
+        const timeout = setTimeout(() => reject(new Error("timeout")), 5000);
```
martinpitt, line 49: If the answer comes back eventually, this won't mess up the protocol and our code? AFAIR calling resolve() after reject() already got  called seems safe. But wrt. the protocol, if getInfo() fails then we don't do any other API call. `</thinking_aloud>`. So should be fine.

test/check-application
```diff
@@ -73,9 +73,6 @@ class TestApplication(testlib.MachineCase):
 
         self.addCleanup(self.machine.execute, "killall -9 podman || true")
 
-        # HACK: https://github.com/containers/podman/issues/8762
-        self.admin_s.execute("timeout 5 curl --unix-socket $XDG_RUNTIME_DIR/podman/podman.sock http://d/v1.24/libpod/info || true")
```
martinpitt, line 77: Curious, isn't that pretty much exactly the same workaround that the previous commit addresses? (info not working the first time)

test/check-application
```diff
@@ -88,6 +88,28 @@ class TestApplication(testlib.MachineCase):
...
+        with self.browser.wait_timeout(30):
+            try:
+                b.wait_visible("#containers-containers")
+                b.wait_not_in_text("#containers-containers", "Loading")
+                b.wait_not_present("#overview div.pf-c-alert")
+            except testlib.Error:
+                if system:
+                    b.wait_in_text("#overview div.pf-c-alert .pf-c-alert__title", "User Podman service is also available")
+                    b.click("#overview div.pf-c-alert .pf-c-alert__action > button:contains(Start)")
+                    b.wait_not_present("#overview div.pf-c-alert")
```
martinpitt, line 108: Ugh, that is really evil.. But I don't have a better idea either. :cry:

martinpitt (commented):
Thanks for addressing these! Please test against the bots PR with the naughty, too.

test/check-application
marusak: it indeed is.
There has been at least 3 times in last few months a problem, where the first call either is stuck or fails. Always slightly differently and the reproducer is always slightly different but takes me at least a full day to get to it. So this PR is me throwing hands in the air and just doing the big and ugly workaround with "if it does not work automatically as it should, just click the button to start it - as users would, if this would happen to them".

test/check-application
marusak: yeah, I don't like it either, but I wasted way too much time with reporting to podman and it always just gets forgotten and then closed with "cannot reproduce anymore"

marusak:
No retries once again.

martinpitt (approved):
Thanks!

PR #675, opened 2021-02-12
Be case-insensitive when checking the error message on checkpoint/restore

KKoukiou:
Upstream podman commit
https://github.com/containers/podman/commit/b53cb57680a6fd7b383636ac2d6cd71003532915
has changed the capitalization.


martinpitt (approved):
Thanks for tracking this down so thoroughly!

PR #678, opened 2021-02-17
package.json: Lock @patternfly/react-styles version

martinpitt:
It's part of patternfly-react, so we should update it in a controlled
fashion. Letting it auto-update freely often breaks the build.

Similar to https://github.com/cockpit-project/cockpit/commit/bbde4c01947

----

This avoids out-of-the-blue failures like [this one](https://logs.cockpit-project.org/logs/pull-1684-20210217-090606-38f52238-rhel-8-4-cockpit-project-cockpit-podman/log.html) in an image refresh.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   "dependencies": {
     "@patternfly/patternfly": "4.87.2",
     "@patternfly/react-core": "4.97.1",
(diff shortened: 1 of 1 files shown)


marusak (approved):
Cheers!

PR #684, opened 2021-02-24
fmf: Split out test from top-level plan

psss:
@martinpitt, this demonstrates how tests can be shared across multiple repositories.
Relevant: https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/32


martinpitt:
Many thanks @psss, this is really helpful ! This makes sense, it'll just require some more fine-tuning. The tests in packit now all fail due to
```
+ cd /ARTIFACTS/work-allCUkAoS/plans/all/discover/default/tests/test
+ git init
+ make bots test/common
make: *** No rule to make target 'bots'.  Stop.
```
It's just running in the wrong directory. test/browser/browser.sh moved one level down in the directory hierarchy, and thus

     SOURCE="$(realpath $TESTS/..)"

is now wrong. I pushed a fixup to your branch.

psss:
> Many thanks @psss, this is really helpful!

Glad to hear that :)

psss:
Hm, interesting, just tested with your latest update of the fedora PR:

```
> tmt run discover -v
/var/tmp/tmt/run-259

/plans/upstream
    discover
        how: fmf
        url: https://github.com/psss/cockpit-podman.git
        ref: shared-test
        hash: d64fd2c
        summary: 1 test selected
            /test/browser
```
Looking at your logs in more detail...

martinpitt:
Hmm, now the packit tests actually did run, but got interrupted in the middle. That now smells like an infra bug.

psss:
Oh I see it: Just don't use `-n` or `--dry` because those will actually skip fetching the repository. The `discover` step should be safe it just clones repo and checks metadata so the dry mode should not be needed.

psss:
> Hmm, now the packit tests actually did run, but got interrupted in the middle. That now smells like an infra bug.

Perhaps you should increase the [duration](https://tmt.readthedocs.io/en/latest/spec/tests.html#duration)?

martinpitt:
@psss : Argh, my bad, thanks!  Confirming that this works.

Interestingly, the test run in https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/32 looks exactly the same: the test stops after one check instead of running all 11. So in that sense the PRs work, upstream and downstream tests are now broken in *exactly* the same way :grin:

martinpitt:
Good call, done. Thanks for all the hand-holding!

psss:
Yes, it was the [timeout](http://artifacts.dev.testing-farm.io/6f4e9158-ae33-4312-988f-cc209f132235/work-upstreamys7kwP/tmt-run.log). The raised duration should help.

martinpitt:
Very cool, all green! Thanks so much @psss! I force-pushed this to squash the commits and clean up the commit message with keeping your authorship. The downstream tests still need improvement, but I'll discuss them in https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/32

martinpitt (approved):

martinpitt:
@marusak , want to have a look as well?

marusak (approved):
Nice!

martinpitt:
This needs a downstream gating test adjustment, I sent https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/33

PR #688, opened 2021-03-04
test: Generalize cgroupsV2() for all rhel-8-* versions

martinpitt:
It looks like RHEL newer 8.y versions won't get cgroupsv2 either, so
generalize the check.


marusak (approved):
Thanks

PR #694, opened 2021-03-25
fmf: Drop rawhide version override

martinpitt:
Our tests don't currently have special cases for fedora. If they ever do
get them again, let's add the rawhide version right away. They will be
gated by packit tests, and we just keep forgetting about this hack.

### DIFF
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -12,10 +12,6 @@ npm install chrome-remote-interface sizzle
 
 . /etc/os-release
 export TEST_OS="${ID}-${VERSION_ID/./-}"
-# HACK: upstream does not yet know about rawhide
-if [ "$TEST_OS" = "fedora-34" ]; then
-    export TEST_OS=fedora-33
-fi
 
 export TEST_AUDIT_NO_SELINUX=1
 


marusak (commented):
Shouldn't we just move it to 35-34 pair?

marusak:
I mean, if we do another release and we run tests in rawhide, this will break, won't it?

martinpitt:
How do you mean?

I meant that if we ever introduce a special case for, say, fedora-34, then the packit rawhide test in that PR will yell at us with "wah wah this doesn't work on fedora-35". I.e. if we do that, we always need to keep the fedora-(n+1) case in mind. I think that's better than forever forgetting about this hack.

marusak:
But don't we use ` test/browser/run-test.sh` in downstream gating already? Because if we do, then the next release will fail already.

martinpitt:
@marusak: Yes, we do -- but why would it fail? packit results show that it obviously works for "fedora-35" (as our hack is obsolete), and our downstream CI tests show that it works for "fedora-34". Our tests don't actually care, as they don't have fedora-* special cases.

So I'm afraid I still don't see the problem -- IRC maybe?

marusak (approved):
Ah I see!
We had this for special cases in tests but we don't have them anymore.

PR #697, opened 2021-03-31
Small test cleanups

martinpitt:
I noticed these while investigating the failures on rhel-9-0.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -78,7 +78,7 @@ class TestApplication(testlib.MachineCase):
         self.allow_journal_messages(".*/run.*/podman/podman.*Connection reset by peer")
 
         self.has_criu = m.image not in ["debian-testing", "ubuntu-stable"]
-        self.has_selinux = m.image not in ["debian-testing", "ubuntu-stable"]
+        self.has_selinux = "debian" not in m.image and "ubuntu" not in m.image
         self.has_cgroupsV2 = m.image not in ["ubuntu-stable"] and not m.image.startswith('rhel-8')
 
     def execute(self, system, cmd):
--- a/test/check-application
+++ b/test/check-application
@@ -40,28 +40,29 @@ class TestApplication(testlib.MachineCase):
 
     def setUp(self):
         super().setUp()
-        self.machine.execute("systemctl stop podman.service; systemctl --now enable podman.socket")
+        m = self.machine
+        m.execute("systemctl stop podman.service; systemctl --now enable podman.socket")
 
         # backup/restore pristine podman state, so that tests can run on existing testbeds
         self.restore_dir("/var/lib/containers", reboot_safe=True)
 
-        self.addCleanup(self.machine.execute, "systemctl stop podman.service")
+        self.addCleanup(m.execute, "systemctl stop podman.service")
 
         # HACK: sometimes podman leaks mounts
-        self.addCleanup(self.machine.execute, "podman rm --force --all && "
+        self.addCleanup(m.execute, "podman rm --force --all && "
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -885,13 +885,11 @@ class TestApplication(testlib.MachineCase):
...
     def testCheckpointRestoreCGroupsV2(self):
         b = self.browser
-        m = self.machine
 
         if not self.has_cgroupsV2:
             return
 
-        # Only Fedora works with CGroupsV2
-        if m.image.startswith('fedora'):
+        if self.has_criu:
```
martinpitt, line 892: @marusak: I'm not 100% sure about this change -- conceptually, hardcoding "fedora" here seems wrong and error prone. In practice, RHEL 9 has cgroups v2 and a recent criu (same as F34), but throws "does not support checkpoint/restore". But I'd consider that a bug (filing now), not a behavior to validate. It also does not help with `testCheckpointRestoreCGroupsV1` where rhel-9 fails in the same way, and this hack doesn't help us.

martinpitt: I filed https://bugzilla.redhat.com/show_bug.cgi?id=1944964 and I'm about to do a naughty for it -- let me know if you think that's the right course of action, or you'd rather hack the tests.

martinpitt: If they don't want to support criu in RHEL 9, then they should drop the `Recommends: criu`, and our image preparation can anticipate that with `rpm -e criu` and setting `has_criu` to False on RHEL-9.

martinpitt:
Interesting - [fedora 34 packit](http://artifacts.dev.testing-farm.io/7f34aaee-3e7c-4666-acc0-09810cf6ff52/) failed `testCheckpointRestoreCGroupsV2`: The [screenshot](http://artifacts.dev.testing-farm.io/7f34aaee-3e7c-4666-acc0-09810cf6ff52//work-alldOdoeO/plans/all/discover/default/tests/TestApplication-testCheckpointRestoreCGroupsV2-fedora-33-127.0.0.1-22-FAIL.png) shows "requires at least CRIU 31100", which is usually the symptom of `criu` not being installed at all. I wonder how this PR broke that -- the previous `if m.image.startswith('fedora')` should have been true previously as well.

I added another commit to make sure criu is installed in FMF.

martinpitt:
That wasn't it. Needs local reproducing/examination, moving to draft.

martinpitt:
Re-dropping the "depend on criu" commit, it didn't help. Also unwinding the "Simplify testCheckpointRestoreCGroupsV2" to confirm if that really was the cause. The other commits really ought to be no-effective-change cleanups, but computers..

martinpitt:
Meh, now I'm surprised -- it [failed again](http://artifacts.dev.testing-farm.io/bdb33e8e-35fa-463c-ba01-96174df91477/). So at least it was not the "Simplify testCheckpointRestoreCGroupsV2" commit. Post-holiday material now..

martinpitt:
A-ha! The FMF packit test is generally broken on F33, it happened in #698 as well. So marking ready for review again, and this needs to be investigated separately.

marusak (approved):
Thanks!
Should this make rhel-9 green?

martinpitt:
@marusak : It helps for rhel-9-0, but there is still some trouble with criu, see https://bugzilla.redhat.com/show_bug.cgi?id=1944964 . We said we wanted to wait until upstream acks the bug or says "sorry, not supported". But they assigned the bug now, so I guess let's treat that as a naughty? So we can land this, and I'll submit a bots PR with naughty + rhel-9 enablement, and I'm also looking into that cgroupsv2 test failure on TF fedora-33.

PR #700, opened 2021-04-06
fmf: Update crun and podman packages

martinpitt:
The initial installation of podman on the Fedora 33 Testing Farm
instances somehow pulls in an old version 0.15.5:
https://github.com/psss/tmt/issues/682

As a quick workaround, ensure that podman and crun are the latest
version, so that criu support is available.

---

This fixes the recent [packit failures on f33](http://artifacts.dev.testing-farm.io/33fd92a3-0d95-4bee-a51b-3faa49fa07a9/).

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -22,6 +22,9 @@ if ! rpm -q chromium; then
     dnf install -y chromium
 fi
 
+# HACK: ensure that critical components are up to date: https://github.com/psss/tmt/issues/682
+dnf update -y podman crun
+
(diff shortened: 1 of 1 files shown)


marusak:
That sounds like bug in TF. It should test the newest updates and not just some random, right?

martinpitt:
Exactly -- I just filed https://github.com/psss/tmt/issues/682 . Test succeded here, so I'll force-push to add the bug reference.

marusak (approved):
Thanks!

PR #702, opened 2021-04-09
Move to webpack 5

martinpitt:
Mostly cherry-picked from recent starter-kit PRs. 

 - [x] prerequisite commits: PR #704
 - [x] Fix scrolling to "used by" container to not crash for stopped containers (regression in f1e7c9b1518dbb): PR #705
 - [x] Get rid of react-scrollable-anchor, which is dead and does not work any more with webpack 5: PR #707 
 - [x] fix terser-built LICENSE file

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "compression-webpack-plugin": "^6.0.0",
     "copy-webpack-plugin": "^6.1.0",
     "css-loader": "^5.2.0",
+    "css-minimizer-webpack-plugin": "^1.3.0",
     "eslint": "^7.10.0",
     "eslint-config-standard": "^16.0.0",
     "eslint-config-standard-jsx": "^10.0.0",
@@ -37,15 +38,14 @@
     "htmlparser": "^1.7.7",
     "jed": "^1.1.1",
     "mini-css-extract-plugin": "^0.11.0",
-    "optimize-css-assets-webpack-plugin": "^5.0.3",
     "po2json": "^1.0.0-alpha",
     "sizzle": "^2.3.3",
     "stdio": "^2.1.0",
-    "string-replace-loader": "^2.3.0",
+    "string-replace-loader": "^3.0.0",
     "svgo": "^1.3.0",
-    "terser-webpack-plugin": "^4.0.0",
-    "webpack": "^4.17.1",
-    "webpack-cli": "^3.1.0"
+    "terser-webpack-plugin": "^3.0.0",
+    "webpack": "^5.31.0",
+    "webpack-cli": "^4.6.0"
   },
   "dependencies": {
     "@patternfly/patternfly": "4.96.2",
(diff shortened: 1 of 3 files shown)


martinpitt:
Oops, the automatic dist/index.js.LICENSE.txt isn't being built any more.

martinpitt:
Hah, so we do need terser-webpack-plugin for the LICENSE bits still. I'll send a PR to starter-kit to enable that, as we had to do that in -podman and -machines.

martinpitt:
This crashes in node_modules/react-scrollable-anchor/lib/utils/func.js, which is shipped (by its own npm module) as:
```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
var _arguments = arguments;
var debounce = exports.debounce = function debounce(func, wait, immediate) {
[...]
```
That `arguments` is not in the [original upstream source](https://github.com/gabergg/react-scrollable-anchor/blob/master/src/utils/func.js), so I suppose some babeling in there somehow broke that.

The same code is present in the webpack in master, but apparently embedded differently. Either way, [react-scrollable-anchor](https://www.npmjs.com/package/react-scrollable-anchor) has not been updated in 3 years, so it's probably time to replace that with something current PF4-ish.

src/Containers.jsx
```diff
@@ -226,12 +225,12 @@ class Containers extends React.Component {
         }
 
         return {
-            expandedContent: <ScrollableAnchor id={container.Id} key={container.Id}>
```
martinpitt, line 229: Wait, I misunderstood what this does -- it was introduced in commit b5f8df880253e2d. I'll check what that actually renders to, and whether it's enough to just replace it with a simple HTML tag.

martinpitt: Lol -- this functionality is broken in master. Once I click on the "used by" container, the page crashes with `TypeError: props.showAll is not a function`.

martinpitt: Fixed in #705, but the actual scrolling is still broken.

martinpitt:
All dependencies are fixed, so this PR is now a rather simple change. Tested locally (also visually) in dev and prod modes.

martinpitt:
Meh -- the internet is full of questions how to *disable* creation of a LICENSE.txt file, but I can't get terser to *build* it now, even though it should.. again needswork.

martinpitt:
Ugh, seems that webpack 5 uses a bit more memory which pushes packit over the line. I can reproduce this locally with
```
git clean -ffdx; podman run -ti -e NODE_OPTIONS=--max-old-space-size=500 --rm --memory 768MB -v $PWD:/src -w /src docker.io/usercont/sandcastle:prod make dist-gzip
```
I experimented with other `--max-old-space-size` values, but there is no sweet spot any more. As webpack is not at all security relevant, I am not willing to pay the price of setting up pre-built -dist for podman right now. Let's just keep webpack 4 for the time being.

At least this wasn't in vain, the recent fixes were all worth the effort :grin:

martinpitt:
Found a workaround -- building in development mode is fine.

martinpitt:
All green! :four_leaf_clover:

marusak (approved):
Danke!

PR #705, opened 2021-04-12
Fix crash on clicking on "Used by" for stopped containers

martinpitt:
This functionality was introduced in commit b5f8df880 and broken by
f1e7c9b1518 (and was previously not tested at all). Partially revert the
latter commit and move the "container filter" state back to Application,
so taht the `<Images>` component can use it. Supply a `showAll` property
to that, as it expects it.

Add an integration test which clicks on the "used by" containers and
cover both the running and the stopped case. The former does not really
do anything right now (that's another bug), but check that the latter
switches to "all containers" view and makes the stopped container
visible.

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -43,7 +43,6 @@ class Containers extends React.Component {
             checkpointInProgress: false,
             restoreInProgress: false,
             width: 0,
-            filter: "running",
         };
         this.renderRow = this.renderRow.bind(this);
         this.onWindowResize = this.onWindowResize.bind(this);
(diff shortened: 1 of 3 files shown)


martinpitt:
@KKoukiou As you wrote that original commit, I'd appreciate a review from you. It wasnt' clear why you moved the state, the commit message doesn't say..

KKoukiou:
@martinpitt I will review tomorrow, but as a quick feedback the reason I had moved it is because the state filtering is done only for the containers list, and therefore it belongs there. Having in the page header, is really misleading, making the user things that images card is affected as well.

martinpitt:
@KKoukiou : I kept the control in the containers list, there is no visual change. I just lifted the state upwards again. Not exactly elegant, but I'm afraid I don't know a better solution.

KKoukiou (approved):
Thanks for fixing this. While testing this I noticed that the `used by` list's items have really no indication that they are clickable. Really, if I would not have seen the test I would have no clue that there is something to click there. That's a thing for another PR of course.

PR #706, opened 2021-04-13
Some pf4 ports and CSS cleanup

KKoukiou:


src/ImageUsedBy.jsx
```diff
@@ -7,7 +7,6 @@ const _ = cockpit.gettext;
 
 const renderRow = (containerStats, container) => {
     const isRunning = container.State == "running";
-
```
KKoukiou, line 10: ups this was removed by accident.

martinpitt:
I was going to complain about broken mobile mode:
Screen Shot 2021-04-14 at 13 59 04

but this doesn't look any better on master:

Screen Shot 2021-04-14 at 13 59 47

So it's not a regression, and :+1: to moving to standard PF and dropping our custom CSS.

martinpitt (approved):
Looks great, thank you! Confirmed that the "used by" scrolling works just as brittle as before, so indeed the extra div was useless. Thanks for cleaning that up!

PR #707, opened 2021-04-13
Some tweaks to "image used by"

martinpitt:


martinpitt:
Fixed `testRunImageUser`.

marusak (approved):
Thanks!

src/Containers.jsx
```diff
@@ -221,12 +220,12 @@ class Containers extends React.Component {
         }
 
         return {
-            expandedContent: <ScrollableAnchor id={container.Id} key={container.Id}>
+            expandedContent: <div id={container.Id}>
```
KKoukiou, line 223: Was this actually even needed?
I think it's not used anywhere.

KKoukiou: I mean the id in the wrapper. Tests don't seem to use it and neither the code somehow.

martinpitt: It is - the jump [happens here](https://github.com/cockpit-project/cockpit-podman/blob/master/src/ImageUsedBy.jsx#L57).

KKoukiou: I sitll don't see how it's used. :/

The row.props.containerId is the props we pass to the row itself here, https://github.com/cockpit-project/cockpit-podman/blob/master/src/ImageUsedBy.jsx#L29

KKoukiou: So I think it's really not used, I removed it here https://github.com/cockpit-project/cockpit-podman/pull/706/commits/3dff474256765fb8b57737bbc775b60e0562dcdf , it's I am misunderstanding something please comment there ;)

PR #711, opened 2021-04-21
Stop explicitly installing cockpit-ws

martinpitt:
It is already present on all our test images, and trying to
install/upgrade it again touches the network. That is often unreliable
and makes tests less reproducible.

For cockpit-podman.rpm itself, upload that and install it with rpm, so
that dnf does not try and update the package metadata. This is the same
approach as with the debs.

-----

Fixes annoying errors like [this one](https://logs.cockpit-project.org/logs/pull-712-20210422-022404-fe2ac86a-rhel-9-0/log.html). The same happened to cockpit-machines today, which I fixed in https://github.com/cockpit-project/cockpit-machines/pull/120 . This is exactly the same fix minus the `--no-network`, as we need networking for the container pulls.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ VM_DEP=$(TARFILE) packaging/debian/rules packaging/debian/control
 VM_PACKAGE=--upload `pwd`/$(TARFILE):/var/tmp/ --upload `pwd`/packaging/debian:/var/tmp/
 else
 VM_DEP=$(RPMFILE)
-VM_PACKAGE=--install cockpit-ws --install `pwd`/$(RPMFILE)
+VM_PACKAGE=--upload `pwd`/$(RPMFILE):/var/tmp/
 endif
 
 ifeq ($(TEST_SCENARIO),rawhide)
--- a/test/vm.install
+++ b/test/vm.install
@@ -48,6 +48,10 @@ if [ -d /var/tmp/debian ]; then
     if systemctl is-enabled docker.service; then
         systemctl disable docker.service
     fi
+
+# install rpms
+elif [ -e /var/tmp/*.rpm ]; then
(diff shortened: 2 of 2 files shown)


martinpitt:
That [fedora-33 packit failure](http://artifacts.dev.testing-farm.io/12699eca-ae66-450b-8147-807bc77ee714/) looks concerning:
```
Error: Transaction test error:
  file /usr/share/man/man5/containers-mounts.conf.5.gz conflicts between attempted installs of podman-2:2.1.1-10.fc33.x86_64 and containers-common-4:1-10.fc33.noarch
```
Retrying once in case it's an unlucky unpack order (and not related to the change here)

martinpitt:
See PR #712, this breaks c-podman as well. This is fairly urgent hence.

martinpitt:
Ugh, the fedora-33 failure is https://github.com/psss/tmt/issues/682 all over again. But I don't see how we could work around this, as this happens early in TMT.

martinpitt:
I [followed up on the tmt issue](https://github.com/psss/tmt/issues/682#issuecomment-824545135) and [reported the containers-common packaging bug](https://bugzilla.redhat.com/show_bug.cgi?id=1900000). At this point I don't have a good workaround, so maybe just let's ignore packit f-33 until this is fixed properly?

martinpitt:
I think I found a fix for the packit failure now, locally it works.

martinpitt:
Ugh, that extra `prepare:` step [breaks the testing-farm](http://artifacts.dev.testing-farm.io/6304cf96-ba4c-43e0-9b49-24712d6845bc/pipeline.log). I revert this extra commit and I'll attempt it separately. Fixing our rhel-9-0 test is urgent, I don't want to stall that.

marusak (approved):
Thanks!

PR #713, opened 2021-04-22
fmf: Install/update critical dependencies earlier

martinpitt:
Updating them in the test script is too late, as now the "prepare" step
runs into a file conflict: https://bugzilla.redhat.com/show_bug.cgi?id=1900000

Introduce an explicit prepare: step which installs current podman/crun
before the test/browser/main.fmf's "require:" preparation. This avoids
the weird dependency resolution in dnf and is also closer to the planned
fix/workaround in tmt to update the testbed as a first step.

### DIFF
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -2,5 +2,10 @@ summary:
     Run all tests
 discover:
     how: fmf
+# HACK: ensure that critical components are up to date
+# https://github.com/psss/tmt/issues/682 and https://bugzilla.redhat.com/show_bug.cgi?id=1946975
+prepare:
+    how: shell
+    script: dnf install -y podman crun
 execute:
     how: tmt
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -22,9 +22,6 @@ if ! rpm -q chromium; then
     dnf install -y chromium
 fi
 
-# HACK: ensure that critical components are up to date: https://github.com/psss/tmt/issues/682
-dnf update -y podman crun
-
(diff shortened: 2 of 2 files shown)


martinpitt:
Draft for now, as the testing farm is currently broken. I'll un-"no-test" this once it works again. This ought to fix [this error](http://artifacts.dev.testing-farm.io/acac7541-cbe8-45d5-b125-f5d8c5f3d39b/), I tested it locally already (`tmt run --all provision --how virtual --image fedora-33`)

martinpitt:
@psss, @thrix: This fix works locally in `tmt` (see above) to work around that tmt/dnf bug, but apparently [still not on the testing farm](http://artifacts.dev.testing-farm.io/6b36baf5-2334-40a6-a0f7-520d0d67c12e/). It looks like it ignores the new `prepare:` step from the test plan? Would you have any other idea how to work around this?

thrix:
@martinpitt testing farm does the artifact installation itself and after that runs tmt, so this indeed is not working, as you would expect. We have ideas to make it possible to run commands before testing farm actually installs the tested artifacts, but it is not yet ready. Actually the real problem here is that those conflicting packages were let into rawhide, and this will solve only gating on installability or rpmdeplint ....

Anyway, it should be possible to workaround, so we plan to implement possibility of doing this ...

martinpitt:
Thanks @thrix, so that explains why this isn't working. So I suppose we'll just ignore the fedora-33 failure for the time being. I tried :wink:

PR #714, opened 2021-04-23
test: Drop workaround for wrong Type= in podman.services

martinpitt:
All supported operating systems have this fixed now.


martinpitt:
The only OS which still has podman 2.x is our current ubuntu-stable image. This gets updated from 20.10 to 21.04 in https://github.com/cockpit-project/bots/pull/1952 , so this may be blocked on that. But then I'll just block this PR and have a reminder to clean up this stuff.

martinpitt:
Some [probably unrelated flakes on packit/f34](http://artifacts.dev.testing-farm.io/e803b803-2cdb-46f3-9507-f9664a3d84af/), retrying for comparison.

marusak (approved):
Danke :)

Issue #710, opened 2021-04-19
Huge gap between container's title and body content


garrett:
There's an enormous gap between the heading of containers and its body content (the list):

![Screenshot_2021-04-19 Podman containers - garrett Rain(2)](https://user-images.githubusercontent.com/10246/115249337-c7ee9e80-a128-11eb-9d99-8eeec600bdf9.png)

Images doesn't have this issue.

marusak:
This is due to using `<Toolbar>` PF4 component.
We use it in Cockpit and we have local overwrites as https://github.com/cockpit-project/cockpit/blob/master/pkg/systemd/terminal.scss#L69 and https://github.com/cockpit-project/cockpit/blob/master/pkg/lib/ct-card.scss#L68

I am looking into this if we can just import it.

marusak:
Aha, we are actually already importing it, just `.ct-card` class is never being defined.

marusak:
Ok, so we need new lib version. It gets a bit better but still not as we want it. It is due to what components we use for showing containers. Right now I am thinking that I don't want to go and fix this up and rather wait for the new design that @garrett is working on. If the design does not change this container items then I would fix it up anyway.

garrett:
The redesign will probably take some time to fix. I would suggest updating the libraries and then add a quick hack to reduce the spacing in the meantime, and then remove the hack when the redesign lands.

marusak:
The gap is smaller now. Is it still too big @garrett ?

![Screenshot from 2021-10-05 11-43-43stilltoobig](https://user-images.githubusercontent.com/12330670/135999874-a1297cd3-dc2e-4847-8322-4627492032c8.png)

garrett:
> The gap is smaller now. Is it still too big?

Yes, still too big of a gap.

It should be more like this (not pixel perfect; rough estimate):

![image](https://user-images.githubusercontent.com/10246/136046671-29494d06-0f77-41ec-8b1c-fc60b1f78c66.png)

jelly:
One issue is [margin top](https://github.com/cockpit-project/cockpit-podman/blob/main/src/Containers.scss#L2) set on container-section for I think seperating pods. The other issue is that `pf-card__body` having a padding of 24px.

PR #719, opened 2021-05-04
Bump cockpit test API and library to 243

martinpitt:
This fixes "DeprecationWarning: Compilation.assets will be frozen in future" as this now uses the updated cockpit-po-plugin.

I visually tested this build, and it looks ok to me (updated PF CSS and such).

I also added a more robust workaround for https://github.com/cockpit-project/bots/issues/1956 (which will solve this naughty).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,14 +183,14 @@ bots:
 # when you start a new project, use the latest release, and update it from time to time
 test/common:
 	flock Makefile sh -ec '\
-	    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 239; \
+	    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 243; \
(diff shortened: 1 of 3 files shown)


marusak:
https://github.com/cockpit-project/cockpit-podman/issues/710#issuecomment-829994505

martinpitt:
I'm afraid this gap isn't much (or any?) better with this version, but at least it's not worse:

image

martinpitt:
Oops, the common components changed selectors a bit, tests need work. [failed tests](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-719-20210504-120102-6e09ff0a-fedora-33/log.html#10-2)

marusak (approved):
Ah right, I wrongly thought it was making it worse.
Thanks!

PR #722, opened 2021-05-18
test: update expected strings in sudo prompt

allisonkarlitskaya:
This got a UI overhaul recently, so just our tests accordingly.

 - [x] https://github.com/cockpit-project/cockpit/pull/15840

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ bots:
 # when you start a new project, use the latest release, and update it from time to time
 test/common:
 	flock Makefile sh -ec '\
-	    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 243; \
+	    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git aed1ad6551f4ac5e5e82a500805111cab8722807; \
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Nice, thank you! (assuming green)

martinpitt (approved):
Danke!

allisonkarlitskaya:
Still not there yet, unfortunately :(

Will test locally to see what's up.

allisonkarlitskaya:
```
    def system_before(self, version):
        try:
            v = self.machine.execute("rpm -q --qf '%{V}' cockpit-system").split(".")
        except subprocess.CalledProcessError:
            return False

        return int(v[0]) < version
```

ugh.

martinpitt:
After the cockpit fix lands, you'll need to bump the test/common checkout in the Makefile to grab that SHA instead of the fixed version there.

martinpitt (approved):

allisonkarlitskaya:
The fail statuses are broken — the test log confirms all tests passed (twice now, even, for both rhel-8-4 and fedora-34/rawhide).  This is an infra issue.

PR #724, opened 2021-06-04
Update terser-webpack-plugin

KKoukiou:
Commit e4a07a moved us back to terser 3.
However when tested against the latest available version from npm:
* dist was not much bigger 
  Master: index.js.gz (283 KiB)
  With new terser: index.js.gz (287 KiB)

* was able to extract LICENSE information with some config options

Unfortunately I don't seem to have full control on the LICENSE file
generated name, there is always gz suffix.

In addition the previous two major releases had a lot of bug fixes, so let's
follow their updates.
https://github.com/webpack-contrib/terser-webpack-plugin/releases

### DIFF
--- a/cockpit-podman.spec.in
+++ b/cockpit-podman.spec.in
@@ -45,7 +45,7 @@ appstream-util validate-relax --nonet %{buildroot}/%{_datadir}/metainfo/*
 
 %files
 %doc README.md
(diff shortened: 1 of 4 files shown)


martinpitt (approved):
Nice, thank you!

PR #732, opened 2021-06-29
tests: wait for image loading to finish before moving forward with downloading new images

KKoukiou:
testDownloadImage was previously quite flaky as when multiple images were
pulled one after another the image list update events would overlap
resulting in a non-deterministic UI updating.

### DIFF
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -247,7 +247,7 @@ class Images extends React.Component {
         }
 
         return (
-            <Card id="containers-images" key="images" className="containers-images">
+            <Card id="containers-images" key="images" className="containers-images" data-updating={this.props.updatingImages}>
                 <CardHeader>
                     <CardTitle><Text component={TextVariants.h2}>{_("Images")}</Text></CardTitle>
                     <CardActions>{getNewImageAction}</CardActions>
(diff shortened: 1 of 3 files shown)


martinpitt:
FWIW, the fix looks okay-ish, although I'd prefer if the tests wouldn't make assumptions about numerical positions of widgets and rather select by container/image name. But looking at the [test log](https://logs.cockpit-project.org/logs/pull-2144-20210629-090520-66c42d6a-rhel-9-0-cockpit-project-cockpit-podman/log.html#5-2) that doesn't actually seem to happen? It's really waiting for `td[data-label=\"Name\"]:contains(\"my-busybox\")")):`, which genuinely never appears. How would a "jumpy page" affect this?

KKoukiou:
So this test also fails on rawhide and according to the stats 86% of the time https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit-podman&test=%2Ftest%2Fcheck-application+TestApplication.testDownloadImage#

KKoukiou:
Asked podman devs for help https://github.com/containers/podman/issues/10812

PR #737, opened 2021-07-15
Port from moment to date-fns

martinpitt:
moment.js has been deprecated for a while now. Move to date-fns'
formatRelative(), which is almost the same as moment's calendar().

Don't use cockpit's timeformat.js, as (1) we already have a centralized
formatting function in c-podman, and (2) timeformat.js currently does
not wrap formatRelative().

----
 - [x] Requires https://github.com/cockpit-project/bots/pull/2209 to get tests green (unrelated to this PR)
 - [x] Requires https://github.com/cockpit-project/cockpit/pull/16106 to unbreak tests (also unrelated)

Screenshots for old, recent, and fresh container with moment:

moment-old
moment-recent
moment-new

The same with date-fns; almost the same, except for the capitalization. But this actually looks better, as we prepend the "Up since":

image

image

image

It also looks fine in German:

image

I didn't make screenshots from the image list, as they look exactly the same.


martinpitt:
The cleanup was alright now, so I pushed the moment → date-fns conversion right away.

martinpitt:
Hmmm, [testDownloadImage failure](https://logs.cockpit-project.org/logs/pull-737-20210715-092023-8e328d38-rhel-9-0/log.html#5) on f34 and rhel-9-0, this smells like https://github.com/cockpit-project/bots/issues/2158 but somehow doesn't match any more? Supposedly fallout from the updated test API, I'll have a closer look.

martinpitt:
I sent https://github.com/cockpit-project/bots/pull/2209 to adjust the naughty, blocking on that.

martinpitt:
A lot of tests now fail like this:
```
1..11
qemu-img create -q -f qcow2 -o backing_file=/work/bots/make-checkout-workdir/test/images/rhel-8-5,backing_fmt=qcow2 /tmp/lwtrlfjp.qcow2
qemu-img create -q -f qcow2 -o backing_file=/work/bots/make-checkout-workdir/test/images/rhel-8-5,backing_fmt=qcow2 /tmp/vhpnjdt0.qcow2
  File "/usr/lib64/python3.9/shutil.py", line 806, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/work/bots/make-checkout-workdir/test/common/pixeldiff.html' -> '/var/tmp/attachments.rimr0jy2/pixeldiff.html'

During handling of the above exception, another exception occurred:

  File "/work/bots/make-checkout-workdir/test/check-application", line 1480, in <module>
    testlib.test_main()
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 1806, in test_main
    attach(os.path.join(TEST_DIR, "common/pixeldiff.html"))
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 99, in attach
    shutil.move(filename, dest)
  File "/usr/lib64/python3.9/shutil.py", line 826, in move
    copy_function(src, real_dst)
  File "/usr/lib64/python3.9/shutil.py", line 435, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib64/python3.9/shutil.py", line 264, in copyfile
    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/work/bots/make-checkout-workdir/test/common/pixeldiff.html'
not ok 7 /work/bots/make-checkout-workdir/test/check-application TestApplication.testLifecycleOperationsUser [ND@1]
```

This is obviously bogus -- why do we *move* pixeldiff.html from the source tree? We can't ever do this twice. We don't even do pixel testing on the OSes other than fedora-34 even.

This isn't new really, that `shutil.move` has existed long before 247. But for now I'll back out the updating of test/common, we don't really need it here. I'll investigate this separately.

martinpitt:
OK, [still happens](https://logs.cockpit-project.org/logs/pull-737-20210715-155154-b9c1ea35-rhel-9-0/log), so I'll put back the test API bump once I fix this. In fact, I'll have to, to pick up the fix.

marusak (approved):
Thank you!

PR #738, opened 2021-07-19
test: test

marusak:
I am having hard time reproducing issues that CI has locally, thus I'll try this PR.
Starting simple.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1322,7 +1322,7 @@ class TestApplication(testlib.MachineCase):
 
         b.click("a:contains('Logs')")
         b.wait_text(".container-logs .xterm-accessibility-tree > div:nth-child(1)", "1")
-        b.wait_text(".container-logs .xterm-accessibility-tree > div:nth-child(6)", "6")
+        b.wait_in_text(".container-logs .xterm-accessibility-tree", "6")
 
         b.click("a:contains('Console')")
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -1322,7 +1322,7 @@ class TestApplication(testlib.MachineCase):
 
         b.click("a:contains('Logs')")
         b.wait_text(".container-logs .xterm-accessibility-tree > div:nth-child(1)", "1")
-        b.wait_text(".container-logs .xterm-accessibility-tree > div:nth-child(6)", "6")
+        b.wait_in_text(".container-logs .xterm-accessibility-tree", "6")
```
martinpitt, line 1325: Any chance the test could log something a little longer/specific than "6"?

martinpitt (approved):
Danke! Tests are still running, but these two look okay (not *thrilled* about the timeouts, but :man_shrugging:

test/check-application
marusak: they surely could. But they log numbers from one onwards. And tests check that we see them from 1 but also that they are being followed. So I could change it for example not to start from one, but from 1000 and then check for 1006 - if that would be better?

martinpitt: I thought about something like `containerNameLog%i`. Either way, fine for a follow-up.

PR #739, opened 2021-07-20
test: Don't check if user or system service failed to start

marusak:
It seems that this can hit system podman as well. Let's see how this behaves.


marusak:
It helped in gating.

martinpitt (approved):
The commit message justification is rather vague. But so is that whole helper function, so -- sure. :grin:

PR #743, opened 2021-08-11
Podman redesign - first pass

KKoukiou:


KKoukiou:
The naughty needs to be adjusted: bots/naughty/rhel-9/2158-podman-api-call-loses-data

KKoukiou:
> Is `Change the color of the running badge to PF's info color` commit replacing #498 ?

#498 is quite obsolete indeed. closed it.

src/Images.jsx
```diff
@@ -127,15 +127,16 @@ class Images extends React.Component {
...
                     onClick={ e => {
                         e.stopPropagation();
                         this.setState({ showRunImageModal: image });
                     } }
-                    aria-label={_("Run image")}
                     isSmall
-                    icon={<PlayIcon />}
-                    data-image={image.Id} />
+                    data-image={image.Id}>
+                {_("Create container")}
```
marusak: I think we need to rename this. `podman create` is not the same as `podman run`. We do the latter in fact.

If I recall correctly, we picked this name but then agreed there will be checkbox in modal 'Immediately start' or so. I don't see that in design. But until that is implemented, I think we need to name this 'Run'.

I'll add comment on the design and also add commit here.

marusak:
Awesome job @KKoukiou !
I've pushed a few commits, the ones starting with `fixup` need to be squashed. Happy to do that. Now running tests.
Do you mind reviewing my commits?

src/Images.jsx
```diff
@@ -151,7 +151,7 @@ class Images extends React.Component {
         let toggleIntermediate = "";
         if (interm) {
             toggleIntermediate = <span className="listing-action">
-                <Button variant="link" onClick={() => this.setState({ intermediateOpened: !intermediateOpened, isExpanded: true })}>
```
KKoukiou: This is needed for the case that there are not images, only intermediate. In this case. clicking show intermediate images, will make the expandable section to appear. I think it makes sense to make it appear expanded.

marusak: ah right, makes sense!

src/Images.jsx
KKoukiou: Makes sense!

marusak (approved):
Awesome, thanks!

PR #744, opened 2021-08-12
build: Use --releasever=rawhide instead of "current + 1"

mvollmer:
"current + 1" might not be rawhide once the next version has been
branched, and "current" isn't the most recent release anymore.


marusak (approved):
Thanks!

PR #745, opened 2021-08-16
test: don't setup ssh key for admin

allisonkarlitskaya:
It's already setup as part of the image these days.


marusak:
This is needed for downstream testing, see failing packit - http://artifacts.dev.testing-farm.io/019c15fa-80fb-4e07-9040-bc58f8f51e51/

allisonkarlitskaya:
Okay: I'll conditionalize it then.  I don't like the error that it throws in the log: it was a red herring when trying to track down another problem yesterday.

marusak (approved):
Thanks!

PR #750, opened 2021-08-30
Redesign images overview

jelly:
Part of the cockpit-podman redesign https://github.com/cockpit-project/cockpit/discussions/16059

### DIFF
--- a/src/ImageDetails.jsx
+++ b/src/ImageDetails.jsx
@@ -9,37 +9,37 @@ const _ = cockpit.gettext;
 
 const ImageDetails = ({ containers, image, showAll }) => {
     return (
-        <DescriptionList isHorizontal className='image-details' columnModifier={{
-            default: '2Col'
-        }}>
+        <DescriptionList className='image-details' isAutoFit>
+            {image.Command !== "" &&
             <DescriptionListGroup>
-                <DescriptionListTerm>{_("ID")}</DescriptionListTerm>
-                <DescriptionListDescription title={image.Id}>{utils.truncate_id(image.Id)}</DescriptionListDescription>
+                <DescriptionListTerm>{_("Command")}</DescriptionListTerm>
+                <DescriptionListDescription>{utils.quote_cmdline(image.Command)}</DescriptionListDescription>
             </DescriptionListGroup>
+            }
+            {image.Entrypoint &&
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Entrypoint")}</DescriptionListTerm>
+                <DescriptionListDescription>{image.Entrypoint.join(" ")}</DescriptionListDescription>
+            </DescriptionListGroup>
+            }
+            {image.RepoTags &&
             <DescriptionListGroup>
                 <DescriptionListTerm>{_("Tags")}</DescriptionListTerm>
                 <DescriptionListDescription>{image.RepoTags ? image.RepoTags.join(" ") : ""}</DescriptionListDescription>
             </DescriptionListGroup>
-            <DescriptionListGroup>
-                <DescriptionListTerm>{_("Entrypoint")}</DescriptionListTerm>
-                <DescriptionListDescription>{image.Entrypoint ? image.Entrypoint.join(" ") : ""}</DescriptionListDescription>
-            </DescriptionListGroup>
-            <DescriptionListGroup>
-                <DescriptionListTerm>{_("Command")}</DescriptionListTerm>
-                <DescriptionListDescription>{image.Command ? utils.quote_cmdline(image.Command) : "" }</DescriptionListDescription>
-            </DescriptionListGroup>
-            <DescriptionListGroup>
-                <DescriptionListTerm>{_("Author")}</DescriptionListTerm>
-                <DescriptionListDescription>{image.Author}</DescriptionListDescription>
-            </DescriptionListGroup>
-            <DescriptionListGroup>
-                <DescriptionListTerm>{_("Ports")}</DescriptionListTerm>
-                <DescriptionListDescription>{image.Ports ? image.Ports.join(', ') : ""}</DescriptionListDescription>
-            </DescriptionListGroup>
+            }
+            {containers &&
             <DescriptionListGroup>
                 <DescriptionListTerm>{_("Used by")}</DescriptionListTerm>
                 <DescriptionListDescription><ImageUsedBy containers={containers} showAll={showAll} /></DescriptionListDescription>
             </DescriptionListGroup>
+            }
+            {image.Ports.length !== 0 &&
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Ports")}</DescriptionListTerm>
+                <DescriptionListDescription>{image.Ports.join(', ')}</DescriptionListDescription>
+            </DescriptionListGroup>
+            }
         </DescriptionList>
     );
 };
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -237,9 +237,10 @@ export class ImageRunModal extends React.Component {
             memoryUnit: 'MiB',
             validationFailed: {},
             volumes: [],
+            runImage: true,
         };
         this.getCreateConfig = this.getCreateConfig.bind(this);
-        this.onRunClicked = this.onRunClicked.bind(this);
+        this.onCreateClicked = this.onCreateClicked.bind(this);
         this.onValueChanged = this.onValueChanged.bind(this);
     }
 
@@ -296,19 +297,24 @@ export class ImageRunModal extends React.Component {
         return createConfig;
     }
 
-    onRunClicked() {
+    onCreateClicked() {
         const createConfig = this.getCreateConfig();
+        const { runImage } = this.state;
 
         client.createContainer(this.state.image.isSystem, createConfig)
                 .then(reply => {
-                    client.postContainer(this.state.image.isSystem, "start", reply.Id, {})
-                            .then(() => this.props.close())
-                            .catch(ex => {
-                                this.setState({
-                                    dialogError: _("Container failed to be started"),
-                                    dialogErrorDetail: cockpit.format("$0: $1", ex.reason, ex.message)
+                    if (runImage) {
+                        client.postContainer(this.state.image.isSystem, "start", reply.Id, {})
+                                .then(() => this.props.close())
+                                .catch(ex => {
+                                    this.setState({
+                                        dialogError: _("Container failed to be started"),
+                                        dialogErrorDetail: cockpit.format("$0: $1", ex.reason, ex.message)
+                                    });
                                 });
-                            });
+                    } else {
+                        this.props.close();
+                    }
                 })
                 .catch(ex => {
                     this.setState({
@@ -417,17 +423,20 @@ export class ImageRunModal extends React.Component {
                                      default={{ envKey: null, envValue: null }}
                                      itemcomponent={ <EnvVar />} />
                 </FormGroup>
+                <FormGroup fieldId='run-image-dialog-start-after-creation' label={_("Start after creation")} hasNoPaddingTop>
+                    <Checkbox isChecked={this.state.runImage} id="start-after-creation" onChange={value => this.onValueChanged('runImage', value)} />
+                </FormGroup>
             </Form>
         );
         return (
             <Modal isOpen
                    position="top" variant="medium"
                    onClose={this.props.close}
-                   title={_("Run image")}
+                   title={_("Create container")}
                    footer={<>
                        {this.state.dialogError && <ErrorNotification errorMessage={this.state.dialogError} errorDetail={this.state.dialogErrorDetail} />}
-                       <Button variant='primary' onClick={this.onRunClicked}>
-                           {_("Run")}
+                       <Button variant='primary' onClick={this.onCreateClicked}>
+                           {_("Create")}
                        </Button>
                        <Button variant='link' className='btn-cancel' onClick={ this.props.close }>
                            {_("Cancel")}
--- a/src/Images.css
+++ b/src/Images.css
@@ -16,3 +16,7 @@
     --pf-c-table__action--PaddingTop: 0.5rem;
     --pf-c-table__action--PaddingBottom: 0.5rem;
 }
+
+.images-grey-text {
+   color: var(--pf-global--Color--200);
+}
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -205,7 +205,7 @@ class Images extends React.Component {
         return (
             <Card id="containers-images" key="images" className="containers-images">
                 <CardHeader>
-                    <CardTitle><Text component={TextVariants.h2}>{_("Images")}</Text></CardTitle>
+                    <CardTitle><Text component={TextVariants.h3}>{_("Images")}</Text></CardTitle>
                     <CardActions>{getNewImageAction}</CardActions>
                 </CardHeader>
                 <CardBody>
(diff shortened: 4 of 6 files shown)


KKoukiou (approved):
The change is straightforward I just have one comment. The same logic exist for VMs as well. In that case the dialog has a checkbox, 'Immediately start VM' instead of 'Start VM after creation' which is the equivalent of the text chosen here. I feel it's nice to keep the texts same, what do you think @garrett? 

No vetoing - I just want to make it a conscious decision.

jelly:
> The change is straightforward I just have one comment. The same logic exist for VMs as well. In that case the dialog has a checkbox, 'Immediately start VM' instead of 'Start VM after creation' which is the equivalent of the text chosen here. I feel it's nice to keep the texts same, what do you think @garrett?
> 
> No vetoing - I just want to make it a conscious decision.
> 
> Screen Shot 2021-08-31 at 09 29 24

There was a discussion about this [here](https://github.com/cockpit-project/cockpit/discussions/16059#discussioncomment-1194000), so that's how I came up with the text.

I am however not sure if it should be the default :thinking:

src/ImageRunModal.jsx
```diff
@@ -237,9 +237,10 @@ export class ImageRunModal extends React.Component {
             memoryUnit: 'MiB',
```
jelly, line 237: Should I rename the whole dialog/file from `run` to `create`?

src/ImageRunModal.jsx
```diff
@@ -417,17 +423,20 @@ export class ImageRunModal extends React.Component {
...
+                    <Checkbox isChecked={this.state.runImage} id="run" onChange={value => this.onValueChanged('runImage', value)} />
+                </FormGroup>
             </Form>
         );
         return (
             <Modal isOpen
                    position="top" variant="medium"
                    onClose={this.props.close}
-                   title={_("Run image")}
+                   title={_("Create image")}
```
marusak, line 435: Not create "image" but "container".

src/ImageRunModal.jsx
marusak: You might, not necessary though

marusak:
No option about "Immediately start" vs "Start after creation".  We can stay consistent with other pages for sure. Or just try to keep the string as short as possible.

jelly:
The last test failures are due to the `tags` missing from the `image details` overview, is this is something missing from the redesign @marusak ?

jelly:
Updated images overview page

image

Updated "create container" modal

image

marusak:
Notice, that when one image has more than one tag, cmdline shows it as two separate lines. (the fedora ones)

Screenshot from 2021-09-01 14-13-01secpmd

You show it as one line with two names. That seems odd. It should be two lines as well.

Screenshot from 2021-09-01 14-11-38

marusak:
Also the common question - do we want MiB or MB units? :) I would say, that MB are better choice in this situation + is consistent with podman cmd.

martinpitt:
> I would say, that MB are better choice in this situation

:+1 -- IMHO, base-2 makes very little sense these days, pretty much only on RAM size.

jelly:
> Notice, that when one image has more than one tag, cmdline shows it as two separate lines. (the fedora ones)
> 
> Screenshot from 2021-09-01 14-13-01secpmd
> 
> You show it as one line with two names. That seems odd. It should be two lines as well.
> 
> Screenshot from 2021-09-01 14-11-38

We can change it to be consistent to with podman output, I do wonder if the column name is then maybe confusing as it says `Image`.

src/Images.jsx
```diff
@@ -60,14 +60,31 @@ class Images extends React.Component {
...
+        }
+    }
+
     renderRow(image) {
         const tabs = [];
+        const { title: usedByText, count: usedByCount } = this.getUsedByText(image);
 
         const columns = [
             { title: image.RepoTags ? image.RepoTags[0] : "<none>:<none>", header: true },
+            { title: image.isSystem ? _("system") : <><span className="images-grey-text">user: </span>{this.props.user}</> },
```
KKoukiou, line 112: This needs to be translated. _("user: ")

src/Images.css
```diff
@@ -16,3 +16,11 @@
     --pf-c-table__action--PaddingTop: 0.5rem;
     --pf-c-table__action--PaddingBottom: 0.5rem;
 }
+
+#containers-images .pf-c-table.pf-m-compact .pf-c-table__toggle {
+    padding-left: 0px;
+}
+
+.images-grey-text {
+   color: var(--pf-global--Color--200);
```
KKoukiou, line 25: Whoa :) Thanks for using PF colors!

src/Images.jsx
```diff
@@ -60,14 +60,31 @@ class Images extends React.Component {
...
     }
 
+    getUsedByText(image) {
+        const { imageContainerList } = this.props;
+        if (imageContainerList === null) {
+            return { title: _("unused"), count: 0 };
+        }
+        const containers = imageContainerList[image.Id + image.isSystem.toString()];
+        if (containers !== undefined) {
+            const fmt = containers.length === 1 ? _("$0 container") : _("$0 containers");
```
KKoukiou, line 69: Better done like this:
const fmt = cockpit.format(cockpit.ngettext("$0 container", "$0 containers", containers.length), containers.length)

src/ImageDetails.jsx
```diff
@@ -9,37 +9,31 @@ const _ = cockpit.gettext;
...
+        <DescriptionList className='image-details' isAutoFit>
+            {image.Command !== "" &&
             <DescriptionListGroup>
-                <DescriptionListTerm>{_("ID")}</DescriptionListTerm>
-                <DescriptionListDescription title={image.Id}>{utils.truncate_id(image.Id)}</DescriptionListDescription>
-            </DescriptionListGroup>
-            <DescriptionListGroup>
-                <DescriptionListTerm>{_("Tags")}</DescriptionListTerm>
-                <DescriptionListDescription>{image.RepoTags ? image.RepoTags.join(" ") : ""}</DescriptionListDescription>
+                <DescriptionListTerm>{_("Default command")}</DescriptionListTerm>
```
KKoukiou, line 15: I know you copied it from the design document - but I feel like Default is not needed here. I would just keep it Command. What do you think @marusak ?

src/ImageRunModal.jsx
```diff
@@ -417,17 +423,20 @@ export class ImageRunModal extends React.Component {
                                      default={{ envKey: null, envValue: null }}
                                      itemcomponent={ <EnvVar />} />
                 </FormGroup>
+                <FormGroup fieldId='run-image-dialog-run' label={_("Start after creation")}>
```
KKoukiou, line 426: The fieldId has to be the same with the id of the field that is nested in the FormGroup. So in this case 'run'.
Please use more complicated ids however, not 'run' as in one page ids should be unique. and single word string tend to become not unique :)

src/Images.jsx
```diff
@@ -202,10 +232,27 @@ class Images extends React.Component {
             </>
         );
 
+        const imageStats = this.calculateStats();
+        const imageTitleStats = (
+            <div className="images-title-stats">
+                <Text component={TextVariants.h5}>
+                    {cockpit.format(_("$0 images total $1"), imageStats.imagesTotal, cockpit.format_bytes(imageStats.imagesSize))}
```
KKoukiou, line 238: The design doc has a comma after total.

src/Images.jsx
```diff
@@ -202,10 +232,27 @@ class Images extends React.Component {
...
+                </Text>
+                {imageStats.unusedTotal !== 0 &&
+                <Text component={TextVariants.h5}>
+                    {cockpit.format(_("$0 unused images $1"), imageStats.unusedTotal, cockpit.format_bytes(imageStats.unusedSize))}
```
KKoukiou, line 242: The design doc has a comma after 'images'

KKoukiou (changes requested):
First quick pass for today - I will do a real one tomorrow :) Pretty amazing work! Thanks for using PF4 components and keeping the commit simple to review.

src/Images.css
```diff
@@ -16,3 +16,24 @@
...
+}
+
+.images-title {
```
KKoukiou, line 28: How about wrapping the title with the stats in a https://www.patternfly.org/v4/layouts/flex/#basic. 

That wil add autopadding as well.

KKoukiou (changes requested):

src/Images.jsx
KKoukiou: Also this breaks in small screens.
Screen Shot 2021-09-01 at 17 45 46

KKoukiou (changes requested):

src/Images.jsx
martinpitt: Please avoid translating trailing (or leading) spaces -- it is pretty much garanteed to get messed up during the translation process. Use CSS or HTML structure to do spacing.

marusak:
> We can change it to be consistent to with podman output, I do wonder if the column name is then maybe confusing as it says Image.

We don't want column 'tag' so repository would not do. It is repo + image name + tag. I think `Image` is then fine, it is obvious whats in the column anyway :)

src/ImageDetails.jsx
marusak: https://docs.docker.com/engine/reference/builder/#cmd

:man_shrugging: It is kind of default command, but also not super important to spell out "default"...

src/Images.jsx
jelly: Oh nice! I should have checked https://cockpit-project.org/guide/latest/cockpit-locale.html I guess!

src/Images.jsx
jelly: Added translations without a trailing space, for the small screens I've put the <span> with the {this.props.user} in a <div> so it doesn't break as it does now. I'm not sure if this is the best solution.

src/ImageRunModal.jsx
jelly: I'll leave it as todo for later, as I feel that the PR is already quite big.

src/Images.css
jelly: Thanks, that works nicely and that means I can remove all the CSS yay!

src/Images.jsx
```diff
@@ -202,10 +232,29 @@ class Images extends React.Component {
             </>
         );
 
+        const imageStats = this.calculateStats();
+        const imageTitleStats = (
+            <>
+                <Text component={TextVariants.h5}>
+                    {cockpit.format(_("$0 images total, $1"), imageStats.imagesTotal, cockpit.format_bytes(imageStats.imagesSize))}
```
KKoukiou, line 238: This would need ngettext - in case there is one image.

src/Images.jsx
```diff
@@ -202,10 +232,29 @@ class Images extends React.Component {
...
+                </Text>
+                {imageStats.unusedTotal !== 0 &&
+                <Text component={TextVariants.h5}>
+                    {cockpit.format(_("$0 unused images, $1"), imageStats.unusedTotal, cockpit.format_bytes(imageStats.unusedSize))}
```
KKoukiou, line 242: This would need ngettext - in case there is one image.

KKoukiou (changes requested):

src/Images.css
```diff
@@ -16,3 +16,11 @@
     --pf-c-table__action--PaddingTop: 0.5rem;
     --pf-c-table__action--PaddingBottom: 0.5rem;
 }
+
+#containers-images .pf-c-table.pf-m-compact .pf-c-table__toggle {
```
KKoukiou, line 20: This will create inconsistent left padding between the images and containers toggles in lists. 
Could you please make it apply also for Containers and move it to podman.scss file?

KKoukiou (changes requested):

src/Images.css
KKoukiou: Something like this I mean:

``` diff
diff --git src/Images.css src/Images.css
index 3caed167..f080943f 100644
--- src/Images.css
+++ src/Images.css
@@ -17,10 +17,6 @@
     --pf-c-table__action--PaddingBottom: 0.5rem;
 }
 
-#containers-images .pf-c-table.pf-m-compact .pf-c-table__toggle {
-    padding-left: 0px;
-}
-
 .images-grey-text {
    color: var(--pf-global--Color--200);
 }
diff --git src/podman.scss src/podman.scss
index 43aa363f..d1ad85af 100644
--- src/podman.scss
+++ src/podman.scss
@@ -8,6 +8,12 @@
     @extend .ct-card;
 }
 
+#containers-images, #containers-containers {
+    .pf-c-table.pf-m-compact .pf-c-table__toggle {
+        padding-left: 0px;
+    }
+}
+
 .container-block {
     display: flex;
     flex-direction: column;

```

KKoukiou: And last thing - because we are all not really good with CSS I like adding comment so that we don't get lost. 
In this case, decrease default padding for the toggle button in list.

jelly:
> Also the common question - do we want MiB or MB units? :) I would say, that MB are better choice in this situation + is consistent with podman cmd.

switched to MB everywhere.

KKoukiou (approved):

src/ImageRunModal.jsx
```diff
@@ -417,17 +423,20 @@ export class ImageRunModal extends React.Component {
                                      default={{ envKey: null, envValue: null }}
                                      itemcomponent={ <EnvVar />} />
                 </FormGroup>
+                <FormGroup fieldId='run-image-dialog-start-after-creation' label={_("Start after creation")}>
```
marusak, line 426: Screenshot from 2021-09-03 09-10-15

It seems misaligned with the label.  If you add `hasNoPaddingTop` prop to this `FormGroup` it should work as expected.

test/check-application
```diff
@@ -1334,7 +1334,7 @@ class TestApplication(testlib.MachineCase):
         b.wait_visible('#containers-images td[data-label="Image"]:contains("busybox:latest")')
         b.click('#containers-containers tbody tr:contains("busybox:latest") td.pf-c-table__toggle button')
         b.click('#containers-images tbody tr:contains("busybox:latest") .ct-container-create')
-        b.wait_visible('div.pf-c-modal-box header:contains("Run image")')
+        b.wait_visible('div.pf-c-modal-box header:contains("Create container")')
```
marusak, line 1337: Can you also test that new checkbox - that you will unset it and make sure it does not start?

marusak (changes requested):
The new checkbox is misaligned and a test would be nice :) Otherwise looks great!

test/check-application
jelly: Added a test.

test/check-application
```diff
@@ -1425,6 +1425,24 @@ class TestApplication(testlib.MachineCase):
...
+
+        b.wait_in_text("#run-image-dialog-image", "busybox:latest")
+        b.set_input_text("#run-image-dialog-name", container_name)
+
+        # Uncheck start container after creation
+        b.click("#start-after-creation")
+        b.click('.pf-c-modal-box__footer button:contains("Create")')
+        b.wait_not_present("div.pf-c-modal-box")
+
+        b.wait_not_in_text("#containers-containers", container_name)
```
marusak, line 1442: This is race-y. It waits for container not to be in the list - which might be just true by default (tests are fast so it might not yet be present), then changes to all containers and then waits for it to be present. It unfortunately does not say anything whether the container is running or not.

It should be safe to just switch to `all` and wait for the container status to be `created`.

jelly: Looking at the test of the tests, it might be better to use `waitContainer`

marusak (approved):
Awesome work, thank you!

garrett:
This is awesome to see! :+1: :partying_face: 

Minor nitpick: We should never have checkboxes without labels. They're inaccessible and the click target is much smaller (as the label itself is part of the click target when done in a proper, accessible way) and that benefits everyone, even people with full sight using a mouse.

Also, I agree we should be consistent where possible with language.

PR #753, opened 2021-09-02
Container card redesign

KKoukiou:


KKoukiou:
Repushed to resolve merge conflicts.

src/Containers.jsx
```diff
@@ -27,112 +28,262 @@ import { PodActions } from './PodActions.jsx';
...
+        this.onWindowResize = this.onWindowResize.bind(this);
+
+        this.cardRef = React.createRef();
 
-    restoreContainer(container) {
-        this.setState({
-            containerWillRestore: container,
-            selectContainerRestoreModal: true,
-        });
+        window.addEventListener('resize', this.onWindowResize);
```
jelly, line 274: should there be debouncing for resizing events?

jelly (approved):
Looks good! I added some minor unrelated comments.

src/ContainerDetails.jsx
```diff
@@ -14,61 +14,68 @@ const render_container_state = (container) => {
...
-);
+const ContainerDetails = ({ container, containerDetail }) => {
+    const ports = render_container_published_ports(container.Ports);
+
+    return (
+        <Flex spaceItems={{ modifier: 'spaceItemsXl' }}>
+            <DescriptionList className='container-details-basic'>
+                <DescriptionListGroup>
+                    <DescriptionListTerm>{_("ID")}</DescriptionListTerm>
+                    <DescriptionListDescription>{utils.truncate_id(container.Id)}</DescriptionListDescription>
```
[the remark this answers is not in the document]
KKoukiou: Correct - I feel it's redundant. The truncated ID should be unique.

src/Containers.jsx
KKoukiou: That should be triggered only once on page load as it's in the component constructor.

src/ContainerDetails.jsx
```diff
@@ -56,10 +56,6 @@ const ContainerDetails = ({ container, containerDetail }) => (
             <DescriptionListTerm>{_("IP address")}</DescriptionListTerm>
             <DescriptionListDescription>{containerDetail ? containerDetail.NetworkSettings.IPAddress : ""}</DescriptionListDescription>
         </DescriptionListGroup>
-        <DescriptionListGroup>
```
marusak, line 59: I am not fan of this commit. If we have that info, why not show it? We also have space for it, it does not cost us anything...

marusak: Oh, should specify that I mean the very first one dropping prefix length

KKoukiou: It will just be 0 most of the cases and it will always shown. How someone should use that anyway?

src/ContainerDetails.jsx
```diff
@@ -27,44 +27,50 @@ const render_container_published_ports = (ports) => {
...
+                <DescriptionListDescription>{containerDetail ? containerDetail.NetworkSettings.MacAddress : ""}</DescriptionListDescription>
+            </DescriptionListGroup>
+        </DescriptionList>
+        <DescriptionList className='container-details-state'>
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Created")}</DescriptionListTerm>
+                <DescriptionListDescription>{utils.localize_time(Date.parse(container.Created) / 1000)}</DescriptionListDescription>
+            </DescriptionListGroup>
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("State")}</DescriptionListTerm>
```
marusak: Do we need state here as well if we have it  the overview already?
Also, how come it shows something different here than in overview?
Screenshot from 2021-09-08 09-33-10huh

marusak: edit - for running it makes sense, as it shows `Running` and in details `Up since <date>`

src/ContainerDetails.jsx
```diff
@@ -27,44 +27,50 @@ const render_container_published_ports = (ports) => {
...
-        <DescriptionListGroup>
-            <DescriptionListTerm>{_("Gateway")}</DescriptionListTerm>
-            <DescriptionListDescription>{containerDetail ? containerDetail.NetworkSettings.Gateway : ""}</DescriptionListDescription>
-        </DescriptionListGroup>
-        <DescriptionListGroup>
-            <DescriptionListTerm>{_("MAC address")}</DescriptionListTerm>
-            <DescriptionListDescription>{containerDetail ? containerDetail.NetworkSettings.MacAddress : ""}</DescriptionListDescription>
-        </DescriptionListGroup>
-    </DescriptionList>
+    <Flex spaceItems={{ modifier: 'spaceItemsXl' }}>
```
marusak: I must say it seems a bit odd to not fill up the card but have items in columns:
Screenshot from 2021-09-08 09-33-10huh

Shouldn;t we put those items on one line if there is space? I know this follows up the design -  @garrett

src/ContainerDetails.jsx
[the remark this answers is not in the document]
marusak: It indeed is unique and I don't think there is ever any use for `show full id`.

src/podman.scss
```diff
@@ -109,3 +109,7 @@ input.pf-c-form-control[type=number] {
 #containers-images .ct-listing-panel-head {
     display: none;
 }
+
+.ct-grey-text {
```
marusak, line 113: Don't we declare this for images already?

src/Containers.jsx
```diff
@@ -28,112 +28,266 @@ import { PodActions } from './PodActions.jsx';
...
-    componentWillUnmount() {
-        window.removeEventListener('resize', this.onWindowResize);
-    }
-
-    deleteContainer(container, event) {
+const ContainerActions = ({ container, onAddNotification, version }) => {
+    const [removeErrorModal, setRemoveErrorModal] = useState(false);
+    const [deleteModal, setDeleteModal] = useState(false);
+    const [checkpointInProgress, setCheckpointInProgress] = useState(false);
+    const [checkpointModal, setCheckpointModal] = useState(false);
```
marusak, line 35: checkpoint and restore should be separate group in the dropdown as well (separated by line)
Screenshot from 2021-09-08 09-49-14

marusak:
`testDownload` fails on `b.click(".pf-c-modal-box button:contains('Force delete')")` but I believe this is flaky test. But maybe it would be worth taking a look anyway :) (outside of this PR)

src/ContainerDetails.jsx
marusak: :man_shrugging: ok, then we can drop it

src/ContainerDetails.jsx
garrett: That doesn't follow the design though... the design has it fill the space.

Here's the design:
podman-redesign-2021

Here's how it would wrap when space gets limited:

podman-redesign-2021-details-wrapping

The idea is that there are groups of similar information and both the contents of the group and the group itself wraps when necessary.

KKoukiou: @garrett the reason I did not follow the design here is that the design looks like it uses a justify-content: space-between

This works nicely when all info are present - but when the whole networking info is missing ( which is very often the case ) then there is a ton of unused in-between space.

garrett: It can use nested PF4 flex elements, each with gaps. I agree about space-between when info isn't there.

Anyway, it's not a big deal, and we can iterate to optimize it later, in another PR.

src/Containers.jsx
KKoukiou: Fixed.

src/ContainerDetails.jsx
KKoukiou: I am not sure how it should behave differently than what we have now. I will let it to you @garrett to work on this when you have time.

marusak (approved):
Super nice one, thank you!

KKoukiou:
The packit f35 has finished, with one failure - which is flake. I am not sure why it's not updated here.

garrett:
Oh, I was working on a few fixes on this branch. I guess I'll send it as a follow-up PR. :wink:

PR #757, opened 2021-09-09
Redesign create container modal

jelly:
First step of converting the create container modal to the new re-design, converting the form into two tabs, details and integration. The source/image/tag fields have been redesign fields have been left out for now

Some minor tweaks are applied to set the default value of CPU shares, tweak width of input fields (so they don't take up the full width).

@garrett  we now show as if the default memory limit is 512 MiB, while the default is to take up all available memory.
@garrett you mentioned the "Start after creation" checkbox has no label, how can we add a label without duplicating text?

## Details

@KKoukiou is there an issue way to adjust the name label to take up less width so it aligns more with the [design](https://github.com/cockpit-project/cockpit/discussions/16059).

image

## Integration
image

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {
     Button, Checkbox, Form, FormGroup, FormSelect, FormSelectOption,
-    InputGroup, Modal, TextInput
+    InputGroup, Modal, TextInput, Tabs, Tab, TabTitleText, Flex, FlexItem,
 } from '@patternfly/react-core';
 import { CloseIcon, PlusIcon } from '@patternfly/react-icons';
 import * as dockerNames from 'docker-names';
@@ -12,6 +12,8 @@ import { FileAutoComplete } from 'cockpit-components-file-autocomplete.jsx';
 import * as utils from './util.js';
 import * as client from './client.js';
 import cockpit from 'cockpit';
+
+import "./ImageRunModal.scss";
 
 const _ = cockpit.gettext;
 
@@ -238,6 +240,7 @@ export class ImageRunModal extends React.Component {
             validationFailed: {},
             volumes: [],
             runImage: true,
+            activeTabKey: 0,
         };
         this.getCreateConfig = this.getCreateConfig.bind(this);
         this.onCreateClicked = this.onCreateClicked.bind(this);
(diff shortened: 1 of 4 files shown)


src/ImageRunModal.scss
```diff
@@ -0,0 +1,18 @@
...
+#run-image-dialog-memory, #run-image-cpu-priority {
+     max-width: 15%;
+}
+
+#memory-unit-select {
+     max-width: 20%;
+}
+
+#run-image-dialog-memory-limit-checkbox, #run-image-dialog-cpu-priority-checkbox {
+     margin-top: 12px;
```
jelly, line 13: This is suboptimal as the patternfly spacing is different. We come up with a better solution for this.

garrett: For spacing, we should try to stick to variables like var(--pf-global--spacer--md), but is this trying to fix alignment or something? (In that case, we should just align it.)

I'll take a look.

garrett:
Are we handlign the design of the "Integration" tab in this PR too? Or is that a follow-up?

Basically this layout, using top labels and sizing the input widths (possibly in something that uses a grid, so we can specify the approximate sizes and let it figure out the rest with an `fr` and such... or something using flex... whatever the PF4 widget would use):

image

src/ImageRunModal.jsx
```diff
@@ -328,104 +331,124 @@ export class ImageRunModal extends React.Component {
...
-                </FormGroup>
-
-                <FormGroup fieldId='run-image-dialog-memory' label={_("Memory limit")}>
-                    <InputGroup className="ct-input-group-spacer-sm" id="run-image-dialog-memory-limit">
-                        <Checkbox id="run-image-dialog-memory-limit-checkbox"
+                        </FormGroup>
+                    </FlexItem>
+                </Flex>
+                <Tabs activeKey={activeTabKey} onSelect={this.handleTabClick}>
+                    <Tab eventKey={0} title={<TabTitleText>Details</TabTitleText>} className="pf-l-grid pf-m-gutter">
```
KKoukiou, line 360: Needs to be translated!

src/ImageRunModal.jsx
```diff
@@ -328,104 +331,124 @@ export class ImageRunModal extends React.Component {
...
+                        </FormGroup>
+
+                        <FormGroup fieldId="run=image-dialog-tty">
+                            <Checkbox id="run-image-dialog-tty"
+                              isChecked={this.state.hasTTY}
+                              label={_("With terminal")}
+                              onChange={checked => this.onValueChanged('hasTTY', checked)} />
+                        </FormGroup>
+
+                        <FormGroup fieldId='run-image-dialog-memory' label={_("Memory limit")}>
```
KKoukiou, line 379: fieldId needs to match nested field Id. The reason is that it uses it for the aria-labelled-by accesibility property.

KKoukiou: In this case you can use the input group's id.

src/ImageRunModal.jsx
```diff
@@ -328,104 +331,124 @@ export class ImageRunModal extends React.Component {
...
+                                <Checkbox id="run-image-dialog-memory-limit-checkbox"
                                   isChecked={this.state.memoryConfigure}
                                   onChange={checked => this.onValueChanged('memoryConfigure', checked)} />
-                        <TextInput type='number'
+                                <TextInput type='number'
                                    value={dialogValues.memory}
                                    id="run-image-dialog-memory"
                                    step={1}
                                    min={0}
                                    isReadOnly={!this.state.memoryConfigure}
```
KKoukiou, line 389: This needs to be discussed with @garrett and adjusted. When disabled the memory limit is unset, having this field disabled is misleading.

src/ImageRunModal.jsx
```diff
@@ -328,104 +331,124 @@ export class ImageRunModal extends React.Component {
...
-                        <Checkbox id="run-image-dialog-cpu-priority-checkbox"
+                                    <FormSelectOption value={units.KiB.name} key={units.KiB.name} label={_("KiB")} />
+                                    <FormSelectOption value={units.MiB.name} key={units.MiB.name} label={_("MiB")} />
+                                    <FormSelectOption value={units.GiB.name} key={units.GiB.name} label={_("GiB")} />
+                                </FormSelect>
+                            </InputGroup>
+                        </FormGroup>
+
+                        { this.state.image.isSystem &&
+                            <FormGroup fieldId='run-image-cpu-priority' label={_("CPU shares")}>
```
KKoukiou, line 404: Same for fieldId.

src/ImageRunModal.jsx
```diff
@@ -328,104 +331,124 @@ export class ImageRunModal extends React.Component {
...
                                   onChange={checked => this.onValueChanged('cpuSharesConfigure', checked)} />
-                        <TextInput type='number'
+                                    <TextInput type='number'
                                    id="run-image-cpu-priority"
                                    value={dialogValues.cpuShares}
                                    step={1}
                                    min={2}
+                                   max={262144}
                                    isReadOnly={!this.state.cpuSharesConfigure}
                                    onChange={value => this.onValueChanged('cpuShares', value === "" ? "" : parseInt(value))} />
```
KKoukiou, line 416: Is the check value == '' needed now?

src/ImageRunModal.jsx
```diff
@@ -328,104 +331,124 @@ export class ImageRunModal extends React.Component {
...
+                                </InputGroup>
+                            </FormGroup>}
 
-                <FormGroup fieldId="run=image-dialog-tty">
-                    <Checkbox id="run-image-dialog-tty"
-                              isChecked={this.state.hasTTY}
-                              label={_("With terminal")}
-                              onChange={checked => this.onValueChanged('hasTTY', checked)} />
-                </FormGroup>
+                        <FormGroup fieldId='run-image-dialog-start-after-creation' label={_("Start after creation")} hasNoPaddingTop>
```
KKoukiou, line 420: fieldId.

src/ImageRunModal.scss
KKoukiou: Use rems not percentage. The percentage works horribly in mobile mode, where you want the element to all of the space (or for sure more)

src/ImageRunModal.scss
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.pf-c-form__group-label {
```
KKoukiou, line 16: When you override pf rules please use a specific className in addition to the PF4 class. In this way you can prevent this from getting applied to all other random places.

KKoukiou (changes requested):

src/ImageRunModal.scss
```diff
@@ -0,0 +1,18 @@
+@import '@patternfly/patternfly/layouts/Grid/grid.css';
+
+// Limit the max-width more the override in podman.scss which is 40%
+#run-image-dialog-memory, #run-image-cpu-priority {
+     max-width: 15%;
+}
+
+#memory-unit-select {
+     max-width: 20%;
+}
```
garrett, line 10: Generally, it's a bad idea to use IDs in CSS. You want to use classes for CSS, IDs for JavaScript, and not mix them.

This is due to specificity in CSS and speed in targeting IDs in JavaScript. There are a few exceptions for one-offs, like this, sometimes. But it's usually still better to duplicate the ID's value as the name of the class (for the resulting HTML... in our case, the elements in JSX which eventually become HTML) and use the class in the CSS.

src/ImageRunModal.scss
```diff
@@ -0,0 +1,18 @@
+@import '@patternfly/patternfly/layouts/Grid/grid.css';
+
+// Limit the max-width more the override in podman.scss which is 40%
+#run-image-dialog-memory, #run-image-cpu-priority {
```
garrett, line 4: It's better if we could use a common class for both of these, like`.run-image-limit` or something.

(See below for ID vs. class.)

src/ImageRunModal.scss
```diff
@@ -0,0 +1,18 @@
...
+     max-width: 15%;
+}
+
+#memory-unit-select {
+     max-width: 20%;
+}
+
+#run-image-dialog-memory-limit-checkbox, #run-image-dialog-cpu-priority-checkbox {
```
garrett, line 12: ...Something like `.dialog-run-checkbox` and apply the class to both elements as a `className` in the JSX would be better if we do have to special case this at the checkbox level.

src/ImageRunModal.scss
garrett: :100: 

This, exactly. I totally agree with @KKoukiou. 

We've hit bugs before due to this. It's often fine to do `.my-item .pf-c-thing`, if you have complete control of the item and it's specific to your page. But if you're doing a 1-off in your page, it's usually better to have a custom class. We usually use `.ct-` as a prefix for reusable components specific to Cockpit and I like to use something unique to the page for other things, just to make it clear what the item applies to. 

But if you can, use a PF4 utility class when it's what you want.

garrett:
It looks like I can click on and focus the memory field, as it's not _really_ disabled, but set to readonly instead.

image

(Note the focus ring.)

At first, I was thinking that you shouldn't be able to click on it, but actually, I think the right pattern is to enable the memory limit checkbox and enable the widgets, so it would treat the entire thing like a toggle on (not a toggle off, of course) if you click on the number input or the dropdown. In this case, it's clear you want to set the value.

(However, you shouldn't be able to tab to them if they're disabled.)

The memory limit checkbox doesn't have a label. It should. Make the "Memory limit"'s "for" point to the checkbox, not the input. And make the input's aria-decribedby point to the label's id.

Did we decide what to do about "Start after creation" yet? Are we going to synchronize the text of the label with the Machines page's option? (I saw something about this mentioned somewhere and don't know if there was a conclusion.)

src/ImageRunModal.scss
garrett: Note, since this is in the ImageRunModal.scss, you might assume it's only going to apply to the image run modal. It won't. It will apply to the whole page where it's imported. (Other people have made that wrong assumption before.) This is why it's important to use custom class names or to scope the CSS to a particular part of the page using CSS itself.

garrett:
I removed the checkbox px hack and use what's basically a line of CSS via a PatternFly utility instead, in 7ac80de.

garrett:
I pushed some fixes, but they're not working becuase someone was extremely specific with other CSS in the page apparently. I don't have time to debug. The issue is with:

`.ct-input-group-spacer-sm.pf-c-input-group > .pf-c-select, .ct-input-group-spacer-sm.pf-c-input-group .pf-c-form-control:not(.pf-c-select__toggle-typeahead)`

Someone's using way too many classes there to not be scoped to a particular element.

KKoukiou:
> I pushed some fixes, but they're not working becuase someone was extremely specific with other CSS in the page apparently. I don't have time to debug. The issue is with:
> 
> `.ct-input-group-spacer-sm.pf-c-input-group > .pf-c-select, .ct-input-group-spacer-sm.pf-c-input-group .pf-c-form-control:not(.pf-c-select__toggle-typeahead)`
> 
> Someone's using way too many classes there to not be scoped to a particular element.

I bet I wrote that. I can make sure you new commit applies tomorrow.

jelly:
> Are we handling the design of the "Integration" tab in this PR too? Or is that a follow-up?
> 
> Basically this layout, using top labels and sizing the input widths (possibly in something that uses a grid, so we can specify the approximate sizes and let it figure out the rest with an `fr` and such... or something using flex... whatever the PF4 widget would use):
> 
> image

I was planning on doing this in a new PR, I have a local branch prepared which uses the existing InputGroup, so it seems we can do some optimisations there and use a grid/flex layour

jelly:
> I removed the checkbox px hack and use what's basically a line of CSS via a PatternFly utility instead, in [7ac80de](https://github.com/cockpit-project/cockpit-podman/commit/7ac80de154e0affd17bb716bb213cc8e4d948c94).

Thanks! I squashed it into the re-design.

jelly:
> > I pushed some fixes, but they're not working becuase someone was extremely specific with other CSS in the page apparently. I don't have time to debug. The issue is with:
> > `.ct-input-group-spacer-sm.pf-c-input-group > .pf-c-select, .ct-input-group-spacer-sm.pf-c-input-group .pf-c-form-control:not(.pf-c-select__toggle-typeahead)`
> > Someone's using way too many classes there to not be scoped to a particular element.
> 
> I bet I wrote that. I can make sure you new commit applies tomorrow.

I think we can remove this after, I've done the integration tab. The max-width: 40% can go away when the inputs are on the same line and can be arranged using flex/grid layout.

src/ImageRunModal.scss
jelly: This is resolved in a new push. Percentages are replaced with `rem` and the checkbox margin is replaced with a patternfly css class.

src/ImageRunModal.jsx
jelly: Fixed everywhere it applies.

src/ImageRunModal.jsx
jelly: Good catch, dropped and simplified.

src/ImageRunModal.scss
jelly: Thanks, I applied your suggestions and changed it to classes everywhere.

src/ImageRunModal.jsx
jelly: I agree that it's misleading but I am not sure what the new UX should be, should we make an improvement issue for this?

src/ImageRunModal.scss
jelly: It was actually not required by anything and dropped.

src/ImageRunModal.jsx
```diff
@@ -328,104 +331,128 @@ export class ImageRunModal extends React.Component {
...
                                      formclass='publish-port-form'
                                      onChange={value => this.onValueChanged('publish', value)}
                                      default={{ IP: null, containerPort: null, hostPort: null, protocol: 'tcp' }}
                                      itemcomponent={ <PublishPort />} />
-                </FormGroup>
+                        </FormGroup>
 
-                <FormGroup fieldId='run-image-dialog-env' label={_("Volumes")}>
-                    <DynamicListForm id='run-image-dialog-volume'
+                        <FormGroup fieldId='run-image-dialog-env' label={_("Volumes")}>
```
KKoukiou, line 438: This fieldId is wrong . I know you just moved it - but if you repush wanna add a fix up commit?

src/ImageRunModal.scss
```diff
@@ -0,0 +1,19 @@
+@import '@patternfly/patternfly/layouts/Grid/grid.css';
+// Note: PatternFly's "Flex" utilities also apply to grid alignment
+@import "@patternfly/patternfly/utilities/Flex/flex.css";
```
KKoukiou, line 3: This is not needed. You use directly the React component. when you import the React component the CSS of it is autoimported.

KKoukiou (approved):

src/ImageRunModal.jsx
jelly: Fixed

src/ImageRunModal.scss
jelly: Fixed

KKoukiou (approved):

PR #758, opened 2021-09-09
Container card redesign follow-ups

garrett:
I've kept the commits separate, for easier reviewing. Between the git commit line and the comments in-file, I hope they're self-explanatory enough.

### DIFF
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -18,6 +18,12 @@
     .ct-listing-panel-body > .pf-l-flex {
         gap: var(--pf-l-flex--spacer--lg);
     }
+
+    // Limit details width to 80 characters or 2/3 the width of the page,
+    // whichever is larger
+    .pf-c-description-list:not(:only-child) {
+        max-width: max(80ch,66vw);
+    }
 }
 
 .container-block {
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -21,10 +21,8 @@
 }
 
 .container-block small {
-    flex-basis: calc(var(--pf-global--LineHeight--sm) * 1rem);
-    line-height: var(--pf-global--LineHeight--sm);
-    opacity: 0.7;
-    overflow: hidden;
+    @include pf-line-clamp("1");
+    color: var(--pf-global--Color--200);
 }
 
 .container-name {
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -26,6 +26,13 @@
     }
 }
 
+@media screen and (max-width: 768px) {
+    // Badges should not stretch in mobile mode
+    .pf-c-table [data-label] > .pf-c-badge {
+        justify-self: start;
+    }
+}
+
 .container-block {
     display: flex;
     flex-direction: column;
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -8,10 +8,15 @@
     @extend .ct-card;
 }
 
-// Decrease padding for the image/container toggle button list
 #containers-images, #containers-containers {
+    // Decrease padding for the image/container toggle button list
     .pf-c-table.pf-m-compact .pf-c-table__toggle {
         padding-left: 0px;
+    }
+
+    // Ensure there's adaquate space between detail groups
+    .ct-listing-panel-body > .pf-l-flex {
+        gap: var(--pf-l-flex--spacer--lg);
     }
 }
 
--- a/src/ContainerDetails.jsx
+++ b/src/ContainerDetails.jsx
@@ -29,6 +29,15 @@ const render_container_published_ports = (ports) => {
 
 const ContainerDetails = ({ container, containerDetail }) => {
     const ports = render_container_published_ports(container.Ports);
+    const networkOptions = (
+        containerDetail &&
+        [
+            containerDetail.NetworkSettings.IPAddress,
+            containerDetail.NetworkSettings.Gateway,
+            containerDetail.NetworkSettings.MacAddress,
+            ports
+        ].some(itm => !!itm)
+    );
 
     return (
         <Flex spaceItems={{ modifier: 'spaceItemsXl' }}>
@@ -46,7 +55,7 @@ const ContainerDetails = ({ container, containerDetail }) => {
                     <DescriptionListDescription>{container.Command ? utils.quote_cmdline(container.Command) : ""}</DescriptionListDescription>
                 </DescriptionListGroup>
             </DescriptionList>
-            <DescriptionList columnModifier={{ default: '2Col' }} className='container-details-networking'>
+            {networkOptions && <DescriptionList columnModifier={{ default: '2Col' }} className='container-details-networking'>
                 {ports && <DescriptionListGroup>
                     <DescriptionListTerm>{_("Ports")}</DescriptionListTerm>
                     <DescriptionListDescription>{ports}</DescriptionListDescription>
@@ -63,7 +72,7 @@ const ContainerDetails = ({ container, containerDetail }) => {
                     <DescriptionListTerm>{_("MAC address")}</DescriptionListTerm>
                     <DescriptionListDescription>{containerDetail.NetworkSettings.MacAddress}</DescriptionListDescription>
                 </DescriptionListGroup>}
-            </DescriptionList>
+            </DescriptionList>}
             <DescriptionList className='container-details-state'>
                 <DescriptionListGroup>
                     <DescriptionListTerm>{_("Created")}</DescriptionListTerm>


garrett:
This PR is a follow-up to #753.

garrett:
Screenshots: 

Screenshot 2021-09-09 at 14-19-02 Podman containers - garrett Bolt

Screen Shot 2021-09-09 at 14 19 54

But the images section is still a disaster on mobile layouts, as I haven't addressed it yet:

Screenshot 2021-09-09 at 14-20-33 Podman containers - garrett Bolt

src/podman.scss
```diff
@@ -8,11 +8,22 @@
...
-// Decrease padding for the image/container toggle button list
 #containers-images, #containers-containers {
+    // Decrease padding for the image/container toggle button list
     .pf-c-table.pf-m-compact .pf-c-table__toggle {
         padding-left: 0px;
     }
+
+    // Ensure there's adaquate space between detail groups
+    .ct-listing-panel-body > .pf-l-flex {
+        gap: var(--pf-l-flex--spacer--lg);
```
jelly, line 19: can't we give the spacer size as an option to a `Flex` component?

src/podman.scss
```diff
@@ -8,11 +8,22 @@
...
+
+        // Don't include empty blocks in the layout
+        // (for example, if networking is there for a non-networked container, don't show it)
+        > :empty {
+          display: none;
+        }
```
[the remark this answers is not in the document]
garrett: Yeah, it's best handled in React. Please add there and remove this.

src/podman.scss
garrett: Nope.

PatternFly does not support a gap option natively, as their browser support is wider than ours and they still want to support an older Safari version (which was the last holdout that didn't support gap where every other browser did for quite some time). Hence, using it here.

PatternFly uses margins instead, which are always present and not contextual like gaps are, so you'd end up with extra space to the side(s) and below.

src/podman.scss
```diff
@@ -24,6 +24,12 @@
           display: none;
         }
     }
+
+    // Labels should not stretch
```
KKoukiou, line 28: This comment is misleading. This affects not really the labels but the while rows. 
In addition it's a really intrusive css override which needs to be explained and if it's generic enough moved to our overrides file and upstreamed.

KKoukiou: This is also only interesting for mobile mode.

src/podman.scss
```diff
@@ -30,6 +30,12 @@
     .pf-m-grid-md.pf-c-table [data-label] {
         justify-items: start;
     }
+
+    // Limit details width to 80 characters or 2/3 the width of the page,
+    // whichever is larger
+    .pf-c-description-list:not(:only-child) {
+        max-width: max(80ch,66ch);
```
KKoukiou, line 37: The comment does not match the code.

KKoukiou (changes requested):

src/podman.scss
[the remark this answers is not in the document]
KKoukiou: You can apply this diff for this:
``` diff

diff --git src/ContainerDetails.jsx src/ContainerDetails.jsx
index 739460c6..a2cb758e 100644
--- src/ContainerDetails.jsx
+++ src/ContainerDetails.jsx
@@ -29,6 +29,14 @@ const render_container_published_ports = (ports) => {
 
 const ContainerDetails = ({ container, containerDetail }) => {
     const ports = render_container_published_ports(container.Ports);
+    const networkOptions = (
+        containerDetail &&
+        [
+            containerDetail.NetworkSettings.IPAddress,
+            containerDetail.NetworkSettings.Gateway,
+            containerDetail.NetworkSettings.MacAddress
+        ].some(itm => !!itm)
+    );
 
     return (
         <Flex spaceItems={{ modifier: 'spaceItemsXl' }}>
@@ -46,7 +54,7 @@ const ContainerDetails = ({ container, containerDetail }) => {
                     <DescriptionListDescription>{container.Command ? utils.quote_cmdline(container.Command) : ""}</DescriptionListDescription>
                 </DescriptionListGroup>
             </DescriptionList>
-            <DescriptionList columnModifier={{ default: '2Col' }} className='container-details-networking'>
+            {networkOptions && <DescriptionList columnModifier={{ default: '2Col' }} className='container-details-networking'>
                 {ports && <DescriptionListGroup>
                     <DescriptionListTerm>{_("Ports")}</DescriptionListTerm>
                     <DescriptionListDescription>{ports}</DescriptionListDescription>
@@ -63,7 +71,7 @@ const ContainerDetails = ({ container, containerDetail }) => {
                     <DescriptionListTerm>{_("MAC address")}</DescriptionListTerm>
                     <DescriptionListDescription>{containerDetail.NetworkSettings.MacAddress}</DescriptionListDescription>
                 </DescriptionListGroup>}
-            </DescriptionList>
+            </DescriptionList>}
             <DescriptionList className='container-details-state'>
                 <DescriptionListGroup>
                     <DescriptionListTerm>{_("Created")}</DescriptionListTerm>

```

src/podman.scss
jelly: Thanks for the explanation!

src/podman.scss
garrett: ...how does the comment not match the code? It explains exactly what it does.

Oh, I did make a typo though, it should be 80ch,66**vw**

src/podman.scss
garrett: > This affects not really the labels

It's literally talking about _data-**label**_:
    
```css
.pf-m-grid-md.pf-c-table [data-label] {
```

> This is also only interesting for mobile mode.

Yes; it does not affect desktop. I figured having a media query here where it doesn't affect desktop doesn't matter.

> it's a really intrusive css override 

It's inside of the following CSS, so it's not intrusive whatsoever (as it only apples to the containers and images lists):
```
#containers-images, #containers-containers {
```

> which needs to be explained

`justify-items` is part of CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items

And the start value is:

"The item is packed flush to each other toward the start edge of the alignment container in the appropriate axis."

garrett: There isn't a `justify-items: start` equivalent on the PF4 utilities @ https://www.patternfly.org/v4/utilities/flex (which are called "flex" but also apply to grid too), so this one CSS line is necessary.

KKoukiou: data-label != label and since the data-label attr applies on the whole table rows, this comment is indeed misleading, maybe not for you that you wrote it of course. Please change it to 'Badges should not stretch on mobile mode' or something that we can all understand what it fixes when looking at the CSS.

In addition - I know that this override exists as is also in cockpit-machines and therefore I asked if it's generic enough to be upstreamed, or at least put into our overrides file.

KKoukiou (changes requested):
This needs at least https://github.com/cockpit-project/cockpit-podman/pull/758/commits/873343d4ce60ac3da0e18deb3ac9b433e29bfdf8#r705427385 addressed for moving it forward.

src/podman.scss
garrett: Aha, I get it. Will do. Thanks!

garrett: I thought you meant it's generic and would be applied where we don't want it. I didn't read it right at first, but yeah, I think you're right that this might be better as a generic fix in Cockpit, as elsewhere will probably have this issue. And, yes, we should report it upstream to PF.

My _guess_ is that they probably have a reason for this, perhaps with widgets in mind, so buttons and dropdowns would take up the space.

src/podman.scss
```diff
@@ -24,6 +24,12 @@
     .pf-m-grid-md.pf-c-table [data-label] {
         justify-items: start;
     }
+
+    // Limit details width to 80 characters or 2/3 the width of the page,
+    // whichever is larger
+    .pf-c-description-list:not(:only-child) {
```
KKoukiou, line 30: I can't see why the default wrapping does not work - please let's clarify this with a screenshot before merging.

src/podman.scss
```diff
@@ -21,10 +21,18 @@
 }
 
 .container-block small {
-    flex-basis: calc(var(--pf-global--LineHeight--sm) * 1rem);
     line-height: var(--pf-global--LineHeight--sm);
     opacity: 0.7;
     overflow: hidden;
+    // line-clamp / max-lines needs webkit prefix, but is supported in modern all browsers like this
```
KKoukiou, line 44: Can't we use .pf-u-text-truncate class - instead of all this? 
We just want to truncate the text from what I understand.

KKoukiou (dismissed):

src/podman.scss
garrett: It needs to be clamped, as it can be arbitrary amounts of lines. This is the only way line-clamp works in browsers, sadly. Anything else is a hack.

garrett:
I'm working on this at the moment, but am in a meeting. I'll have to get back to it after our standup.

src/podman.scss
KKoukiou: PF has a mixin for line clamping - see  pf-line-clamp in src/patternfly/sass-utilities/mixins.scss

KKoukiou: Seem you can use it like:
@include pf-line-clamp("1");

garrett: Oh, that's interesting! Does it work for us though? (Or does it use inline CSS?)

garrett: OK, I've looked into it and the CSS that's generated is similar, but it'll keep us from rewriting CSS hacks (which work across browser, but it still depends on nasty prefixed CSS even if it's supported well). 

I'll switch to using it in our CSS here instead.

garrett:
@KKoukiou: The clamping include you pointed out is essentially producing the same resulting code when it compiles down, but it's nicer in our source SCSS. I also removed the remains of our hacky clamping (which made assumptions and no longer worked properly) along with the actual clamping (which is a webkit prefix hack, even if all browsers support it) and converted the opacity to an actual PF color equivalent (also used for `user:`)... all in f3baec98839db974de79fca725412ccfebb52b37.

Thanks for pointing this out! Too bad PatternFly doesn't document the include. Is it meant to be used internally only? It seems that something like this should be officially exposed for projects. And they could make some helper utilities (as classes) that basically use this CSS but vary the lines, so it could be directly used from within JSX.

src/podman.scss
garrett: This is to allow the command to wrap when needed, so the layout flows as intended.

This is what it looks like on main:
Screenshot 2021-09-14 at 17-35-53 Podman containers - garrett Bolt

And here's this PR:
Screenshot 2021-09-14 at 17-31-43 Podman containers - garrett Bolt

You can see the command line is really long in this case, so this tells it to wrap when it needs to, so the layout works. The only child limiting is saying that if there are no other details, then this one can take the whole space.

garrett: Additionally, in smaller sizes (this is a narrower desktop mode with just the content, but it applies to mobile too), the layout is messed up as it doesn't use gap but tries to fake things with margins:

On main:
Screenshot 2021-09-14 at 17-36-12 Podman containers - garrett Bolt

With this PR:
Screenshot 2021-09-14 at 17-38-22 Podman containers - garrett Bolt

garrett:
Oh, I do see a side effect with the badge/label alignment:

Screenshot 2021-09-14 at 17-41-38 Podman containers - garrett Bolt

Here's what it's like on main:
Screenshot 2021-09-14 at 17-41-55 Podman containers - garrett Bolt

(The images kebab button layout is weird, but that's for another PR.)

garrett:
@KKoukiou: I dropped the previous solution to not have badges stretch in mobile mode and implemented a different version in 5b81cde.

We should probably convert this to a PF4 override and suggest it upstream, as you suggested.

Here's what it looks like, currently in `podman.scss`:

```scss
@media screen and (max-width: 768px) {
    // Badges should not stretch in mobile mode
    .pf-c-table [data-label] > .pf-c-badge {
        justify-self: start;
    }
}
```

This solution specifically targets a badge within a collapsed table. We would want to use the proper max-width variable instead of 768px. (It's getting late here.) And we should move it, if you agree about it.

WDYT?

We can completely drop this one commit from this particular PR and move it to another, as it would affect shared code and I don't want to block the rest of the changes for this.

src/podman.scss
```diff
@@ -21,10 +21,8 @@
 }
 
 .container-block small {
-    flex-basis: calc(var(--pf-global--LineHeight--sm) * 1rem);
-    line-height: var(--pf-global--LineHeight--sm);
-    opacity: 0.7;
-    overflow: hidden;
+    @include pf-line-clamp("1");
```
KKoukiou, line 42: Thanks - much nicer.

src/podman.scss
KKoukiou: I reverted this commit - and kept all others in this PR  and I can get the issues you see here with long commands.

src/podman.scss
garrett: Thanks for pointing it out! :+1: 

It basically does the same thing after compilation, but it is so much nicer in the source code here. Plus, I removed the rest of our previous bad hack (which I didn't completely remove before).

jelly (approved):

PR #761, opened 2021-09-13
.gitignore: add ./bots/vm-run temporary files

jelly:
Ignore the temporary qcow2 image from vm-run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,4 @@ test/common/
 test/images/
 *.pot
 POTFILES*
-tmp/weblate-repo
+tmp/


.gitignore
```diff
@@ -15,3 +15,4 @@ test/images/
 *.pot
 POTFILES*
 tmp/weblate-repo
+tmp/run
```
martinpitt, line 18: How about we just ignore tmp/ entirely? 

After this lands, can you please apply this to starter-kit as well? Thanks!

jelly: Sounds good to me!

martinpitt (approved):
Thanks!

PR #762, opened 2021-09-13
Build node cache on release, move to xz release tarballs

martinpitt:
Ported most commits from https://github.com/cockpit-project/starter-kit/pull/488 , except for the one that actually rebuilds the webpack during RPM. We don't currently have pressure to do this for existing projects, and with the commits from these PRs (and thus having node cache tarballs available for releases), we could enable this quickly if/when needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 *~
 *.retry
-*.tar.gz
+*.tar.xz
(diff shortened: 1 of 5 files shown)


marusak (approved):
Thanks!

PR #766, opened 2021-09-20
Port integraton tab in run image modal to the new design

KKoukiou:


src/ImageRunModal.scss
```diff
@@ -1,17 +1,18 @@
-@import '@patternfly/patternfly/layouts/Grid/grid.css';
+.dynamic-form-group {
+    .pf-c-empty-state {
+        padding: 0;
+    }
```
jelly, line 4: When I was working on the design I used:
```
 <EmptyState variant='xs'>
```

src/ImageRunModal.jsx
```diff
@@ -32,135 +38,177 @@ const units = {
...
+                </FormGroup>
+                <FormGroup className="pf-m-1-col-on-sm remove-button-group">
+                    <Button variant='secondary'
+                                className="btn-close"
+                                id={id + "-btn-close"}
+                                isSmall
+                                aria-label={_("Remove item")}
+                                icon={<MinusIcon />}
+                                onClick={() => removeitem(idx)} />
+                </FormGroup>
```
jelly, line 103: This button should be red

src/ImageRunModal.jsx
```diff
@@ -32,135 +38,177 @@ const units = {
...
+    if (itemCount) {
+        return (
+            <Grid hasGutter id={id}>
+                <FormGroup className="pf-m-4-col-on-sm"
+                    label={_("IP address")}
+                    fieldId={id + "-ip-address"}
+                    labelIcon={
+                        <Popover aria-label={_("IP address help")}
+                            enableFlip
+                            bodyContent={_("If host IP is set to 0.0.0.0 or not set at all, the port will be bound on all IPs on the host.")}>
```
jelly, line 51: Should the default value be 0.0.0.0 then?

src/ImageRunModal.jsx
```diff
@@ -32,135 +38,177 @@ const units = {
...
+                    <TextInput id={id + "-container-path"}
+                        value={item.containerPath || ''}
+                        onChange={value => onChange(idx, 'containerPath', value)} />
+                </FormGroup>
+                <FormGroup className="pf-m-2-col-on-sm" label={_("Mode")} fieldId={id + "-mode"}>
+                    <Checkbox id={id + "-mode"}
+                        label={_("Writable")}
+                        value={item.mode == "rw"}
+                        onChange={value => onChange(idx, 'mode', value ? "rw" : "ro")} />
+                </FormGroup>
```
jelly, line 172: Toggling the button, didn't work for me, nor is it checked by default.

jelly (changes requested):

src/ImageRunModal.jsx
KKoukiou: I actually don't agree with the design on this. Red buttons imply dangerous operations, that can't be reverted etc. Removing options in the dialog does not fall into this category.
@garrett What do you think?

src/ImageRunModal.jsx
```diff
@@ -32,135 +38,177 @@ const units = {
...
+                                enableFlip
+                                bodyContent={_("If the host port is not set the container port will be randomly assigned a port on the host.")}>
+                                <button onClick={e => e.preventDefault()} className="pf-c-form__group-label-help">
+                                    <OutlinedQuestionCircleIcon />
+                                </button>
+                            </Popover>
+                        }>
+                    <TextInput id={id + "-host-port"}
+                                value={item.hostPort || ''}
+                                onChange={value => onChange(idx, 'hostPort', value)} />
```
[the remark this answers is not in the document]
KKoukiou: Done

src/ImageRunModal.jsx
```diff
@@ -32,135 +38,177 @@ const units = {
...
+                </FormGroup>
+                <FormGroup className="pf-m-3-col-on-sm"
+                            label={_("Container port")}
+                            fieldId={id + "-container-port"} isRequired>
+                    <TextInput id={id + "-container-port"}
+                                value={item.containerPort || ''}
+                                onChange={value => onChange(idx, 'containerPort', value)} />
```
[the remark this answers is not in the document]
KKoukiou: Done

src/ImageRunModal.scss
KKoukiou: Done

src/ImageRunModal.jsx
KKoukiou: Fixed.

src/ImageRunModal.jsx
```diff
@@ -32,135 +38,177 @@ const units = {
...
-        </>
-    );
+const PublishPort = ({ id, item, onChange, idx, removeitem, itemCount }) => {
+    if (itemCount) {
+        return (
+            <Grid hasGutter id={id}>
+                <FormGroup className="pf-m-4-col-on-sm"
+                    label={_("IP address")}
+                    fieldId={id + "-ip-address"}
+                    labelIcon={
```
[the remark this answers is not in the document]
jelly: I checked out the docs and couldn't find an example with a hostname, so I'm not sure where it's taken from.

src/ImageRunModal.jsx
KKoukiou: I don't like the idea of prefilling with default values.

jelly (changes requested):

src/ImageRunModal.jsx
jelly: Ok :+1:

src/ImageRunModal.jsx
```diff
@@ -32,135 +38,152 @@ const units = {
...
+                        value={item.IP || ''}
+                        onChange={value => onChange(idx, 'IP', value)} />
+            </FormGroup>
+            <FormGroup className="pf-m-2-col-on-sm"
+                    label={_("Host port")}
+                    fieldId={id + "-host-port"}
+                    type='number'
+                    step={1}
+                    min={1}
+                    max={65535}
```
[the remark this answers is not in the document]
KKoukiou: Thanks for catching this!

src/ImageRunModal.jsx
```diff
@@ -32,135 +38,152 @@ const units = {
...
+                <TextInput id={id + "-host-port"}
+                            value={item.hostPort || ''}
+                            onChange={value => onChange(idx, 'hostPort', value)} />
+            </FormGroup>
+            <FormGroup className="pf-m-3-col-on-sm"
+                        label={_("Container port")}
+                        type='number'
+                        step={1}
+                        min={1}
+                        max={65535}
```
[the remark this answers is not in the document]
KKoukiou: We can use it fine.

marusak:
The position of `Writable` checkbox seems a bit odd? I would say it should align to baseline?

Screenshot from 2021-09-27 11-25-41

marusak:
Thanks! Other than that alignment it looks great!

KKoukiou:
> The position of `Writable` checkbox seems a bit odd? I would say it should align to baseline?
> 

I agree but anything like align-self: center or basine does not seem to take effect.
@garrett can you please take a look?

garrett:
After going through many dozens of grid and flex permutations on various widgets (with associated styles), I stepped back to think about the problem. 

The check widget isn't the same height as the inputs here. That's the root of the problem.

As the check widget happens to be a grid itself, I played around with changing the height to 36px (what widgets are computed to) and clanging the align-content to center. That worked. But I didn't like that I was hardcoding the height, so I adjusted it a bit more. 

I also played around with an even simpler solution: If I just add a spacer-sm to the top and bottom and set the height to 20px (as it's 20.8px now and gets rounded up), that would also work. But in this layout, we just need to bump the widget down, right? So I simplified it to a spacer-sm on the top padding minus 1px because of a PF rounding error. That also does it. But it's probably a little more hackish.

As pf-c-check doesn't have access to the pf-c-input variables, I have to use the global spacer for this. It's what the others compute to anyway.

So, we have two main options:

1. Hardcode the height to be the same as other widgets (we could use a calc to base it on the variables inputs use), resolving to 36px, let alignment work.

   a. Precomputed

    ```scss
    .dynamic-form-group .pf-c-form__group-control > .pf-c-check {
        // Set height to the same as inputs
        height: 36px;
        align-content: center;
    }
    ```

   b. Calc

    ```scss
    .dynamic-form-group .pf-c-form__group-control > .pf-c-check {
        // Set height to the same as inputs
        // Widgets use 5px and have 1px border on top and bottom;
        // (5 + 1) * 2 = 12
        height: calc(var(--pf-global--FontSize--md) * var(--pf-global--LineHeight--md) + 12px)
        align-content: center;
     }
     ```

2. Set padding to the top.

    ```scss
    .dynamic-form-group .pf-c-form__group-control > .pf-c-check {
        // Add vertical padding to the top similar to inputs, for alignment
        // -1px due to rounding error in PF's checkbox height computation
        padding-top: calc(var(--pf-global--spacer--sm) - 1px);
    }
    ```

While 2 is the simplest, 1b is more correct (and should be the most robust), so I've committed it as the solution here and will push in a minute.

jelly (approved):

PR #770, opened 2021-09-22
Adjust tests for rhel-8-6 as this uses cgroups-v2

KKoukiou:


KKoukiou:
Just needs to copy an existing naugty

PR #771, opened 2021-09-23
Migrate to new stats endpoint to fix incorrect memory limit

### LINKED ISSUE #756: Inconsistency on showing memory limits
After last updates on Fedora 34, i see inconsistency in showing memory limits. See this :

podman stats :
```
 ID                NAME                    CPU %   MEM USAGE / LIMIT      MEM %     NET IO    BLOCK IO                PIDS        CPU TIME    AVG CPU %
7e4947c192d3    3e4e4901fd6f-infra          --       155.6kB / 33.57GB    0.00%     -- / --    -- / --                  1          2.297ms    4.14%
88bf4930199a    node-odm_1                  --       4.451GB / 33.57GB    13.26%    -- / --    16.38kB / 5.59GB         1     52m35.41304s  301.61%
a569070430c4    db                         0.21%     307.1MB / 33.57GB    0.91%    -- / --    983kB / 2.118GB           7        6.103455s    0.41%
b55d103d198a    broker                     0.07%     3.813MB / 33.57GB    0.01%    -- / --    0B / 2.54MB               5        768.803ms    0.09%
d4a5ed5fd53b    worker                     0.30%     858.9MB / 33.57GB    2.56%    -- / --    0B / 690.4MB             12        7.094489s    0.64%
e990d3218cdd    webapp                     0.58%     1.807GB / 33.57GB    5.38%    -- / --    0B / 1.019GB             54       35.024928s    0.49%
```

In cockpit podman app it shows only 16GB memory limits (i have 32GB RAM).

KKoukiou:
Fixes #756


jelly:
I noticed a few things when playing around with the new stats API. It now returns an object with stats for all containers so we could choose to call the stats API once with a filter of running containers.

Secondly it seems that the new stats endpoint's interval is by default 1 second and the interval parameter does not work until [this commit in podman](https://github.com/containers/podman/commit/3a922cbc99f49b996c0379a9ec90ccc47a325018). Which is not released yet.

KKoukiou:
@jelly I rethought about the point regarding making one API call for all running containers.

Making one API call sounds indeed cheaper operation.
However the multiple (per container) API calls is our current implementation, thus keeping this won't be a regressoin.
 
The issue I see with batching all the stat in one call, is that when a container get's started, we need to cancel the previous API call, and run a new one with the new container added.

This sounds already a bit ugly. What do you think?

jelly:
> @jelly I rethought about the point regarding making one API call for all running containers.
> 
> Making one API call sounds indeed cheaper operation.
> However the multiple (per container) API calls is our current implementation, thus keeping this won't be a regressoin.

I'm fine with merging it as is to keep from regressing,

> The issue I see with batching all the stat in one call, is that when a container get's started, we need to cancel the previous API call, and run a new one with the new container added.
> 
> This sounds already a bit ugly. What do you think?

Aha, yeah I thought about not filtering the stats and just received all of them as I thought the stats endpoint only returns active running containers. Then our code would when we start cockpit-podman just listen for all stats and keep them in state with an image id to stats mapping, when a container is started we can obtain the stats by looking it up in our state without subscribing to something explicitly.

jelly:
The tests might fail because the default timeout is 1 second before podman 3.4.0, which most likely causes the test failures. Manually upgrading to podman 3.4.0 seems to resolve the test failures in `testLifecycleOperationsUser` for me.

PR #773, opened 2021-09-23
Implement creating a container without an local image

jelly:
In the current behaviour a user first has to download an image to create
a container. The new `Create container` button in the containers view
shows an images dropdown where a user can filter local or remote
containers from registries defined by podman. When a user clicks on
create the image is downloaded and started if that is desired.

This is a draft PR and lacks the following:

- [ ] Filter local containers on owner
- [ ] Close the dialog when we are downloading the image and show a container with status "Downloading.." in the Containers view
- [ ] The tags dropdown, as we are unable to fetch the image tags. This has to be researched and in my opinion is something for a follow up, for now the current behaviour is retained by having a tags input field.
- [ ] Searching in a specific registry does not exists yet, will be created when we decide that the current search/dropdown behaviour is good enough


garrett:
Filter should be to the left, action should be to the right

It should look similar to the same pattern as used the Machines page:

Screenshot from 2021-09-27 14-57-48-mod

(I also trimmed that extra space that shouldn't be there at the top of the list. But that bug is in main too.)

garrett:
I get an oops when clicking the create container button:

Screenshot from 2021-09-28 09-39-46

I realized it's because the container and image lists were both not fully loaded yet when I clicked the button. (This should be fixed, of course.)

garrett:
I get an oops when typing "f" and hitting backspace:

I'm sorry I don't really have a better way of sharing the backtraces other than a screenshot (as copy/pasting loses the formatting):

Screenshot from 2021-09-28 09-42-54

garrett:
After searching all registries, the dropdown is propagated with the results and yet still has 
search all registries".

image

garrett:
Typing something and hitting down adds the "search all registries: ..." as text in the entry:

Screenshot from 2021-09-28 09-48-56

garrett:
I was looking into the various select options and found footer. While it doesn't scroll by default (so it'll happily go offscreen), I noticed that if we set the overflow to auto and max-height to whatever, it could scroll the list and keep footer actions on top, like so:

image

I'm saying we could consider a filter switcher thing in the dropdown, perhaps. Like this concept:

image-search-footer

It would let us immediately show local matches and immediately (after a _tiny_ delay for typing) kick off a search without having to click on search.

garrett:
Here's what auto-searching could look like (with local images first)

image-search-footer(1)

jelly:
After a meeting today, we came to the following improvements/conclusions:

- [ ] Don't show search for ... when we already have a search result
- [ ] Don't show the option to select tags and make the latest tag implicit
- [ ] Always display the registry when an image is selected

https://github.com/patternfly/patternfly-react/issues/6364

React warnings in console are from patternfly https://github.com/patternfly/patternfly-react/issues/6365

garrett:
Today's meeting notes:

## Popover text

Info popover with something of the following (better explained):

(i) popover text for `host[:port]/[user]/container[:tag]
`
Examples for:
- container search
- search a registry
- specify an image
- add an optional tag (defaults to :latest)

Placeholder in the dropdown: "Search string or container location"


## Helper text

Below the dropdown

- "" (empty string at first, so form doesn't jump? create button disabled)
- Local image (from user/system?) will be used (success, create button enabled)
- Checking remote (validating that the image exists, button disabled while the check is happening)
- Image will downloaded (success, create button enabled )
- Remote image does not exist (error, create button disabled)
- Random error, like an API limit (enable create button anyway?)

Verification like this: `curl -X GET -s -g --no-buffer --unix-socket /run/user/1000/podman/podman.sock 'http://localhost/v1.12/libpod/images/search?term=docker.io/library/redis&limit=10&listTags=true' | jq`

## Minimum for dropdown

- Show local images first
- Search icon and blue text for starting a search

## User versus system images...

- Move system/user selector right above the image?
- Show both? Or an expander for others?


## Future

- download progress indicator
- automatic search? (although it'd be nice to get in, if we can)
- filtering
- tag browsing for validated images (typing `:` would switch to tag mode)
- how do we show official images? ("label"/badge at the side? what about when selected?)
- option to pull the latest version (when there's a local image match)
- mockups for tag handling (later)
- mockup for "recommended" images during a search via shortnames (eventually)

jelly:
> Today's meeting notes:
> 
> ## Popover text
> Info popover with something of the following (better explained):
> 
> (i) popover text for `host[:port]/[user]/container[:tag] ` Examples for:
> 
> * container search
> * search a registry
> * specify an image
> * add an optional tag (defaults to :latest)

How should this look like? Pushed something but I'm not sure if that's the same and surely the text needs to be expanded.

> Placeholder in the dropdown: "Search string or container location"

Done
 
> ## Helper text
> Below the dropdown
> 
> * "" (empty string at first, so form doesn't jump? create button disabled)
> * Local image (from user/system?) will be used (success, create button enabled)
> * Checking remote (validating that the image exists, button disabled while the check is happening)
> * Image will downloaded (success, create button enabled )
> * Remote image does not exist (error, create button disabled)
> * Random error, like an API limit (enable create button anyway?)
> 
> Verification like this: `curl -X GET -s -g --no-buffer --unix-socket /run/user/1000/podman/podman.sock 'http://localhost/v1.12/libpod/images/search?term=docker.io/library/redis&limit=10&listTags=true' | jq`

 
> ## Minimum for dropdown
> * Show local images first
> * Search icon and blue text for starting a search

Done
 
> ## User versus system images...
> * Move system/user selector right above the image?
> * Show both? Or an expander for others?
> 
> ## Future
> * download progress indicator
> * automatic search? (although it'd be nice to get in, if we can)
> * filtering
> * tag browsing for validated images (typing `:` would switch to tag mode)
> * how do we show official images? ("label"/badge at the side? what about when selected?)
> * option to pull the latest version (when there's a local image match)
> * mockups for tag handling (later)
> * mockup for "recommended" images during a search via shortnames (eventually)

garrett:
Oh, I forgot to push up some pending changes... one adds a vertical divider between filter and actions (I think I mentioned this above? or in IRC?), the other relaxes the image names in the dropdown so they can wrap (we talked about this in the call a little).

garrett:
Nice to see that hitting enter does a search. However, pressing the down key ("debia", down-arrow) still places the action text into the entry:

Screenshot 2021-09-30 at 12-09-19 Podman containers - garrett Bolt

garrett:
Hitting escape when the autocomplete entry is shown should dismiss the entry... but it dismisses the entire dialog at the moment.

garrett:
Wow, this is awesome! It's just about ready, I think!?

This main issue I see is when it's downloading, there's no name for the container.

jelly:
> When I'm typing something and have it filtered down, how can I use the keyboard to change the source filter? I tried tab (dismisses the popdown), arrow keys (down doesn't do anything), etc.
> 
> image
> 
> I also cannot tab to the dropmarker icon... This thing: image
> 
> Otherwise, it's feeling quite nice.

I've found out that the dropmaker icon has an explicit [tabIndex=-1](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Select/SelectToggle.tsx#L314)

jelly:
@garrett the new push fixes the tab behaviour on Firefox/Epiphany. On chromium it still works a bit weird. I've created an issue upstream for it https://github.com/patternfly/patternfly-react/issues/6471 but it still works.

src/Containers.jsx
```diff
@@ -284,6 +289,23 @@ class Containers extends React.Component {
         window.removeEventListener('resize', this.onWindowResize);
     }
 
+    onDownloadContainer = (container) => {
+        const { downloadingContainers } = this.state;
```
KKoukiou, line 293: Here is a more safe and the react way to do this:
``` 
this.setState(prevState => ({
    downloadingContainers: [...prevState.downloadingContainers, container]
}))
```

src/Containers.jsx
```diff
@@ -284,6 +289,23 @@ class Containers extends React.Component {
...
+        this.setState({
+            downloadingContainers: [...downloadingContainers,
+                container
+            ]
+        });
+    }
+
+    onDownloadContainerFinished = container => {
+        const { downloadingContainers } = this.state;
+        const newState = downloadingContainers.filter(entry => entry.name !== container.name);
```
KKoukiou, line 303: Do the same as above with the prevState, otherwise this is racy.

src/Containers.jsx
```diff
@@ -433,7 +486,28 @@ class Containers extends React.Component {
                             <FormSelectOption value='all' label={_("All")} />
                         </FormSelect>
                     </ToolbarItem>
+                    <Divider isVertical />
```
KKoukiou, line 489: This will need a row-gap in custom CSS:
Screen Shot 2021-10-21 at 09 44 17

or actually ideally an upstream issue.

src/client.js
```diff
@@ -273,3 +273,15 @@ export function pullImage(system, reference) {
                 .catch(reject);
     });
 }
+
+export function imageExists(system, id) {
```
marusak, line 277: Couldn't this be written as:
```
...
.then(resolve)
.catch(reject);
```

or even:
```py
export function imageExists(system, id) {
    return podmanCall("libpod/images/" + id + "/exists", "GET", {}, system);
}
```

test/check-application
```diff
@@ -1139,6 +1139,85 @@ class TestApplication(testlib.MachineCase):
...
+        # Untag busybox image which duplicates the image we are about to download
+        self.execute(True, "podman rmi -f quay.io/libpod/busybox localhost:5000/my-busybox localhost:6000/my-busybox")
+
+        self.login(auth)
+
+        b = self.browser
+        container_name = "busybox-downloaded"
+        # No local results found
+        b.wait_visible("#containers-containers button.pf-c-button.pf-m-primary:not([disabled])")
+        b.click("#containers-containers button.pf-c-button.pf-m-primary")
```
marusak, line 1167: Why doing this? `b.click()` already waits for element to be visible and not disabled.

src/ImageRunModal.jsx
```diff
@@ -355,6 +395,65 @@ export class ImageRunModal extends React.Component {
                 });
     }
 
+    onCreateClicked() {
+        const createConfig = this.getCreateConfig();
+        const { owner, runImage } = this.state;
+
+        // TODO: correct?
+        let isSystem = false;
```
KKoukiou, line 403: I think you can simplify this by pre-setting the state variable owner like this:
`this.state = { owner:  this.state.image.isSystem ? systemOwner : whaterverowner } `

Then here you can just do:
` const isSystem = owner === systemOwner`

src/ImageRunModal.jsx
```diff
@@ -253,8 +261,13 @@ DynamicListForm.propTypes = {
 export class ImageRunModal extends React.Component {
     constructor(props) {
         super(props);
+        let command = "sh";
```
marusak, line 264: Do we want a "default" command? Why not just leaving it empty?

src/ImageRunModal.jsx
```diff
@@ -367,14 +466,255 @@ export class ImageRunModal extends React.Component {
...
+        if (regexString.includes('/')) {
+            regexString = searchText.replace(searchText.split('/')[0], '');
+        }
+        const input = new RegExp(regexString, 'i');
+
+        const results = imageRegistries
+                .map((reg, index) => {
+                    const filtered = (reg in images ? images[reg] : [])
+                            .filter(image => {
+                                if ('isSystem' in image && image.isSystem && !isSystem) {
```
KKoukiou, line 621: If image is an object you don't need this 'isSystem' in image check.

src/ImageRunModal.jsx
```diff
@@ -355,6 +395,65 @@ export class ImageRunModal extends React.Component {
                 });
     }
 
+    onCreateClicked() {
+        const createConfig = this.getCreateConfig();
+        const { owner, runImage } = this.state;
+
+        // TODO: correct?
```
marusak, line 402: ? what is the question?

KKoukiou (changes requested):
Clicking on the registries toggle buttons in the dialog footer closes the modal for me, using firefox.

src/ImageRunModal.scss
```diff
@@ -27,10 +27,19 @@
...
+
 // Make sure the footer is visible with more then 5 results.
 .pf-c-select__menu-list {
-    max-height: 20rem;
-    overflow-y: scroll;
+    // 35% viewport height is for 1280x720;
+    // since it picks the min of the two, it works everywhere
+    max-height: min(20rem, 35vh);
+    overflow: hidden scroll;
 }
```
garrett: I don't like how these apply to PF directly without being custom classed (like either existing _under_ or in _addition to_ or _just using_  a Cockpit or page specific class).

That said, These are probably generic enough fixes that we might want to actually consider these as PatternFly fixes, so they might need to be overrides and pushed upstream to PatternFly.

garrett: (Yes, I modified and added this SCSS myself :grin:, but as the menu is popping up outside of the modal, we can't really subclass it, as far as I can tell. But it is really probably generic enough for an upstream fix.)

src/ImageRunModal.jsx
```diff
@@ -355,6 +395,65 @@ export class ImageRunModal extends React.Component {
...
+        }
+        if (!this.state.image && owner === systemOwner) {
+            isSystem = true;
+        }
+
+        client.imageExists(isSystem, createConfig.image).then(reply => {
+            this.createContainer(isSystem, createConfig, runImage);
+        })
+                .catch(ex => {
+                    console.info('image does not exists pull');
```
marusak, line 415: This is rather cryptic message :) Also if it fails, would make sense to show it in alert as we show other errors?

src/ImageRunModal.jsx
```diff
@@ -367,14 +466,255 @@ export class ImageRunModal extends React.Component {
         });
     };
 
+    onSearchTriggered = value => {
+        // Do not call the SearchImage API if the input string  is not at least 2 chars,
+        // unless Enter is pressed, which should force start the search.
```
marusak, line 471: > Unless enter is pressed

Is that implemented here?

src/ImageRunModal.jsx
```diff
@@ -367,14 +466,255 @@ export class ImageRunModal extends React.Component {
...
+        }
+
+        return results;
+    }
+
+    // Similiar to the output of podman search and podman's /libpod/images/search endpoint only show the root domain.
+    truncateRegistryDomain = (domain) => {
+        const parts = domain.split('.');
+        if (parts.length > 2) {
+            // TODO: nicer?
```
marusak, line 666: I think it is fine :)

src/ImageRunModal.jsx
```diff
@@ -386,12 +726,67 @@ export class ImageRunModal extends React.Component {
                                onChange={value => this.onValueChanged('containerName', value)} />
                         </FormGroup>
                     </FlexItem>
+                    { this.props.userServiceAvailable && this.props.systemServiceAvailable &&
+                    <FlexItem align={{ default: 'alignRight' }}>
+                        <FormGroup fieldId='run-image-dialog-owner' label={_("Owner")}>
+                            <ToggleGroup aria-label="Default with single selectable">
```
marusak, line 732: Please localize also aria-label

PR #779, opened 2021-10-04
Set key for children for the DynamicListForm

jelly:
Resolves the react warning that every child needs to have an unique
prop.


marusak (approved):
Thanks!

PR #788, opened 2021-10-06
Test on Arch Linux

jelly:
Include Arch Linux in the cockpit-podman testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,8 +123,13 @@ ifneq ($(filter debian-% ubuntu-%,$(TEST_OS)),)
 VM_DEP=$(TARFILE) packaging/debian/rules packaging/debian/control
 VM_PACKAGE=--upload `pwd`/$(TARFILE):/var/tmp/ --upload `pwd`/packaging/debian:/var/tmp/
 else
+ifneq ($(filter arch,$(TEST_OS)),)
+VM_DEP=$(TARFILE) packaging/arch/PKGBUILD
+VM_PACKAGE=--upload `pwd`/$(TARFILE):/var/tmp/ --upload `pwd`/packaging/arch:/var/tmp/
+else
 VM_DEP=$(RPMFILE)
 VM_PACKAGE=--upload `pwd`/$(RPMFILE):/var/tmp/
+endif
 endif
 
 ifeq ($(TEST_SCENARIO),rawhide)
--- a/packaging/arch/PKGBUILD
+++ b/packaging/arch/PKGBUILD
@@ -1,0 +1,15 @@
+pkgname=cockpit-podman
+pkgver=VERSION
+pkgrel=1
+pkgdesc='Cockpit UI for podman containers'
+arch=('any')
+url='https://github.com/cockpit-project/cockpit-podman'
+license=(LGPL)
+depends=(cockpit podman)
+source=("SOURCE")
+sha256sums=('SKIP')
+
+package() {
+  cd $pkgname
+  make DESTDIR="$pkgdir" install
+}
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -122,6 +122,10 @@ $(RPMFILE): $(TARFILE) $(RPM_NAME).spec
 ifneq ($(filter debian-% ubuntu-%,$(TEST_OS)),)
 VM_DEP=$(TARFILE) packaging/debian/rules packaging/debian/control
 VM_PACKAGE=--upload `pwd`/$(TARFILE):/var/tmp/ --upload `pwd`/packaging/debian:/var/tmp/
+endif
+ifneq ($(filter arch,$(TEST_OS)),)
+VM_DEP=$(TARFILE) packaging/arch/PKGBUILD
+VM_PACKAGE=--upload `pwd`/$(TARFILE):/var/tmp/ --upload `pwd`/packaging/arch:/var/tmp/
```
jelly, line 128: not 100% sure if I did if, if else else 100% correct

jelly: It's wrong debian tests fail.

packaging/arch/PKGBUILD
```diff
@@ -0,0 +1,15 @@
+pkgname=cockpit-podman
+pkgver=VERSION
+pkgrel=1
+pkgdesc='Cockpit UI for podman containers'
+arch=('any')
+url='https://github.com/cockpit-project/cockpit-podman'
+license=(LGPL)
+depends=(cockpit podman)
```
martinpitt, line 8: I assume you copied this from the official arch repo? Do you have a pointer to the cockpit one? In Debian/Ubuntu/Fedora/RHEL we split the cockpit source package into multiple binaries: cockpit-bridge (arch dependent), cockpit-ws (webserver), cockpit-system (login, shell, most important pages), and a bunch of others like cockpit-networkmanager. "cockpit" binary package is just an empty meta-package that depends/recommends the former.

Usually "extension" packages like c-podman should only depend on cockpit-bridge, not the full cockpit meta-package, to avoid dragging in the webserver on every installation (you can use it on a remote host through ssh).

No blocker for this PR of course, but worth discussing.

test/check-application
```diff
@@ -821,10 +821,12 @@ class TestApplication(testlib.MachineCase):
               .selectImageAndDownload() \
               .expectDownloadErrorForNonExistingTag()
 
+    @testlib.skipImage("Arch Linux does not support checkpoint/restore", "arch")
```
martinpitt, line 824: This reason is misleading -- the lifecycle operations don't test checkpoint/restore. Sure, the test might be broken for a different reason, but maybe call it something like "TODO: investigate"?

test/vm.install
```diff
@@ -55,6 +55,19 @@ if [ -d /var/tmp/debian ]; then
...
+    # build source package
+    cd /var/tmp
+    TAR=$(ls cockpit-podman-*.tar.xz)
+    VERSION="${TAR#cockpit-podman-}"
+    VERSION="${VERSION%.tar.xz}"
+
+    cp arch/PKGBUILD .
+    sed -i "s/VERSION/$VERSION/" PKGBUILD
+    sed -i "s/SOURCE/$TAR/" PKGBUILD
+    sudo -i -u builder extra-x86_64-build
```
martinpitt, line 68: We normally use `su` to drop privs from root to a user, as sudo has quite a complex setup, and trying to run it as user might hang eternally due to prompting for a password.

martinpitt (dismissed):
Nice one! I have a few points to discuss, but they don't need to block landing this. After this lands, please move the arch `_manual` test maps to `main`.

test/check-application
jelly: Ah, will do! I thought it was related to checkpoint and I need to investigate what the test really does honestly.

packaging/arch/PKGBUILD
martinpitt: I see, thanks! So only cockpit-pcp is split out, then this is correct.

PR #789, opened 2021-10-07
FMF/packit test improvements (backported from starter-kit), fix tests for firefox

martinpitt:
See https://github.com/cockpit-project/starter-kit/pull/500

This should fix the FMF tests due to chromium crashes on F34/rawhide

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -20,3 +20,4 @@ jobs:
     metadata:
       targets:
       - fedora-all
+      - centos-stream-8
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -12,15 +12,12 @@ LOGS="$(pwd)/logs"
 mkdir -p "$LOGS"
 chmod a+w "$LOGS"
 
-# install browser; on RHEL, use chromium from epel
-# HACK: chromium-headless ought to be enough, but version 88 has a crash: https://bugs.chromium.org/p/chromium/issues/detail?id=1170634
-if ! rpm -q chromium; then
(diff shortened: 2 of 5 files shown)


martinpitt:
I pushed a firefix :grin: Works locally, let's see what TF has to say.

martinpitt:
So rawhide now [fails differently](http://artifacts.dev.testing-farm.io/22c738eb-cf48-4958-97b3-85694e166282/) -- not with chromium crashes, but some more tangible failing UI checks. These could happen with chromium as well (if it wouldn't crash), or are some firefox specific issue, although it does not look like the latter at first sight. This needs some deeper investigation, so my proposal would be to land this and investigate rawhide separately. WDYT?

martinpitt:
Interesting failure on [centos-8-stream](http://artifacts.dev.testing-farm.io/d416d17d-369f-4a62-bdaf-82296d654ebb/):

> Error: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "sysfs" to rootfs at "/sys" caused: mount through procfd: operation not permitted: OCI permission denied

This *is* a blocker for this PR, as it enabled C8S. 

I investigated this in 

    tmt run --until report provision --how virtual --image centos-stream-8

which also fails, but totally differently.

When I ssh in as `admin@`, running podman user containers work. That should be the same as `self.admin_s.execute()`, so this is a bit more subtle.

I don't want to block these other fixes for too long, so I'll back out the "enable c8s" commit. I need a new PR for this, as packit can't reconfigure existing test COPRs.

PR #793, opened 2021-10-08
Prevent form submission in Search image dialog

jelly:
There is nothing preventing the form from being submitted once enter is
pressed resulting in the podman page being reloaded.


martinpitt (approved):
This is really weird indeed. I open the image search dialog, type some letters (like "busybox"), press enter, the dialog goes away and the page reloads. But the *second* time it just triggers a new search. With this fix it works correctly now. Thanks!

jelly:
> This is really weird indeed. I open the image search dialog, type some letters (like "busybox"), press enter, the dialog goes away and the page reloads. But the _second_ time it just triggers a new search. With this fix it works correctly now. Thanks!

I noticed the same behaviour, but I am clueless what influences it to not close the dialog a second time. My conclusion was that it has to be some browser quirk/intended behaviour.

PR #794, opened 2021-10-12
webpack: Use default resolve path for npm 7 compatibility

martinpitt:
npm 7 changed how it resolves dependencies, and projects fail to
build with lots of unresolved peer dependencies of PatternFly.

With an absolute path, `resolve.modules` will only look in that
directory; the default is a relative path "node_modules" that just
works [1]. Use that default, as we don't use `$SRCDIR` in this project
anyway.

[1] https://webpack.js.org/configuration/resolve/#resolvemodules

Cherry-picked from starter-kit commit 0abeda35284b84e497.

-----

Curiously, current c-podman main actually *does* build with npm 7 right now. But I have a gut feeling that this is sheer luck, and will bite us at some point. So let's apply the same thing as I did to starter-kit, cockpit, and c-machines.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -10,8 +10,6 @@ const CockpitPoPlugin = require("./src/lib/cockpit-po-plugin");
 
 const webpack = require("webpack");
 
-const nodedir = path.resolve((process.env.SRCDIR || __dirname), "node_modules");
-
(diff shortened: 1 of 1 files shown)


marusak (approved):
Thanks!

PR #799, opened 2021-10-20
Implement prune unused images dialog

garrett:
Here's a quick mockup:

Screenshot 2021-10-20 at 17-05-59 PatternFly 4 • Modal

It's really not much different from what you have in the PR; it's just a few tweaks.

---

# Podman: Prune unused images

Unused images can now be cleaned up in cockpit-podman. This behaves similar to `podman image prune -a` and can delete unused system _and_ user container images at the same time.

139233630-1ca43c26-a276-4f9a-9ac5-5caf0d56b277

140961475-7dbd2501-5a3b-49e0-8edc-f5d5f1f7df53

### DIFF
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -17,6 +17,7 @@ import ImageDetails from './ImageDetails.jsx';
 import { ImageRunModal } from './ImageRunModal.jsx';
 import { ImageSearchModal } from './ImageSearchModal.jsx';
 import { ImageDeleteModal } from './ImageDeleteModal.jsx';
+import PruneUnusedImagesModal from './PruneUnusedImagesModal.jsx';
 import ForceRemoveModal from './ForceRemoveModal.jsx';
 import * as client from './client.js';
 import * as utils from './util.js';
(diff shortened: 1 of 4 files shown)


marusak:
I have two unused images, yet the button to Prune is disabled:
Screenshot from 2021-10-21 11-30-47

garrett:
Is the screenshot up to date? The language differs from the mockup. It should say "Delete..." like the mockup, _not_ "The following..."

It's action-oriented wording, as it's a checkbox. You're not enabling/disabling "the following", you're enabling/disabling the deletion. Plus, having a shorter phrase with bolding is intentional, as it makes it easier to scan and understand what the list is. "The following" is also redundant, as the list is following anyway and there's a colon there.

If the language is like it is in the mockup, then the screenshots need to be updated. If it differs, we need a follow-up ASAP (so it goes in before the release) and would still need the screenshots to be updated in this PR, so the release notes are correct.

jelly:
> Is the screenshot up to date? The language differs from the mockup. It should say "Delete..." like the mockup, _not_ "The following..."

Yes, it's up to date.

> It's action-oriented wording, as it's a checkbox. You're not enabling/disabling "the following", you're enabling/disabling the deletion. Plus, having a shorter phrase with bolding is intentional, as it makes it easier to scan and understand what the list is. "The following" is also redundant, as the list is following anyway and there's a colon there.

Ok, that's an easy fix except getting bold text in a translation string, that's not so easy with our translations so we decided against it.

> If the language is like it is in the mockup, then the screenshots need to be updated. If it differs, we need a follow-up ASAP (so it goes in before the release) and would still need the screenshots to be updated in this PR, so the release notes are correct.

Ok.

PR #801, opened 2021-10-29
Add example for the registry / search entry

jelly:


src/ImageRunModal.jsx
```diff
@@ -731,7 +731,7 @@ export class ImageRunModal extends React.Component {
                           labelIcon={
                               <Popover aria-label={_("Image selection help")}
                                 enableFlip
-                                bodyContent={_("host[:port]/[user]/container[:tag].")}>
+                                bodyContent={_("hostname[:port]/[user]/container[:tag] For example: quay.io/cockpit")}>
```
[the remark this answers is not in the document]
jelly: I've now switched to using Flex's column layout so it shoud be multiple lines.

The example with a tag does not work (yet), as searching with a tab is not possible yet.

src/ImageRunModal.jsx
```diff
@@ -731,7 +731,12 @@ export class ImageRunModal extends React.Component {
                           labelIcon={
                               <Popover aria-label={_("Image selection help")}
                                 enableFlip
-                                bodyContent={_("host[:port]/[user]/container[:tag].")}>
+                                bodyContent={
+                                    <Flex direction={{ default: 'column' }}>
+                                        <FlexItem>{_("hostname[:port]/[user]/container[:tag]")}</FlexItem>
+                                        <FlexItem>{_("For example: quay.io/cockpit")}</FlexItem>
```
martinpitt, line 737: Hah, clever! Can you please take out the container name from the translatable part, and also use a container name which actually exists?

    cockpit.format(_("Example: $0"), "quay.io/libpod/busybox")

That way, translators don't have to precisely copy&paste the container name, and we can re-use the translation for other examples in the future.

martinpitt (changes requested):

src/ImageRunModal.jsx
jelly: Good point, did you mean to suggest that we change the example container search as well?

martinpitt: You mean the first line in the tooltip on line 736? That seems fine to me, it gives the general  pattern.

martinpitt: Your recent force push still has `quay.io/cockpit`, which doesn't exist; also, our own containers are not a particularly good example.

jelly: ok, easy to resolve. But I wanted to give a "searching" example. so what about quay.io/busybox

martinpitt: Oh, I see what you mean -- yes, that sounds fine then. It feels a little strange, as it's not a substring of the actual container names, but that's just the magic first slash :)

jelly: I've now added two examples, for searching and specifying the whole image, what do you think?

martinpitt (approved):
Cheers!

PR #803, opened 2021-11-03
test: Add test for selection options when pruning

jelly:
An superuser has the option to delete his or users images, test both
options in a new test so we have images to prune.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1628,6 +1628,42 @@ class TestApplication(testlib.MachineCase):
         b.click("#image-actions-dropdown")
         b.wait_visible("button:contains(Prune unused images).pf-m-disabled")
 
+    def testPruneUnusedImagesSystemSelections(self):
+        ''' Test the prune unused images selection options'''
+        b = self.browser
+        self.login(True)
+
+        b.click("#image-actions-dropdown")
+        b.click("button:contains(Prune unused images)")
+
+        # Deselect both
+        b.click("#deleteSystemImages")
+        b.click("#deleteUserImages")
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -1628,6 +1628,42 @@ class TestApplication(testlib.MachineCase):
...
+        # Pruning again, should delete all system images
+        b.click("#image-actions-dropdown")
+        b.click("button:contains(Prune unused images)")
+        b.wait_js_func("ph_count_check", ".pf-c-modal-box__body .pf-c-list li", 3)
+        b.click(".pf-c-modal-box button:contains(Prune)")
+        b.wait_js_func("ph_count_check", "#containers-images table tbody", 0)
+
+        # Prune button should now be disabled
+        b.click("#image-actions-dropdown")
+        b.wait_visible("button:contains(Prune unused images).pf-m-disabled")
```
KKoukiou, line 1665: Try to avoid using 'contains' as it won't work in the browser inspector.

KKoukiou (approved):

test/check-application
marusak: > as it won't work in the browser inspector.

That is correct, but I still find it nicer to user `:contains()` as it makes tests that much more readable.

KKoukiou: I guess it's more of a taste thing then. :whatever

jelly: @KKoukiou do you prefer using an hardcoded id? I guess that makes things more predictable?

KKoukiou: This would work nicely when copying pasting css selectors to the browser inspector, but really it's not a big deal.

PR #805, opened 2021-11-05
Containers and Images reduce gaps between title and content

jelly:


KKoukiou (commented):
I think this can be done in a more elegant way, there is already too much custom CSS here, taking a look.

src/Containers.scss
```diff
@@ -35,3 +39,7 @@
 .containers-containers .pf-c-toolbar__content-section {
     gap: var(--pf-global--spacer--sm);
 }
+
+.containers-containers .pf-c-card__header:not(:last-child) {
```
KKoukiou, line 43: I actually disagree with removing the spacer after the header elements. So this rule here and same in the other commit.

The card has a layout that makes sense, that it header -> body -> footer and all these elements should be spaced.

 What we want to do instead here is make the content of these elements not add extra spacers.
There is currently a lot of custom CSS here, which increased overtime, and some of it even is there without being applied. 

One of the interesting facts is that we really don't need to use CardBody which introduces a lot of these spacers that we overwrite, as it's totally fine to add the elements directly into the Card as children.

So I think something closer to https://github.com/KKoukiou/cockpit-podman/commit/634d82a5bfaa7133fd272d7170a52bbab57c89ab would be what I would expect here.

KKoukiou (changes requested):

src/Containers.scss
jelly: This works, but I've found one issue

image

KKoukiou: Hm it looks actually different for me both on chromium and firefox. That missing is padding-bottom coming from the pf-c-card__header element. It can't be browser specific. Did you maybe have some local changes as well?

KKoukiou (approved):

PR #806, opened 2021-11-05
npm: Drop unused svgo dependency

martinpitt:
This depends on "coa", which recently got a trojan horse:
https://github.com/advisories/GHSA-73qr-pfmq-6rp8

As it's not being used at all, just drop it.


martinpitt:
FTR: Aside from [cockpit-certificates](https://github.com/skobyda/cockpit-certificates/pull/77) and c-podman, all our other projects don't pull in svgo.

jelly (approved):

PR #809, opened 2021-11-08
Update deletion text for prune unused images

jelly:


martinpitt:
LGTM and matches @garrett 's original design, but I'll let him do the review. Thanks!

garrett (approved):
Spacing is incorrect. The lists should be indented.

What it looks like:

image


What it should look like:

image

This is after adding `margin-left: var(--pf-global--spacer--md);` to the `11.pf-c-list` element. It would be better to group them, but we kind of can't natively here, I think? 

So the second best thing would probably mean using a PF utility class to add the left margin, such as `pf-u-ml-md` on the list. This comes from the PF utilities https://www.patternfly.org/v4/utilities/spacing/ which have to be included once (and only once) in the CSS.

I looked at the generated HTML and there are empty `<div>`s for some reason. Shouldn't the code use `<></>` instead of `<div></div>` here? (That `<>` is shorthand for `<React.Fragment>`).

Also, there's no association between the list and the checkbox for screen. It should probably use an ARIA association (https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships) as we don't want to nest the list inside of the checkbox (as that would make the list itself clickable).

I'd suggest using `aria-owns="id-of-the-list"` in the checkbox input element.

---

However, it's more important to get the text fixes in ASAP for the release. The rest of this can be another follow-up PR, so I'm marking this as approve. Please still fix the above however.

garrett:
Also, on first glance the modal is a bit too wide in most cases... but these names can get long and having the space for the long strings is a good thing, so it's not really an issue (and is better than always wrapping the names when they get a bit too long). :wink:

jelly:
> Spacing is incorrect. The lists should be indented.
> 
> What it looks like:
> 
> image
> 
> What it should look like:
> 
> image
> 
> This is after adding `margin-left: var(--pf-global--spacer--md);` to the `11.pf-c-list` element. It would be better to group them, but we kind of can't natively here, I think?
> 
> So the second best thing would probably mean using a PF utility class to add the left margin, such as `pf-u-ml-md` on the list. This comes from the PF utilities https://www.patternfly.org/v4/utilities/spacing/ which have to be included once (and only once) in the CSS.

Ok, I've included the css fix locally.
 
> I looked at the generated HTML and there are empty `<div>`s for some reason. Shouldn't the code use `<></>` instead of `<div></div>` here? (That `<>` is shorthand for `<React.Fragment>`).

Aha, that comes from a <FlexItem> actually, which can be dropped.

> Also, there's no association between the list and the checkbox for screen. It should probably use an ARIA association (https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships) as we don't want to nest the list inside of the checkbox (as that would make the list itself clickable).
> 
> I'd suggest using `aria-owns="id-of-the-list"` in the checkbox input element.

Sounds good.

> However, it's more important to get the text fixes in ASAP for the release. The rest of this can be another follow-up PR, so I'm marking this as approve. Please still fix the above however.

Merged, working on a follow up.

PR #810, opened 2021-11-08
Add fallback registry support

jelly:


src/ImageRunModal.jsx
```diff
@@ -737,6 +752,7 @@ export class ImageRunModal extends React.Component {
                                   </button>
                               </Popover>
                           }
+                          helperText={Object.keys(this.props.registries).length === 0 ? _("No search registries configured in /etc/containers/registries.conf") : ""}
```
jelly, line 755: @martinpitt @garrett I am not 100% satisfied with this text, but it is the simplest I could thing of.

martinpitt: I'd honestly just leave this out. The UI provides a fallback, the text is not actionable ("so what do I do? is this an error?" → it's not), and IMHO we don't really want to encourage people to change the default config file.

IMHO it is significantly more important to make it easy and discoverable to search in a new registry which is specified right in the dialog.

If you really want to keep this, please change the fallback to `: undefined`, to avoid rendering an empty string as helper text (possibly the component already DTRT, but it still looks cleaner that way). The standard pattern is `length > 0 && _("..")`

martinpitt (changes requested):
Thanks!

src/ImageRunModal.jsx
```diff
@@ -474,19 +474,33 @@ export class ImageRunModal extends React.Component {
 
         this.setState({ searchFinished: false, searchInProgress: true });
         this.activeConnection = rest.connect(client.getAddress(this.state.isSystem), this.state.isSystem);
+        let searches = [];
+
+        if (Object.keys(this.props.registries).length !== 0) {
```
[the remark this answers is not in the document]
jelly: Registries comes from https://github.com/cockpit-project/cockpit-podman/blob/main/src/app.jsx#L437 so shouldn't be undefined. With no `unqualified-search-registries` it returns an empty dict.

src/ImageRunModal.jsx
```diff
@@ -474,19 +474,33 @@ export class ImageRunModal extends React.Component {
...
+                }
+            }));
+        } else {
+            searches = searches.concat(utils.fallbackRegistries.map(registry =>
+                this.activeConnection.call({
+                    method: "GET",
+                    path: client.VERSION + "libpod/images/search",
+                    body: "",
+                    params: {
+                        term: registry + "/" + value
```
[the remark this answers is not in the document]
jelly: Interesting question, tested it with searching for "/foo" and this actually works.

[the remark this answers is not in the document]
martinpitt: So if I select the "docker.io" registry and search for "quay.io/busybox", it just wouldn't get any results?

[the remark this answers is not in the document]
jelly: Resolved this now, but making searches with a "/" not prefix the registries we have.

src/ImageRunModal.jsx
```diff
@@ -474,19 +474,33 @@ export class ImageRunModal extends React.Component {
...
-            method: "GET",
-            path: client.VERSION + "libpod/images/search",
-            body: "",
-            params: {
-                term: value,
-            },
-        };
-        this.activeConnection.call(options)
+        Promise.all(searches)
                 .then(reply => {
```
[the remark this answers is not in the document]
jelly: Done.

src/ImageRunModal.jsx
garrett: Yeah, it isn't an error and the registries we have by default are fine. We could, as an enhancement, later add a popover somewhere that contains text something to the effect of:

> Searching for images uses system-configured registries in `/etc/containers/registries.conf`. If no registries are configured, `docker.io` and `quay.io` will be used as defaults.

Or we could just leave it out, which is what we probably should do for now.

We can and probably should mention it in the release notes, however.

garrett: Yeah, on trying this locally, it really feels like there's an error with that text (even though it's not indicated as an error or warning). It's better to just leave it out, at least for now.

jelly: Ok! I have left it out, it also feels weird to say, "No configured registries, falling back to docker.io, quay.io".

src/ImageRunModal.jsx
```diff
@@ -516,7 +543,8 @@ export class ImageRunModal extends React.Component {
                             imageResults: images || {},
                             searchFinished: true,
                             searchInProgress: false,
-                            dialogError: ""
+                            dialogError,
+                            dialogErrorDetail,
                         });
                     }
                 })
```
martinpitt, line 550: You can delete the following `.catch()` -- allSettled() never fails.

src/ImageRunModal.jsx
```diff
@@ -474,19 +474,46 @@ export class ImageRunModal extends React.Component {
...
+                        let imageResults = [];
+                        let dialogError = "";
+                        let dialogErrorDetail = "";
+
+                        for (const result of reply) {
+                            if (result.status === "fulfilled") {
+                                imageResults = imageResults.concat(JSON.parse(result.value));
+                            } else {
+                                dialogError = _("Failed to search for new images");
+                                dialogErrorDetail = cockpit.format(_("Failed to search for images: $0"), result.reason);
```
martinpitt, line 514: It'd be nice to have the registry in the error message. Perhaps it's already part of result.reason. Can be a follow-up for sure.

src/ImageSearchModal.jsx
```diff
@@ -75,22 +76,45 @@ export class ImageSearchModal extends React.Component {
...
+                            }
+                        }
+
+                        this.setState({
+                            imageList: results || [], searchInProgress: false,
+                            searchFinished: true, dialogError, dialogErrorDetail
+                        });
+                    }
                 })
                 .catch(ex => {
```
martinpitt, line 119: Dito, allSettled() cannot fail.

martinpitt (changes requested):
This works excellently, many thanks! Just a little cleanup to do still.

martinpitt:
Now the main issue is that searching for "fedoraproject.org/fedora" gives no results, plus it generates this unsightly error message:

image

I also get spurious "malformed XML" and similar error messages while typing; they tend to go away after a second, though. However, that already happens on main (and on Fedora), so not a regression.

jelly:
> Now the main issue is that searching for "fedoraproject.org/fedora" gives no results, plus it generates this unsightly error message:
> 
> image
> 
> I also get spurious "malformed XML" and similar error messages while typing; they tend to go away after a second, though. However, that already happens on main (and on Fedora), so not a regression.

Ah, the real registry lives here registry.fedoraproject.org/fedora not sure hmm why podman gives such a weird error

podman search is better in this regard:
```
[jelle@t14s][~/projects/cockpit-podman]%podman search fedoraproject.org/fedora
Error: 1 error occurred:
	* couldn't search registry "fedoraproject.org": pinging container registry fedoraproject.org: invalid status code from registry 404 (Not Found)
```

src/ImageRunModal.jsx
jelly: That would be nice, but not included in the reply from podman.

martinpitt: No, but we know which registry we put into the search API call.

martinpitt (approved):
Cheers!

PR #813, opened 2021-11-10
npm: Avoid broken eslint-plugin-jsx-a11y 6.5.0 version

martinpitt:
That latest version has a grave regression:

    TypeError: Error while loading rule 'jsx-a11y/alt-text': rule.create is not a function

Downgrade back to 6.4.1 for the time being.

See https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/824

---

This just [broke the release](https://github.com/cockpit-project/cockpit-podman/runs/4162488559?check_suite_focus=true). c-machines is affected as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "eslint-config-standard-react": "^11.0.1",
     "eslint-plugin-flowtype": "^5.2.0",
     "eslint-plugin-import": "^2.22.1",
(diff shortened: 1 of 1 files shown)


martinpitt:
I applied the same hack to c-machines here: https://github.com/cockpit-project/cockpit-machines/pull/403

KKoukiou (approved):

PR #821, opened 2021-11-16
Commit modal improvements and bug fixes

### LINKED ISSUE #818: commit image dialog needs UI attention
141472940-e3e8db06-c815-424d-b7c5-72f8663bef9c

0. Which image is this commit image dialog affecting? It's unclear, as the title is generic.
1. Can OCI vs. Docker format be switched between commits or change from the original image? The radio buttons here imply that. Also, why isn't one selected by default (the current image format type)? Additionally, the case is "OCI" (an initialism) and "Docker" (a proper noun).
2. Can the image name be changed? Or does it have to be changed versus what is currently used?
3. How many of these fields can be prepopulated with the current values?
4. What is the message? Is it a notes/memo field, where the value doesn't matter, but it's freeform text to explain what changed? How long can it be? (Is it a few words, a sentence, paragraph, or more?)
5. What does pausing and variables have to do with "command"? Right now, the UI implies that they're related. They should be separate fields if they're not related.
6. What does pausing do? Does that affect the container, not the image?
7. What does "set container on build variables" mean? Does it set variables that are used by default when a container is created by this image?
8. Are previously existing variables shown in this commit image dialog?

I can help with the UI once I understand the above a bit better.

marusak:
Fixes #818

Screenshot from 2021-11-23 16-48-30

### DIFF
--- a/src/ContainerCommitModal.jsx
+++ b/src/ContainerCommitModal.jsx
@@ -21,10 +21,11 @@ class ContainerCommitModal extends React.Component {
             tag: "",
             author:"",
             command: props.container.Command ? utils.quote_cmdline(props.container.Command) : "",
-            pause: true,
+            pause: false,
             selectedFormat: "oci",
             commitInProgress: false,
             useDocker: false,
+            nameError: "",
         };
 
         this.handleInputChange = this.handleInputChange.bind(this);
@@ -32,14 +33,26 @@ class ContainerCommitModal extends React.Component {
     }
 
     handleInputChange(targetName, value) {
-        this.setState({
-            [targetName]: value
-        });
+        const newState = { [targetName]: value };
+
+        if (targetName === "imageName" || targetName === "tag")
+            newState.nameError = "";
+
+        this.setState(newState);
     }
 
-    handleCommit() {
-        if (!this.state.imageName) {
-            this.setState({ dialogError: "Image name is required" });
+    handleCommit(force) {
+        if (!force && !this.state.imageName) {
+            this.setState({ nameError: "Image name is required" });
+            return;
+        }
+
+        let full_name = this.state.imageName + ":" + (this.state.tag ? this.state.tag : "latest");
+        if (full_name.indexOf("/") < 0)
+            full_name = "localhost/" + full_name;
+
+        if (!force && this.props.localImages.some(image => image.isSystem === this.props.container.isSystem && image.Name === full_name)) {
+            this.setState({ nameError: "Image name is not unique" });
             return;
         }
 
@@ -69,7 +82,7 @@ class ContainerCommitModal extends React.Component {
             commitData.changes.push(cmdData);
         }
 
-        this.setState({ commitInProgress: true });
+        this.setState({ commitInProgress: true, nameError: "", dialogError: "" });
         client.commitContainer(this.props.container.isSystem, commitData)
                 .then(() => this.props.onHide())
                 .catch(ex => {
@@ -84,9 +97,12 @@ class ContainerCommitModal extends React.Component {
     render() {
         const commitContent =
             <Form isHorizontal>
-                <FormGroup fieldId="commit-dialog-image-name" label={_("New image name")}>
+                <FormGroup fieldId="commit-dialog-image-name" label={_("New image name")}
+                           validated={this.state.nameError ? "error" : "default"}
+                           helperTextInvalid={this.state.nameError}>
                     <TextInput id="commit-dialog-image-name"
                                value={this.state.imageName}
+                               validated={this.state.nameError ? "error" : "default"}
                                onChange={value => this.handleInputChange("imageName", value)} />
                 </FormGroup>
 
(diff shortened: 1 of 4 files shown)


src/ContainerCommitModal.jsx
```diff
@@ -112,113 +93,73 @@ class ContainerCommitModal extends React.Component {
...
                 <FormGroup fieldId="commit-dialog-image-tag" label={_("Tag")}>
                     <TextInput id="commit-dialog-image-tag"
+                               placeholder="latest" // Do not translate
                                value={this.state.tag}
                                onChange={value => this.handleInputChange("tag", value)} />
                 </FormGroup>
 
                 <FormGroup fieldId="commit-dialog-author" label={_("Author")}>
                     <TextInput id="commit-dialog-author"
+                               placeholder={_("Example, Your Name <yourname@exmaple.com>")}
```
garrett, line 117:
```suggestion
                               placeholder={_("Example, Your Name <yourname@example.com>")}
```

garrett: Typo fix

src/ContainerCommitModal.jsx
```diff
@@ -112,113 +93,73 @@ class ContainerCommitModal extends React.Component {
...
-                    <Checkbox id="commit-dialog-setonbuild" isDisabled={this.state.onbuildDisabled}
-                              isChecked={this.state.setonbuild}
-                              onChange={value => this.handleInputChange("setonbuild", value)}
-                              label={_("Set container on build variables")} />
-                    {this.state.setonbuild && onbuilds}
+                              label={_("Pause container when creating image")} />
+                    <Checkbox id="commit-dialog-docker"
+                              isChecked={this.state.useDocker}
+                              onChange={value => this.handleInputChange("useDocker", value)}
+                              description={_("Docker format is useful when sharing the image with Docker or Mody Engine")}
```
garrett, line 136:
```suggestion
                              description={_("Docker format is useful when sharing the image with Docker or Moby Engine")}
```

Another typo fix

src/ContainerCommitModal.jsx
```diff
@@ -112,113 +93,73 @@ class ContainerCommitModal extends React.Component {
...
                 </FormGroup>
             </Form>;
 
         return (
             <Modal isOpen
                    showClose={false}
                    position="top" variant="medium"
-                   title={_("Commit image")}
+                   title={_("Commit container")}
+                   description={cockpit.format(_("Create a new image based on the current state of the $0 container."), this.props.container.Names)}
```
garrett, line 146: Are we able to wrap the container name with a `<strong>` element? Or do descriptions only support text-only?

garrett:
Is the tag without the `:`? What happens when you do `:foo`? Does it tag it with `:foo`, `::foo`, or just fail?

marusak:
> Is the tag without the :? What happens when you do :foo? Does it tag it with :foo, ::foo, or just fail?

> (It's good and fine without :, but it should accept :tagname too as the tagname tag, as some people would be used to typing :, right?)

The tag is without `:`. `:` is just separator, not part of tag itself. If you do `:foo` it will fail. But indeed, we could check for leading `:` and drop it to make it more user friendly.

src/ContainerCommitModal.jsx
martinpitt: Marius has a clever [fmt_to_fragments()](https://github.com/cockpit-project/cockpit/blob/main/pkg/storaged/utilsx.jsx) helper which can help with translating strings with markup. We could move that to pkg/lib for sure.

marusak:
> Is the tag without the :? What happens when you do :foo? Does it tag it with :foo, ::foo, or just fail?

> (It's good and fine without :, but it should accept :tagname too as the tagname tag, as some people would be used to typing :, right?)

The tag is without `:`. `:` is just separator, not part of tag itself. If you do `:foo` it will fail. But indeed, we could check for leading `:` and drop it to make it more user friendly.

src/ContainerCommitModal.jsx
```diff
@@ -112,113 +93,73 @@ class ContainerCommitModal extends React.Component {
...
+                           helperTextInvalid={this.state.nameError}>
                     <TextInput id="commit-dialog-image-name"
                                value={this.state.imageName}
+                               validated={this.state.nameError ? "error" : "default"}
                                onChange={value => this.handleInputChange("imageName", value)} />
                 </FormGroup>
 
                 <FormGroup fieldId="commit-dialog-image-tag" label={_("Tag")}>
                     <TextInput id="commit-dialog-image-tag"
+                               placeholder="latest" // Do not translate
```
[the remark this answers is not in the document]
jelly: Ah, it's the default value, so I think we should just set it as value instead of a placeholder?

jelly (changes requested):

src/ContainerCommitModal.jsx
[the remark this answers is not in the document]
marusak: @garrett I followed here the design, WDYT?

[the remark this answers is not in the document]
garrett: "latest" is not an example; it is what you'd get by default if you don't enter anything.

https://www.patternfly.org/v4/components/form/design-guidelines/#placeholder-text-0

> If your placeholder text demonstrates a correct input, introduce it as an example so that users can distinguish impermanent examples from field labels and their own inputted text.

If we think it looks too inconsistent, then we can drop the latest and add some helper text saying that if no tag is given, then it defaults to "latest".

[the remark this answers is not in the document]
garrett: What happens if author is blank?

src/ContainerCommitModal.jsx
marusak: Done

src/util.js
```diff
@@ -174,3 +176,17 @@ export function compare_versions(a, b) {
...
+
+    function replace(part) {
+        if (part[0] == "$") {
+            return args[parseInt(part.slice(1))];
+        } else
+            return part;
+    }
+
+    return React.createElement.apply(null, [React.Fragment, { }].concat(fmt.split(/(\$[0-9]+)/g).map(replace)));
+}
```
jelly, line 192: This is already in src/lib with this commit https://github.com/cockpit-project/cockpit-podman/commit/47a97efebdd036b0f45f2a7388beea8dfbaac7cd

jelly (changes requested):

src/util.js
marusak: Ooops, sorry, fixed

jelly (approved):

PR #827, opened 2021-11-19
fmf: Skip TestApplication.testCheckpointRestoreCGroupsV2 on Fedora

martinpitt:
This has started to fail a while ago, and thus now all Fedora packit
tests are constantly red. Until we get to the root cause (which is a bit
subtle), let's not introduce more regressions and make tests green
again.

### DIFF
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -30,6 +30,11 @@ fi
 
 EXCLUDES=""
 
+if [ "${TEST_OS#fedora-}" != "$TEST_OS" ]; then
+    # FIXME: this started to fail on the testing farm, need to investigate
+    EXCLUDES="$EXCLUDES TestApplication.testCheckpointRestoreCGroupsV2"
+fi
+
 RC=0
 test/common/run-tests --nondestructive --machine 127.0.0.1:22 --browser 127.0.0.1:9090 $EXCLUDES || RC=$?
 


martinpitt:
[example failure](http://artifacts.dev.testing-farm.io/0c3c8995-d70f-46d3-ab36-0aa205ff0ba1/). I can't try to reproduce it with local TMT right now (:bullettrain_front: ), but if anyone wants to, https://cockpit-project.org/blog/fmf-unified-testing.html has some details.

marusak:
I see it in my last PR on [fedora-35](https://logs.cockpit-project.org/logs/pull-821-20211118-112447-25a54de3-fedora-35/log.html#4) and rhel-9. I can barely reproduce it locally, but I am fighting with it. Not smarter yet, but glad to see, it is not related to my changes. It feels mostly like if the VM runs long enough, then this becomes a problem.

martinpitt:
F34/35 (and both centos) are green, but rawhide [failed](http://artifacts.dev.testing-farm.io/b312cce8-b511-40c0-9ed2-c7b8d3557d42/) on a temporary quay.io networking . Retrying.

martinpitt:
Oh come on, Friday. Now all Fedoras succeeded, but now [c9s failed](http://artifacts.dev.testing-farm.io/4ff9e599-02ad-4822-a9e8-c81e319e7a5b/) in `testRunImageUser()`. Third time's the charm!

martinpitt:
@marusak , @jelly : I'll let you decide on this, based on your progress with trying to reproduce/investigate this. I'm happy to try myself next week too -- this shouldn't stay for long for sure.

marusak:
So the issue is, that `descriptors.json` file contains `["/dev/pts/0","/dev/pts/0","/dev/pts/0"]]` instead of `["/dev/pts/0","/dev/pts/0","/dev/pts/0"]` (trailing `]`). This is [generated by podman](https://github.com/containers/crun/commit/21624359392108697732a99a6e402d91e9daba51) and points to some bug that they have.
This file is located in `<static>//checkpoint/descriptors.json` where static is `podman inspect --format '{{.StaticDir}}' <container_name>`.

I would love to get some reproducer and report this to podman, but I can reproduce this only under very specific conditions using our tests. I'll look more if there is something specific which could help us.

marusak:
My theory - We do restore twice. First time on stopped container and then on running one.
It works in this case: (first line is content of `descriptors.json` before first restore and second line before second (often failing) one)
```
["/dev/null","/dev/null","pipe:[93613]"]
["/dev/pts/0","/dev/pts/0","/dev/pts/0"]
```
but not in this case:
```
["/dev/null","/dev/null","pipe:[100769]"]
["/dev/pts/0","/dev/pts/0","/dev/pts/0"]]
```

So it seems, that it does not clear the file and therefore it leaves there part of the old file - which makes the json invalid. And it also explains, why it fails only in gating and in my new PR - we need to get pids over 100000 and if this test runs too late, there were enough processes already.

martinpitt:
@marusak :clap: eeexcellent debugging fu, thanks for figuring this out! This explanation makes total sense now.

PR #829, opened 2021-11-22
Allow CPU shares to be set when owner is system

jelly:


src/ImageRunModal.jsx
```diff
@@ -837,8 +837,19 @@ export class ImageRunModal extends React.Component {
...
-                        {(this.state.owner === systemOwner || (this.state.image && this.state.image.isSystem)) &&
-                            <FormGroup fieldId='run-image-cpu-priority' label={_("CPU shares")}>
+                        {this.state.owner === systemOwner &&
+                            <FormGroup
+                              fieldId='run-image-cpu-priority'
+                              label={_("CPU shares")}
+                              labelIcon={
+                                  <Popover aria-label={_("CPU Shares help")}
+                                      enableFlip
+                                      bodyContent={_("CPU shares determine how much CPU time a container can use relative to other running containers. Set it to a lower number for lower priority or a higher number for higher.")}>
```
martinpitt, line 847: IMHO the second sentence is rather redundant, and makes the string harder to translate. Can we trim this a bit?

    The more shares a container has, the more CPU time it gets relative to other running containers.

src/ImageRunModal.jsx
```diff
@@ -837,8 +837,19 @@ export class ImageRunModal extends React.Component {
                             </Flex>
                         </FormGroup>
 
-                        {(this.state.owner === systemOwner || (this.state.image && this.state.image.isSystem)) &&
-                            <FormGroup fieldId='run-image-cpu-priority' label={_("CPU shares")}>
+                        {this.state.owner === systemOwner &&
```
martinpitt, line 840: Did you deliberately change the condition here?

martinpitt (changes requested):

src/ImageRunModal.jsx
jelly: Yes, for the new create container dialog the cpu shares option was not shown.

jelly: Ah woops, I put it in the wrong commit :)

src/ImageRunModal.jsx
garrett: How's this?

```suggestion
                                      bodyContent={_("CPU shares determine priority of running containers. Default priority is 1024. Higher numbers prioritize containers. Lower numbers decrease priority.")}>
```

garrett:
We also talked about showing "default priority" / "higher priority" / "lower priority" next to the number. (This can be added outside of this PR.)

src/ImageRunModal.jsx
jelly: We already shown the default as 1024 as greyed out, should we mention it in the information popover a well.

garrett: It does help to explain what's going on.

jelly: Ok,`Higher numbers prioritize containers` sounds a bit vague to me, as one could read it as prioritizes all containers.

martinpitt: How about singular? "A higher number prioritizes this container"?

jelly: Sounds good, I made the suggested changes.

src/ImageRunModal.jsx
```diff
@@ -844,8 +844,19 @@ export class ImageRunModal extends React.Component {
...
-                        { this.state.image && this.state.image.isSystem &&
-                            <FormGroup fieldId='run-image-cpu-priority' label={_("CPU shares")}>
+                        {this.state.owner === systemOwner &&
+                            <FormGroup
+                              fieldId='run-image-cpu-priority'
+                              label={_("CPU shares")}
+                              labelIcon={
+                                  <Popover aria-label={_("CPU Shares help")}
+                                      enableFlip
+                                      bodyContent={_("CPU shares determine priority of running containers. Default priority is 1024. A higher number prioritize this container. Lower numbers decrease priority.")}>
```
garrett, line 854: The statements should be internally consistent with each other.

For example:

```suggestion
                                      bodyContent={_("CPU shares determine the priority of running containers. Default priority is 1024. A higher number prioritizes this container. A lower number decreases priority.")}>
```

Alternate:

```suggestion
                                      bodyContent={_("CPU shares determine the priority of running containers. Default priority is 1024. Higher numbers prioritize this container. Lower numbers decrease the priority of this container.")}>
```

(FWIW: I like the first option better.)

garrett (dismissed):
We need to fix the text a bit. It wasn't consistent with itself. I've provided some solutions.

martinpitt:
Looks good, thanks! This will need a rebase/updated pixel test, though. Will formally approve after that.

martinpitt (approved):
Cheers!

PR #830, opened 2021-11-22
Remove created container if it doesn't start

jelly:
When a user tries to create a container of which the port is already
used the container doesn't start and is not removed. If the user
corrects the host port and clicks create again the container won't be
created as the name is already in use so removed it on error.

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -382,10 +382,21 @@ export class ImageRunModal extends React.Component {
                         client.postContainer(isSystem, "start", reply.Id, {})
                                 .then(() => this.props.close())
                                 .catch(ex => {
-                                    this.setState({
-                                        dialogError: _("Container failed to be started"),
-                                        dialogErrorDetail: cockpit.format("$0: $1", ex.reason, ex.message)
-                                    });
+                                    // If container failed to start remove it, so a user can fix the settings and retry and
+                                    // won't get another error that the container name is already taken.
+                                    client.delContainer(isSystem, reply.Id, true)
+                                            .then(() => {
+                                                this.setState({
+                                                    dialogError: _("Container failed to be started"),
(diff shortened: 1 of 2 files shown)


src/ImageRunModal.jsx
```diff
@@ -378,10 +378,20 @@ export class ImageRunModal extends React.Component {
                         client.postContainer(isSystem, "start", reply.Id, {})
                                 .then(() => this.props.close())
                                 .catch(ex => {
-                                    this.setState({
-                                        dialogError: _("Container failed to be started"),
-                                        dialogErrorDetail: cockpit.format("$0: $1", ex.reason, ex.message)
-                                    });
+                                    // If container failed to start remove it, so a user can fix the settings and retry and
+                                    // won't get another error that the container name is already taken.
+                                    client.delContainer(isSystem, reply.Id, true).then(() => {
```
martinpitt, line 383: Please move the .then() to the next line. It should (1) be on the same intentation as the corresponding .catch(), and it's hard to read these chained calls.

src/ImageRunModal.jsx
```diff
@@ -378,10 +378,20 @@ export class ImageRunModal extends React.Component {
...
+                                        this.setState({
+                                            dialogError: _("Failed to clean up container"),
```
martinpitt, line 385: I think you mixed up the error messages. Here the start failed, but the cleanup succeeded, and below the opposite.

martinpitt (changes requested):
Thanks!

src/ImageRunModal.jsx
jelly: Yes

src/ImageRunModal.jsx
jelly: Done

martinpitt (approved):
Thanks!

marusak (approved):

PR #831, opened 2021-11-23
Make the system/user text bold in unused images

jelly:
As per design, let the system/user images stand out when pruning as an
administrator. As in https://github.com/cockpit-project/cockpit-podman/pull/799#issuecomment-948586523


martinpitt (approved):
Thanks!

martinpitt (approved):
Thanks! There is a potential gotcha that e.g. "user" and "system" have two different grammatical genders in languages, and it might be impossible to translate that properly. But at least French is fine (the main language that I know which is prone to that kind of problem), so I think we can land that for now.

PR #832, opened 2021-11-23
fmf: Treat rawhide like Fedora 35

marusak:


martinpitt (changes requested):
Now I remember why I did *not* do this the last time -- I actually created a naughty/fedora-36 in bots when I ran into a regression on rawhide. I thought that would be cleaner than adding these hacks, which keep getting outdated. I think this would work for you as well.

Similarly, if we add some `m.image` special case to the tests, we should add "fedora-36" right away. All our tests run in packit, so we have them all covered anyway (unlike cockpit, where a lot of tests are not nondestructive).

Sorry for not thinking about this earlier!

marusak:
Sent https://github.com/cockpit-project/bots/pull/2656

PR #839, opened 2021-11-29
tests: autoconvert all python string formatting to fstring

KKoukiou:
Let's make this string formatting method the only one to use from now
on.

These files were autoconverted with flynt.


martinpitt (approved):
Yay death to .format()!

PR #841, opened 2021-11-30
test: Drop installing basic cockpit packages on Debian/Ubuntu

martinpitt:
Since https://github.com/cockpit-project/bots/commit/b06bd37b0ad our
Debian/Ubuntu test VMs have the basic cockpit packages pre-installed.
Drop the apt calls to install them, so that we stop getting surprise
breakages due to new cockpit versions. These will now properly appear on
image refreshes.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -4,9 +4,6 @@ set -eu
 
 # for Debian based images, build and install debs; for RPM based ones, the locally built rpm gets installed separately
 if [ -d /var/tmp/debian ]; then
-    apt-get update
-    eatmydata apt-get install ${APT_INSTALL_OPTIONS:-} -y cockpit-ws cockpit-system podman
-
     # build source package
     cd /var/tmp
     TAR=$(ls cockpit-podman-*.tar.xz)


test/vm.install
```diff
@@ -4,9 +4,6 @@ set -eu
 
 # for Debian based images, build and install debs; for RPM based ones, the locally built rpm gets installed separately
 if [ -d /var/tmp/debian ]; then
-    apt-get update
-    eatmydata apt-get install ${APT_INSTALL_OPTIONS:-} -y cockpit-ws cockpit-system podman
```
KKoukiou, line 8: This was not only installing basic cockpit packages as the commit message suggests, but also podman package.
Do you know why that was there in the first place? Do we pre-install podman in our images actually or we rely on this here for cockpit-podman dependencies to exist?

martinpitt: Our Debian/Ubuntu images have pre-installed podman since February: https://github.com/cockpit-project/bots/commit/1d94c0e7a88d -- for the same reason. I am happy to re-push to adjust the commit message, or we just do it on squash&merge?

KKoukiou: No need - thanks for the clarification.

KKoukiou (approved):

Issue #816, opened 2021-11-13
Option to select text mode for Environmental Variable


firmanjml:
It would be nice to have an option to switch to text field or just plain text mode to copy and paste existing .env text

jelly:
Instead of adding another field, I've implemented a PoC which supports pasting an env like file like:|

```
COCKPIT=12
PODMAN=true
```

https://user-images.githubusercontent.com/67428/145414425-bd7d4540-60bb-45d5-ba4b-9304a93cdd0b.mp4

firmanjml:
@jelly This is great! I would love to have this feature on production because it’s useful

garrett:
@jelly: I like this! It's nice and simple!

ohsix:
the fix for this made it impossible to enter an equals sign in the value field

PR #850, opened 2021-12-09
Allow inserting multiple environment variables

jelly:
Podman users might have file or a list of environment variables they
want to add to a new container. As adding them one by one is quite
cumbersome handle inserting FOO=bar and multiple lines of such formatted
env variables.

Closes: #816

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -6,6 +6,7 @@ import {
     Form, FormGroup, FormFieldGroup, FormFieldGroupHeader,
     FormSelect, FormSelectOption,
     Grid,
+    HelperText, HelperTextItem,
     Modal, Select, SelectVariant,
     SelectOption, SelectGroup,
     TextInput, Tabs, Tab, TabTitleText,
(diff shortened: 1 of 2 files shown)


src/ImageRunModal.jsx
```diff
@@ -119,18 +120,37 @@ const PublishPort = ({ id, item, onChange, idx, removeitem, itemCount }) =>
         </Grid>
     );
 
+const handleEnvValue = (key, value, idx, onChange, additem, itemCount) => {
+    // Allow the input of KEY=VALUE seperated value pairs for bulk import
+    if (value.includes('=')) {
+        const parts = value.trim().split(" ");
```
marusak, line 126: Why splitting on space? Shouldn;t we split on new line? (although it works)

src/ImageRunModal.jsx
```diff
@@ -119,18 +120,37 @@ const PublishPort = ({ id, item, onChange, idx, removeitem, itemCount }) =>
...
+        let index = idx;
+        for (const part of parts) {
+            const [envKey, envVar] = part.split('=', 2);
```
marusak, line 129: Can you please check if both values are defined? If for example you would have empty line in between items or spaces after item, it leads to many empty items. We should skip those.
Screenshot from 2021-12-10 12-52-43
Screenshot from 2021-12-10 12-52-06

test/check-application
```diff
@@ -1360,8 +1360,8 @@ class TestApplication(testlib.MachineCase):
         b.set_input_text('#run-image-dialog-env-2-value', 'GRAPE')
         b.click('#run-image-dialog-env-2-btn-close')
         b.click('.env-form .btn-add')
-        b.set_input_text('#run-image-dialog-env-2-key', 'RHUBARB')
-        b.set_input_text('#run-image-dialog-env-2-value', 'STRAWBERRY')
+        # Test inputting an key=var entry
+        b.set_input_text('#run-image-dialog-env-2-value', "RHUBARB=STRAWBERRY", value_check=False)
```
marusak, line 1364: Can you please test it with more values at the same time? To test the "bulk" part as well.

marusak (changes requested):
Looks great! One question, one improvement and making tests happy :)

src/ImageRunModal.jsx
jelly: It seems that the browser for a single line input converts newlines to spaces.

test/check-application
jelly: Sure, done.

marusak (approved):
Thank you!

PR #851, opened 2021-12-10
Add restart policy to create new container

jelly:
Implement the option to configure a restart policy for containers, the
default is no restart policy which means the container exits.
'on-failure' means it's restarted when the container exists with a
non-zero exit code for a maximum of retries. Always means it is always
restarted, to make this work on reboot the podman-restart service needs
to be enabled. For user containers this means `loginctl enable-linger`
has to set if a user wants his container to start without logging in
after a reboot.

---

# Podman restart policy

For system containers it is now possible to set a restart policy for new containers.

image

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -5,7 +5,7 @@ import {
     EmptyState, EmptyStateBody,
     Form, FormGroup, FormFieldGroup, FormFieldGroupHeader,
     FormSelect, FormSelectOption,
-    Grid,
+    Grid, GridItem,
     HelperText, HelperTextItem,
     Modal, Select, SelectVariant,
     SelectOption, SelectGroup,
@@ -317,6 +317,8 @@ export class ImageRunModal extends React.Component {
             validationFailed: {},
             volumes: [],
             runImage: true,
+            restartPolicy: "no",
+            restartTries: 5,
             activeTabKey: 0,
             owner: this.props.systemServiceAvailable ? systemOwner : this.props.user,
             /* image select */
@@ -402,6 +404,18 @@ export class ImageRunModal extends React.Component {
                             record.options.push(volume.selinux);
                         return record;
                     });
+        }
+
+        if (this.state.restartPolicy !== "no") {
+            createConfig.restart_policy = this.state.restartPolicy;
+            if (this.state.restartPolicy === "on-failure" && this.state.restartTries !== null) {
+                createConfig.restart_tries = parseInt(this.state.restartTries);
+            }
+            // Enable podman-restart.service for system containers, for user
+            // sessions enable-linger needs to be enabled for containers to start on boot.
+            if (this.state.restartPolicy === "always" && this.props.systemServiceAvailable) {
+                this.enablePodmanRestartService();
+            }
         }
 
         return createConfig;
@@ -721,6 +735,15 @@ export class ImageRunModal extends React.Component {
         return domain;
     }
 
+    enablePodmanRestartService = () => {
+        const argv = ["systemctl", "enable", "podman-restart.service"];
+
+        cockpit.spawn(argv, { superuser: "require", err: "message" })
+                .catch(err => {
+                    console.warn("Failed to start podman-restart.service:", JSON.stringify(err));
+                });
+    }
+
     render() {
         const { image } = this.props;
         const dialogValues = this.state;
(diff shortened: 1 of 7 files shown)


src/ImageRunModal.jsx
```diff
@@ -908,6 +931,46 @@ export class ImageRunModal extends React.Component {
                             <Checkbox isChecked={this.state.runImage} id="run-image-dialog-start-after-creation"
                                       onChange={value => this.onValueChanged('runImage', value)} label={_("Start after creation")} />
                         </FormGroup>
+                        <Grid hasGutter md={6}>
+                            { this.props.userServiceAvailable && this.props.systemServiceAvailable &&
```
marusak, line 935: Why is this guarded by both services being present?

src/ImageRunModal.jsx
```diff
@@ -908,6 +931,46 @@ export class ImageRunModal extends React.Component {
...
+                            <GridItem span={dialogValues.restartPolicy === "on-failure" ? 6 : 12}>
```
marusak, line 936: This logic behaves a bit weirdly on mobile:
Screen Shot 2021-12-14 at 09 16 36

I think we want the first select always be about half of the screen. I think we can do something like `max-width: 70px` or something - that way it is not through the whole dialog just for `No`. And then let PF doing its wrapping on mobile.

src/ImageRunModal.jsx
```diff
@@ -908,6 +931,46 @@ export class ImageRunModal extends React.Component {
...
+                                enableFlip
+                                bodyContent={_("Restart policy to follow when containers exit.")}>
+                                  <button onClick={e => e.preventDefault()} className="pf-c-form__group-label-help">
+                                      <OutlinedQuestionCircleIcon />
+                                  </button>
+                              </Popover>
+                          }
+                                >
+                                    <FormSelect id="run-image-dialog-restart-policy"
+                              aria-label={_("Restart Policy")}
```
marusak, line 949: Can we keep sentence case here as well? It will ease translators job.

test/check-application
```diff
@@ -1326,6 +1326,13 @@ class TestApplication(testlib.MachineCase):
         # Set up command line
         b.set_input_text('#run-image-dialog-command', "sh -c 'for i in $(seq 20); do sleep 1; echo $i; done; sleep infinity'")
 
+        if auth:
+            # Set restart policy to 8 retries
+            b.set_val("#run-image-dialog-restart-policy", "on-failure")
+            b.set_input_text('#run-image-dialog-restart-retries', '3')
+        else:
+            b.wait_not_present("#run-image-dialog-restart-policy")
+
```
marusak, line 1335: Hmm, so enabling system service also works for user containers?

test/check-application
```diff
@@ -1326,6 +1326,13 @@ class TestApplication(testlib.MachineCase):
         # Set up command line
         b.set_input_text('#run-image-dialog-command', "sh -c 'for i in $(seq 20); do sleep 1; echo $i; done; sleep infinity'")
 
+        if auth:
+            # Set restart policy to 8 retries
```
marusak, line 1330: Also you say `to 8 retries` but set it to 3 :)

test/check-application
marusak: It should also test that the service is actually started

test/check-application
```diff
@@ -1419,6 +1426,9 @@ class TestApplication(testlib.MachineCase):
             cpuShares = self.execute(auth, "podman inspect --format '{{.HostConfig.CpuShares}}' busybox-with-tty || podman inspect --format '{{.HostConfig.CPUShares}}' busybox-with-tty").strip()
             self.assertEqual(cpuShares, '512')
 
+            restartPolicy = self.execute(auth, "podman inspect --format '{{.HostConfig.RestartPolicy}}' busybox-with-tty").strip()
```
marusak, line 1429: Can you please also add test that on some other container it is not set (or set to `no`) and then also check `always` option? Thanks!

marusak (changes requested):
Looks good! Just some concerns and questions

src/ImageRunModal.jsx
jelly: hmm actually it should most likely follow the cpuShares logic `this.state.owner === systemOwner`

test/check-application
jelly: > Hmm, so enabling system service also works for user containers?

No? It should only be enabled for user containers.

src/ImageRunModal.jsx
garrett: > max-width: 70px

Never use arbitrary pixels for things that contain text. It should be sized via PatternFly, CSS variables, or CSS units that are based on font size (or in some cases, screen size).

This layout is lacking breakpoints for whatever reason. 

In this case, it's because the gridItem has variable spanning that's hardcoded without respect to mobile, even though the grid itself is using md.

If we were to use PF sizing, it would need breakpoints. Something like `pf-u-w-50` would be `pf-u-w-50-on-md` (for example, although the size and breakpoint varies).

garrett (changes requested):
I'm playing around with it locally and it's strange that the dropdown changes widths. It should be fixed to half when there's space.

Also: Oddly, the width doesn't line up with other splits above.

image

It'd be less problematic if we sized the memory limit and CPU shares to be approximately to the expected size. Right now, they look like they're supposed to line up with the grid... but they don't.

And options should be the last item, not restart policy.

This is what I mean:

image

image

But, really, the maximum retries should also be constrained like the rest:

image


The vertical ryhytm of the form is also inconsistent. For something like command, which has an additional line, it's expected, but for the rest after with one line, there's inconsistent spacing between each line.

Here's what I mean visually... the light orange boxes are all the same size. I sized it based on the baseline of options to restart policy. But then, going up, options to CPU shares has less space and CPU shares to memory limit has _more_ space. Weird, right?
 
Screenshot from 2021-12-14 10-16-22

src/ImageRunModal.jsx
```diff
@@ -902,8 +925,49 @@ export class ImageRunModal extends React.Component {
...
+                                <FormGroup fieldId='run-image-dialog-restart-retries'
+                                  label={_("Maximum retries")}>
+                                    <TextInput
+                              type="number"
+                              id="run-image-dialog-restart-retries"
+                              className="run-image-dialog-restart-retries"
+                              value={dialogValues.restartTries}
+                              step={1}
+                              min={1}
+                              max={5}
```
marusak, line 964: I don't see any limit from podman side - so lets not introduce any arbitrary value ourselves.

marusak (approved):

garrett:
The screenshot in the description is lacking context. It should show the entire dialog.

Additionally, shouldn't the options come after the restart policy?

PR #852, opened 2021-12-13
Bump test/common API to 259

marusak:
This will fix the tests when running against cockpit 259, where the
shell got changed to not render the `#machine-reconnect` button all the
time any more.

This should fix broken rawhide tests


martinpitt (approved):
Heh, same thing that I stumbled over in https://github.com/cockpit-project/cockpit-machines/pull/478 . Thanks!

PR #854, opened 2021-12-13
Enable checkbox when input is focussed

### LINKED ISSUE #849: create modal: Clicking in the input in the checkbox form combos does not activate the input
When the memory limit or CPU shares is unchecked and someone clicks in the disabled areas, it should check and enable the option.

Screenshot:
image

After clicking 512, it should look like this:
image

It should be the same for CPU shares.

Additionally: Clicking on MiB should also activate the memory limit.

jelly:
When a user clicks on the cpu shares / memory limit input automatically
enable the corresponding checkboxes. For a <FormSelect> the onFocus
method does not trigger when a user clicks on it, making it impossible
to enable the checkbox.

Closes #849

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1313,6 +1313,12 @@ class TestApplication(testlib.MachineCase):
 
         # CPU shares work only with system containers
         if auth:
+            # Check that the checkbox is enabled when clicked on the field
+            b.wait_visible("#run-image-dialog-cpu-priority-checkbox:not(:checked)")
+            b.click('#run-image-cpu-priority')
+            b.wait_visible("#run-image-dialog-cpu-priority-checkbox:checked")
+            b.set_checked("#run-image-dialog-cpu-priority-checkbox", False)
+
             b.set_checked("#run-image-dialog-cpu-priority-checkbox", True)
             b.wait_visible("#run-image-dialog-cpu-priority-checkbox:checked")
             b.wait_visible('#run-image-dialog-cpu-priority input[value="1024"]')
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -862,6 +862,7 @@ export class ImageRunModal extends React.Component {
                                    step={1}
                                    min={0}
                                    isReadOnly={!this.state.memoryConfigure}
+                                   onClick={() => !this.state.memoryConfigure && this.onValueChanged('memoryConfigure', true)}
                                    onChange={value => this.onValueChanged('memory', value)} />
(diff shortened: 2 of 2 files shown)


marusak (changes requested):
Now it gets enabled when you just move with `<Tab>` through the dialog. That does not seem like intended behavior, is it?

If you change it to `onClick` it seems to behave better.

Also can you please rebase (to fix packit tests) and write some test?

jelly:
> Now it gets enabled when you just move with `<Tab>` through the dialog. That does not seem like intended behavior, is it?
> 
> If you change it to `onClick` it seems to behave better.
> 
> Also can you please rebase (to fix packit tests) and write some test?

Thanks, oops forgot to rebase.

garrett:
> Now it gets enabled when you just move with <Tab> through the dialog. That does not seem like intended behavior, is it?

Correct, it should not enable on tab.

If someone clicks or starts typing (non-tab) keys, it should focus.

Related: This makes me wonder if we need some sort of ARIA relationship to indicate that it is controlled by the checkbox. :thinking: Perhaps aria-owns on the checkboxes? https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships#aria-owns :shrug:

marusak:
Can we get some test for this?
Something like:
1. Check that the field is disabled
2. Click on the field
3. Check that the field is enabled

jelly:
> Can we get some test for this? Something like:
> 
> 1. Check that the field is disabled
> 2. Click on the field
> 3. Check that the field is enabled

Added a test for the cpu shares. Does the same but a tiny bit different.

test/check-application
```diff
@@ -1313,6 +1313,11 @@ class TestApplication(testlib.MachineCase):
 
         # CPU shares work only with system containers
         if auth:
+            # Check that the checkbox is enabled when clicked on the field
+            b.click('#run-image-cpu-priority')
+            b.wait_visible("#run-image-dialog-cpu-priority-checkbox:checked")
+            b.set_checked("#run-image-dialog-cpu-priority-checkbox", False)
+
```
marusak, line 1320: We should check that it is actually disabled before, so we are sure it did something. And I don't think we need to then unset it and set it back, surely there is some different place that checks these checkboxes.

marusak (changes requested):

test/check-application
jelly: This is the only place the cpu priority is tested, so I'll leave the unsetting in.

marusak (approved):
Thank you!

PR #856, opened 2021-12-14
Fix spelling errors found by codespell

garrett:


src/Images.jsx
```diff
@@ -207,7 +207,7 @@ class Images extends React.Component {
 
         const imageRows = filtered.map(id => this.renderRow(this.props.images[id]));
 
-        const interm = this.props.images && Object.keys(this.props.images).some(id => {
+        const interim = this.props.images && Object.keys(this.props.images).some(id => {
```
garrett, line 210: This is a code change. But it's also changed lower in the file where it's referenced.

martinpitt: Supposedly it was meant to be an abbreviation for "intermediate" (image), but it's fine like that.

martinpitt (approved):
Thanks!

src/Images.jsx
marusak: correct, but indeed, if this should always jump out when we do codespell, we can use `interim`

PR #859, opened 2021-12-14
Add always fetch latest version

jelly:
---

# Always pull the latest image when creating a new container

When a local image `fedora:latest` is available but a newer version is available on the image registry, checking the "pull latest image" checkbox will pull the latest version before creating the new container

image

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1202,6 +1202,11 @@ class TestApplication(testlib.MachineCase):
         self._testCreateContainer(False)
 
     def _testCreateContainer(self, auth):
+        new_container = 'new-container'
+        self.execute(True, f"podman run -d --name {new_container} quay.io/libpod/busybox touch /latest")
+        self.execute(True, f"podman commit {new_container} newimage")
+        new_image_sha = self.execute(True, "podman inspect --format '{{.Id}}' newimage").strip()
+
         self.execute(True, "podman run -d -p 5000:5000 --name registry registry:2")
         self.execute(True, "podman run -d -p 6000:5000 --name registry_alt registry:2")
         # Add local insecure registry into registries conf
@@ -1241,10 +1246,45 @@ class TestApplication(testlib.MachineCase):
         # Select image
         b.click('button.pf-c-select__menu-item:contains("localhost:5000/my-busybox")')
 
+        # Remote image, no pull latest image option
+        b.wait_not_present("#run-image-dialog-pull-latest-image")
+
         # Create Container, image is pulled and should end up being "running"
         b.click('.pf-c-modal-box__footer button:contains("Create")')
         sel = "> span:not(.downloading)"
         b.wait(lambda: self.getContainerAttr(container_name, "State", sel) in 'Running')
+        output = self.execute(auth, f"podman exec {container_name} ls -lh /latest || true").strip()
+        self.assertEqual('', output)
+
+        # Now that we have downloaded an image, verify that selecting download latest image
+        # downloads the latest image we now push to the registry. Note this image has a /latest file
+        # to differnatiate it from the other local image.
+        self.execute(True, f"podman push {new_image_sha} localhost:5000/my-busybox")
+        self.execute(True, f"podman push {new_image_sha} localhost:6000/my-busybox")
+        self.execute(True, f"podman rmi {new_image_sha}")
+
+        container_name = "busybox-latest"
+
+        b.click("#containers-containers button.pf-c-button.pf-m-primary")
+        b.set_input_text("#run-image-dialog-name", container_name)
+
+        # Local registry
+        b.set_input_text("#create-image-image-select-typeahead", "my-busybox")
+        b.click('button.pf-c-toggle-group__button:contains("Local")')
+
+        # Select image
+        b.click('button.pf-c-select__menu-item:contains("localhost:5000/my-busybox")')
+
+        # Pull the latest image
+        b.set_checked("#run-image-dialog-pull-latest-image", True)
+
+        # Create Container, image is pulled and should end up being "running"
+        b.click('.pf-c-modal-box__footer button:contains("Create")')
+        sel = "> span:not(.downloading)"
+        b.wait(lambda: self.getContainerAttr(container_name, "State", sel) in 'Running')
+        # Verify that the latest file exists
+        output = self.execute(auth, f"podman exec {container_name} ls -lh /latest").strip()
+        self.assertNotIn("No such file or directory", output)
 
         # Test creating a container with
         if auth:
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -308,6 +308,18 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
     );
 };
 
+export let onDownloadContainer = function funcOnDownloadContainer(container) {
+    this.setState(prevState => ({
+        downloadingContainers: [...prevState.downloadingContainers, container]
+    }));
+};
+
+export let onDownloadContainerFinished = function funcOnDownloadContainerFinished(container) {
+    this.setState(prevState => ({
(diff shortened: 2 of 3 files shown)


test/check-application
```diff
@@ -1241,6 +1246,39 @@ class TestApplication(testlib.MachineCase):
...
+        # Local registry
+        b.set_input_text("#create-image-image-select-typeahead", "my-busybox")
+        b.click('button.pf-c-toggle-group__button:contains("localhost:5000")')
+        b.wait_text("button.pf-c-select__menu-item:not(.pf-m-disabled)", "localhost:5000/my-busybox")
+
+        # Select image
+        b.click('button.pf-c-select__menu-item:contains("localhost:5000/my-busybox")')
+
+        # Pull the latest image
+        b.click("#run-image-dialog-pull-latest-image")
```
marusak, line 1273: `b.set_checked("#run-image-dialog-pull-latest-image", True)`

test/check-application
```diff
@@ -1769,7 +1807,7 @@ class TestApplication(testlib.MachineCase):
         b.click('.pf-c-modal-box__footer button:contains("Create")')
         self.waitContainerRow(container_name)
 
-    @testlib.skipImage("podman-restart not available in debian/ubuntu", *DISTROS_WITHOUT_PODMAN_RESTART)
+    @ testlib.skipImage("podman-restart not available in debian/ubuntu", *DISTROS_WITHOUT_PODMAN_RESTART)
```
marusak, line 1810: Why these spaces in decorators?

marusak (changes requested):
Do we also test, that when you don't check in the new checkbox then it does not pull?

marusak (approved):
Thanks!

garrett:
Initial reactions:

- There's not enough vertical space between the checkboxes. Is it using the correct elements? (Not saying space should be manually added... it's just weird that there isn't more. There should be, so this points to incorrect usage of elements... or a bug.)
- Start should be after pull, conceptually.

garrett:
So this is weird on a create dialog when there is no local image. Of course it'd pull the latest image, right? So then it's confusing why it's there.

I am pretty sure when we talked about this that it would probably only exist when there's an existing image and I think we talked about it showing up in the image section, under the image selector, and only when there's a match with a local image.

(Note: Between the holidays and currently being sick including a headache, I could be mistaken. But still, I think it's confusing when there's no local image, so I think this probably still needs work regardless of what we did or didn't talk about in the past.)

jelly:
> 

That's now implemented, not 100% sure if we discussed it before the implementation but it's certainly something which should be in the PR. Thanks for noticing!

jelly:
> So this is weird on a create dialog when there is no local image. Of course it'd pull the latest image, right? So then it's confusing why it's there.
> 
> I am pretty sure when we talked about this that it would probably only exist when there's an existing image and I think we talked about it showing up in the image section, under the image selector, and only when there's a match with a local image.

Moved the pull on create checkbox under the image select, but to avoid it from making the whole form jump it's always shown but only enabled for local images.

garrett:
> Moved the pull on create checkbox under the image select, but to avoid it from making the whole form jump it's always shown but only enabled for local images.

But we know when the image is being used in the image flow and should show it by default and when create container is selected, it's usually not going to be using a local image.

When we have some kind of "cloning" / "rebuild" option, it would also use a local image by default too.

Having a pull option always present even when it doesn't apply would be confusing.

(I'll look at this PR on my computer in a little bit. It's next in the queue now. :wink:)

jelly:
> > Moved the pull on create checkbox under the image select, but to avoid it from making the whole form jump it's always shown but only enabled for local images.
> 
> But we know when the image is being used in the image flow and should show it by default and when create container is selected, it's usually not going to be using a local image.

That's sort of true, if you create an image from a local container from the `Images` dialog sure it shouldn't be shown. But if you create it from the Containers dialog you can select local and remote images there I disabled it.

So, good point we should hide it when you create it from the `Images` section and keep it disabled/enabled for the `Containers` flow.

garrett:
> So, good point we should hide it when you create it from the Images section and keep it disabled/enabled for the Containers flow.

I was suggesting the opposite of that, actually. Show it for the images section (if there's a remote image available... that is, show it if it's not a locally created image) and don't show it by default for the create containers flow, unless it matches a local image in the selector.

jelly:
> > So, good point we should hide it when you create it from the Images section and keep it disabled/enabled for the Containers flow.
> 
> I was suggesting the opposite of that, actually. Show it for the images section (if there's a remote image available... that is, show it if it's not a locally created image) and don't show it by default for the create containers flow, unless it matches a local image in the selector.

Seems my brain mixed things up, sorry. We can do that, do we accept that the dialog will "jump" if you select a local image?

garrett:
> Seems my brain mixed things up, sorry. 

I'm grappling with a headache and feeling icky from being sick still (better, but still ugh)... so I'm just as likely to mix things up or suggest the wrong thing.

> We can do that, do we accept that the dialog will "jump" if you select a local image or can we keep some spacing so it doesn't occur.

We could have a statusline using a helpertext item there in place of a checkbox. ("The image will be downloaded from XYZ." or something like it could work... although it's probably redundant.) The helptertext widget, as it uses small text, might have a different line-height and we might need to adjust it to be the same height to not have a jump. But it'd be better than just a big gap there still... and this is a way to not have a jump.

...Or we just accept that the checkbox might jump into existence in some cases. We do have dialogs that change height already, as the result of interacting widgets above the content that changes.

jelly:
> > Seems my brain mixed things up, sorry.
> 
> I'm grappling with a headache and feeling icky from being sick still (better, but still ugh)... so I'm probably even more likely to mix things up or suggest the wrong thing or not be clear enough. wink
> 
> > We can do that, do we accept that the dialog will "jump" if you select a local image or can we keep some spacing so it doesn't occur.
> We could have a statusline using a helpertext item there in place of a checkbox. ("The image will be downloaded from XYZ." or something like it could work... although it's probably redundant.) The helptertext widget, as it uses small text, might have a different line-height and we might need to adjust it to be the same height to not have a jump. But it'd be better than just a big gap there still... and this is a way to not have a jump.

I think that adds to much redundant information, something which we should avoid.

> ...Or we just accept that the checkbox might jump into existence in some cases. We do have dialogs that change height already, as the result of interacting widgets above the content that changes.

That's the best compromise in my opinion, the new updated PR now shows it by default for the `Images` overview and it is shown when required for the "Create container" workflow under "Containers", the jumping seems acceptable and maybe something we can improve on later with new ideas.

jelly:
> Here's what the autocomplete is covering for me:
> 
> image
> 
> Additionally (out of scope of this PR): What is the `<none:none>`?

I have <none:none> too hmm, not sure where it comes from but it's also shown in `podman images`.

marusak:
`none:none` are intermediate images. We show them in the listing, but there is button on the bottom to show them.

marusak (changes requested):
Looks good, thanks!
Can you please add test that we don't show this checkbox when pulling the image?

jelly:
> Looks good, thanks! Can you please add test that we don't show this checkbox when pulling the image?

added!

test/check-application
```diff
@@ -1241,10 +1246,46 @@ class TestApplication(testlib.MachineCase):
...
+
+        # Local registry
+        b.set_input_text("#create-image-image-select-typeahead", "my-busybox")
+        b.click('button.pf-c-toggle-group__button:contains("Local")')
+
+        # Select image
+        b.click('button.pf-c-select__menu-item:contains("localhost:5000/my-busybox")')
+
+        # Pull the latest image
+        # testlib.sit()
```
marusak, line 1279: :)

marusak (approved):
Awesome, thank you!
Can you please update screenshot as well for the release notes?

jelly:
> Awesome, thank you! Can you please update screenshot as well for the release notes?

done!

Issue #487, opened 2020-08-04
[pod] Add container to a pod


marusak:
See https://raw.githubusercontent.com/cockpit-project/cockpit-design/master/containers/podman-pod-groups.png
https://github.com/cockpit-project/cockpit-podman/issues/238#issuecomment-662482073

jelly:
Adding an existing [container to a pod](https://github.com/containers/podman/issues/3979) is not supported in podman so that's not something we can implement yet.

garrett:
The action is to create a container inside the pod, not add an existing one. I guess the (old) pod mockup says "add container" instead of "create container".

The redesign mockup has "create" in the button and also includes pods:

![podman redesign](https://user-images.githubusercontent.com/10246/124738736-8694b080-df19-11eb-91e3-dae599569683.png)

jelly:
The "Create container" modal changes the title to "Create container in $podname"

PR #867, opened 2021-12-17
Create a container in a pod

jelly:
This adds the ability to create a new container in a pod group, a
container cannot be moved out or into a pod once it has been created.

Closes: #487

---

# Create container in pod

It is now possible to create a container in an existing pod using the `Create container in pod` button.

image

Create container dialog now shows in which pod the container is created

image

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -273,6 +273,7 @@ class Containers extends React.Component {
         this.state = {
             width: 0,
             showCreateContainerModal: false,
+            createPod: null,
             downloadingContainers: [],
         };
         this.renderRow = this.renderRow.bind(this);
@@ -502,7 +503,8 @@ class Containers extends React.Component {
                 <ImageRunModal
                 user={this.props.user}
                 localImages={localImages}
-                close={() => this.setState({ showCreateContainerModal: false })}
+                close={() => this.setState({ showCreateContainerModal: false, createPod: null })}
+                pod={this.state.createPod}
                 registries={this.props.registries}
                 selinuxAvailable={this.props.selinuxAvailable}
                 podmanRestartAvailable={this.props.podmanRestartAvailable}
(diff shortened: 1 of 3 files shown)


jelly:
Once everything is reviewed, I think it makes sense to add a pixel test for the creation modal and overview

marusak (changes requested):
So I was somewhat confused when I open the dialog for the first time. I was expecting to see there select with available pods. When you would click on `Create container` it would not preselect any pod, but would be `No pod` or something like that. And if you would click on `Create container` in pod area, it would preselect that given one. (Note, that I don't like `Create container in pod` string, I would always just keep it `Create container`).
@jelly WDYT?

jelly:
> So I was somewhat confused when I open the dialog for the first time. I was expecting to see there select with available pods. When you would click on `Create container` it would not preselect any pod, but would be `No pod` or something like that. And if you would click on `Create container` in pod area, it would preselect that given one. (Note, that I don't like `Create container in pod` string, I would always just keep it `Create container`). @jelly WDYT?

In the discussion I had with Garrett, we concluded to not add another option to the already quite extensive options we have and instead change the title of the creation modal.

Can we do something to improve the button text maybe, "create a container in this pod?" It seems that the context for the button is not clear enough?

marusak:
> Can we do something to improve the button text maybe, "create a container in this pod?" It seems that the context for the button is not clear enough?

That is not an issue. I understand the difference between those buttons. I just thought that the workflow would be more clear with having the option, but if two of you agreed not to do it, then I am fine with it as well :)

marusak (approved):
Thanks!

garrett:
The status label is between actions:

image

This should be (running) [Create container in pod] ...

marusak:
@jelly can you please update here screenshot to accommodate change from #878 ? Thanks!

PR #870, opened 2021-12-23
Add pause/resume to containers

jelly:
Bring containers to the similiar feature parity to podman pods by adding
pause/resume options. Also make the order of the menu options the same
for pods and containers and add "Force restart" to the pod actions.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -22,7 +22,7 @@ registries = ['localhost:5000', 'localhost:6000']
 registries = ['localhost:5000', 'localhost:6000']
 """
 
-NOT_RUNNING = ["exited", "stopped"]
+NOT_RUNNING = ["Exited", "Stopped"]
 
 # Ubuntu stable has 3.2.1 while we need podman 3.3.
 DISTROS_WITHOUT_PODMAN_RESTART = ['ubuntu-stable']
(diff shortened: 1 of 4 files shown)


jelly:
@garrett What do you think about the status text, for pods it shows as `Running` while for containers it's `running`, I've changed pods status to be display as lowercase as well.

test/check-application
```diff
@@ -1837,6 +1837,34 @@ class TestApplication(testlib.MachineCase):
...
+
+    def _testPauseResumeContainer(self, auth):
+        b = self.browser
+        container_name = "pauseresume"
+
+        self.execute(auth, f"podman run -dt --name {container_name} alpine")
+        self.login(auth)
+
+        self.waitContainerRow(container_name)
+        self.toggleExpandedContainer(container_name)
```
marusak, line 1857: Here I would add check, that `Resume` is not present.

test/check-application
```diff
@@ -1837,6 +1837,34 @@ class TestApplication(testlib.MachineCase):
...
+
+        self.performContainerAction(container_name, "Pause")
+
+        # show all containers and check status
+        self.filter_containers('all')
+
+        b.wait(lambda: self.getContainerAttr(container_name, "State") == "paused")
```
marusak, line 1864: And here, that `Pause` is not present.

marusak (dismissed):
Nice, thank you!
I am approving, but if you feel like adding two more check into test, that would be appreciated :)

garrett:
> @garrett What do you think about the status text, for pods it shows as `Running` while for containers it's `running`, I've changed pods status to be display as lowercase as well.

We're supposed to use "Sentence case":
- https://www.patternfly.org/v4/ux-writing/capitalization/#capitalization-types
- https://www.patternfly.org/v4/ux-writing/capitalization/#capitalization-across-red-hat-uis (this is directly below the other link)

marusak (approved):
Thank you!

PR #874, opened 2022-01-03
test: Update pixel test references for new Fedora 35 image

mvollmer:
The content area got slightly smaller due to a change in the shell.

In the future, the content area size will no longer change when the
shell changes, but we have to catch up with that one last
time.  (Hopefully.)


martinpitt (approved):
Thanks! Retrying [r85 flake](https://logs.cockpit-project.org/logs/pull-874-20220103-133949-28e1d94f-rhel-8-5/log.html)

PR #877, opened 2022-01-04
test: Debian testing now has container restart capability

martinpitt:
Needs to land in lockstep with the image refresh in https://github.com/cockpit-project/bots/pull/2741


marusak (approved):
Danke :)

PR #883, opened 2022-01-06
Rsync webpack plugin

jelly:


webpack.config.js
```diff
@@ -1,3 +1,4 @@
+const fs = require("fs");
```
martinpitt, line 1: Nice, thank you! Can you please cherry-pick the corresponding README files as well, to show how to use this?

martinpitt (changes requested):
Thanks!

webpack.config.js
jelly: Ah, I was thinking about it, but we seem to miss a lot so wasn't sure if I should have added. But as you agree I'll add it!

jelly: Added a section of the install section of the starterkit README, as that information is in c-podman in HACKING.md

HACKING.md
```diff
@@ -8,6 +8,33 @@ Here's where to get the code:
 The remainder of the commands assume you're in the top level of the
 Cockpit Podman git repository checkout.
 
+For development, you usually want to run your module straight out of the git
+tree. To do that, link that to the location were `cockpit-bridge` looks for packages:
+
+```
+mkdir -p ~/.local/share/cockpit
+ln -s `pwd`/dist ~/.local/share/cockpit/starter-kit
```
KKoukiou, line 16: s/startet-kit/cockpt-podman

jelly: should be podman right? 

```
[jelle@t14s][~/projects/starter-kit]%ls -lh ~/.local/share/cockpit/
drwxr-xr-x 1 jelle jelle  688 Dec  8 09:03 kdump/
drwxr-xr-x 1 jelle jelle  344 Dec 22 12:04 lib/
lrwxrwxrwx 1 jelle jelle   42 Dec 22 17:35 machines -> /home/jelle/projects/cockpit-machines/dist/
```

KKoukiou: Correct.

martinpitt (approved):
Dankje wel!

PR #885, opened 2022-01-13
Fix jumping headers in container details

jelly:
The calculation for the padding in the card and table was incorrect
leading to the table headers to "jump" when switching between details
and logs as the terminals does not fit in the card.


martinpitt:
This [rawhide failure](http://artifacts.dev.testing-farm.io/377343b8-ff18-466c-98bf-df5ba47f3817/) could still be fallout from yesterday's Firefox snafu, retrying

marusak (approved):
nice one! Thanks

PR #888, opened 2022-01-17
Turn eslint warnings into errors

jelly:


martinpitt (approved):
Nice, thanks! I wonder why this falls over so hard in starter-kit, but works fine here..

PR #891, opened 2022-01-21
Fix regression this.state.isSystem undefined

jelly:
With the commit 107f060 this.state.isSystem was referenced with it being
initialized. This was never caught in our test setup as we never
explicitly login as root and then fallback to the admin's podman socket.

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -538,6 +538,7 @@ class Containers extends React.Component {
                     <Divider isVertical />
                     <ToolbarItem>
                         <Button variant="primary" key="get-new-image-action"
+                        id="containers-containers-create-container-btn"
                         isDisabled={localImages === null}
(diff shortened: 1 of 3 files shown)


jelly:
Should we include a test with `self.login_and_go("/podman", superuser=False)` to not get sudo behaviour?

marusak:
> Should we include a test with self.login_and_go("/podman", superuser=False) to not get sudo behaviour?

If we can reproduce this, then we definitely should add test for it.

jelly:
> > Should we include a test with self.login_and_go("/podman", superuser=False) to not get sudo behaviour?
> 
> If we can reproduce this, then we definitely should add test for it.

Seems testBasicSystem is the way to go.

marusak (approved):
Thanks!

PR #893, opened 2022-01-25
test: Drop obsolete hack

marusak:


marusak:
Names of port items has changed: https://docs.podman.io/en/latest/_static/api.html#operation/ContainerListLibpod
`hostIP` -> `host_ip`
`hostPort` -> `host_port`
`containerPort` -> `container_port`

I'll just fix it with simple `||` to support all versions :man_shrugging:

marusak:
`json_object_get_string_member: assertion 'node != NULL' failed` hmm - I thought this was already fixed in https://github.com/cockpit-project/cockpit/pull/16765

martinpitt:
@marusak : It was, but podman's tests don't upgrade cockpit-ws to the latest version (particularly for ubuntu-stable, which still has 252). Arch should have 260, but the image refresh has failed for a long time -- so our current image would still have an old version. For the time being I suggest to just add that as an expected message to the tests, with a HACK comment. Like we [did for distropkg](https://github.com/cockpit-project/cockpit/blob/main/test/common/testlib.py#L1411).

[F35 failure](https://logs.cockpit-project.org/logs/pull-893-20220125-133513-0f78510d-fedora-35/log.html#9) is different, that's a pixel test

marusak:
F35 packit is failing due to the selinux issue. rawhide failed with http://artifacts.dev.testing-farm.io/e0e07dd7-f29f-492f-9031-d88ce6430106/work-allARaS6x/plans/all/execute/data/test/browser/logs/TestApplication-testCheckpointRestoreCGroupsV2-fedora-36-127.0.0.1-22-FAIL.png but it is less helpful than I thought. Lets figure it out some other time

src/ContainerDetails.jsx
```diff
@@ -17,13 +17,18 @@ const render_container_published_ports = (ports) => {
     if (!ports)
         return null;
 
-    const result = [];
+    const result = ports.map(port => {
+        const protocol = port.protocol;
+        const host_port = port.hostPort || port.host_port;
+        const container_port = port.containerPort || port.container_port;
+        const host_ip = port.hostIP || port.host_ip || '0.0.0.0';
```
jelly, line 24: Do we want to have a comment on when this was introduced and for podman 4.00

jelly (approved):

src/ContainerDetails.jsx
marusak: Added comment

martinpitt (approved):
Looks great, thank you!

Issue #845, opened 2021-12-01
mobile: Tabs overflow in small sizes


garrett:
![Screen Shot 2021-12-01 at 17 47 20](https://user-images.githubusercontent.com/10246/144276827-71e09ef1-143e-46ba-bcfc-7476ed2abf6b.png)

It's also impossible to see and switch tabs at this sizes.

This is the iPhone 5 / SE size.

garrett:
It's still bad on the iPhone 13 Pro Max size (which is huge)... just not quite as bad, as you can see a little bit of the second tab:

![Screen Shot 2021-12-01 at 17 49 06](https://user-images.githubusercontent.com/10246/144277101-1ab63484-165b-4aca-895e-778fe9b02749.png)

garrett:
(These are the 2x "retina" sizes. Sorry for them being so huge.)

jelly:
I can see the issue, I am not sure how we can improve the situation, do we need to switch to a different component or can we improve the Tab component to be more user friendly?

Should we consider switching to a `vertical` view https://www.patternfly.org/v4/components/tabs#vertical ?

garrett:
No, the vertical one would be even more problematic. This has been an issue with PF for a while. I had an override for tertiary navigation (which looks much like the tabs and was removed in favor of tabs) and had some upstream issues for it and tabs at:

- https://github.com/patternfly/patternfly-design/issues/840
- https://github.com/patternfly/patternfly-design/issues/1034

We'll probably have to do some CSS override to make them wrap here.

garrett:
The fix in #848 helps here, but we could still hit the same issue with longer strings (different languages, when this is translated, as it apparently is not) or more strings (as in more tabs).

PR #897, opened 2022-01-30
Support healthcheck

### LINKED ISSUE #345: Support healthcheck
Definitely needs design, but should at least allow:
1. Show if some healthcheck is available
2. Enable setting up healthcheck (command, interval, retries, start-period, timeout)
3. Show in the overview if the healthcheck failed
4. Show when the healthcheck run the last time (if possible show history of those runs?)
5. Enable to trigger healthcheck run

marusak:
Fixes #345
One known limitation is reported in https://github.com/containers/podman/issues/13083

TODO - if image has healthcheck defined, prefill the values

---

# Support for health checks

It is now possible to set up, inspect and trigger health checks for container. Also the most recent history as well as overall status are presented.

Screenshot from 2022-02-04 10-41-59

### DIFF
--- a/src/ContainerHealthLogs.jsx
+++ b/src/ContainerHealthLogs.jsx
@@ -1,0 +1,127 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import cockpit from 'cockpit';
+import * as utils from './util.js';
+import * as client from './client.js';
+
+import { ListingTable } from "cockpit-components-table.jsx";
+import {
+    Button,
+    DescriptionList, DescriptionListGroup, DescriptionListTerm, DescriptionListDescription,
+    Flex, FlexItem,
+} from '@patternfly/react-core';
+import { CheckCircleIcon, ErrorCircleOIcon } from "@patternfly/react-icons";
+
+const _ = cockpit.gettext;
+
+const format_nanoseconds = (ns) => {
+    const seconds = ns / 1000000000;
+    return cockpit.format(cockpit.ngettext("$0 second", "$0 seconds", seconds), seconds);
+};
+
+const ContainerHealthLogs = ({ container, containerDetail, onAddNotification, state }) => {
+    let healthCheck = {};
+    let failingStreak = 0;
+    let logs = [];
+    if (containerDetail) {
+        healthCheck = containerDetail.Config.Healthcheck || containerDetail.Config.Health;
+        const healthState = containerDetail.State.Healthcheck || containerDetail.State.Health;
+        failingStreak = healthState.FailingStreak || 0;
+        logs = [...(healthState.Log || [])].reverse();
+    }
+
+    return (
+        <>
+            <Flex alignItems={{ default: "alignItemsFlexStart" }}>
+                <FlexItem grow={{ default: 'grow' }}>
+                    <DescriptionList isAutoFit id="container-details-healthcheck">
+                        <DescriptionListGroup>
+                            <DescriptionListTerm>{_("Status")}</DescriptionListTerm>
+                            <DescriptionListDescription>{state}</DescriptionListDescription>
+                        </DescriptionListGroup>
+                        <DescriptionListGroup>
+                            <DescriptionListTerm>{_("Command")}</DescriptionListTerm>
+                            <DescriptionListDescription>{utils.quote_cmdline(healthCheck.Test)}</DescriptionListDescription>
+                        </DescriptionListGroup>
+                        {healthCheck.Interval && <DescriptionListGroup>
+                            <DescriptionListTerm>{_("Interval")}</DescriptionListTerm>
+                            <DescriptionListDescription>{format_nanoseconds(healthCheck.Interval)}</DescriptionListDescription>
+                        </DescriptionListGroup>}
+                        {healthCheck.Retries && <DescriptionListGroup>
+                            <DescriptionListTerm>{_("Retries")}</DescriptionListTerm>
+                            <DescriptionListDescription>{healthCheck.Retries}</DescriptionListDescription>
+                        </DescriptionListGroup>}
+                        {healthCheck.StartPeriod && <DescriptionListGroup>
+                            <DescriptionListTerm>{_("Start period")}</DescriptionListTerm>
+                            <DescriptionListDescription>{format_nanoseconds(healthCheck.StartPeriod)}</DescriptionListDescription>
+                        </DescriptionListGroup>}
+                        {healthCheck.Timeout && <DescriptionListGroup>
+                            <DescriptionListTerm>{_("Timeout")}</DescriptionListTerm>
+                            <DescriptionListDescription>{format_nanoseconds(healthCheck.Timeout)}</DescriptionListDescription>
+                        </DescriptionListGroup>}
+                        {failingStreak !== 0 && <DescriptionListGroup>
+                            <DescriptionListTerm>{_("Failing streak")}</DescriptionListTerm>
+                            <DescriptionListDescription>{failingStreak}</DescriptionListDescription>
+                        </DescriptionListGroup>}
+                    </DescriptionList>
+                </FlexItem>
+                { container.State === "running" &&
+                    <FlexItem>
+                        <Button variant="secondary" onClick={() => {
+                            client.runHealthcheck(container.isSystem, container.Id)
+                                    .catch(ex => {
+                                        const error = cockpit.format(_("Failed to run health check on container $0"), container.Names);
+                                        onAddNotification({ type: 'danger', error, errorDetail: ex.message });
+                                    });
+                        }}>
+                            {_("Run health check")}
+                        </Button>
+                    </FlexItem>}
+            </Flex>
+            <ListingTable aria-label={_("Logs")}
+                          variant='compact'
+                          columns={[_("Last 5 runs"), _("Started at")]}
+                          rows={
+                              logs.map(log => {
+                                  const id = "hc" + log.Start + container.Id;
+                                  return {
+                                      expandedContent: log.Output ? <pre>{log.Output}</pre> : null,
+                                      columns: [
+                                          {
+                                              title: <Flex flexWrap={{ default: 'nowrap' }} spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
+                                                  {log.ExitCode === 0 ? <CheckCircleIcon className="green" /> : <ErrorCircleOIcon className="red" />}
+                                                  <span>{log.ExitCode === 0 ? _("Passed health run") : _("Failed health run")}</span>
+                                              </Flex>
+                                          },
+                                          utils.localize_time(Date.parse(log.Start) / 1000)
+                                      ],
+                                      props: {
+                                          key : id,
+                                          "data-row-id": id,
+                                      },
+                                  };
+                              })
+                          } />
+        </>
+    );
+};
+
+export default ContainerHealthLogs;
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -7,7 +7,8 @@ import {
     FormSelect, FormSelectOption,
     Grid, GridItem,
     HelperText, HelperTextItem,
-    Modal, Radio, Select, SelectVariant,
+    InputGroup, Radio, InputGroupText, InputGroupTextVariant,
+    Modal, NumberInput, Select, SelectVariant,
     SelectOption, SelectGroup,
     TextInput, Tabs, Tab, TabTitleText,
     ToggleGroup, ToggleGroupItem,
@@ -330,6 +331,13 @@ export class ImageRunModal extends React.Component {
             imageResults: {},
             isImageSelectOpen: false,
             searchByRegistry: 'all',
+            /* health check */
+            healthcheck_command: "",
+            healthcheck_interval: 30,
+            healthcheck_timeout: 30,
+            healthcheck_start_period: 0,
+            healthcheck_retries: 3,
+
         };
         this.getCreateConfig = this.getCreateConfig.bind(this);
         this.onValueChanged = this.onValueChanged.bind(this);
@@ -420,6 +428,16 @@ export class ImageRunModal extends React.Component {
             if (this.state.restartPolicy === "always" && this.props.systemServiceAvailable) {
                 this.enablePodmanRestartService();
             }
+        }
+
+        if (this.state.healthcheck_command !== "") {
+            createConfig.healthconfig = {
+                Interval: parseInt(this.state.healthcheck_interval) * 1000000000,
+                Retries: this.state.healthcheck_retries,
+                StartPeriod: parseInt(this.state.healthcheck_start_period) * 1000000000,
+                Test: utils.unquote_cmdline(this.state.healthcheck_command),
+                Timeout: parseInt(this.state.healthcheck_timeout) * 1000000000,
+            };
         }
 
         return createConfig;
@@ -518,6 +536,14 @@ export class ImageRunModal extends React.Component {
 
     onValueChanged(key, value) {
         this.setState({ [key]: value });
+    }
+
+    onPlusOne(key) {
+        this.setState(state => ({ [key]: state[key] + 1 }));
+    }
+
+    onMinusOne(key) {
+        this.setState(state => ({ [key]: state[key] - 1 }));
     }
 
     handleTabClick = (event, tabIndex) => {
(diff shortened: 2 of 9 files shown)


marusak:
@garrett WDYT of the design?

jelly:
Do we really need to have the health check details this prominent in the details view? Can't they be part of the `Health check` logs?

marusak:
> Do we really need to have the health check details this prominent in the details view? Can't they be part of the Health check logs?

Something for design review. But I think there is enough space for it so I would not try to hide it.

src/ImageRunModal.jsx
```diff
@@ -1034,6 +1052,99 @@ export class ImageRunModal extends React.Component {
...
+                              }>
+                            <InputGroup>
+                                <TextInput type='number'
+                                        id="run-image-healthcheck-interval"
+                                        className="healthcheck-seconds"
+                                        value={dialogValues.healthcheck_interval}
+                                        step={1}
+                                        min={0}
+                                        max={262144}
+                                        onChange={value => this.onValueChanged('healthcheck_interval', parseInt(value))} />
```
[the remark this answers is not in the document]
marusak: 0 means disabled

marusak:
@jelly I updated your comments. I also changed the layout - now everything is under new `Health check` tab. I updated screenshot in the first comment.
Commits are mess, they need squashing and also there are some unrelated changes. I'll cherry-pick those. Needs also rebase, but for that I want to first stabilize design.

src/ContainerHealthLogs.jsx
```diff
@@ -0,0 +1,117 @@
...
+                                    onAddNotification({ type: 'danger', error, errorDetail: ex.message });
+                                });
+                    }}>
+                        {_("Run health check")}
+                    </Button>
+                </FlexItem>
+            </Flex>
+            <ListingTable aria-label={_("Logs")}
+                          variant='compact'
+                          columns={['Last 5 runs', _("Started at")]}
```
jelly, line 91: `Last 5 Runs` should be translated as well right?

webpack.config.js
```diff
@@ -25,14 +28,14 @@ const copy_files = [
...
-        failOnWarning: true,
-    }),
     new CockpitPoPlugin(),
     new CockpitRsyncPlugin({dest: packageJson.name}),
 ];
 
+if (eslint) {
+    plugins.push(new ESLintPlugin({ extensions: ["js", "jsx"], failOnWarning: true, }));
+}
+
```
jelly, line 38: hmm why are these changes here? Ideally this should go through starterkit first right?

jelly (changes requested):

garrett:
Oh, apologies for not getting back to this. I need to help out with the design a bit. I guess next week?

webpack.config.js
marusak: well, it is already in cockpit anyway :) We can add it into starter kit, but it seems like pattern that you don't want people to use, it is just good hack when you quickly iterate over the code

src/ContainerHealthLogs.jsx
marusak: Indeed!

marusak:
> When there's no output, we probably should probably not have expanders on the rows:

Can take a look. Just mind that right now due to https://github.com/containers/podman/issues/13083 no runs have any output, but it should be very soon available. (I am not sure if we can mix and match these rows with and without expanders)

> What happens after 5 runs? We just show the latest ones? Would it be useful to know when something went wrong?

That is limitation from podman itself. It only stores and exposes last 5 runs. If you have 5 runs shown and a new one happens, the oldest one disappears.

> The order is going from first to last, which is quite wrong if we're truncating the information. It should be latest first, historical after. (Flip the order.)

Will do.

> Should we consider something other than a list, especially if it's just a bunch of successes? I guess if there's detailed information, it might be interesting... but a health check is probably only interesting if it passes or when it goes wrong, correct?

Right, if it is bunch of successes you probably don't care that much - maybe just to check if it actually does something or it dies silently. But at the same time, when you see your container is `Healthy` in overview, you likely won't even visit the page.

marusak:
@garrett I've updaed your comments. It does not have toggle if there is not output, I switched the order and only show buttons to run healtcheck when the container is running.

Screenshot from 2022-02-23 08-26-43

garrett:
The defaults for health check are every 30 seconds. Is this a default from somewhere? Is this what we want? The logs have duplicate times as a result:

image

The labels should be in a label group:

https://www.patternfly.org/v4/components/label-group/

image

We also still have the overflow issue, which is even more noticeable with the badge-y labels like this.

marusak:
> The defaults for health check are every 30 seconds. Is this a default from somewhere? 

Yes, from `man podman run`:
```
   --health-interval=interval
       Set an interval for the healthchecks. An interval of disable results in no automatic timer setup. The default is 30s.
```

> Is this what we want?

Yes, we don't want to create some arbitrary defaults on our own.

> The logs have duplicate times as a result:

Let me take a look if showing seconds makes sense.

marusak:
> > The logs have duplicate times as a result:

> Let me take a look if showing seconds makes sense.

Hmm, we use https://date-fns.org/v2.28.0/docs/formatRelative which does not show seconds. We would have to switch to some different method and would loose the relative part.

marusak:
> The labels should be in a label group:

Updated

garrett (changes requested):
Firstly, I was playing around with the idea of not stretching the items and making the column count 4 (which can't be done in PF as a modifier, but can be adjusted via a variable).

```scss
.container-details-healthcheck {
    --pf-c-description-list--GridTemplateColumns--count: 4;
}
```

There are a few odd things here:

1. "Starting health check" wraps as it's a long label. We should consider something smaller or do something else.
2. Health check status isn't shown outside of the icon. We should have success / fail text.

Screenshot 2022-03-01 at 14-02-01 Podman containers - garrett Bolt

(The names were generated; the naughty one happens to be naughty.)

If I bump the column count to 5, the items don't wrap. This even lets the items line up across expanded containers:

Screenshot 2022-03-01 at 14-06-34 Podman containers - garrett Bolt

So we should definitely fix the "starting" label and the health check status. And optionally consider not stretching the healthcheck columns.

garrett:
Also, I noticed that the alignment of the flex that wraps the health check info and the button isn't correct. The items should be aligned to the start with an `align-items: start` or `align-items: stretch`, the latter of which I think can be set in PF?

garrett:
Does this pre-fill the suggested healthcheck from the container?

`Dockerfile` / `Containerfile` supports a `HEALTHCHECK` directive; it's not just at container creation time with command-line flags:

https://docs.docker.com/engine/reference/builder/#healthcheck

marusak:
I tried to prepare such image for myslef, but it says:
```
WARN[0000] Healthcheck is not supported for OCI image format and will be ignored. Must use `docker` format.
```

Adding `--format docker` option to `podman build` now contains it, so I guess I could show it if it is defined, but I think it is going to be a very rear occasion.

marusak:
> Adding --format docker option to podman build now contains it, so I guess I could show it if it is defined, but I think it is going to be a very rear occasion.

Also API does not expose it :man_shrugging:

marusak:
> "Starting health check" wraps as it's a long label. We should consider something smaller or do something else.

Right, it depends on screen size. It is never gonna be ideal as different translation etc still might be too long. We maybe could prevent manually any kind of wrapping?

marusak:
> Health check status isn't shown outside of the icon. We should have success / fail text.

Added

garrett:
Hi! What's the status of this PR?

---

I built it and I just get an "**Ooops!**" when trying it out.

Error:

```js
TypeError: containerDetail.State.Healthcheck is undefined
    renderRow Containers.jsx:389
    rows Containers.jsx:653
    card Containers.jsx:652
    render Containers.jsx:650
    React 12
    unstable_runWithPriority scheduler.development.js:468
    React 6
    updateState app.jsx:118
    inspectContainerDetail app.jsx:147
[app.jsx:149:36](webpack://podman/src/app.jsx)
```

This is after doing a git clean, making sure the branch was updated, etc.

marusak:
Oh my, the API changed and in fedora-36/rhel-8-6 it is `State.Health` but in all previous ones it is `State.Healthcheck`. $#@! stable podman API! :laughing: 

https://github.com/containers/podman/commit/1199733754b1185d0d2a7a3d832c57565864d0f4

src/client.js
```diff
@@ -114,6 +114,14 @@ export function commitContainer(system, commitData) {
     });
 }
 
+export function runHealthcheck(system, id) {
+    return new Promise((resolve, reject) => {
+        podmanCall("libpod/containers/" + id + "/healthcheck", "GET", {}, system)
```
jelly, line 119: This is already a promise so you can return it directly.

marusak: Updated

marusak:
ubuntu-stable does not get event as other OSes when new health check is run. Looking if there is some way how we can recognize this.

src/ImageRunModal.scss
```diff
@@ -77,3 +77,11 @@
...
     }
 }
+
+.healthcheck-seconds {
+    max-width: 25ch;
+}
+
+.healthcheck-retries {
+    max-width: 9ch;
+}
```
marusak, line 87: I hate this. I will use `<NumberInput>` where the width can be controlled.

marusak: changed

marusak:
@garrett I could not find any PF options to make the list wrap and not fill the full width. I put down a small hack and now it behaves as we expected it to. Is this ok?

Screenshot from 2022-05-16 12-35-17

Also now the modal looks like this:
Screenshot from 2022-05-16 12-35-32

garrett:
I'm hitting a few errors when testing out this PR. I think some are not from this PR, but I do see that when a container is running and has a health check. I'll try to see what's causing the issues here in a bit. If I can reproduce, I'll open an issue or post here (if it's relevant to this PR).

When you stop the container, the health check tab goes away. I guess it's only available when containers are running? Is there any way to know? Ideally, we'd show the past health checks, but even if health checks exist and we can't read them, then we could keep the tab around and say something about how health checks are not available unless the container is running.

garrett:
Screenshot 2022-05-16 at 14-49-07 Podman containers - garrett@Bolt

Tabs and content, including the current and list, are too wide and flowing off the side. We really need the tab overflow that I've requested from PF for quite some time.

There's an override somewhere in Cockpit — not sure if it's in the normal overrides or a local override — for tabs in the expanded tables. I think it's in the normal overrides, which we're not using in Cockpit-Podman at the moment (but should be), as seen by the gutters in mobile mode (which shouldn't be there).

Fixing the tabs _might_ fix the content. It may be flowing to be too wide as a result of the tabs making the content too wide. Or it could be another issue.

The tabs issue is also a problem for the create dialog in mobile sizes:

Screenshot 2022-05-16 at 14-53-12 Podman containers - garrett@Bolt

Screenshot 2022-05-16 at 14-53-32 Podman containers - garrett@Bolt

marusak:
> When you stop the container, the health check tab goes away. I guess it's only available when containers are running? Is there any way to know? Ideally, we'd show the past health checks, but even if health checks exist and we can't read them, then we could keep the tab around and say something about how health checks are not available unless the container is running.

Screenshot from 2022-05-17 13-16-24


seems to work here just fine :thinking: Can you reproduce this?

garrett:
> seems to work here just fine thinking Can you reproduce this?

Yes, it happens for every container I've tried here. I can try again now.

garrett:
All the issues I could quickly find related to this specific tab overflow problem I've brought up with PF over the past several years (and there might be more than just this):

- https://github.com/patternfly/patternfly/issues/1625
- https://github.com/patternfly/patternfly/pull/2757
- https://github.com/patternfly/patternfly/issues/4800
- https://github.com/patternfly/patternfly-design/issues/840
- https://github.com/patternfly/patternfly-design/issues/1034
- https://github.com/cockpit-project/cockpit-podman/issues/845

And _none_ of these have a solution for this particular problem yet. (Some solve some other tab related problems in different ways for other teams.)

We side-stepped the problem for Cockpit-Machines by _completely redesigning and overhauling the page_ as to not use tabs in expandable lists, but sub-pages instead. (Yes, this was one of the reasons, but it wasn't the only one. There were other good reasons as to why, mainly related to making a machine visible and editable all in one go instead of forcing the user to hunt around for information and for us to be constrained layout-wise.) And in other places in Cockpit, we've managed to keep the number low, like with two tabs in most other place. So it became less of a priority. But here it is again.

[Our workaround is for tertiary navigation](https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/patternfly/patternfly-4-overrides.scss#L5-L23) which looks the same, but was split and merged into tabs & nav upstream. Meanwhile, I think we're also [overriding it in our listing panel widget, with ct-listing-panel](https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/cockpit-components-listing-panel.scss#L3
), which looks the same as tabs in an expanded list.

garrett:
I added a patch that fixes tabs in Cockpit-Podman @ https://github.com/cockpit-project/cockpit-podman/pull/897/commits/4422e1a87c4446c8f5eb5491690ebcd056935dd3.

It's literally a one liner that can be summarized as:

```css
.pf-c-tabs__list { flex-wrap: wrap; }
```

It says "if there isn't enough space in the flex direction, then wrap if you need to. In the PF4 implementation of tabs, this is in the x direction (left-to-right) and the chonky overflow arrows don't show up as the content doesn't get truncated, so, instead, things overflow.

_Ideally_, we'd have this in PF4 overrides file and pull that in to Cockpit-Podman. (We should still have this in our overrides file as well, though. And it should apply in other places, such as Cockpit-Machines.)

Here's a stress test by using iPhone SE widths:

Screenshot 2022-05-17 at 16-05-28 Podman containers - garrett@Bolt

Screenshot 2022-05-17 at 16-06-11 Podman containers - garrett@Bolt

garrett:
For a direct comparison of before and after the fix, here are crops from the create modal in an iPhone SE width.

PF behavior for overflow:

image

Our wrap behavior:

image


PF for the expanded content (and the content runs over, where some isn't even visible)

image

Our wrap behavior (tabs are all visible and content fits):

image

jelly:
PR looks fine to me just needs some fixes.

marusak:
Rabased, dropped commit from Garrett and replaced it with checkout of cockpit overrides.

> PR looks fine to me just needs some fixes.

Could you be more specific what fixes? :)

test/check-application
```diff
@@ -1290,8 +1294,8 @@ class TestApplication(testlib.MachineCase):
         b.wait_not_present("#run-image-dialog-pull-latest-image")
 
         # Create Container, image is pulled and should end up being "running"
-        b.click('.pf-c-modal-box__footer #create-image-create-run-btn')
-        sel = "> span:not(.downloading)"
+        b.click('.pf-c-modal-box__footer button:contains("Create and run")')
```
jelly, line 1297: What makes you prefer using `:contains`?

test/check-application
```diff
@@ -1853,6 +1859,82 @@ class TestApplication(testlib.MachineCase):
...
+
+        self.toggleExpandedContainer("sick")
+        b.click(".pf-m-expanded button:contains('Health check')")
+        b.wait_visible(".pf-m-expanded .ct-listing-panel-body tbody tr:nth-child(1)")
+        b.wait_visible(".pf-m-expanded .ct-listing-panel-body tbody tr:nth-child(4)")
+        b.wait_visible(".pf-m-expanded .ct-listing-panel-body tbody tr:nth-child(2) svg.red")
+        b.wait_visible('#container-details-healthcheck dt:contains("Failing streak")')
+        self.assertGreater(int(b.text('#container-details-healthcheck dt:contains("Failing streak") + dd')), 3)
+
+    @testlib.skipImage("Ubuntu-stable does not get healt check run signals", "ubuntu-stable")
```
jelly, line 1930: health

test/check-application
```diff
@@ -1853,6 +1859,82 @@ class TestApplication(testlib.MachineCase):
...
+    def testHealthcheckSystem(self):
+        self._testHealthcheck(True)
+
+    @testlib.skipImage("Ubuntu-stable does not get healt check run signals", "ubuntu-stable")
```
jelly, line 1934: health

test/check-application
jelly: Is there an issue somewhere about this or is ubuntu-stable using an older podman?

test/check-application
```diff
@@ -1853,6 +1859,82 @@ class TestApplication(testlib.MachineCase):
...
+
+        b.set_input_text("#run-image-dialog-name", "healthy")
+
+        b.click("#pf-tab-2-create-image-dialog-tab-healthcheck")
+        b.set_input_text('#run-image-dialog-healthcheck-command', 'true')
+        b.set_input_text('#run-image-healthcheck-interval input', '325')
+        b.set_input_text('#run-image-healthcheck-timeout input', '35')
+        b.set_input_text('#run-image-healthcheck-start-period input', '5')
+        b.click('#run-image-healthcheck-retries button:nth-child(1)')
+        b.wait_val("#run-image-healthcheck-retries input", 2)
```
jelly, line 1885: should we make a new pixel test here?

jelly (approved):
Only found some small nitpicks, overall it works well!

test/check-application
marusak: bad rebase? :thinking:

test/check-application
marusak: yeah, I did not report this as this is related to just stable version having older podman

martinpitt:
FTR, the fcos failure is not the fault of this PR -- it still worked yesterday in https://github.com/cockpit-project/bots/pull/3390 , and spontaneously broke. We see it in #987 and #978 as well. Must be something about these three digits..

jelly (approved):

garrett (changes requested):
Tabs are broken again...

image

I did delete and re-check out the branch again and they're still broken. (Screenshot is after that.)

marusak:
> I did delete and re-check out the branch again and they're still broken. (Screenshot is after that.)

Did you delete with `git clean -dffx`? (double f). You need a new `src/lib` checkout for it to work. They wrap here just fine.

garrett (approved):
Ah, I thought I did a clean during the rebuild. I guess I didn't? Anyway, after doing the ff clean, it's working.

I still think we should do a follow-up to improve the layout in smaller sizes, but that's probably more on me, once we finally make the gutter get dropped on mobile (like we do on the rest of Cockpit), as this will already solve some of the layout woes.

For this, PR: :+1: Approved!

PR #899, opened 2022-01-30
Merge image-customize calls for rawhide scenario

martinpitt:
With image-customize now processing its arguments in the given order
[1], we can do the rawhide upgrade and the other VM preparation in one
step. This saves a now unnecessary shutdown/reboot.

[1] https://github.com/cockpit-project/bots/commit/78e199f600f

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,14 +133,13 @@ endif
 endif
 
 ifeq ($(TEST_SCENARIO),rawhide)
-# needs to run as a separate command, as --run-command is executed before --script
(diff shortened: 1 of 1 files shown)


martinpitt:
@allisonkarlitskaya FYI, a nice first gain from the recent image-customize work.

martinpitt:
I am not actually sure why we want the rawhide upgrade to happen *after* installing cockpit-podman.rpm. But the [build/install log](https://logs.cockpit-project.org/logs/pull-899-20220130-145829-dd9369a3-fedora-35-rawhide/log) shows that it maintains the same order as before.

marusak (approved):
Nice, thanks!

martinpitt:
Argh @ TF f-35 -- *still* that selinux bug? sigh

PR #900, opened 2022-01-30
fmf: Test fixes

martinpitt:

PR #901, opened 2022-01-31
Use image-customize --build to build/install distribution packages

martinpitt:
This centralizes/factorizes the rpm/deb/arch package builds, and now also
builds RPM packages in the VM instead of on the host, which is cleaner.

Drop the `make srpm` rule, as it's not very useful. Keep the `make rpm`
rule, as sometimes developers do this manually. This might be replaced
later on with another image-customize feature which copies the built rpm 
out of the VM. 

----

Dependencies:
 - https://github.com/cockpit-project/bots/pull/2875
 - https://github.com/cockpit-project/bots/pull/2876

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -1,5 +1,5 @@
 #!/bin/sh
-# image-customize script to enable cockpit in test VMs
+# image-customize script to prepare a bots VM for cockpit-podman testing
 set -eu
 
 # for Debian based images, build and install debs; for RPM based ones, the locally built rpm gets installed separately
@@ -35,6 +35,7 @@ if [ -d /var/tmp/debian ]; then
     fi
 
     # tuned is installed for testing cockpit; but it causes funny bugs, and we are not testing this here
+    # https://launchpad.net/bugs/1774000 https://launchpad.net/bugs/1925765
     systemctl disable tuned
 
 # install rpms
--- a/test/vm.install
+++ b/test/vm.install
@@ -2,29 +2,7 @@
 # image-customize script to prepare a bots VM for cockpit-podman testing
 set -eu
 
-# for Debian based images, build and install debs; for RPM based ones, the locally built rpm gets installed separately
-if [ -d /var/tmp/debian ]; then
-    # build source package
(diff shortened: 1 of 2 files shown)


martinpitt:
This almost works, but requires `appstream-glib` in our [pre-built mocks on rhel-8-6](https://logs.cockpit-project.org/logs/pull-901-20220131-104343-a3acf221-rhel-8-6-bots-2875/log.html).

martinpitt:
@allisonkarlitskaya FYI

test/vm.install
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-# image-customize script to enable cockpit in test VMs
+# image-customize script to prepare a a bots VM for cockpit-podman testing
```
marusak, line 2: `a`nonymous `a`lcoholics bots?

martinpitt:
Wait, what? The double "a a bots" was apparently vitally important -- there are a lot of weird test failures now. Investigating..

marusak:
> Wait, what? The double "a a bots" was apparently vitally important -- there are a lot of weird test failures now. Investigating..

Of course, if that fixes the issue I am fine with keeping it :smile:

marusak:
I just finished reviewing this PR (got interrupted by the standup)  and these changes are all fine for me, thanks! Of course not approving now as tests are very sad.

martinpitt:
The [TF f35 failure](https://github.com/cockpit-project/cockpit-podman/pull/901/checks?check_run_id=5053686829) is still good old "broken SELinux version". The [TF rawhide failure](https://github.com/cockpit-project/cockpit-podman/pull/901/checks?check_run_id=5052944185) is proof that someone doesn't understand package dependencies, and I worked around it in #900. Both are unrelated here.

Also, this isn't the kind of change that could break individual tests -- either preparing the VM fails, or the package would somehow be broken and tests wouldn't work at all. The cockpit failures were some networking flake, all fine now.

So, ready for review.

marusak (approved):
Thanks!

Issue #864, opened 2021-12-15
mobile: Images list is broken


garrett:
When the images list is in mobile, it looks like this (on a very large phone even; this is the iPhone 13 Pro Max size):

![image](https://user-images.githubusercontent.com/10246/146195673-c332b817-136f-42ab-9e86-2187d910a85d.png)

This is related to #844, but is additionally even more problematic due to the action button. I suggest reworking the list with the same technique in the (currently yet-to-land) storage page:

- https://github.com/cockpit-project/cockpit/pull/16681

mvollmer:
The code in the storage page is impossible to re-use, unfortunately.  We could factor it out into a reusable component and put it into lib/. Or should we wait for a PF fix?

PR #902, opened 2022-01-31
Images: Improve mobile layout

mvollmer:
- Don't use a Flex to group button and menu, this avoids some extra
  space between the button and the menu.

- Don't give the actions column a title, this avoids some extra space
  between the first field and the menu in mobile layout.

- Let the image name be broken at every character, this prevents the
  table from being wider than its container at certain widths.

- Move the "Create container" button into the menu in mobile layout.
  This is done by always having both button and menu entry in the DOM,
  and only showing one of them as appropriate.

- Take control over the card header layout, so that the actions stay
  in the top-right corner.  Also, the title stats stay together as
  long as possible.

Fixes: #864

New pixel test references: https://github.com/cockpit-project/pixel-test-reference/compare/8ae010f8744dc2eacef7f860459b4f3aa57ea0bc..e45318da3c227623af5062326fab086d87f7e9cd

### DIFF
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -1,5 +1,6 @@
 @use "ct-card.scss";
 @use "page.scss";
+@import "global-variables";
 // For pf-line-clamp
 @import "@patternfly/patternfly/sass-utilities/mixins.scss";
 
@@ -79,4 +80,21 @@
 .container-section.pf-m-plain tbody,
 .containers-images tbody {
     border: var(--pf-c-card--m-flat--BorderWidth) solid var(--pf-c-card--m-flat--BorderColor);
+}
+
+.content-action {
+    text-align: right;
+    white-space: nowrap !important;
+}
+
+@media ( max-width: $pf-global--breakpoint--md - 1 ) {
+    .show-only-when-wide {
+        display: none;
+    }
+}
+
+@media ( min-width: $pf-global--breakpoint--md ) {
+    .show-only-when-narrow {
+        display: none;
+    }
 }
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -1,9 +1,9 @@
 import React, { useState } from 'react';
 import {
     Button,
-    Card, CardBody, CardHeader, CardTitle, CardActions, CardFooter,
+    Card, CardBody, CardHeader, CardFooter,
     Dropdown, DropdownItem,
-    Flex,
+    Flex, FlexItem,
     ExpandableSection,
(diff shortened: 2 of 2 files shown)


mvollmer:
Note that the header is still not great:

image

mvollmer:
So this doesn't actually use a general component to do the button/menu dance, but does it all very explicitly... are we okay with that?

marusak:
> Note that the header is still not great:

Plan on fixing it as well in this PR?

> So this doesn't actually use a general component (which would have to be written) to do the button/menu dance, but does it all very explicitly... are we okay with that?

I am. Maybe eventually we want to think of some more general approach, but I am absolutely fine with this. It is simple enough and I am a bit afraid the component itself would have to be rather complex to deal with all situations.

mvollmer:
> I am a bit afraid the component itself would have to be rather complex to deal with all situations.

Yes, it will have to worry about menus that become empty in a wide layout, and separators between menu items that might or might not be visible.  The storage code doe most of that, but is not fully general.  (It can only handle a single separator.)  A fully general code will be non-trivial, but I think it will be quite clean.

My idea was to make a component that takes the PF [IAction](https://github.com/patternfly/patternfly-react/blob/94aea84036e26f4cc472b7d445ee49b088eb6a47/packages/react-table/src/components/Table/TableTypes.tsx#L149) description.

mvollmer:
> > Note that the header is still not great:
> 
> Plan on fixing it as well in this PR?

Let me try...

mvollmer:
> Note that the header is still not great:

I had a go, looks like this at our mobile width of 400ish pixels:

image

And like this when slightly wider:

image

I can't say I know what I am doing, though.  I just replaced CardTitle and CardActions with a custom Flex, and then wrapped the stats in another Flex to get the line to break between "Images" and the stats first, instead of between the stats.

src/podman.scss
```diff
@@ -1,5 +1,6 @@
 @use "ct-card.scss";
 @use "page.scss";
+@import "global-variables";
```
marusak, line 3: What is this file and what do we need it for? Does it define `$pf-global--breakpoint--md`?

marusak: nvm, just tested it and indeed it defines these variables

marusak (approved):
Looks great, thank you!

src/podman.scss
mvollmer: Yes, exactly.

PR #909, opened 2022-02-05
Build valid arch PKGBUILD and Debian changelog in dist tarballs

martinpitt:
Similarly how we already create a valid spec file, rename PKGBUILD and
debian/changelog to *.in and replace the VERSION and SOURCE macros at
`make dist` time.

With that we can drop these specific hacks from `image-customize --build`.


Makefile
```diff
@@ -81,12 +87,12 @@ dist: $(TARFILE)
...
 # node_modules/ can be reconstructed if necessary)
 $(TARFILE): export NODE_ENV=production
-$(TARFILE): $(WEBPACK_TEST) $(SPEC)
+$(TARFILE): $(WEBPACK_TEST) $(SPEC) packaging/arch/PKGBUILD packaging/debian/changelog
 	touch -r package.json $(NODE_MODULES_TEST)
 	touch dist/*
 	tar --xz -cf $(TARFILE) --transform 's,^,cockpit-$(PACKAGE_NAME)/,' \
-		--exclude packaging/$(SPEC).in --exclude node_modules \
-		$$(git ls-files) src/lib/ package-lock.json $(SPEC) dist/
+		--exclude *.in --exclude node_modules \
```
martinpitt, line 94: whoops, this glob should be quoted, in case we ever get an `*.in` file in the root directory.

martinpitt: Fixed.

marusak (approved):
Danke :)

PR #910, opened 2022-02-08
Use image-customize's new --fresh function

martinpitt:
See https://github.com/cockpit-project/bots/commit/468ee1b39e2

This avoids hardcoding the test image overlay paths.


allisonkarlitskaya (approved):
deja vu :)

PR #919, opened 2022-02-16
test: Update pixel references for Chromium 96

martinpitt:
Our recent tasks container refresh updated Chromium from 94 to 96, which
changes the pixels a bit.

[pixel diffs](https://github.com/cockpit-project/pixel-test-reference/compare/44585b3d10aebacd9d543c3efbb72e4d77102e32..a82718624031fc91eec3ddf0f90b110f29d38a0e)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,13 +115,9 @@ rpm: $(TARFILE) $(SPEC)
 	rm -r "`pwd`/rpmbuild"
 	rm -r "`pwd`/output" "`pwd`/build"
 
-ifeq ($(TEST_SCENARIO),rawhide)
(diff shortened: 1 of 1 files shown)


martinpitt:
This [GPG/package download error](https://logs.cockpit-project.org/logs/pull-919-20220216-060801-cd0b8481-fedora-35-rawhide/log.html) in the rawhide scenario reproduces locally. Possibly related to the Fedora 36 branching and changing GPG keys?

However, as we now test F36 and rawhide in packit anyway (and know that it fails), perhaps it's time to drop that scenario.

marusak (approved):
Thanks!

PR #920, opened 2022-02-16
test: Accept 'Created' for not started containers

marusak:
We only create container and not start it. This status makes more sense
in general. It must have been updated in v4.

FWIW https://github.com/containers/podman/commit/1ef96364fe
It seems that there is precedence for mapping configured to created.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -922,14 +922,18 @@ class TestApplication(testlib.MachineCase):
         with b.wait_timeout(5):
             self.waitContainer(container_sha, auth, name='swamped-crate', image='busybox:latest', cmd='sh',
                                state='Running', owner="system" if auth else "admin")
+
+        def get_cpu_usage(sel):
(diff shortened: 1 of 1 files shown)


marusak:
ok, also `testLifecycleOperationsSystem` fails on rawhide

test/check-application
```diff
@@ -1657,7 +1657,7 @@ class TestApplication(testlib.MachineCase):
 
         b.wait_val("#containers-containers-filter", "all")
         sha = self.execute(auth, "podman inspect --format '{{.Id}}' " + container_name).strip()
-        self.waitContainer(sha, auth, name=container_name, image='busybox:latest', state='Configured')
+        self.waitContainer(sha, auth, name=container_name, image='busybox:latest', state=['Configured', 'Created'])
```
jelly, line 1660: I'd love a comment here explaining why it was changed, but it's fine for now.

jelly (approved):

PR #930, opened 2022-02-23
Fix flex and improve table alignment for Container & Image lists

garrett:


garrett:
I didn't see where PF overrides are located, so I added some to podman.scss for now (in the first comment). For the second commit, I changed the way things are aligned for various fields in the images and containers lists... but this could be done better, probably inside the table itself (if this is even possible) or by using some more generic local classes.

I wanted to share what I have so far, so it's not just sitting on my computer, and to get feedback.

We'll probably want to change the implementation of 9553424 and we'll definitely want to upstream 543249b to both Cockpit's workarounds and also to PatternFly itself.

garrett:
I just converted this to a draft to indicate that it _probably_ needs some adjustments before being merged, as I stated above.

garrett:
Here's what it looks like:

image


This alignment makes it more obvious which are user and which are system containers and makes it easier to compare percentages, space, etc.

Ideally, we'd want to:

- turn on PatternFly's improve RedHatFont support
   - including the tabular numbers
- switch to GB instead of GiB
- fix the rounding to two significant digits at max (as it's absurdly specific now) for space and make percentages 1 (or even 0) significant digit

...but these all will require some code changes and could be in a follow-up PR.

src/podman.scss
```diff
@@ -86,4 +86,39 @@
...
+            &[data-label=Owner], &[data-label=CPU], &[data-label=Memory], &[data-label="Disk space"], &[data-label="Created"] {
+                text-align: right;
+            }
+
+            &[data-label=CPU] {
+                min-width: 6ch;
+            }
+
+            &[data-label=Memory] {
+                min-width: 14ch;
```
marusak, line 105: This would not work when localized. Data labels are localized as well. But also, cannot we just pass this as property through JS. We define these columns so it should be possible to tell them, if they are for example left or right aligned.

src/podman.scss
```diff
@@ -86,4 +86,39 @@
...
+            }
+        }
+    }
+}
+
+/* PatternFly workarounds */
```
marusak, line 111: Do we have report for this in BZ?

src/podman.scss
garrett: Yeah, I was afraid of that.

It's more than being right or left aligned, of course. But for whatever we can't represent with PF, we could add classes.

src/podman.scss
garrett: Not yet. It was already late when I posted this PR.

I need to port it to our overrides file in Cockpit and open a PR. I intend on doing this today.

garrett: Also, what I _didn't_ mention above is that the `> :empty {}` part is to work around an empty block which we should just not show in the JSX. I'll just port over the gap & margin change (not the `:empty` one... as we should probably remove it in the CSS and fix it in the JSX) to PF and suggest altering it upstream by moving the margin to gap instead of overriding the margin (which is what we have to do in a workaround, of course).

garrett: Additionally, Cockpit Podman should pull in the PF workarounds from Cockpit as part of the library sync.

garrett: I found that there's an issue about gap in flex. It's not a 100% match, as it wants to do more. But fixing the gap to use the margin is an obvious first step and definitely part of the issue.

https://github.com/patternfly/patternfly/issues/4523

garrett: Added in Cockpit @ https://github.com/cockpit-project/cockpit/pull/17043

src/podman.scss
garrett: I mentioned it last night when I was hastily pushing up this draft PR (as it was late): "but this could be done better, probably inside the table itself".

I may need help with this to locate where that should be done. I ran out of time last night to do this properly, so sending up a draft to demonstrate how it should be is better than keeping it to myself.

src/podman.scss
marusak: > Also, what I didn't mention above is that the > :empty {} part is to work around an empty block which we should just not show in the JSX

Can you please point me to html where this happens and I can take a look why we render empty elements there

garrett: It's the line that has a highlight. It's in the details view:

Screenshot from 2022-02-24 16-31-41

The first div has the id, image, and command. The third div has the created and state. The middle div probably should have content in some cases, but doesn't here.

src/podman.scss
```diff
@@ -86,4 +86,44 @@
...
+
+    > * {
+        margin-right: unset;
+    }
+
+    /* Flex should ignore empty blocks in the flow */
+    // Ideally: Do not render empty children in JSX
+    > :empty {
+        display: none;
+    }
```
jelly, line 128: This totally breaks the displaying of disabled inputs.

jelly (changes requested):

src/podman.scss
marusak: @garrett I think you are running some older version.
I dont see here `<div>` but `<dl>` and empty ones are hidden.

Screenshot from 2022-02-25 12-24-07


Can you please make sure you are fully rebased and building from a clean tree and that your checkout is actually being used? I tried to find where this was updated and I think it is https://github.com/cockpit-project/cockpit-podman/commit/bf1de36cc554c826295a4d777064d859b675ee92 but that is half a year old so not sure.

marusak:
Right, so I think here are more questions then answers.
The `:empty` pattern seems unnecessary and even actively breaks stuff.
The other change should be imported from cockpit through library - I am not certain if we include overwrites - if not, then we should add it and then check how it affects the whole plugin.
Finally the label approach is not usable and should be preferably written in jsx.

So I think we need to scrape all that is here and do this from scratch. I don't think though that we need to block on this releases as I don't want to rush this since it has potential to break something somewhere (as for example it now breaks disabled inputs).

marusak:
Screenshot from 2022-02-28 10-11-10

@garrett this PR looks broken to me - the `text-align: right` seems to move the columns all over the place. Should the column labels be also align the same way?

Also, I wrote it in JS - it is no biggie. The issue is, if we want to condition this only for some screen width then it won't be possible to do in JS. But if we want to apply these changes for all screen resolutions, then it will be fine. I have changes pending, but I posted here two questions, so we need to resolve them first.

> The other change should be imported from cockpit through library - I am not certain if we include overwrites - if not, then we should add it and then check how it affects the whole plugin.

We do, we just need to bump the library in makefile.

garrett:
> The :empty pattern seems unnecessary and even actively breaks stuff.

It _shouldn't_ break anything. If it does, then whatever it "breaks" is already broken. We shouldn't render out elements without content.

But, agreed; the proper place to do this is not in CSS, but fix it in JSX.

I sent this as-is as a draft to show how things should act, not to be the final code.

> The other change should be imported from cockpit through library - I am not certain if we include overwrites - if not, then we should add it and then check how it affects the whole plugin.

Agreed, it should be in overrides... but there is no overrides. It should be brought in.

The override change for flex started in this PR and was expanded. It's a good idea to drop it from this PR and pull in the fix via overrides.

> Finally the label approach is not usable and should be preferably written in jsx.

Agreed; I said that at the beginning. That's why this PR is still set as draft. I wanted to show how things should look, but it needs to be implemented elsewhere.

> So I think we need to scrape all that is here and do this from scratch. I don't think though that we need to block on this releases as I don't want to rush this since it has potential to break something somewhere (as for example it now breaks disabled inputs).

OK, sure.

(It shouldn't break disabled inputs... But anyway, it was a draft to demonstrate how the columns should act, not to be considered to merge as-is.)

garrett:
Rebased. This still needs someone to pick it up and implement it correctly.

marusak:
Sent https://github.com/cockpit-project/cockpit-podman/pull/942 for implementation in JS.

marusak:
Everything already landed except for right aligned columns - that is now in #942

PR #936, opened 2022-03-02
Always use base 10 units

marusak:


martinpitt (approved):
So you... rebased this branch to 10? :stuck_out_tongue_winking_eye: 

Thanks!

src/ImageRunModal.jsx
```diff
@@ -373,7 +373,7 @@ export class ImageRunModal extends React.Component {
         }
         const resourceLimit = {};
         if (this.state.memoryConfigure && this.state.memory) {
-            const memorySize = this.state.memory * (1024 ** units[this.state.memoryUnit].base1024Exponent);
+            const memorySize = this.state.memory * (1000 ** units[this.state.memoryUnit].baseExponent);
```
garrett, line 376: Oh, I thought we were using base 10 on everything except memory?

martinpitt: Just "total (physical) memory size". Container memory can be arbitrary, and even on the CLI with `--memory` you specify base-10 prefixes. So IMHO that's fine.

marusak: This one is from the modal, where users can pick amount. I think we should put there base 10 units.

But I noticed on overview:


My PC has 32GB RAM, so this is wrong.
Before it was:
Screenshot from 2022-03-02 18-53-39

which is not great, but at least correct. Looking where these data come from

marusak: Or do I? Everywhere is written that I have 31.1 GiB which is 33.4GB indeed :man_shrugging: So I am confused

marusak:
```
$ grep MemTotal /proc/meminfo
MemTotal:       32640988 kB
```
that is 32.64 Gb unless the unit should be `kiB` in which case it is `31.1 GiB`. Seems the units are all over the place.

```
$ free --si -g
               total        used        free      shared  buff/cache   available
Mem:              33          12           1           1          19          18
Swap:             25           0          25
```

```
               total        used        free      shared  buff/cache   available
Mem:              31          12           1           1          18          16
Swap:             23           0          23
```

So I guess I really have 33.4 GB.

martinpitt: Plausible. Laptop with builtin graphics card, which takes a bit off the 32 GiB for its own memory?

garrett: A quick and dirty way to determine video memory size:
```bash
lspci -v | grep " prefetch"
```
(Assuming the video card is the only prefetchable one and the rest are non-prefetchable.)

A more robust way would be to check the identifier:

```bash
lspci  -v -s  $(lspci | grep VGA | cut -d" " -f 1) | grep " prefetchable"
```

(Drop the $ for fish shell.)

---

For my T14s, that means 512M is used for the GPU. I think it can be altered in BIOS/EFI settings.

However, things don't add up still — or I just need more coffee before I attempt math in the morning. :wink:

I have 31Gi when running `free` and it lists 32615484 bytes total.

Extremely fuzzy rounding math, ignoring reported units: of 31Gi + 0.5 GB ≅ 32 with rounding... sure, I guess.

---

A wild guess: There could even be reserved memory space for the kernel that nothing else could touch. It would probably be a security feature.

Perhaps it's a side effect of Exec Shield or something like it? https://en.wikipedia.org/wiki/Exec_Shield

> The first Exec Shield patch **attempts to flag data memory** as non-executable and program memory **as non-writeable**. This suppresses many security exploits, such as those stemming from buffer overflows and other techniques relying on overwriting data and inserting code into those structures. Exec Shield also supplies some address space layout randomization for the mmap() and heap base. 

(Emphasis mine)

marusak:
Indeed, even `podman stats` outputs these numbers:
```
ID            NAME        CPU %       MEM USAGE / LIMIT  MEM %       NET IO      BLOCK IO    PIDS        CPU TIME    AVG CPU %
0329cfac88b0  frown       --          5.091MB / 33.42GB  0.02%       -- / --     -- / --     1           24.617ms    7.47%
5ac5a8410164  heartzy     --          5.825MB / 33.42GB  0.02%       -- / --     -- / --     1           26.2ms      7.96%
b7b63a68ed89  unhealthy   --          6.246MB / 33.42GB  0.02%       -- / --     -- / --     1           47.281839s  1.44%
```

and we agreed that we will show base 2 for physical memory (like on HW page) but that does not apply for containers.

PR #937, opened 2022-03-07
packit: build SRPM in Copr

TomasTomecek:
and be able to specify the precise list of deps needed to create the
SRPM

just testing without any extra deps


TomasTomecek:
uhhh, this was meant to be my fork, not upstream; sorry for the noise

PR #938, opened 2022-03-09
packit: build SRPM in Copr

TomasTomecek:
and be able to specify the precise list of deps needed to create the
SRPM

Details: https://packit.dev/posts/copr-srpms/

~~The current implementation requires the key `srpm_build_deps` to be set and since cockpit-podman doesn't require any extra, the value is an empty list.~~ make is needed :)


martinpitt (approved):
Danke!

PR #939, opened 2022-03-09
Move Owner option to Details tab

jelly:
Align the podman create image design with the in progress machines
dialog redesign. The owner option is now moved to the details tab and
changed to a radio button so it's more clear it's a form option. The
container name input can now take up the full input which is nicer on
mobile devices.

---

# Redesign

image

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -7,8 +7,8 @@ import {
     FormSelect, FormSelectOption,
     Grid, GridItem,
     HelperText, HelperTextItem,
-    Modal, Select, SelectVariant,
-    SelectOption, SelectGroup, Stack,
+    Modal, Radio, Select, SelectVariant,
+    SelectOption, SelectGroup,
     TextInput, Tabs, Tab, TabTitleText,
     ToggleGroup, ToggleGroupItem,
     Flex, FlexItem,
@@ -317,7 +317,6 @@ export class ImageRunModal extends React.Component {
             memoryUnit: 'MB',
             validationFailed: {},
             volumes: [],
-            runImage: true,
             restartPolicy: "no",
             restartTries: 5,
             pullLatestImage: false,
@@ -333,7 +332,6 @@ export class ImageRunModal extends React.Component {
             searchByRegistry: 'all',
         };
         this.getCreateConfig = this.getCreateConfig.bind(this);
-        this.onCreateClicked = this.onCreateClicked.bind(this);
         this.onValueChanged = this.onValueChanged.bind(this);
     }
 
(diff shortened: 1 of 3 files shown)


garrett (changes requested):
It looks great!

There's an alignment issue, however:

Screenshot from 2022-03-09 16-11-14

garrett:
Oh, and it's called "Location" in Machines. Should it be the same here too? We should make the two consistent when it makes sense.

garrett:
Ahem...

image

jelly:
> It looks great!
> 
> There's an alignment issue, however:
> 
> Screenshot from 2022-03-09 16-11-14

That's missing `hasNoPaddingTop`, fixed.

jelly:
> Ahem... (Name entry appears _below_ its label.)
> 
> image

Is that on mobile? Firefox? Web?

garrett:
> Is that on mobile? Firefox? Web?

Firefox, just switched tabs to integration and this happened.

garrett:
So I have an idea and it seems to work...

Locally recreate `.pf-c-form__group.pf-m-horizontal`, but have it apply at the FormGroup level instead of Form.

```patch
diff --git src/ImageRunModal.jsx src/ImageRunModal.jsx
index e438f4a6..8a897145 100644
--- src/ImageRunModal.jsx
+++ src/ImageRunModal.jsx
@@ -810,8 +810,8 @@ export class ImageRunModal extends React.Component {
         );
 
         const defaultBody = (
-            <Form isHorizontal={activeTabKey == 0}>
-                <FormGroup fieldId='run-image-dialog-name' label={_("Name")}>
+            <Form>
+                <FormGroup fieldId='run-image-dialog-name' label={_("Name")} className="ct-m-horizontal">
                     <TextInput id='run-image-dialog-name'
                            className="image-name"
                            placeholder={_("Container name")}
diff --git src/ImageRunModal.scss src/ImageRunModal.scss
index 51fb65f6..ccde9424 100644
--- src/ImageRunModal.scss
+++ src/ImageRunModal.scss
@@ -1,3 +1,5 @@
+@import "global-variables";
+
 .dynamic-form-group {
     .pf-c-empty-state {
         padding: 0;
@@ -65,3 +67,13 @@
 .run-image-dialog-restart-retries {
     max-width: 5ch;
 }
+
+// HACK: A local copy of pf-m-horizontal (as ct-m-horizontal),
+// but applied at the FormGroup level instead of Form
+@media (min-width: $pf-global--breakpoint--md) {
+    .pf-c-form__group.ct-m-horizontal {
+        display: grid;
+        grid-column-gap: var(--pf-c-form--m-horizontal__group-label--md--GridColumnGap);
+        grid-template-columns: var(--pf-c-form--m-horizontal__group-label--md--GridColumnWidth) var(--pf-c-form--m-horizontal__group-control--md--GridColumnWidth);
+    }
+}
```

If you like this (ping me on IRC about it), I can push up my change.

garrett (approved):
Looks good after our changes.

PR #940, opened 2022-03-10
Port AppStream i18n from starter-kit

martinpitt:
Tested by `make update-po`, adding German translations locally, `make install DESTDIR=/tmp/x` and inspecting /tmp/x/usr/share/metainfo/org.cockpit-project.podman.metainfo.xml . This works fine.

After that lands, and the next POT refresh, I'm happy to add the translations to weblate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ test/images/
 *.pot
 POTFILES*
 tmp/
+/po/LINGUAS
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ export TEST_OS
 TARFILE=cockpit-$(PACKAGE_NAME)-$(VERSION).tar.xz
 NODE_CACHE=cockpit-$(PACKAGE_NAME)-node-$(VERSION).tar.xz
 SPEC=$(RPM_NAME).spec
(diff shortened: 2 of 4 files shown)


martinpitt:
Not obvious why, but breaks a lot of tests in packit. → Draft

martinpitt:
@marusak : Thanks for the hint! (I didn't really look yet). As first step I added showing the package versions in the log, similar to what I recently did for c-machines and cockpit. If it's version confusion, then we apply the same old recipe, but I first want to confirm this.

martinpitt:
Got the [logs](https://artifacts.dev.testing-farm.io/851614aa-1298-401a-8aed-5c8de7fe13b8/):
```
cockpit-podman-43-1.20220310152120999659.pr940.9.gd72967a.fc35.noarch
cockpit-bridge-264-1.fc35.x86_64
```

This seems alright. We didn't change cockpit yet. But indeed my branch is 2 commits behind, so maybe that just was it (I thought packit would rebase these days?)

martinpitt:
Ah, that was it - glad that it's not yet another breakage :grinning:

marusak (approved):
Thanks!

Issue #597, opened 2020-11-17
Alignment issues


garrett:
1. Memory wraps (in containers). A unit should _never_ wrap away from what it is describing.
2. Size (in images) is misaligned. Sizes should be right-aligned, so the units line up.
3. Expander arrow is vertically too low by a few pixels.
4. Spinner is not vertically aligned; it is far too low when searching on the "Get new image" window. (Not pictured.)

![Screenshot_2020-11-17 Podman Containers - garrett Rain](https://user-images.githubusercontent.com/10246/99409207-687a3180-28f1-11eb-9506-4a8eb929e203.png)

Screenshot:

garrett:
Dates are mm/dd/yyyy here; that's not an ideal format. It would be better if they would be human-friendly relative with a mouseover/focus with the full date in the local date.

![Screenshot_2020-11-17 Podman containers - garrett Rain](https://user-images.githubusercontent.com/10246/99414601-42579000-28f7-11eb-8842-87570c2871ab.png)

Still, formatted like this, they're out of alignment and would need to be right-aligned.

marusak:
> Dates are mm/dd/yyyy here; that's not an ideal format. 

That is what moment offers us as it is default format for english. We use it all over the place in cockpit, like in selinux:
![selinuxdate](https://user-images.githubusercontent.com/12330670/99417636-6c5e8180-28fa-11eb-976a-eab9f3b32b0f.png)

It is being localized, so for example in czech it looks like this:
![localized](https://user-images.githubusercontent.com/12330670/99417687-7da78e00-28fa-11eb-8806-6f15937e669f.png)

Also it shows this date only for older images, for anything newer it has more verbose form:
![freshbuild](https://user-images.githubusercontent.com/12330670/99417939-c3fced00-28fa-11eb-98bf-643b736c99a3.png)

> need to be right-aligned

I though that just the last column should be right aligned and all other left aligned. I think it makes it easier to read.

> we could alternately make the numbers monospaced, I guess.

That would make things like "Today at" awkward to read and inconsistent with rest of the text.

garrett:
There's not too much we can do currently with the date alignment, unless we right-align the dates, as we would need a monospaced font otherwise, which would align at least the years.

> I though that just the last column should be right aligned and all other left aligned. I think it makes it easier to read.

PatternFly has alignment classes, if you want to stick to PF4 in React:
https://www.patternfly.org/v4/utilities/alignment/

[...]

PR #942, opened 2022-03-17
Right align some columns

marusak:
It behaves the same as #930 but does not look great:
Screenshot from 2022-03-17 07-29-53

@garrett mentioned that columns should  constrain to their content. Not sure how to do that. By any chance, do you know if we use this right-alight and constraining anywhere else when using ListingTable?

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -378,9 +378,9 @@ class Containers extends React.Component {
         const containerState = container.State.charAt(0).toUpperCase() + container.State.slice(1);
         const columns = [
             { title: info_block },
-            { title: container.isSystem ? _("system") : <div><span className="ct-grey-text">{_("user:")} </span>{this.props.user}</div>, props: { modifier: "nowrap" } },
-            { title: proc, props: { modifier: "nowrap" } },
-            { title: mem, props: { modifier: "nowrap" } },
+            { title: container.isSystem ? _("system") : <div><span className="ct-grey-text">{_("user:")} </span>{this.props.user}</div>, props: { modifier: "nowrap", style: { "text-align": "right" } } },
+            { title: proc, props: { modifier: "nowrap", style: { "text-align": "right" } } },
+            { title: mem, props: { modifier: "nowrap", style: { "text-align": "right" } } },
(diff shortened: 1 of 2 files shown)


marusak:
The wrapping commit seems fine so I picked it into https://github.com/cockpit-project/cockpit-podman/pull/943 also bumped pkg/lib to bring in animations of new items and pull in https://github.com/cockpit-project/cockpit-podman/pull/930/commits/46e94f0de60d000480a9fb266ec39cc2c891a8fc from #930

garrett (changes requested):
I have the CSS in the other PR constrained to the content. But there was no way to attach the CSS to the rows except by how I did it. (Using offset numbers is not a valid method, as that can easily change and doesn't explain what's going on. There needs to be a way to target the cells by the type.)

Additionally:

- Legacy tables have _some_ ways to control the text: https://www.patternfly.org/v4/components/table/react-legacy#controlling-text
- Composable tables seem to have noWrap as an option too? https://www.patternfly.org/v4/components/table/#composable-custom-row-wrapper-header-tooltips--popovers

But I don't think the PF React methods are suitable enough. 

Elsewhere in Cockpit, either we don't have such dense information or we add classes to the `th` and `td` elements and use CSS to specify the approximate sizes.

garrett:
Also related to alignment and displaying numbers: I also mentioned in https://github.com/cockpit-project/cockpit-podman/pull/930#issuecomment-1049043237 about fixing the rounding so it's not absurd like this. I opened an issue about this for Metrics, but it would be useful across all of Cockpit:

- https://github.com/cockpit-project/cockpit/issues/17769

I provided an example function in that issue that does the job. It handles floats. If we're just interested in whole, positive numbers, we could even simplify it down to `num ? Math.round(num) || 1 : 0` (but I think we want decimal places and probably want to display a `< ` too).

KKoukiou:
@marusak this can't be  solved like this unfortunately. It's quite more complicated this closing this PR as it's also quite state. 
We have https://github.com/cockpit-project/cockpit-podman/issues/597 to track.

PR #943, opened 2022-03-17
Don't wrap selected columns and bring in animations

marusak:


marusak:
Hmm, we lost spacing in front of unused images: https://logs.cockpit-project.org/logs/pull-943-20220317-065919-fd9e74e1-fedora-35/pixeldiff.html#TestApplication-testBasicSystem-overview

Looking

marusak:
So this is actually fallback from https://github.com/cockpit-project/cockpit/pull/17043

If you do something like:
```
<Flex id="first">
    <FlexItem>
        <Flex id="second">
            <FlexItem id="a" />
            <FlexItem id="b" />
        </Flex>
    </FlexItem>
</Flex>
```
then elements "a" and "b" won't have any gap in between them.

Screenshot from 2022-03-17 09-19-08

The second flex is actually `:last-child` so for it is defined `--pf-l-flex--spacer: 0;` and that is then used for `gap`.

@garrett should we re-define also the `:last-child` rule from PF in our hack?

jelly:
> Hmm, we lost spacing in front of unused images: https://logs.cockpit-project.org/logs/pull-943-20220317-065919-fd9e74e1-fedora-35/pixeldiff.html#TestApplication-testBasicSystem-overview
> 
> Looking

Good that we have pixel tests \o/

garrett:
> The second flex is actually :last-child so for it is defined --pf-l-flex--spacer: 0; and that is then used for gap.

I don't think that matters at all. Gap is set at the `display: flex` level. Changing it in a child wouldn't change the value in the parent. At the child level, that variable is only used for margin and we don't want that anyway. The workaround sets it to 0, but reuses the value on the parent for gap.

And, unless I'm doing something wrong or if there's some local fix, I don't see a problem with the lack of a gap locally:

Screenshot 2022-03-21 at 18-06-48 Podman containers - garrett@Bolt

...but on closer inspection, I don't see the workaround; I still see margin spacing... so I guess there's something wrong here.

But, still: You can't change a value in a child and have it reflected in the parent. In other words, "CSS rolls downhill." :wink: So it'd be something else.

Now... I wonder what's going wrong here. I'm on the branch, I built it, and... nada. No imported CSS. I guess I'll git clean and try again.

garrett:
Eureka! I have the fix. In overrides, change `*` to `:not(.pf-l-flex)`. So that part of the snippet would look like this instead:

```scss
    // Negate the margin hack used by immediate flex children
    > :not(.pf-l-flex) {
        --pf-l-flex--spacer-base: 0;
    }
```

So it fixes all children except for immediate nested flex.

I guess we need to apply this to Cockpit first, then import the fix here?

marusak:
> OK, doing a git clean fixes the issue and I see what you're seeing now.

Yes, you needed to do that, as we include `/src/lib` just by checking it out and only control if this directory exists. So when we update the checkout version, we need to remove this directory. `git clean -dffx` usually works the best.

> Eureka! I have the fix. 

Awesome!! Thank you

marusak:
@garrett So I updated the checkout to include the new change and now there is one more difference: https://logs.cockpit-project.org/logs/pull-943-20220322-090639-0cff4967-fedora-35/pixeldiff.html#TestApplication-testBasicSystem-overview-mobile

Can you please check out if that is fine? Thanks!

marusak:
Right, I tried to do this but with no success - when I define either `spaceItems` or `spacer` it does reduce space when it wraps, but it also reduces space when it does not wrap - which is then very hard to read.
@KKoukiou do you know by any chance how to do that in `<Flex>`?

garrett:
Ah, we'd need to be able to set it up on horizontal and vertical, which PF doesn't seem to do, oddly. (For HTML + CSS this is no problem whatsoever.)

It's either done with shorthand like:

```css
gap: Y X;
```
(Where Y and X are values for vertical and horizontal.)

or with:

```css
row-gap: Y;
column-gap: X;
```

We could set one of these explicitly here. There isn't a way for PF to use it, as they don't use gap in flex for some really odd reason (which is why we have the override fix).

All the browsers except Safari supported gap in flex for years. Safari added support a few versions ago, but PatternFly never updated it. Their margin hack causes a lot of problems with the layout, hence our local fix. But you can't properly use the margin hack when things wrap (which is our problem) or for vertical spacing (for the same reason).

So, for right now, I guess we can add `row-gap: var(--pf-global--spacer--sm);` (or xs) locally?

marusak:
Ok, so now here is space between card title and details. But I think that is actually correct so I'll update the screenshots.
https://logs.cockpit-project.org/logs/pull-943-20220323-100009-b6a31df1-fedora-35/pixeldiff.html#TestApplication-testBasicSystem-overview-mobile

martinpitt:
oh-uh @ [c9s packit test](https://artifacts.dev.testing-farm.io/4cf350da-66d2-4001-811f-7d0ae786916a/): `(src/bridge/cockpithttpstream.c:942):cockpit_http_stream_close: code should not be reached`. But I trust that's unrelated to this PR.

martinpitt (approved):
Dakujem!

PR #944, opened 2022-03-17
Run Fedora podman tests offline

jelly:


Makefile
```diff
@@ -131,9 +131,15 @@ rpm: $(TARFILE) $(SPEC)
 	rm -r "`pwd`/rpmbuild"
 	rm -r "`pwd`/output" "`pwd`/build"
 
+NETWORK=
+# fedora can be built offline; the rest does not pull podman containers in image-create yet
+ifeq ($(filter fedora,$(TEST_OS)),)
```
martinpitt, line 136: Don't you want `ifneq` here? I.e. add `--no-network` for Fedora images, not "everything but Fedora images"?

test/vm.install
```diff
@@ -27,10 +27,17 @@ if type firewall-cmd >/dev/null 2>&1; then
 fi
 
 # grab a few images to play with; tests run offline, so they cannot download images
-podman rmi --all
-podman pull quay.io/libpod/busybox
-podman pull quay.io/libpod/alpine
-podman pull quay.io/cockpit/registry:2
+if grep -q ID.*fedora /etc/os-release; then
+    version=$(. /etc/os-release && echo "$VERSION_ID")
```
martinpitt, line 31: Please use /usr/lib/os-release, and you can source this unconditionally. Just check `[ "$ID" = "fedora" ]`

martinpitt (changes requested):

Makefile
```diff
@@ -131,9 +131,15 @@ rpm: $(TARFILE) $(SPEC)
 	rm -r "`pwd`/rpmbuild"
 	rm -r "`pwd`/output" "`pwd`/build"
 
+NETWORK=--no-network
+# fedora can be built offline; the rest does not pull podman containers in image-create yet
+ifneq ($(filter fedora,$(TEST_OS)),)
```
martinpitt, line 136: oops, `filter` is wrong here; TEST_OS is not a list of strings, but a single string, and you want substring match. You probably want `find`; see https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

Makefile
jelly: I've found https://www.gnu.org/software/make/manual/make.html#Testing-Flags and man this is confusing.

martinpitt (approved):
Yay, thanks! Looks good, and our CI is green. Let's wait for packit for good measure, but the changes here should not have any influence there.

PR #947, opened 2022-03-24
test: Adjust for ubuntu-2204

martinpitt:
This still has a too old criu (inherited from Debian testing).

 - Requires new image: https://github.com/cockpit-project/bots/pull/3131


martinpitt:
[two failures](https://logs.cockpit-project.org/logs/pull-947-20220324-065944-06607665-ubuntu-2204-bots-3131/log.html#9-1) which I didn't see in my initial local run, will investigate. Ignore the `#3132` context, that was just the wrong PR. But the image works in general, so I'll land it to make it easier to test.

marusak (approved):
Thank you!

PR #948, opened 2022-03-24
Makefile: run tests fully offline

jelly:
All our test images now pull the requires podman test images during
image creation.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -20,5 +20,7 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-      - fedora-all
+      - fedora-35
+      - fedora-36
+      - fedora-development
       - centos-stream-9-x86_64
--- a/Makefile
+++ b/Makefile
@@ -131,15 +131,9 @@ rpm: $(TARFILE) $(SPEC)
 	rm -r "`pwd`/rpmbuild"
 	rm -r "`pwd`/output" "`pwd`/build"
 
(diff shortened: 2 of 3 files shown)


martinpitt:
This requires a corresponding change in test/vm.install to not untag and re-pull the images. Essentially just dropping the `else` branch.

jelly:
> This requires a corresponding change in test/vm.install to not untag and re-pull the images. Essentially just dropping the `else` branch.

Yup, I forgot we had special fedora handling, this is now done. But requires image refreshes for rhel-9-0 and rhel-8-6

martinpitt:
Sad to see some [f34 regression](https://artifacts.dev.testing-farm.io/f1d0d253-4969-4722-9547-1f07983e369b/) -- but we don't actually support/release to F34 any more, so we could just disable that.

Otherwise, nice to see that this works everywhere now! :+1:

martinpitt (approved):

packit.yaml
```diff
@@ -20,5 +20,6 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-      - fedora-all
+      - fedora-35
+      - fedora-36
```
martinpitt, line 24: Please also add `- fedora-development` to continue to cover Rawhide.

martinpitt (changes requested):

martinpitt (approved):
Thanks! (Please wait on green, obviously)

PR #950, opened 2022-03-30
test: Collect code coverage

mvollmer:
- [x] https://github.com/cockpit-project/cockpit/pull/17078

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,10 @@ packaging/debian/changelog: packaging/debian/changelog.in
 
 $(WEBPACK_TEST): $(COCKPIT_REPO_STAMP) $(shell find src/ -type f) package.json webpack.config.js
 	$(MAKE) package-lock.json && NODE_ENV=$(NODE_ENV) node_modules/.bin/webpack
+	# In development mode terser does not run and so no LICENSE.txt.gz is generated, so we explictly create it as it is required for building rpm's
+	if [ "$$NODE_ENV" = "development" ]; then \
+		gzip </dev/null >dist/index.js.LICENSE.txt.gz; \
+	fi
 
 watch:
 	NODE_ENV=$(NODE_ENV) node_modules/.bin/webpack --watch
@@ -129,10 +133,10 @@ TEST_NPMS = \
 dist: $(TARFILE)
 	@ls -1 $(TARFILE)
 
-# when building a distribution tarball, call webpack with a 'production' environment
+# when building a distribution tarball, call webpack with a 'production' environment by default
 # we don't ship most node_modules for license and compactness reasons, only the ones necessary for running tests
 # we ship a pre-built dist/ (so it's not necessary) and ship package-lock.json (so that node_modules/ can be reconstructed if necessary)
-$(TARFILE): export NODE_ENV=production
+$(TARFILE): export NODE_ENV ?= production
 $(TARFILE): $(WEBPACK_TEST) $(SPEC) packaging/arch/PKGBUILD packaging/debian/changelog
 	if type appstream-util >/dev/null 2>&1; then appstream-util validate-relax --nonet *.metainfo.xml; fi
 	tar --xz $(TAR_ARGS) -cf $(TARFILE) --transform 's,^,$(RPM_NAME)/,' \
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,11 @@ COCKPIT_REPO_STAMP=pkg/lib/cockpit-po-plugin.js
 # common arguments for tar, mostly to make the generated tarballs reproducible
 TAR_ARGS = --sort=name --mtime "@$(shell git show --no-patch --format='%at')" --mode=go=rX,u+rw,a-s --numeric-owner --owner=0 --group=0
 
+ifeq ($(TEST_COVERAGE),yes)
+RUN_TESTS_OPTIONS+=--coverage
+NODE_ENV=development
+endif
+
 all: $(WEBPACK_TEST)
 
 # checkout common files from Cockpit repository required to build this project;
--- a/test/run
+++ b/test/run
@@ -6,6 +6,12 @@ set -eu
 
 export RUN_TESTS_OPTIONS=--track-naughties
 
+TEST_SCENARIO=${TEST_SCENARIO:-}
+
+if [ "$TEST_SCENARIO" == "devel" ]; then
+    export TEST_COVERAGE=yes
+fi
+
 make codecheck
 make check
 make po/podman.pot


marusak:
I like the report! Just looking into it I can see two things:
1. We define `compare_versions` and `getCommitArr` functions but not use them. We can drop this dead code
2. We don't test at all pod actions

Makefile
```diff
@@ -60,6 +60,7 @@ packaging/debian/changelog: packaging/debian/changelog.in
 
 $(WEBPACK_TEST): $(NODE_MODULES_TEST) $(LIB_TEST) $(shell find src/ -type f) package.json webpack.config.js
 	NODE_ENV=$(NODE_ENV) node_modules/.bin/webpack
+	[ -f dist/index.js.LICENSE.txt.gz ] || gzip </dev/null >dist/index.js.LICENSE.txt.gz
```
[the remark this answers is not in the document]
martinpitt: I liked your idea about telling terser to only do the LICENSE bit, but not minification. Is that too difficult? This is oh-kay, but as it's so brittle and error prone I'd then really like to see some integration test that the file shipped by the RPM/deb is actually valid.

jelly:
This is really cool!

Should genhtml set `--title` so my browser tab shows what project it generated coverage for, it would require some argument to be passed to --coverage or is there 

Should we exclude src/lib? It's something which comes from cockpit so we aren't really interested in it's coverage in cockpit-podman. Does --coverage support a way to exclude data?

mvollmer:
> Should we exclude src/lib? It's something which comes from cockpit so we aren't really interested in it's coverage in cockpit-podman.

Without thinking too much about it, I would say let's keep it.  it's easy enough to ignore, and might be interesting in certain cases.

mvollmer:
> Should genhtml set `--title`

Yeah, let's try to make this work.

jelly:
Shall we pick this up again? As Cockpit already has the integration fo ra few weeks it might be nice to add it here now as well.

mvollmer:
> Shall we pick this up again? As Cockpit already has the integration for a few weeks it might be nice to add it here now as well.

Sure.  I think the blocker is still how to get a development build of cockpit-podman, no?

marusak:
> how to get a development build of cockpit-podman

`NODE_ENV=development make` should works, right?

mvollmer:
> > how to get a development build of cockpit-podman
> 
> `NODE_ENV=development make` should works, right?

I guess, but `NODE_ENV=development make check` does not.

Makefile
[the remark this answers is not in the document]
jelly: Technically this should only run when NODE_ENV == development, so we can make the Makefile conditional, it's a bit of a hack. But then we don't need to test if the license is empty as it will not run.

jelly:
@martinpitt I think the PR is ready for review now? Probably my commit should be squashed into the touching of the LICENSE file? Do you find the solution I propose acceptable?

src/manifest.json
```diff
@@ -1,4 +1,6 @@
 {
+    "name": "podman",
```
martinpitt, line 2: This should not be necessary. If it is for some reason, please split out to a new commit with a rationale.

Makefile
```diff
@@ -93,7 +93,9 @@ packaging/debian/changelog: packaging/debian/changelog.in
 
 $(WEBPACK_TEST): $(COCKPIT_REPO_STAMP) $(shell find src/ -type f) package.json webpack.config.js
 	$(MAKE) package-lock.json && NODE_ENV=$(NODE_ENV) node_modules/.bin/webpack
-	[ -f dist/index.js.LICENSE.txt.gz ] || gzip </dev/null >dist/index.js.LICENSE.txt.gz
+	if [ "$$NODE_ENV" = "development" ]; then \
```
martinpitt, line 96: I didn't like the first commit, as it is brittle. This explicit condition is better, please squash.

Makefile
```diff
@@ -93,7 +93,9 @@ packaging/debian/changelog: packaging/debian/changelog.in
...
+		[ -f dist/index.js.LICENSE.txt.gz ] || gzip </dev/null >dist/index.js.LICENSE.txt.gz; \
```
martinpitt, line 97: As you said, we don't need the `-f` test here, please drop.

martinpitt (changes requested):
Thanks! Nice to see coverage coming to life!

src/manifest.json
mvollmer: `write_lcov` in test/common/lvoc.py needs this.  I forgot the details, something about constructing filenames in dist/, I think.

Makefile
jelly: Yup, this was a test commit. Dropped `-f`.

jelly:
@mvollmer added a small comment about the LICENSE file.

Makefile
```diff
@@ -88,6 +88,10 @@ packaging/debian/changelog: packaging/debian/changelog.in
 
 $(WEBPACK_TEST): $(COCKPIT_REPO_STAMP) $(shell find src/ -type f) package.json webpack.config.js
 	$(MAKE) package-lock.json && NODE_ENV=$(NODE_ENV) node_modules/.bin/webpack
+	# In development mode no LICENSE.txt.gz is generated by terser, so we explictly create it as it is required for building rpm's
```
martinpitt, line 96: More precisely, terser doesn't run at all for devel mode, as it's part of the `optimization:` webpack config.

src/manifest.json
martinpitt: Eww, what? does that mean that we only get code coverage for the "systemd" webpack in cockpit? That's the only one that has an explicit "name:", due to some historical error/opinion change etc. (I'd really love to just rename "systemd" to "system"..)

martinpitt (commented):
Thanks! Looks fine to me by and large, and really happy to see this! Just two cleanups.

src/manifest.json
jelly: This is rather curious indeed, however coverage on main seems to work.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-17835-20221024-133642-ee252b8c-fedora-36-devel/Coverage/index.html

Maybe we should instead fallback to `package.json`'name?

martinpitt: Does coverage work without this hack? This really bothers me. If this was really required, coverage would be broken for almost all cockpit webpacks as well, and I'd rather fix that in cockpit tests/common then.

martinpitt: ... in particular as I'd really like to drop/discourage this `name:` thing in manifests.

jelly: Ah, well this really broke coverage, see here https://cockpit-logs.us-east-1.linodeobjects.com/pull-950-20221025-142943-da2a82ca-fedora-36-devel/log.html

jelly:
Blocked on a coverage not working without a `name` field in `src/manifest.json`.

src/manifest.json
mvollmer: The coverage machinery needs to find files in `dist/` when giving their URL.  For example, the machinery needs to make this association:

```
http://127.0.0.2:9091/cockpit/[CHECKSUM]/podman/index.js
  ->
/home/mvo/work/cockpit-podman/dist/index.js
/home/mvo/work/cockpit-podman/dist/index.js.map
```

The URL uses `podman/index.js` while webpack produces just `dist/index.js`.  The coverage machinery needs to know that `podman/index.js` can be found in `dist`, but that for example `base1/cockpit.js` should be ignored.

We can change the build system to produce `dist/podman/index.js`.  This would be quite elegant, but a bit of work.  The starterkit would also need to change.

The "podman" string does in fact come from package.json. (It turns into PACKAGE_NAME in the Makefile, and "make install" puts everything into `/usr/share/cockpit/$(PACKAGE_NAME)`. So we can change lcov.py to use package.json instead if dist/manifest.json doesn't contain "name". (We need to keep looking at dist/*/manifest.json for cockpit-project/cockpit itself.) I think this is a bit magic, but so is the whole distmap thing.

Wdyt? Using package.json seems best for starterkit-based projects.

jelly: Using package.json sounds good to me.

mvollmer: Done, see https://github.com/cockpit-project/cockpit/pull/17849

mvollmer:
> Eww, what? does that mean that we only get code coverage for the "systemd" webpack in cockpit?

No, cockpit uses subdirectories in dist/, and without a "name" in the manifest we can assume that the directory in the URL is the same as the directory in dist/.  (We have a "name" in many manifests, not just in systemd, no?)

Starterkit-derived projects don't use subdirs in dist/, but they still have a directory in their URL, which either comes from the manifest, or is the same as their subdirectoy in /usr/share/cockpit.

jelly:
> The coverage part of this works. I don't know what's up with the pixel test failures...

Weird, should try that in a separate PR as well

martinpitt:
Thanks @mvollmer ! Pixel test failure is most likely due to the newer PF in cockpit; it *could* also be due to a newer browser in our tasks container, it was updated today.

jelly:
Same pixel changes in https://cockpit-logs.us-east-1.linodeobjects.com/pull-650-20221026-131526-25e8c1e1-fedora-36/pixeldiff.html#TestMachinesConsoles-testSerialConsole-vm-details-console-serial so I assume this is because of PF updates, @garrett do you have an idea if we should just accept this new shadow border?

jelly (approved):

jelly:
> Thanks @mvollmer ! Pixel test failure is most likely due to the newer PF in cockpit; it _could_ also be due to a newer browser in our tasks container, it was updated today.

Pixel test failure is now resolved, all green :)

jelly:
As follow up, make a PR to bots to update `lib/testmap.py` to add `devel` to our standard test set.

martinpitt (approved):

PR #952, opened 2022-04-04
cockpituous: Build in cockpit-preview COPR

martinpitt:
All our other projects release to that. For a user it does not make too
much sense to have each project in its own COPR. All Cockpit packages
have similar QA standards, after all.


marusak (approved):
Thanks!

PR #953, opened 2022-04-05
workflows: Drop livetest

martinpitt:
We have not used that in a long time, and it stopped working a while
ago.


marusak (approved):
Thanks!

PR #957, opened 2022-04-12
Volumes support

strzinek:
Basic support for Podman volumes management:
* Container details - Mounts
* Container details - ENV
* Volumes list
* Volume details
* Create and remove volume


garrett (changes requested):
Nice! Thanks for the PR.

It's a great start and seems mostly implemented feature-wise, but it's currently incomplete and still needs a bit of work.

---

As-is, this doesn't fit the design of Cockpit Podman, as you flipped the order of images and containers. As containers are derived from images, the images need to be on top. Additionally, the containers list can get very long, so everything below it might not be seen. This is why the design has images at the top and is expandable (and collapsed by default).

(Design doc @ https://github.com/cockpit-project/cockpit/discussions/16059)

Here's what it looks like in the PR right now:

Screenshot 2022-04-12 at 10-36-32 Podman containers - garrett@Bolt

We might want to use side-by-side cards at the top, similar to Cockpit-Machines. In widescreen, they're side-by-side. When space is constrained, they wrap and stack. (This can be done with [PatternFly Flex](https://www.patternfly.org/v4/utilities/flex/) with breakpoints.):

Screenshot 2022-04-12 at 10-51-00 Virtual Machines - garrett@Bolt

Screenshot 2022-04-12 at 10-51-16 Virtual Machines - garrett@Bolt


---

While it works when maximized on a high-res screen and also collapses down on mobile, the list doesn't work in standard desktop sizes:

Screenshot 2022-04-12 at 10-36-46 Podman containers - garrett@Bolt

Here it is in huge, just for reference of others viewing the PR:

Screenshot 2022-04-12 at 10-37-36 Podman containers - garrett@Bolt

---

Create volume says "Create container" and has tabs. (It's not obvious what this does. I guess it creates volumes in a standard location, mainly for use in containers later?)

It shouldn't have tabs.

Screenshot 2022-04-12 at 10-37-05 Podman containers - garrett@Bolt

I'm assuming you'll add various options as listed @ https://docs.podman.io/en/latest/markdown/podman-volume-create.1.html too? (This could be handled in follow-up PRs.)

---


Just as a point of reference, prune and delete are the only options at the moment:

Screenshot 2022-04-12 at 10-38-37 Podman containers - garrett@Bolt

Screenshot 2022-04-12 at 10-38-45 Podman containers - garrett@Bolt

strzinek:
Thanks @garrett for your review, I'll incorporate your valid comments.

jelly:
As volumes allow for purning unused volumes, shouldn't we also show which volume is unused and used by a container similar to images?

strzinek:
I agree, @jelly, but there is no volume usage info present in API, unlike for images. API does not even return volume ID's, only names. List of unused volumes could be reconstructed by going through all the containers on client side, I think. Or is there any easier way?

garrett:
Volumes details are empty. Should there be anything there? If not, they shouldn't be expandable. And if there's only one tab of information, there shouldn't be a tab.

Screenshot 2022-05-04 at 11-53-16 Podman containers - garrett@Bolt

---

Also, PF Cards now have the ability to be expandable. Perhaps we should use this for both Images and Volumes. But we can address that in a follow-up. It's just something to keep in mind as it would impact the way volumes look on the page.

https://www.patternfly.org/v4/components/card#expandable
https://www.patternfly.org/v4/components/card/react-demos/#horizontal-card-grid

marusak:
@strzinek Hey! Seems this fall of our radar and now has conflicts. Is this still something you are interested in and able to adjust to currect podman?

strzinek:
> @strzinek Hey! Seems this fall of our radar and now has conflicts. Is this still something you are interested in and able to adjust to currect podman?

I resolved merge conflicts and removed changes to container details, as these have since been implemented by someone else.

TODO: write basic tests for volumes

jelly:
Work continued in https://github.com/cockpit-project/cockpit-podman/pull/1959

PR #960, opened 2022-04-13
workflows: Split npm-update to PF and non PF tasks

marusak:
Run PF update every Monday and anything else try to update on Tue, Thu and Sat.


martinpitt (approved):
Dakujem!

PR #966, opened 2022-04-19
container rename support

strzinek:
Allow to rename a container, if it is not running or paused.

There is an issue with container not being refreshed after renamed, as no rename event is fired. This should be fixed in podman v4.1 according to https://github.com/containers/podman/issues/13919

### DIFF
--- a/src/ContainerRenameModal.jsx
+++ b/src/ContainerRenameModal.jsx
@@ -1,0 +1,103 @@
+import React, { useState } from 'react';
+import {
+    Button,
+    Form, FormGroup,
+    Modal, TextInput
+} from '@patternfly/react-core';
+import cockpit from 'cockpit';
+
+import * as client from './client.js';
+import { ErrorNotification } from './Notification.jsx';
+
+const _ = cockpit.gettext;
+
+const ContainerRenameModal = ({ container, version, onHide, updateContainerAfterEvent }) => {
+    const [name, setName] = useState(container.Names[0]);
+    const [nameError, setNameError] = useState(null);
+    const [dialogError, setDialogError] = useState(null);
+    const [dialogErrorDetail, setDialogErrorDetail] = useState(null);
+
+    const handleInputChange = (targetName, value) => {
+        if (targetName === "name") {
+            setName(value);
+            if (value === "") {
+                setNameError(_("Container name is required."));
+            } else if (/^[a-zA-Z0-9][a-zA-Z0-9_\\.-]*$/.test(value)) {
+                setNameError(null);
+            } else {
+                setNameError(_("Invalid characters. Name can only contain letters, numbers, and certain punctuation (_ . -)."));
+            }
+        }
+    };
+
+    const handleRename = () => {
+        if (!name) {
+            setNameError(_("Container name is required."));
+            return;
+        }
+
+        setNameError(null);
+        setDialogError(null);
+        client.renameContainer(container.isSystem, container.Id, { name })
+                .then(() => {
+                    onHide();
+                    // HACK: This is a workaround for missing API rename event in Podman versions less than 4.1.
+                    if (version.localeCompare("4.1", undefined, { numeric: true, sensitivity: 'base' }) < 0) {
+                        updateContainerAfterEvent(container.Id, container.isSystem);
+                    }
+                })
+                .catch(ex => {
+                    setDialogError(cockpit.format(_("Failed to rename container $0"), container.Names[0]));
+                    setDialogErrorDetail(cockpit.format("$0: $1", ex.message, ex.reason));
+                });
+    };
+
+    const handleKeyPress = (event) => {
+        if (event.key === "Enter") {
+            event.preventDefault();
+            handleRename(name);
+        }
+    };
+
+    const renameContent =
+        <Form isHorizontal>
+            <FormGroup fieldId="rename-dialog-container-name" label={_("New container name")}
+                    validated={nameError ? "error" : "default"}
+                    helperTextInvalid={nameError}>
+                <TextInput id="rename-dialog-container-name"
+                        value={name}
+                        validated={nameError ? "error" : "default"}
+                        type="text"
+                        aria-label={nameError}
+                        onChange={value => handleInputChange("name", value)} />
+            </FormGroup>
+        </Form>;
+
+    return (
+        <Modal isOpen
+            position="top" variant="medium"
+            onClose={onHide}
+            onKeyPress={handleKeyPress}
+            title={cockpit.format(_("Rename container $0"), container.Names[0])}
+            footer={<>
+                <Button variant="primary"
+                        className="btn-ctr-rename"
+                        id="btn-rename-dialog-container"
+                        isDisabled={nameError}
+                        onClick={handleRename}>
+                    {_("Rename")}
+                </Button>
+                <Button variant="link"
+                        className="btn-ctr-cancel-commit"
+                        onClick={onHide}>
+                    {_("Cancel")}
+                </Button>
+            </>}
+        >
+            {dialogError && <ErrorNotification errorMessage={dialogError} errorDetail={dialogErrorDetail} onDismiss={() => setDialogError(null)} />}
+            {renameContent}
+        </Modal>
+    );
+};
+
+export default ContainerRenameModal;
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -25,6 +25,7 @@ import ForceRemoveModal from './ForceRemoveModal.jsx';
 import * as utils from './util.js';
 import * as client from './client.js';
 import ContainerCommitModal from './ContainerCommitModal.jsx';
+import ContainerRenameModal from './ContainerRenameModal.jsx';
 
 import './Containers.scss';
 import { ImageRunModal } from './ImageRunModal.jsx';
@@ -32,7 +33,7 @@ import { PodActions } from './PodActions.jsx';
 
 const _ = cockpit.gettext;
 
-const ContainerActions = ({ container, onAddNotification, version, localImages }) => {
+const ContainerActions = ({ container, onAddNotification, version, localImages, updateContainerAfterEvent }) => {
     const [removeErrorModal, setRemoveErrorModal] = useState(false);
     const [deleteModal, setDeleteModal] = useState(false);
     const [checkpointInProgress, setCheckpointInProgress] = useState(false);
@@ -40,6 +41,7 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
     const [restoreInProgress, setRestoreInProgress] = useState(false);
     const [restoreModal, setRestoreModal] = useState(false);
     const [commitModal, setCommitModal] = useState(false);
+    const [renameModal, setRenameModal] = useState(false);
     const [isActionsKebabOpen, setActionsKebabOpen] = useState(false);
     const isRunning = container.State == "running";
     const isPaused = container.State === "paused";
@@ -112,6 +114,11 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
                 });
     };
 
+    const renameContainer = () => {
+        setRenameModal(container.State !== "running" || version.localeCompare("3.0.1", undefined, { numeric: true, sensitivity: 'base' }) >= 0);
+        setActionsKebabOpen(false);
+    };
+
     const handleRemoveContainer = () => {
         const id = container ? container.Id : "";
 
@@ -170,6 +177,15 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
                 });
     };
 
+    const addRenameAction = () => {
+        actions.push(
+            <DropdownItem key="rename"
+                        onClick={() => renameContainer()}>
+                {_("Rename")}
+            </DropdownItem>
+        );
+    };
+
     const actions = [];
     if (isRunning || isPaused) {
         actions.push(
@@ -225,6 +241,9 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
                 {_("Start")}
             </DropdownItem>
         );
+        if (version.localeCompare("3", undefined, { numeric: true, sensitivity: 'base' }) >= 0) {
+            addRenameAction();
+        }
         if (container.isSystem && container.hasCheckpoint) {
             actions.push(
                 <DropdownSeparator key="separator-0" />,
@@ -233,6 +252,10 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
                     {_("Restore")}
                 </DropdownItem>
             );
+        }
+    } else { // running or paused
+        if (version.localeCompare("3.0.1", undefined, { numeric: true, sensitivity: 'base' }) >= 0) {
+            addRenameAction();
         }
     }
 
@@ -296,10 +319,19 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
             localImages={localImages}
         />;
 
+    const containerRenameModal =
+        <ContainerRenameModal
+            container={container}
+            version={version}
+            onHide={() => setRenameModal(false)}
+            updateContainerAfterEvent={updateContainerAfterEvent}
+        />;
+
     return (
         <>
             {kebab}
             {deleteModal && containerDeleteModal}
+            {renameModal && containerRenameModal}
             {checkpointModal && containerCheckpointModal}
             {restoreModal && containerRestoreModal}
             {removeErrorModal && containerRemoveErrorModal}
@@ -385,7 +417,7 @@ class Containers extends React.Component {
         ];
 
         if (!container.isDownloading) {
-            columns.push({ title: <ContainerActions version={this.props.version} container={container} onAddNotification={this.props.onAddNotification} localImages={localImages} />, props: { className: "pf-c-table__action" } });
+            columns.push({ title: <ContainerActions version={this.props.version} container={container} onAddNotification={this.props.onAddNotification} localImages={localImages} updateContainerAfterEvent={this.props.updateContainerAfterEvent} />, props: { className: "pf-c-table__action" } });
         }
 
         const tty = containerDetail ? !!containerDetail.Config.Tty : undefined;
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -367,6 +367,7 @@ class Application extends React.Component {
         case 'sync':
         case 'unmount':
         case 'unpause':
+        case 'rename': // rename event is available starting podman v4.1; until then the container does not get refreshed after renaming
             this.updateContainerAfterEvent(event.Actor.ID, system);
             break;
         case 'remove':
@@ -673,6 +674,7 @@ class Application extends React.Component {
                 registries={this.state.registries}
                 selinuxAvailable={this.state.selinuxAvailable}
                 podmanRestartAvailable={this.state.podmanRestartAvailable}
+                updateContainerAfterEvent={this.updateContainerAfterEvent}
             />;
 
         const notificationList = (
--- a/src/client.js
+++ b/src/client.js
@@ -81,6 +81,8 @@ export function inspectContainer(system, id) {
 
 export const delContainer = (system, id, force) => podmanCall("libpod/containers/" + id, "DELETE", { force }, system);
 
+export const renameContainer = (system, id, config) => podmanCall("libpod/containers/" + id + "/rename", "POST", config, system);
+
 export function createContainer(system, config) {
     return new Promise((resolve, reject) => {
         podmanCall("libpod/containers/create", "POST", {}, system, JSON.stringify(config))
--- a/test/check-application
+++ b/test/check-application
@@ -1949,6 +1949,36 @@ class TestApplication(testlib.MachineCase):
         self.performContainerAction(container_name, "Resume")
         b.wait(lambda: self.getContainerAttr(container_name, "State") == "Running")
 
+    def testRenameContainerSystem(self):
+        self._testRenameContainer(True)
+
+    def testRenameContainerUser(self):
+        self._testRenameContainer(False)
+
+    def _testRenameContainer(self, auth):
+        b = self.browser
+        container_name = "rename"
+        container_name_new = "rename-new"
+
+        self.execute(auth, f"podman container create -t --name {container_name} quay.io/libpod/busybox:latest")
+        self.login(auth)
+
+        self.filter_containers('all')
+
+        self.waitContainerRow(container_name)
+        self.toggleExpandedContainer(container_name)
+        self.performContainerAction(container_name, "Rename")
+
+        # the container name should be in the "Rename container" header
+        b.wait_in_text("#pf-modal-part-1", container_name)
+
+        b.set_input_text("#rename-dialog-container-name", container_name_new)
+        b.click('#btn-rename-dialog-container')
+        b.wait_not_present("#rename-dialog-container-name")
+
+        self.execute(auth, f"podman inspect --format '{{{{.Id}}}}' {container_name_new}").strip()
+        self.waitContainerRow(container_name_new)
+
 
 if __name__ == '__main__':
     testlib.test_main()


martinpitt:
Thanks @strzinek for all your work! I just wanted to explain the delays -- most of us had some holidays in the last week. We'll go through your PRs one by one, but they will reach require some work: All features, fixes, and other changes need to be covered with corresponding tests (functional, and where appropriate, the "pixel tests", i.e. screenshots). E.g. for this PR, testing the renaming should be added to `_testLifecycleOperations()`, or a new `testRename()`. There will likely be some fallout on some operating systems, which we'll need to handle.

Do you want to give it a try to write a test? Please start with trying to [run the existing ones](https://github.com/cockpit-project/cockpit-podman/blob/main/HACKING.md#running-tests-locally), to make sure you have everything set up. Then you can add a new one for renaming.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+import cockpit from 'cockpit';
+
+import * as client from './client.js';
+
+import { ErrorNotification } from './Notification.jsx';
+import * as dockerNames from 'docker-names';
+
+const _ = cockpit.gettext;
+
+class ContainerRenameModal extends React.Component {
```
martinpitt, line 16: In new code we try to use functional components and `useState()`, see [React hooks docs](https://reactjs.org/docs/hooks-state.html). If you don't feel comfortable with them, we are happy to help with this.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+    constructor(props) {
+        super(props);
+        this.state = {
+            name: dockerNames.getRandomName(),
+            nameError: "",
```
martinpitt, line 21: "empty" values should be `null`.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+import { ErrorNotification } from './Notification.jsx';
+import * as dockerNames from 'docker-names';
+
+const _ = cockpit.gettext;
+
+class ContainerRenameModal extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            name: dockerNames.getRandomName(),
```
martinpitt, line 20: The dialog input should start with the *current* container name. Often you merely want to tweak it a bit (perhaps add a suffix). I don't think we should use random names here at all -- the whole point of renaming is to give it some descriptive user-selected name.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+            nameError: "",
+        };
+
+        this.handleInputChange = this.handleInputChange.bind(this);
+        this.handleRename = this.handleRename.bind(this);
+    }
+
+    componentDidMount() {
+        this._isMounted = true;
```
martinpitt, line 29: `_isMounted` is not used anywhere. Please drop this function and the assignment in WillUnmount().

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+        if (this.activeConnection)
+            this.activeConnection.close();
+    }
+
+    handleInputChange(targetName, value) {
+        const newState = { [targetName]: value };
+
+        if (targetName === "name")
+            newState.nameError = "";
```
martinpitt, line 43: `null`

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    handleRename() {
+        if (!this.state.name) {
+            this.setState({ nameError: "Container name is required" });
+            return;
+        }
+
+        const renameData = {};
+        renameData.name = this.state.name;
```
martinpitt, line 55: Please merge into the initialization.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+        this.setState({ nameError: "", dialogError: "" });
+        client.renameContainer(this.props.container.isSystem, this.props.container.Id, renameData)
+                .then(() => this.props.onHide())
+                .catch(ex => {
+                    this.setState({
+                        dialogError: cockpit.format(_("Failed to rename container $0"), this.props.container.Names),
```
martinpitt, line 62: "Names" sounds like this is a list, not a string. Pick the first element to be safe.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+            <Form isHorizontal>
+                <FormGroup fieldId="rename-dialog-image-name" label={_("New container name")}
+                           validated={this.state.nameError ? "error" : "default"}
+                           helperTextInvalid={this.state.nameError}>
+                    <TextInput id="rename-dialog-image-name"
+                               value={this.state.name}
+                               validated={this.state.nameError ? "error" : "default"}
+                               onChange={value => this.handleInputChange("name", value)} />
+                </FormGroup>
+
```
martinpitt, line 79: stray empty line

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+            </Form>;
+
+        return (
+            <Modal isOpen
+                   showClose={false}
+                   position="top" variant="medium"
+                   title={cockpit.format(_("Rename container $0 "), this.props.container.Names)}
```
martinpitt, line 86: Please drop the trailing space in the string

src/Containers.jsx
```diff
@@ -112,6 +114,15 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
                 });
     };
 
+    const renameContainer = () => {
+        if (container.State == "running") {
+            setRenameModal(false);
+        } else {
+            setRenameModal(true);
+        }
```
martinpitt, line 122: `setRenameModal(container.State !== "running")`

src/client.js
```diff
@@ -98,6 +98,14 @@ export function delContainer(system, id, force) {
     });
 }
 
+export function renameContainer(system, id, config) {
+    return new Promise((resolve, reject) => {
+        podmanCall("libpod/containers/" + id + "/rename", "POST", config, system)
+                .then(resolve)
+                .catch(reject);
+    });
+}
```
martinpitt, line 107: I know this is just copying the previous pattern, but that Promise() wrapper is completely useless. Please use something like this:

```js
export const renameContainer = (system, id, config) => podmanCall("libpod/containers/" + id + "/rename", "POST", config, system);
```

or if you prefer, just change the function body to
```js
return podmanCall("libpod/containers/" + id + "/rename", "POST", config, system);
```

martinpitt (changes requested):
Thanks!

src/client.js
strzinek: Thanks @martinpitt for the review, I will fix the code and prepare the tests when I have some free time to code. In my case it's usually during the holidays. ;-)

src/ContainerRenameModal.jsx
jelly: I agree, starting editing with the old container name makes more sense to me.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+        if (targetName === "name")
+            newState.nameError = "";
+
+        this.setState(newState);
+    }
+
+    handleRename() {
+        if (!this.state.name) {
+            this.setState({ nameError: "Container name is required" });
```
jelly, line 50: Missing translations `_("")`

src/app.jsx
```diff
@@ -367,6 +367,7 @@ class Application extends React.Component {
         case 'sync':
         case 'unmount':
         case 'unpause':
+        case 'rename': // rename event is available starting podman v4.1; until then the container does not get refreshed after renaming
             this.updateContainerAfterEvent(event.Actor.ID, system);
```
jelly, line 371: I've tested this on an older podman and it's a bit annoying that there is no visual feedback from the rename action. I think we should do something about that before merging the PR.

When /rename returns 204 did the rename already happen? If so we can manually refresh the container status I suppose until everything is podman 3.4

strzinek: I agree and I would refresh it manually only if someone helped me how to call Application instance method correctly from handler code and not breaking architecture? I am new to react.

src/ContainerRenameModal.jsx
strzinek: Fixed in commit https://github.com/cockpit-project/cockpit-podman/pull/966/commits/0c3706e1009fabc1fee2a17b247175993469a9c4

src/ContainerRenameModal.jsx
strzinek: Fixed in commit https://github.com/cockpit-project/cockpit-podman/commit/0c3706e1009fabc1fee2a17b247175993469a9c4

src/ContainerRenameModal.jsx
strzinek: Thanks, fixed in commit https://github.com/cockpit-project/cockpit-podman/commit/0c3706e1009fabc1fee2a17b247175993469a9c4

garrett:
- I can't rename running containers. I would have assumed they would be stopped and restarted, but still would be able to be renamed. (Show rename, show modal with a message at the top that mentions that the container will be renamed and restarted, then make the action "Rename and restart".)
- Hitting enter reloads the page. The default form action needs to be caught so it does the right thing, instead of doing an HTTP submit.
- Escape doesn't close the modal.
- The error is in the wrong place. It should, ideally, be directly underneath the field, not an alert. It should also be shorter and to the point.
   - https://www.patternfly.org/v4/components/form/design-guidelines#errors-and-validation
- When a string doesn't match the regular expression, it should say something about invalid characters... or it should just filter the characters like the Machines page does for naming. (Where invalid ones are stripped and spaces are changed to dashes or underscores, whichever is supported, unless spaces are simply supported in the names... then they shouldn't be stripped.) It should _not_ be an alert.

image

src/app.jsx
strzinek: I figured out how to refresh the status of renamed container manually, it should be fixed now.

garrett:
> About renaming running containers: I suppose it could be implemented as you assumed, but I tried to follow the use cases of podman itself.

Yeah. I just found it surprising. I think we'll probably want to implement it, but as a follow-up, not in this PR, as it's more complex.

> It would get even more complicated if container would have been paused - in such case it would have to be resumed first and stopped to rename the container. Or use force-stop in such cases? Let user decide in two steps then? I like to keep things simple.

We'd need to mention in the rename modal that that container will be stopped and restarted after the rename for running and paused containers.

These are a lot of things to think about, and I think it reinforces that this should be handled in a follow-up PR.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,97 @@
...
+            setNameError(_("Container name is required"));
+            return;
+        }
+
+        const renameData = {
+            name: name
+        };
+        setNameError(null);
+        setDialogError(null);
+        client.renameContainer(props.container.isSystem, props.container.Id, renameData)
```
martinpitt, line 38: Please do this inline, with `renameData` → `{ name }`; more succinct, and avoids the indirection with the single-use constant.

`{ x }` is JS shorthand for `{ x: x }`, BTW, as it's a very common use case.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,97 @@
...
+                .then(() => {
+                    props.onHide();
+                    // This is a workaround for missing API rename event in Podman.
+                    // It should be fixed in Podman v4.1, then next line and reference from props can be removed
+                    props.updateContainerAfterEvent(props.container.Id, props.container.isSystem);
```
martinpitt, line 43: Very good! :clap: A clean and documented workaround. Can you please add a "HACK:" to the comment? This is our standard way to mark such workarounds, easily greppable.

test/check-application
```diff
@@ -1913,6 +1913,36 @@ class TestApplication(testlib.MachineCase):
...
+        self.login(auth)
+
+        self.filter_containers('all')
+
+        self.waitContainerRow(container_name)
+        self.toggleExpandedContainer(container_name)
+        self.performContainerAction(container_name, "Rename")
+
+        # the container name should be in the "Rename container" header
+        self.assertIn(container_name, b.text("#pf-modal-part-1"))
```
martinpitt, line 1937: Please use `b.wait_in_text("#pf-modal-part-1", container_name)` for this, both for consistency and also to ensure that there are no race conditions with async rendering.

test/check-application
```diff
@@ -1913,6 +1913,36 @@ class TestApplication(testlib.MachineCase):
...
+
+        b.set_input_text("#rename-dialog-container-name", container_name_new)
+        b.click('#btn-rename-dialog-container')
+        b.wait_not_present("#rename-dialog-container-name")
+
+        container_sha = self.execute(auth, f"podman inspect --format '{{{{.Id}}}}' {container_name_new}").strip()
+        self.waitContainer(container_sha, auth, name=container_name_new, image='busybox:latest', cmd='sh',
+                           state='Configured', owner="system" if auth else "admin")
```
martinpitt, line 1945: Well done on the test! Can you please now start the container, and pin down the behaviour for running containers? Judging by the code, the intention is that there should not be a "Rename" action, so the test should ensure that the kebab does not have that menu entry any more. Then, if at some point running container names are supported/get implemented, we have a reminder to update this.

martinpitt (changes requested):
Nice progress! A few more small things, but well done!

src/ContainerRenameModal.jsx
strzinek: Added.

test/check-application
strzinek: Added.

strzinek: >     * When a string doesn't match the regular expression, it should say something about invalid characters... or it should just filter the characters like the Machines page does for naming. (Where invalid ones are stripped and spaces are changed to dashes or underscores, whichever is supported, unless spaces are simply supported in the names... then they shouldn't be stripped.) It should _not_ be an alert.

I added a regexp validator for name, have a look. Machines names can now have non-ASCII characters, unlike Containers.

strzinek:
@garrett wrote:
>     * When a string doesn't match the regular expression, it should say something about invalid characters... or it should just filter the characters like the Machines page does for naming. (Where invalid ones are stripped and spaces are changed to dashes or underscores, whichever is supported, unless spaces are simply supported in the names... then they shouldn't be stripped.) It should _not_ be an alert.

I added a regexp validator for name, have a look. Machines names can now have non-ASCII characters, unlike Containers.

garrett (changes requested):
It's looking nice.

Can we drop the `[ ]` surrounding the characters? It looks odd — ideally, it'd be some kind of `kbd` element which would make it styled similar to `pre` (but is inline instead). However, I know that having elements in translated strings can be tricky. I'd suggest just removing the squared off braces and show the characters as-is... and perhaps in parenthesis?

It's also missing commas separates after the text and each character.

"Invalid characters. Name must only contain letters, numbers, and certain punctuation (_ . -)."

image

garrett:
I tried reviewing this today, but kept hitting a strange error where containers wouldn't load. I tried rebooting my system, switching between versions of Fedora (you can do that pretty easily with Silverblue), and so on.

Podman couldn't find a container and was just very, very upset. Even when I finally gave up and ran `podman system reset`, it couldn't remove some related file. I had to `rm -rf ~/.local/share/containers/` afterward. I nuked everything and started over. 

Do you think this could possibly be related to renaming and that I hit some sort of bug when testing this earlier?

I just tried renaming a container and it seemed to work. But running it fails with:

```
Failed to start container why_hello_there_
plugin type="bridge" failed (add): failed to find plugin "bridge" in path [/usr/local/libexec/cni /usr/libexec/cni /usr/local/lib/cni /usr/lib/cni /opt/cni/bin]
```

Screenshot 2022-05-03 at 17-45-57 Podman containers - garrett@Bolt


Is Podman's rename buggy? Or could it be code in the PR? Or both? This seems really suspicious — both the corruption earlier and now having a problem running the container.

Otherwise, the UI is pretty good. I think we still probably want to convert spaces to dashes or underscores instead of just throwing an error with a space. Otherwise, it's good UI-wise. However, if Podman is really causing problems when renaming containers, I think that might negate this entire PR, sadly, until Podman fixes its bugs (_if_ it is buggy with renaming).

garrett:
After renaming another container, I got an error about an improper container state:

Screenshot 2022-05-03 at 17-51-12 Podman containers - garrett@Bolt

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,100 @@
...
+const ContainerRenameModal = ({ container, onHide, updateContainerAfterEvent }) => {
+    const [name, setName] = useState(container.Names[0]);
+    const [nameError, setNameError] = useState(null);
+    const [dialogError, setDialogError] = useState(null);
+    const [dialogErrorDetail, setDialogErrorDetail] = useState(null);
+
+    const handleInputChange = (targetName, value) => {
+        if (targetName === "name") {
+            setName(value);
+            if (/^[a-zA-Z0-9][a-zA-Z0-9_\\.-]*$/.test(value)) {
```
jelly, line 23: If it's empty, it shows the wrong message:

image

test/check-application
```diff
@@ -1913,6 +1913,40 @@ class TestApplication(testlib.MachineCase):
...
+        self._testRenameContainer(True)
+
+    def testRenameContainerUser(self):
+        self._testRenameContainer(False)
+
+    def _testRenameContainer(self, auth):
+        b = self.browser
+        container_name = "rename"
+        container_name_new = "rename-new"
+
```
jelly, line 1926: This line has some extraneous whitespaces.

test/check-application
```diff
@@ -1913,6 +1913,40 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_in_text("#pf-modal-part-1", container_name)
+
+        b.set_input_text("#rename-dialog-container-name", container_name_new)
+        b.click('#btn-rename-dialog-container')
+        b.wait_not_present("#rename-dialog-container-name")
+
+        container_sha = self.execute(auth, f"podman inspect --format '{{{{.Id}}}}' {container_name_new}").strip()
+        self.waitContainer(container_sha, auth, name=container_name_new, image='busybox:latest', cmd='sh',
+                           state='Configured', owner="system" if auth else "admin")
+
```
jelly, line 1946: This seems to have some extraneous whitespaces.

jelly (changes requested):
Triggered the tests, I think we still need to handle older podman versions without rename support. So let's see what the bots think about it.

jelly:
```
test/check-application:1926:1: W293 blank line contains whitespace
test/check-application:1946:1: W293 blank line contains whitespace
test/check-application:1951:1: E305 expected 2 blank lines after class or function definition, found 1
```
For the CI to run please fix these issues and I'll re-trigger the bots.

garrett:
> handle older podman versions without rename support

Don't show rename. I think we should only support rename where Podman does.

It looks like some form of rename is in Podman 3.0 (2021-02-11) and above, according to the history of the first commit @ https://github.com/containers/podman/commit/997de2f8e9e5453a99108bde012aa6c41d7323ec

[The release notes for 3.0](https://github.com/containers/podman/releases/tag/v3.0.0) say:

- Added the `podman rename` command, which allows containers to be renamed after they are created.

src/ContainerRenameModal.jsx
```diff
@@ -0,0 +1,100 @@
...
+            title={cockpit.format(_("Rename container $0"), container.Names[0])}
+            footer={<>
+                {dialogError && <ErrorNotification errorMessage={dialogError} errorDetail={dialogErrorDetail} onDismiss={() => setDialogError(null)} />}
+                <Button variant="primary"
+                        className="btn-ctr-rename"
+                        id="btn-rename-dialog-container"
+                        isDisabled={nameError}
+                        onClick={handleRename}>
+                    {_("Rename")}
+                </Button>{' '}
```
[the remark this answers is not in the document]
strzinek: I followed pattern used in ContainerDeleteModal.jsx. I assume I can just remove the spacing.

test/check-application
strzinek: Fixed.

strzinek:
> Podman couldn't find a container and was just very, very upset. Even when I finally gave up and ran `podman system reset`, it couldn't remove some related file. I had to `rm -rf ~/.local/share/containers/` afterward. I nuked everything and started over.
> 
> Do you think this could possibly be related to renaming and that I hit some sort of bug when testing this earlier?
> 

I have never seen this. I had no problems running containers after renaming them. The code I wrote is calling standard Podman API, so if there is a problem, this is definitely bug on Podman side.

jelly:
> > > handle older podman versions without rename support
> > 
> > 
> > Don't show rename. I think we should only support rename where Podman does.
> 
> Rename action showed only for v3+ now.

Ok, running the tests, let's see what they make of it :)

Thanks for the updates!

strzinek:
@jelly :
> Please resolve this, then I can trigger the tests:
> 
> ```
> python3 -m pycodestyle --max-line-length=195 test/check-application # TODO: Fix long lines
> ```

Fixed, thanks.

src/Containers.jsx
```diff
@@ -223,8 +230,16 @@ const ContainerActions = ({ container, onAddNotification, version, localImages }
             <DropdownItem key="start"
                           onClick={() => startContainer()}>
                 {_("Start")}
-            </DropdownItem>
+            </DropdownItem>,
```
martinpitt, line 233: Why the trailing comma here? previous stanzas don't use it either, and it looks like an error. This isn't a list.

test/check-application
```diff
@@ -1913,6 +1913,40 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_in_text("#pf-modal-part-1", container_name)
+
+        b.set_input_text("#rename-dialog-container-name", container_name_new)
+        b.click('#btn-rename-dialog-container')
+        b.wait_not_present("#rename-dialog-container-name")
+
+        self.execute(auth, f"podman inspect --format '{{{{.Id}}}}' {container_name_new}").strip()
+        self.waitContainerRow(container_name_new)
+
+        # test if rename is hidden for running container. This should change with rename events fixed in Podman.
```
martinpitt, line 1946: Getting an event for a rename seems independent from supporting renames for running containers. Is that comment actually true? I thought events were already fixed in podman 4.1.

martinpitt (changes requested):
Some tiny issues, but LGTM overall. I started the tests. Many thanks!

test/check-application
jelly: I did some digging, and indeed the rename event is added in 4.1 https://github.com/containers/podman/releases/tag/v4.1.0

About renaming a running container, this works surprisingly well?

```
[root@fedora-36-127-0-0-2-2201 ~]# podman ps
CONTAINER ID  IMAGE                          COMMAND     CREATED       STATUS            PORTS       NAMES
359ac580010e  quay.io/libpod/busybox:latest  /bin/sh     1 second ago  Up 2 seconds ago              foo
[root@fedora-36-127-0-0-2-2201 ~]# podman rename foo bar
[root@fedora-36-127-0-0-2-2201 ~]# podman ps
CONTAINER ID  IMAGE                          COMMAND     CREATED        STATUS             PORTS       NAMES
359ac580010e  quay.io/libpod/busybox:latest  /bin/sh     9 seconds ago  Up 10 seconds ago              bar
[root@fedora-36-127-0-0-2-2201 ~]# podman -v
podman version 4.0.2
```

jelly: And also on 3.0.1?

```
root@debian:~# podman ps
CONTAINER ID  IMAGE    COMMAND  CREATED       STATUS            PORTS   NAMES
cdc59dfb3e96  busybox  /bin/sh  1 second ago  Up 2 seconds ago          foo
root@debian:~# podman rename foo bar
root@debian:~# podman ps
CONTAINER ID  IMAGE    COMMAND  CREATED        STATUS            PORTS   NAMES
cdc59dfb3e96  busybox  /bin/sh  7 seconds ago  Up 7 seconds ago          bar
root@debian:~# podman -v
podman version 3.0.1
```

garrett (changes requested):
I think this is almost there. Thanks for answering about the bugs I found (which really could be — and probably are — specific to my system).

1. Inline alerts should be at the top of modals, according to PatternFly guidelines.

   image

2. (_Optional; could be follow-up_) If we could focus the entry when the modal comes up, that would be nice. 

3. (_Optional; could be follow-up_) If we can rename running containers, then that'd be great to have in here.

Reference for errors in form validations on PatternFly's website:
https://www.patternfly.org/v4/components/form/design-guidelines/#form-sections:

> When a form field submission results in an error, let users know as soon as possible and as close as possible to the error. Default to presenting error states on a form using field level errors whenever possible.
>
> In some use cases, you may choose to use inline errors at the top of a form to emphasize errors within it. These inline errors should always be a supplemental component used in addition to field level errors.

...and then, a little further down on the page:

> For validation on submission, supplement field level errors with an [inline alert](https://www.patternfly.org/v4/components/alert/design-guidelines#inline-alerts) to make it simple for the user to quickly identify and address areas that need attention. The inline alert should be placed below the form’s header and description, and it should equal the total width of the form.

Since the alert's details are rather long, we have the option of either using a more specific error message (about the container already existing) and using an expandable alert (to hide the extra information, if we do think it's necessary, even with a more precise error message) so that the alert isn't so huge and doesn't push everything down by default. I think it just needs to have an `isExpandable` attribute in the `<Alert>` component.

More info on expandable alerts: https://www.patternfly.org/v4/components/alert#expandable

strzinek:
@garrett wrote:
>     1. Inline alerts should be at the top of modals, according to PatternFly guidelines.
Changed.

>     2. (_Optional; could be follow-up_) If we could focus the entry when the modal comes up, that would be nice.
When I tried to auto focus on input element in modal rename form, this error was reported during make: `error  The autoFocus prop should not be used, as it can reduce usability and accessibility for users  jsx-a11y/no-autofocus`

>     3. (_Optional; could be follow-up_) If we can rename running containers, then that'd be great to have in here.
Done. I added conditions for versions 3.0.1+ for renaming running/paused containers and reverted test.

jelly (approved):
Everything looks good to me, except the one extraneous comma

strzinek:
> Everything looks good to me, except the one extraneous comma

I deleted all of them, even the ones that were there originally.

jelly:
> > Everything looks good to me, except the one extraneous comma
> 
> I deleted all of them, even the ones that were there originally.

Oh that was wrongly suggested by me, the commas should be there for pushing the elements on the list. Sorry about that.

strzinek:
> Oh that was wrongly suggested by me, the commas should be there for pushing the elements on the list. Sorry about that, if you re-push please squash everything into one commit.

Should I squash the whole PR?

jelly:
> > Oh that was wrongly suggested by me, the commas should be there for pushing the elements on the list. Sorry about that, if you re-push please squash everything into one commit.
> 
> Should I squash the whole PR?

Yes, please the whole feature should be one commit

strzinek:
> Yes, please the whole feature should be one commit

But isn't this done while merging the PR to main?

martinpitt:
Build fails:
```
ERROR in /tmp/tmp0mvkw23n/src/Containers.jsx
  196:12  error  Parsing error: Adjacent JSX elements must be wrapped in an enclosing tag. Did you want a JSX fragment <>...</>? (196:12)
```

As long as the result is just a single commit, we can squash&merge, yes. We don't usually do this, but if you prefer to work with such a huge "fixup" commit list, you can keep doing that.

garrett (approved):
Looks good to me from a design standpoint! :+1: Thanks for adding rename support!

jelly:
It seems you have set up some sort of pull request syncing mechanism when something is merged in main, this doesn't work well with our CI as it rebases automatically which it can't.

I've pushed your old commit and rebased it locally, so now the tests run and it's good to be merged when everything is green.

jelly (approved):

jelly:
Thanks for creating the PR and the follow up fixes :tada:

strzinek:
I haven't had time to pay much attention to it today. Thank you for sorting it out.

jelly:
> I haven't had time to pay much attention to it today. Thank you for sorting it out.

No worries, I am happy it's merged. It's a nice feature to have!

PR #975, opened 2022-05-12
warning dialogs: remove "Please"

jelly:
image

### DIFF
--- a/src/ContainerDeleteModal.jsx
+++ b/src/ContainerDeleteModal.jsx
@@ -9,7 +9,7 @@ const ContainerDeleteModal = (props) => {
         <Modal isOpen
                position="top" variant="medium"
                onClose={props.handleCancelContainerDeleteModal}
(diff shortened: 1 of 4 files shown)


marusak (dismissed):
Looks good to me, thanks!
I did `git grep -i please pkg` in cockpit and there is bunch of such messages. Do we want to adjust these as well?

jelly:
> Looks good to me, thanks! I did `git grep -i please pkg` in cockpit and there is bunch of such messages. Do we want to adjust these as well?

Fixed.

garrett:
Changed the title to reflect what this PR is about. (We should try to not have generic headings; it makes it harder to understand what the changes are supposed to be, and it makes it even more difficult to come back to PRs that are open for a while, which this one isn't... as I'll approve it too. :wink:)

garrett (approved):
_Please_ accept my +1 :+1: :wink:

PR #976, opened 2022-05-12
Don't wrap a promise with a promise

jelly:
Most functions in client.js wrap a promise around a podmanCall which is
already a promise.

Just noticed this occurs a few times in a new PR, I've also been guilty of doing such copy pasta so let's clean it up.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -66,15 +66,7 @@ export function getContainers(system, id) {
     });
 }
 
-export function getContainerStats(system, callback) {
-    return new Promise((resolve, reject) => {
-        const options = {
-            stream: true,
-        };
-        podmanMonitor("libpod/containers/stats", "GET", options, callback, system)
-                .then(resolve, reject);
-    });
-}
+export const getContainerStats = (system, callback) => podmanMonitor("libpod/containers/stats", "GET", { stream: true }, callback, system);
 
 export function inspectContainer(system, id) {
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -67,13 +67,10 @@ export function getContainers(system, id) {
...
-        const options = {
-            stream: true,
-        };
-        podmanMonitor("libpod/containers/stats", "GET", options, callback, system)
-                .then(resolve, reject);
-    });
+    const options = {
+        stream: true,
+    };
+    return podmanMonitor("libpod/containers/stats", "GET", options, callback, system);
```
martinpitt, line 73: I'd clean this up even further; declaring `options` here seems overkill. You can just make this all functional with something like
```js
export const getContainerStats = (system, callback) => podmanMonitor("libpod/containers/stats", "GET", { stream: true }, callback, system);
```

which is nice and concise IMHO.

src/client.js
```diff
@@ -106,39 +99,17 @@ export function createContainer(system, config) {
...
 }
 
-export function commitContainer(system, commitData) {
-    return new Promise((resolve, reject) => {
-        podmanCall("libpod/commit", "POST", commitData, system)
-                .then(resolve)
-                .catch(reject);
-    });
-}
+export const commitContainer = (system, commitData) => podmanCall("libpod/commit", "POST", commitData, system);
```
martinpitt, line 102: oh yes, exactly like this :grin:

martinpitt (dismissed):
Thanks! Could be cleaned up even more, but definitively an improvement!

src/client.js
jelly: Fine by me, I wasn't sure if we wanted to go so far but happy to delete more lines :)

marusak: Of course this is a good cleanup. This is just gonna be copy-pasta from when this whole file has been introduced. So please feel free to clean it up as well :)

marusak (approved):
Thanks, nice cleanup!

martinpitt:
Nice, thanks!

PR #982, opened 2022-05-16
Support url parameters for filtering

jelly:
Since Cockpit 269 the service top CPU usage lists podman containers and
links to the podman page without filtering on a specific container name.
For users it would be more useful if they could directly spot the
container, the newly introduced url parameters can filter based on owner
and a search string.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -65,12 +65,14 @@ class Application extends React.Component {
             podmanRestartAvailable: false,
             currentUser: _("User"),
             privileged: false,
+            location: {},
         };
         this.onAddNotification = this.onAddNotification.bind(this);
         this.updateState = this.updateState.bind(this);
         this.onDismissNotification = this.onDismissNotification.bind(this);
         this.onFilterChanged = this.onFilterChanged.bind(this);
         this.onOwnerChanged = this.onOwnerChanged.bind(this);
+        this.onContainerFilterChanged = this.onContainerFilterChanged.bind(this);
         this.updateImagesAfterEvent = this.updateImagesAfterEvent.bind(this);
         this.updateContainerAfterEvent = this.updateContainerAfterEvent.bind(this);
         this.updateContainerStats = this.updateContainerStats.bind(this);
@@ -79,6 +81,7 @@ class Application extends React.Component {
         this.handleImageEvent = this.handleImageEvent.bind(this);
         this.handleContainerEvent = this.handleContainerEvent.bind(this);
         this.checkUserService = this.checkUserService.bind(this);
+        this.onNavigate = this.onNavigate.bind(this);
     }
 
     onAddNotification(notification) {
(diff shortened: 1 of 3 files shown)


src/app.jsx
```diff
@@ -65,6 +65,7 @@ class Application extends React.Component {
             podmanRestartAvailable: false,
             currentUser: _("User"),
             privileged: false,
+            location: {},
```
jelly, line 68: maybe it should be renamed to `url_parameters`?

src/app.jsx
```diff
@@ -511,6 +533,29 @@ class Application extends React.Component {
...
+        cockpit.removeEventListener("locationchanged", this.onNavigate);
+    }
+
+    onNavigate() {
+        // HACK: Use usePageLocation when this is rewritten into a functional component
+        const { options, path } = cockpit.location;
+        this.setState({ location: options });
+        // only use the root path
+        if (path.length === 0) {
+            if (options.container) {
```
jelly, line 551: We should probably change this to `name`

garrett (changes requested):
Ah, this is a great start!

It doesn't support all/running for containers.

Also, does this support targeting specific containers? Or just their names? (I guess container names are unique.) I'm specifically wondering about how this will be used to link from the metrics page.

marusak:
> It doesn't support all/running for containers.

Maybe we should add that as well. I know that when you get here from metrics page the container needs to be running (to consume resources) but it might just die as you go to the podman page and then you would see empty page. going to `?name=foo&owner=system&containers=all` would be more robust.

src/app.jsx
marusak: :man_shrugging:

garrett (approved):
:+1: Looks great!

---

I think we should add a filter clear link when filters are active. As we had filtering prior to this PR, I guess that could be done in a follow-up. But it would be nice to have it in an _immediate_ follow-up, especially if we want to talk about filtering in the release notes.

Here's a screenshot of something being filtered. It's not entirely obvious that things are filtered at the moment, especially if one follows a link that filters by default, from another page.

Screenshot 2022-05-17 at 18-01-00 Podman containers - garrett@Bolt

https://www.patternfly.org/v4/guidelines/filters/#clearing-filters:

> **Clearing filters**
>
> The option to clear all filters will be displayed after the last filter chip. It should read “Clear filters” in sentence case. When clicked, it should remove all applied filters and collapse the toolbar back into a single row.

It doesn't show a screenshot, but there is a filter clear link @ https://www.patternfly.org/v4/components/table#composable-empty-state (although that's an empty state, and this wouldn't necessarily be so).

jelly:
> +1 Looks great!
> 
> I think we should add a filter clear link when filters are active. As we had filtering prior to this PR, I guess that could be done in a follow-up. But it would be nice to have it in an _immediate_ follow-up, especially if we want to talk about filtering in the release notes.
> 
> Here's a screenshot of something being filtered. It's not entirely obvious that things are filtered at the moment, especially if one follows a link that filters by default, from another page.
> 
> Screenshot 2022-05-17 at 18-01-00 Podman containers - garrett@Bolt
> 
> https://www.patternfly.org/v4/guidelines/filters/#clearing-filters:
> 
> > **Clearing filters**
> > The option to clear all filters will be displayed after the last filter chip. It should read “Clear filters” in sentence case. When clicked, it should remove all applied filters and collapse the toolbar back into a single row.
> 
> It doesn't show a screenshot, but there is a filter clear link @ https://www.patternfly.org/v4/components/table#composable-empty-state (although that's an empty state, and this wouldn't necessarily be so).

I understand the concern but I also feel that it doesn't per se fit the design as filtering is usually applied to a table. We can add a clear filters button but that also feels weird as it clears three things then?

I was looking at the logs page as comparison btw.

PR #972, opened 2022-05-04
Add testing on fedora-coreos

martinpitt:
- https://github.com/cockpit-project/bots/pull/3331

PR #986, opened 2022-05-18
test: Fix TestApplication.testBasicSystem on fedora-coreos

martinpitt:
The test tries to log in as root with password, which is not allowed by
default through SSH. Enable that temporarily.

---

Fixes [this failure](https://logs.cockpit-project.org/logs/pull-3390-20220517-131823-342f3f53-fedora-coreos-cockpit-project-cockpit-podman/log.html). It still worked in PR #972 two weeks ago, but I forgot to enable fedora-coreos in the testmap. I am doing this in https://github.com/cockpit-project/bots/pull/3390

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -354,8 +354,14 @@ class TestApplication(testlib.MachineCase):
         # Checking images is harder but if there would be more than one this would fail
         b.wait_visible("#containers-images:contains('quay.io/cockpit/registry:2')")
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Hmm, [weird failure](https://logs.cockpit-project.org/logs/pull-986-20220518-060247-8d3cae7f-debian-testing/log.html#1), but looks unrelated. Retrying.

jelly (approved):

PR #988, opened 2022-05-19
test: Adjust tests for coreos

marusak:
fedora-coreos contains two additional images - cockpit/ws and intermediate one.


jelly:
Do we know where the image came from, as the tests did succeed on coreos at some point?

martinpitt:
See vm.install -- image-prepare fedora-coreos builds a new container out of cockpit/ws with installing the locally built .rpms. So the question is rather, why did it work before?

martinpitt (approved):
Thanks! This could use some more factorization, but let's fix this regression on main ASAP.

PR #991, opened 2022-05-25
Build translation cleanup

jelly:
Use the html2po and manifest2po tools that we already check out from
cockpit repo pkg/lib.

These tools switched to a new argument parser library (`argparse`) in
cockpit-project/cockpit#16271, so add it to our package.json and drop
`stdio`.

Drop our old copies.

I checked the result of building `po/podman.pot` before and after the
change, and aside from the timestamp, it's identical.

---

PR based on https://github.com/cockpit-project/cockpit-machines/pull/681


marusak:
Did you try and run this to see it actually works?

martinpitt (approved):
Thanks! I diffed po/podman.pot before/after and it LGTM.

PR #993, opened 2022-05-27
Makefile: Update default OS and cockpit checkout

marusak:
Checkout sha is tag 270.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PACKAGE_NAME := $(shell awk '/"name":/ {gsub(/[",]/, "", $$2); print $$2}' packa
 RPM_NAME := cockpit-$(PACKAGE_NAME)
 VERSION := $(shell T=$$(git describe 2>/dev/null) || T=1; echo $$T | tr '-' '.')
 ifeq ($(TEST_OS),)
-TEST_OS = fedora-35
+TEST_OS = fedora-36
 endif
 export TEST_OS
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -185,7 +185,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
-COCKPIT_REPO_COMMIT = 47539d6d31520da8b0ed916ee6d78e5aefc1ea87
+COCKPIT_REPO_COMMIT = 8f48c4740593e4eb08e1f40ffd98dbeb09f5c14a
```
martinpitt, line 188: I'd really prefer this was a tag name, or at least get an accompanying comment that says "# 270". SHAs are not for humans.

martinpitt (approved):
.. but we can do this the next time we bump the SHA.

Makefile
marusak: https://github.com/cockpit-project/cockpit-podman/commit/cfc05e730d6f5ddf4244d111bf2d7d0b0c24cc4b

as per commit message:
> ...
> We'll hardcode shas from here on out: it's more secure (ie:
tamperproof), and more performant (since we don't have to check the ref
if we already have the object), in addition to being easier to work with
in terms of the required git commands.

martinpitt: Sure, but that doesn't prevent us from adding a comment.

PR #996, opened 2022-05-30
Revert "test: Adjust tests for coreos", run tests offline

martinpitt:
The intermediate images were the previous versions of the images which
`podman pull` in test/vm.install updated. That was a time bomb, as
sometimes there would be a new version and sometimes not.

Now that we never refresh the images any more during tests, these
transient intermediate images should be gone.
    
This reverts commit fb0437f5045cd03346d16b282235ce7bc7a233b4.

---

Spotted by/blocks https://github.com/cockpit-project/bots/pull/3437


marusak (approved):
ah, nice! Thanks :)

PR #1001, opened 2022-05-31
Correctly preselect and disable owner when creating container in pod

marusak:
Otherwise when you click on `Create container in pod` for user pod, it
selects by default `system` and then fails with `No such pod`.


marusak:
Needs naughty, sent https://github.com/cockpit-project/bots/pull/3455

PR #1009, opened 2022-06-03
workflows: Tag our releases in node-cache

martinpitt:
Unmodified copy from cockpit.git. Copying this is a bit awkward, but
let's discuss how we can share these (composite actions or otherwise)
without time pressure.


jelly:
So the idea of this workflow is that tags the node-cache when there is a release and this runs on tag?

martinpitt:
@jelly : Right, so that we can easily reconstruct the exact node_modules/ at the time of the release. Same as in https://github.com/cockpit-project/cockpit/blob/main/.github/workflows/release.yml#L136

allisonkarlitskaya (approved):
I have some comments on the content of the script, but since it's a straight copy...

PR #1013, opened 2022-06-09
Re-drop node_modules from release tarballs, Include Cockpit repo files in release tarballs

martinpitt:
commit c06bf5cfd6c8 had the unintended side effect of including
node_modules/ into the dist tarball, as `git ls-files` now includes
`node_modules. This made the tarball balloon from 0.8 to 65 MB, and make
`tar` take ages to compress.

Explicitly ignore the directory.

----

Compare the tar size for [49](https://github.com/cockpit-project/cockpit-podman/releases/tag/49) to [48](https://github.com/cockpit-project/cockpit-podman/releases/tag/48).

I noticed this when trying to package the new releases for Debian. It took painfully long to process the tarball, and I got a gazillion lintian warnings about the unaccounted files in n_m/. After this I'd like to tag a 49.1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,28 @@ PYEXEFILES=$(shell git grep -lI '^#!.*python')
 
 all: $(WEBPACK_TEST)
 
+# checkout common files from Cockpit repository required to build this project;
+# this has no API stability guarantee, so check out a stable tag when you start
+# a new project, use the latest release, and update it from time to time
+COCKPIT_REPO_FILES = \
+	pkg/lib \
+	test/common \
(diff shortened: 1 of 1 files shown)


martinpitt:
This is even more annoying.. tmt rsyncs the tree to and from the VM, multiple times, and each round takes several minutes. So this also increased our test runtime quite a bit.

martinpitt:
Like in https://github.com/cockpit-project/cockpit-machines/pull/737 I added another fix.

marusak (approved):
Danke!

PR #1014, opened 2022-06-09
packit: Drop FMF hack, enable downstream job notifications

martinpitt:
tmt 1.14.0 can finally discover tests from the dist-git source, instead
of having them to checkout from upstream git.

https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/37
dropped the hardcoded `ref:`, so we can drop all that hackery.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -6,19 +6,8 @@ downstream_package_name: cockpit-podman
 copy_upstream_release_description: true
 
 actions:
-  post-upstream-clone:
-    - make cockpit-podman.spec
-    # HACK: until FMF uses tests from dist-git source tarball: https://github.com/teemtee/tmt/issues/585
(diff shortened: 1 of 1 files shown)


martinpitt:
@allisonkarlitskaya You'll be pleased :grin:

packit.yaml
```diff
@@ -1,4 +1,6 @@
 upstream_project_url: https://github.com/cockpit-project/cockpit-podman
+# enable notification of failed downstream jobs as issues
+issue_repository: https://github.com/cockpit-project/cockpit-podman
```
marusak, line 3: Hmm, lets see how verbose this is gonna be :)

marusak (approved):
Thanks!

packit.yaml
martinpitt: On failures you'll get something like https://github.com/martinpitt/umockdev/issues/178 . But ideally of course it should be silent :grinning:

PR #1016, opened 2022-06-14
Keep .spec in git

martinpitt:
Give up on .spec.in and directly keep the .spec in git. This aligns a
lot better with what packit wants to do, and avoids having explicit
`post-upstream-clone:` actions.

We still want to build tarballs with a correct .spec (for using with
`rpmbuild -bb` and VM image preparation), so make sure that they get the
patched file.

 - [x] fix or work around https://github.com/packit/packit/issues/1621
 - [x] fix or work around https://github.com/packit/packit/issues/1637
 - [ ] Test release on my fork

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -1,14 +1,11 @@
 upstream_project_url: https://github.com/cockpit-project/cockpit-podman
 # enable notification of failed downstream jobs as issues
 issue_repository: https://github.com/cockpit-project/cockpit-podman
-specfile_path: cockpit-podman.spec
-upstream_package_name: cockpit-podman
-downstream_package_name: cockpit-podman
+specfile_path: packaging/cockpit-podman.spec
 # use the nicely formatted release description from our upstream release, instead of git shortlog
 copy_upstream_release_description: true
 
 actions:
-  post-upstream-clone: make cockpit-podman.spec
   create-archive: make dist
 
 srpm_build_deps:
@@ -30,7 +27,6 @@ jobs:
     project: "cockpit-preview"
     preserve_project: True
     actions:
-      post-upstream-clone: make cockpit-podman.spec
       # HACK: tarball for releases (copr_build, koji, etc.), copying spec's Source0; this
       # really should be the default, see https://github.com/packit/packit-service/issues/1505
       create-archive:
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ endif
 export TEST_OS
 TARFILE=$(RPM_NAME)-$(VERSION).tar.xz
 NODE_CACHE=$(RPM_NAME)-node-$(VERSION).tar.xz
-SPEC=$(RPM_NAME).spec
 APPSTREAMFILE=org.cockpit-project.$(PACKAGE_NAME).metainfo.xml
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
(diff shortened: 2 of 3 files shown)


martinpitt:
packit srpm [failed](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-podman-1016/srpm-builds/04531979/builder-live.log.gz) with this:

    ValueError: '/tmp/tmp48d6s1r0/cockpit-podman-49.1.5.gaedec28.tar.xz' is not in the subpath of '/tmp/tmp48d6s1r0/packaging' OR one path is relative and the other is absolute.

It already [failed the same way](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-podman-1016/srpm-builds/04531965/builder-live.log.gz) in my previous attempt when I dropped `specfile_path:`. I [put that back](https://github.com/cockpit-project/cockpit-podman/compare/68321b85de4e1cfc6cc44b0ca/54ff3b193bb4a13..aedec2890f8a459b404ea2fa35f5026d0f9a835b) as I thought it might maybe require it (even though the [docs](https://packit.dev/docs/configuration/) say that it's optional), but that didn't help.

@LachmanFrantisek, @TomasTomecek : What am I doing wrong here? This PR is actually meaning to go *towards* the approach that packit wants, i.e. we want to reduce our packit.yaml. I.e. I'm happy to structure this otherwise, in some "that's the standard way". Thank you in advance!

packit.yaml
```diff
@@ -1,14 +1,12 @@
 upstream_project_url: https://github.com/cockpit-project/cockpit-podman
 # enable notification of failed downstream jobs as issues
 issue_repository: https://github.com/cockpit-project/cockpit-podman
-specfile_path: cockpit-podman.spec
```
martinpitt, line 4: See https://github.com/packit/packit-service/issues/1511 -- there's a bug that propose-downstream fails without an explicit `specfile_path:`, so we'll most likely have to bring that back anyway (with a bug ref). That will surface when testing the releasing, but until then we first need to sort out the `packit srpm` build.

martinpitt:
Same issue in https://github.com/cockpit-project/cockpit/pull/17450. I can reproduce this locally with `packit srpm`, will have a look after lunch.

martinpitt:
@LachmanFrantisek, @TomasTomecek : Isolated, reproduced, filed https://github.com/packit/packit/issues/1621 about it.

martinpitt:
Still fails srpm build, I followed up to https://github.com/packit/packit/issues/1621

martinpitt:
The packit issue was fixed, so let's try this again.

martinpitt:
Argh, this is *still* broken. I can reproduce this locally:
```
❱❱❱ packit srpm
2022-08-31 17:37:51.075 command_handler.py INFO   Nothing to clean.
2022-08-31 17:37:51.075 utils.py          ERROR  Preparation of the repository for creation of an SRPM failed: Specfile /var/home/martin/upstream/cockpit-podman/cockpit-podman.spec not found on ref spec.
```

So it seems that once again violates the "Else recursively search the tree and use the first spec file found." [documentation](https://packit.dev/docs/configuration/#specfile_path).

```
❱❱❱ find -name '*.spec'
./packaging/cockpit-podman.spec
```

martinpitt:
Not important, and not worth the hassle.

PR #1020, opened 2022-06-20
.github: add reposchutz workflow

jelly:
Verify the commits in the node_modules repository using the similiar
workflow as cockpit's with a small adjustment to checkout the
tools/node-modules script.

Tested it with a my main:

https://github.com/jelly/cockpit-podman/runs/6972626867?check_suite_focus=true

And broke it:

https://github.com/jelly/cockpit-podman/runs/6972679595?check_suite_focus=true 

If this is correct, I'll make a similar PR to cockpit-machines.

### DIFF
--- a/.github/workflows/reposchutz.yml
+++ b/.github/workflows/reposchutz.yml
@@ -1,0 +1,65 @@
+name: repository
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
+
+jobs:
+  check:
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! LGTM, and it's pretty much an exact copy, isn't it?

jelly:
> 

Yes, except it needs to pull in `tools/node-modules` as that's not cloned by default.

PR #1023, opened 2022-06-23
all: Use WithDialogs protocol

mvollmer:
This also makes sure that the dropdowns are closed after selecting an
action from them.  This was a bit inconsistent.

Also, the error in the PodDeleteModal is now outside of the container
Stack element.  Deleting a empty pod would never shown an error.

### DIFF
--- a/src/ContainerCheckpointModal.jsx
+++ b/src/ContainerCheckpointModal.jsx
@@ -3,6 +3,8 @@ import { Button, Checkbox, Form, Modal } from '@patternfly/react-core';
 import { DialogsContext } from "dialogs.jsx";
 import cockpit from 'cockpit';
 
+import * as client from './client.js';
+
 const _ = cockpit.gettext;
 
 class ContainerCheckpointModal extends React.Component {
@@ -25,6 +27,21 @@ class ContainerCheckpointModal extends React.Component {
             this.setState({ [event.target.name]: event.target.checked });
     }
 
+    handleCheckpointContainer(args) {
+        const Dialogs = this.context;
+        const container = this.props.containerWillCheckpoint;
+        this.setState({ inProgress: true });
+        client.postContainer(container.isSystem, "checkpoint", container.Id, args)
+                .catch(ex => {
+                    const error = cockpit.format(_("Failed to checkpoint container $0"), container.Names);
+                    this.props.onAddNotification({ type: 'danger', error, errorDetail: ex.message });
+                    this.setState({ inProgress: false });
+                })
+                .finally(() => {
+                    Dialogs.close();
+                });
+    }
+
     render() {
         const Dialogs = this.context;
         return (
@@ -36,8 +53,7 @@ class ContainerCheckpointModal extends React.Component {
                        <Button variant="primary" isDisabled={this.state.inProgress}
                                isLoading={this.state.inProgress}
                                onClick={() => {
-                                   this.setState({ inProgress: true });
-                                   this.props.handleCheckpointContainer({
+                                   this.handleCheckpointContainer({
                                        keep: this.state.keep,
                                        leaveRunning: this.state.leaveRunning,
                                        tcpEstablished: this.state.tcpEstablished,
--- a/src/ContainerCheckpointModal.jsx
+++ b/src/ContainerCheckpointModal.jsx
@@ -1,13 +1,17 @@
 import React from 'react';
 import { Button, Checkbox, Form, Modal } from '@patternfly/react-core';
+import { DialogsContext } from "dialogs.jsx";
 import cockpit from 'cockpit';
 
 const _ = cockpit.gettext;
 
 class ContainerCheckpointModal extends React.Component {
+    static contextType = DialogsContext;
+
     constructor(props) {
         super(props);
         this.state = {
+            inProgress: false,
             keep: false,
             leaveRunning: false,
             tcpEstablished: false,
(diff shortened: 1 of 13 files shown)


src/ContainerCheckpointModal.jsx
```diff
@@ -22,19 +26,28 @@ class ContainerCheckpointModal extends React.Component {
...
                    position="top" variant="medium"
                    title={cockpit.format(_("Checkpoint container $0"), this.props.containerWillCheckpoint.Names)}
                    footer={<>
-                       <Button variant="primary" isDisabled={this.props.checkpointInProgress}
-                               isLoading={this.props.checkpointInProgress}
-                               onClick={() => this.props.handleCheckpointContainer(this.state)}>
+                       <Button variant="primary" isDisabled={this.state.inProgress}
+                               isLoading={this.state.inProgress}
+                               onClick={() => {
+                                   this.setState({ inProgress: true });
```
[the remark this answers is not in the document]
mvollmer: Yes, I noticed that, too... hmm.

mvollmer:
Many of the action functions are a parameter of the dialogs, but they close the dialogs themselves. That could be cleaned up, by moving action functions into the dialogs if possible, or turning them into promises and having the dialog code watch them and close the themselves.
That would then also extend to the "in progress" state.

src/ContainerCheckpointModal.jsx
```diff
@@ -21,20 +27,43 @@ class ContainerCheckpointModal extends React.Component {
...
+        const container = this.props.containerWillCheckpoint;
+        this.setState({ inProgress: true });
+        client.postContainer(container.isSystem, "checkpoint", container.Id, args)
+                .catch(ex => {
+                    const error = cockpit.format(_("Failed to checkpoint container $0"), container.Names);
+                    this.props.onAddNotification({ type: 'danger', error, errorDetail: ex.message });
+                })
+                .finally(() => {
+                    this.setState({ inProgress: false });
+                    Dialogs.close();
```
martinpitt, line 41: The `inProgress: false` should be in the error path (catch) when the dialog stays open, to show the inline error. Setting it right before closing the dialog (in the success case) is a no-op, isn't it?

martinpitt (changes requested):
Thanks! I still have some difficulty with understanding the inProgress handling.

src/ContainerCheckpointModal.jsx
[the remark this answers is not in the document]
martinpitt: better, but let's continue in the new thread below.

src/ContainerCheckpointModal.jsx
mvollmer: Yes, `this.setState({ inProgress: false });` is a no-op and its presence is probably too confusing. The dialog always closes, even when there was an error.  The error message appears in a global notification, not an inline alert.

We have two basic patterns for what happens after the user clicks "Apply" in a dialog (see below), and this dialog here is a mix of them. I learned about this from @garrett just recently, so I might be wrong, as always. :-) Let me try to write this down, I'll copy into some more offical place laterish.

1) The action triggered by "Apply" is carried out while the dialog is open.  All inputs/buttons/checkboxes etc in the dialog are disabled, except` for a "Cancel" or "Stop" button if the action can be cancelled.  A spinner appears in the "Apply" button, and maybe some progress messages appear in the footer. When the action finishes successfully, the dialog is closed.  When there is an error, the error appears in the dialog at the top in a inline alert (and the dialog stays open, of course).

2) The dialog closes as soon as "Apply" is clicked (and input validation has succeeded) and the action runs in the background. There is some immediate indication in the page that the action has started, and some obvious effect when the action has finished successfully. There is no potential for problems when the user interacts with the page while the action is still running. There is no potential for problems when multiple actions run concurrently. If there is an error, a global notification appears. In the case of an error, it is trivial for the user to try again (without having to fill out a big dialog again from scratch).

Pattern 2) is preferred and should be used when all the conditions are fulfilled.

Here, checkpointing a container does not leave a immediate indication in the page when it has started and there is no obvious effect when it has finished. (The content of the dropdown menu changes, but that is not normally visible). Also, it might be problematic to checkpoint a container again that is currently being checkpointed or try to restore a container that hasn't finished checkpointing, depending on how the podman API behaves in these cases. I guess that's why we have the progress indicator here. (The dialog for deleting a container does not have a progress indicator, probably because deletion happens instantly and has an obvious effect.) 

However, the error of a checkpoint action is then shown as a global notification. And retrying is not totally trivial since the user has to ponder all the checkboxes again.

So I'd say we either need to fix the page so that checkpointing can use pattern 2 (put some visible indications somewhere that a container is being checkpointed or has a checkpoint), or use an inline error.

I didn't want to do that here, though.

I guess I simply take out the `this.setState({ inProgress: false });`, okay?

mvollmer: > If you take out the inProgress: false from finally(), then pelase move it to .catch() -- then it will DTRT if we ever actually _do_ move the alert into the dialog, and it's the correct thing to do as a best practice anyway.

Excellent idea!

mvollmer: Done.

martinpitt (approved):
Thanks! A lot of the dialogs still don't set inProgress back to false, but this should become obvious if/when we actually change them to inline alerts. At least we have *a* good role model in the code now.

PR #1026, opened 2022-06-28
test pyhon3.11 tryout

jelly:


martinpitt:
See #1028 where I rope this in.

PR #1028, opened 2022-06-30
Makefile: Sync with starter-kit; bring in Python 3.11 test fix

martinpitt:


martinpitt:
[rawhide failed](https://artifacts.dev.testing-farm.io/afa7e51b-060b-4afd-b71d-4ae324d97243/) with "connection refused" errors on firefox. chrome-remote-interface is current (0.31.2), and the exact same testlib worked for starter-kit already in https://github.com/cockpit-project/starter-kit/pull/590 . Retrying once.

martinpitt:
Meh, failed again. Trying without commit 0d45a8ab147f50541816fc6ed81e40650f137f1b (syncing to starter-kit Makefile) for now.

martinpitt:
Still the [same failure](https://artifacts.dev.testing-farm.io/df1f4796-1710-431e-82c3-519cb869bcee/), so syncing the Makefile was not it (that also would have been weird..). I'm putting this back, to avoid losing it.

I suspect it's still using the wrong node_modules somehow? The rawhide test run does *not* run npm install:

    + '[' '!' -d node_modules/chrome-remote-interface ']'

but then again our tarball does not contain node_modules/, so WTH. is this existing directory coming from? c-machine has exactly the same test/browser/run-test.sh, and [its rawhide run](https://artifacts.dev.testing-farm.io/278549d8-c609-44d7-a85f-5907bb987e86/) also does not install node_modules/.

Adding some debugging.

test/browser/run-test.sh
```diff
@@ -7,15 +7,9 @@ git init
...
-    # copy package.json temporarily otherwise npm might try to install the dependencies from it
-    rm -f package-lock.json  # otherwise the command below installs *everything*, argh
-    mv package.json .package.json
-    # only install a subset to save time/space
-    npm install chrome-remote-interface sizzle
-    mv .package.json package.json
-fi
+# ensure current node_modules for testing
+rm -rf node_modules
+tools/node-modules checkout
```
jelly, line 12: Still weird that there is a node_modules, but this looks much better then the hacky workaround code we had.

martinpitt: Right, I just gave up trying to understand it, sorry. I have not the slightest idea where it's coming from :cry:

jelly (approved):

PR #1030, opened 2022-07-04
.github: don't depend on COCKPIT_REPO_STAMP target

jelly:
Instead of relying on COCKPIT_REPO_STAMP which changed in
fca470b5135228fcd8865 use the existing make target.

### DIFF
--- a/.github/workflows/reposchutz.yml
+++ b/.github/workflows/reposchutz.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Check for node_modules availability and package.json consistency
         run: |
           # Make tools/node-modules available
-          make pkg/lib/cockpit.js
+          make tools/node-modules
           # for each commit in the PR which modifies package.json or node_modules...
           for commit in $(git log --reverse --full-history --format=%H \
                               "${HEAD_SHA}" --not "${BASE_SHA}" -- package.json node_modules); do


jelly:
@allisonkarlitskaya do I need to make my PR from `origin`?

.github/workflows/reposchutz.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Check for node_modules availability and package.json consistency
         run: |
           # Make tools/node-modules available
-          make pkg/lib/cockpit.js
+          make tools/node_modules
```
allisonkarlitskaya, line 59: I think you meant `tools/node-modules`?

allisonkarlitskaya (changes requested):

allisonkarlitskaya:
> @allisonkarlitskaya do I need to make my PR from `origin`?

No.  But you need the label set.  I just created it.

.github/workflows/reposchutz.yml
jelly: Yes, danke

jelly:
> > @allisonkarlitskaya do I need to make my PR from `origin`?
> 
> No. But you need the label set. I just created it.

Aha, but then it still fails as it takes the current workflow from main?

allisonkarlitskaya:
> > > @allisonkarlitskaya do I need to make my PR from `origin`?
> > 
> > 
> > No. But you need the label set. I just created it.
> 
> Aha, but then it still fails as it takes the current workflow from main?



> > > @allisonkarlitskaya do I need to make my PR from `origin`?
> > 
> > 
> > No. But you need the label set. I just created it.
> 
> Aha, but then it still fails as it takes the current workflow from main?

Correct.  I propose we merge this and then test with another PR.

allisonkarlitskaya (approved):

PR #1033, opened 2022-07-09
Remove non-formatted text

Fowindev:
msgstr should be:
"Suppression d’images non utilisées"
and was: "$0 Suppression d’images non utilisées$1"


marusak:
Thank you for the finding!
If we would fix it here as you suggested, it would get overwritten with the next sync from Weblate. I did this update there. 
https://translate.fedoraproject.org/translate/cockpit-podman/main/fr/?checksum=be8c3a856d71e79e#history

PR #1038, opened 2022-07-15
packit.yaml: drop deprecated metadata key

jelly:
The packit docs no longer mention the metadat key as required.
https://packit.dev/docs/configuration/#supported-jobs


martinpitt (approved):
Cheers!

PR #1039, opened 2022-07-15
Restart policy users

jelly:


src/ImageRunModal.jsx
```diff
@@ -353,6 +354,8 @@ export class ImageRunModal extends React.Component {
     componentDidMount() {
         this._isMounted = true;
         this.onSearchTriggered(this.state.searchText);
+        cockpit.file(`/var/lib/systemd/linger/${this.props.user}`).read()
+                .then((content, tag) => this.setState({ userLingeringEnabled: content !== null && tag !== '-' }));
```
jelly, line 358: Alternative if we don't want to depend on this file existing, call

```
[jelle@t14s][~/projects/systemd/build]%loginctl show-user jelle | grep Linger
Linger=yes
```

Or use the DBUS interface:

But we need to get an uid then.
```
[jelle@t14s][~/projects/systemd/build]%sudo busctl call org.freedesktop.login1 /org/freedesktop/login1 org.freedesktop.login1.Manager GetUser u  1000
o "/org/freedesktop/login1/user/_1000"

[jelle@t14s][~/projects/systemd/build]%busctl get-property org.freedesktop.login1  /org/freedesktop/login1/user/_1000 org.freedesktop.login1.User  Linger
b true
```

jelly: Although, maybe this should watch? As now a user has to reload the browser to notice changes.

garrett: Yeah, with very few exceptions we don't require a user to reload the browser for any other changes in Cockpit, really. (The only one I can think of is rebooting, and we should really fix that with a little bit of "PWA" magic.)

martinpitt: > But we need to get an uid then

what do you mean? you need the uid for building the path as well, and that path isn't a public API. D-Bus property would really be better, and you should also get a PropertyChanged message for it?

jelly: Looking at https://freedesktop.org/wiki/Software/systemd/logind/ I am not sure if we even get a PropertyChanged event as the page only mentions PropertyChanged where it seems to emit it?

jelly: Tried:
```
┃         const loginmanager = client.proxy();
┃         loginmanager.addEventListener("changed", (changes) => console.log(changes));
┃         client.subscribe({
E              interface: "org.freedesktop.DBus.Properties",     ■ Expected indentation of 12 spaces but found 13.
E              member: "PropertiesChanged"     ■ Expected indentation of 12 spaces but found 13.
E         }, (changes) => console.log(changes));     ■ Block must not be padded by blank lines.
```

But does not seem to work

martinpitt: OK, meh. If you want to implement reacting to changes, can you please file a systemd issue about the missing PropertiesChanged signal, and describe our workaround? Then we can add a file watch to that directory  with a `# HACK` coomment. This then needs a test to tell us when it stops working (like, the file moves someplace else).

martinpitt: FTR, https://www.freedesktop.org/wiki/Software/systemd/logind/ explicitly documents the signals that *do* generate a PropertiesChanged. So it's "intended" in some sense. We can still ask for this if it's useful, of course.

test/check-application
```diff
@@ -24,6 +24,9 @@ registries = ['localhost:5000', 'localhost:6000']
 
 NOT_RUNNING = ["Exited", "Stopped"]
 
+# Ubuntu 2204 lacks user systemd units https://github.com/containers/podman/commit/9312d458b4254b48e331d1ae40cb2f6d0fec9bd0
```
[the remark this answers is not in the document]
jelly: This is what I get on our image:

```
admin@ubuntu:~$ cat /etc/os-release  | grep VERSION
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
admin@ubuntu:~$ systemctl --user start podman-restart
Failed to start podman-restart.service: Unit podman-restart.service not found.
```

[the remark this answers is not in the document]
jelly: I can't find a bug report here https://launchpad.net/ubuntu/+source/libpod/+bugs should I report a new one?

src/ImageRunModal.jsx
jelly: I added a new test, but maybe that's unrequired and we can just check if the file exists? It feels a bit expensive to keep adding tests.

test/check-application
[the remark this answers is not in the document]
martinpitt: Yes please. From the changelog it doesn't look intended.

test/check-application
```diff
@@ -2028,15 +2031,28 @@ class TestApplication(testlib.MachineCase):
     def testHealthcheckUser(self):
         self._testHealthcheck(False)
 
-    def testPodmanRestartEnabled(self):
+    @testlib.skipImage("podman-restart not available in debian/ubuntu", *DISTROS_WITHOUT_PODMAN_USER_RESTART)
```
martinpitt, line 2034: It does work fine in Debian, right? so somehow the package gets misbuilt in Ubuntu.

src/app.jsx
```diff
@@ -513,6 +514,20 @@ class Application extends React.Component {
                 });
     }
 
+    getUserLingeringEnabled(uid) {
+        const client = cockpit.dbus("org.freedesktop.login1");
+        client.call("/org/freedesktop/login1", "org.freedesktop.login1.Manager", "GetUser",
+                    [uid], { type: "u" })
+                .then(([userpath]) => {
+                    client.call(userpath, "org.freedesktop.DBus.Properties", "Get",
+                                ["org.freedesktop.login1.User", "Linger"], { type: "ss" }).then(([result]) => {
```
martinpitt, line 523: Please always break the line before .then, so that the .catch aligns and this becomes readable.

test/check-application
```diff
@@ -2081,6 +2083,21 @@ class TestApplication(testlib.MachineCase):
...
+        self.login(False)
+
+        b.click("#containers-images button.pf-c-expandable-section__toggle")
+        self.machine.execute("loginctl enable-linger $(id -u admin)")
+        self.addCleanup(self.machine.execute, "loginctl disable-linger $(id -u admin)")
+
+        b.wait_visible('#containers-images td[data-label="Image"]:contains("busybox:latest")')
+        b.click('#containers-images tbody tr:contains("busybox:latest") .ct-container-create')
+        b.wait_visible('div.pf-c-modal-box header:contains("Create container")')
+        b.wait_visible("#run-image-dialog-restart-policy")
```
martinpitt, line 2099: Hmm, this is visible even if podman-restart.service isn't available?

src/ImageRunModal.jsx
martinpitt: Not sure what you mean by this, but indeed this whole PR is a lot of effort. Honestly I don't quite like that the linger file hack is now at *two* different places. This shouldn't be necessary -- .watch() always fires its callback immediately with the current file contents.

(FWIW, I still consider this whole feature "unrequired" -- I guess you better land this while I'm on PTO and not watching :grin: ))

src/app.jsx
```diff
@@ -550,6 +565,15 @@ class Application extends React.Component {
 
         cockpit.user().then(user => {
             this.setState({ currentUser: user.name || _("User") });
+            this.getUserLingeringEnabled(user.id);
+            // HACK: https://github.com/systemd/systemd/issues/22244#issuecomment-1210357701
+            cockpit.file(`/var/lib/systemd/linger/${user.name}`).watch((content, tag) => {
```
martinpitt, line 570: There's no point in querying d-bus if we still have to assume this file location anyway. This watch alone ought to suffice, so let's just drop the dbus property query.

martinpitt (dismissed):

test/check-application
jelly: Good point, we do check for it for the system service, but not the user service yet. Also a good point is to check when we can drop that..

src/app.jsx
```diff
@@ -545,11 +547,28 @@ class Application extends React.Component {
                     }
                 });
 
+        cockpit.spawn(["systemctl", "--user", "show", "--value", "-p", "LoadState", "podman-restart"], { environ: ["LC_ALL=C"], error: "ignore" })
```
martinpitt, line 550: How would this look as a d-bus query? calling systemctl is very heavy compared to a d-bus call.

src/app.jsx
```diff
@@ -545,11 +547,28 @@ class Application extends React.Component {
...
+                .then(out => {
+                    if (out.trim() === "loaded") {
+                        this.setState({ userPodmanRestartAvailable: true });
+                    } else {
+                        this.setState({ userPodmanRestartAvailable: false });
```
martinpitt, line 555: Unify to `userPodmanRestartAvailable: out.trim() === "loaded"`

jelly: Looks good, thanks!

src/app.jsx
jelly: Hmm we first have to look up the unit path and then get the LoadState property. I am not entirely sure how this works for a user session.

marusak: This will be called even when root logs in. I think it makes more sense in `checkUserService`?

jelly: Fixed.

marusak (approved):
Thank you!

PR #1041, opened 2022-07-18
Fix system logic in prune unused images dialog

### LINKED ISSUE #1022: Pruning unused images doesn't work as root
When using cockpit-podman as root, "Prune unused images" always results in this error for me:

    The endpoint must be either a unix path or port number

I think this happens because we try to prune user images, but that doesn't work for root.  The Browser console shows:

    $XDG_RUNTIME_DIR is not present. Cannot use user service.

jelly:
Check if the user or system service is avaiable instead of checking if
both are true. Which does not work as userServiceAvailbe is null when
logging in as root.

Closes #1022

### DIFF
--- a/src/PruneUnusedImagesModal.jsx
+++ b/src/PruneUnusedImagesModal.jsx
@@ -52,10 +52,9 @@ function ImageOptions({ images, checked, isSystem, handleChange, name, showCheck
 class PruneUnusedImagesModal extends React.Component {
     constructor(props) {
         super(props);
-        const isSystem = this.props.userServiceAvailable && this.props.systemServiceAvailable;
         this.state = {
-            deleteUserImages: true,
-            deleteSystemImages: isSystem,
+            deleteUserImages: this.props.userServiceAvailable !== null && this.props.userServiceAvailable,
+            deleteSystemImages: this.props.systemServiceAvailable,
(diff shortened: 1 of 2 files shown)


marusak:
Can you please also add test where you reproduce #1022 ?

src/PruneUnusedImagesModal.jsx
```diff
@@ -52,10 +52,9 @@ function ImageOptions({ images, checked, isSystem, handleChange, name, showCheck
 class PruneUnusedImagesModal extends React.Component {
     constructor(props) {
         super(props);
-        const isSystem = this.props.userServiceAvailable && this.props.systemServiceAvailable;
         this.state = {
-            deleteUserImages: true,
-            deleteSystemImages: isSystem,
+            deleteUserImages: this.props.userServiceAvailable !== null && this.props.userServiceAvailable,
```
marusak, line 56: Couldn't this just be `deleteUserImages: this.props.userServiceAvailable`?

marusak (approved):
Thank you! Just one question, mostly OOI

martinpitt:
> jelle | pitti: I kinda like the explicit null check, it being null was what bit us
> jelle | pitti: we had null && bool  which resuls to null :D

PR #1045, opened 2022-07-19
test: Fix pgrep invocation

martinpitt:
pgrep does not have an `-e` option. "echo" conceptually only makes sense
for pkill, as pgrep's raison d'être is to show the pids..

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -57,7 +57,7 @@ class TestApplication(testlib.MachineCase):
             systemctl reset-failed podman.service podman.socket
             podman system reset --force
             pkill -e -9 podman || true
-            while pgrep -e podman; do sleep 0.1; done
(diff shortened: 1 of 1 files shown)


martinpitt:
See https://github.com/cockpit-project/cockpit-podman/pull/1017#issuecomment-1189386116

@jelly: This also shows that the loop never actually worked -- the loop and the sync just introduced enough delay to mitigate the delayed writes on the overlay.

jelly:
The debian-testing issue seems consistent but retrying at least works. Still curious that it still happens.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-1045-20220719-190330-f95397ee-debian-testing/log.html#15-2

jelly (approved):

PR #1053, opened 2022-07-25
Makefile: Use rsync-friendly directory path

martinpitt:
We want to move bots API from deprecated scp to rsync. rsync uses
trailing slashes to tell apart copying a directory as a different name
in the target (trailing slash) from copying a directory *into* the given
target path (no trailing slash). Here we want to do the former.

---

This blocks https://github.com/cockpit-project/bots/pull/3640


marusak (approved):
Thanks!

PR #1061, opened 2022-08-01
Ship npm integration test dependencies in dist tarball

martinpitt:
This avoids having to run `npm`, and thus a potentially brittle internet
access, during FMF test runs. It also avoids the npm test dependency,
which tends to be uninstallable in Rawhide.

We have successfully used this approach in cockpit for a long time.

----

Avoids [this failure](https://artifacts.dev.testing-farm.io/b3b0d33e-b8f6-4cb5-80da-5a94c5a07095/), which has been going on for a week [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2110750). This already hit us in the past.

By and large the same as https://github.com/cockpit-project/cockpit-machines/pull/784

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,19 +118,28 @@ devel-uninstall:
 print-version:
 	@echo "$(VERSION)"
 
+# required for running integration tests; commander and ws are deps of chrome-remote-interface
+TEST_NPMS = \
+       node_modules/chrome-remote-interface \
+       node_modules/commander \
+       node_modules/sizzle \
+       node_modules/ws \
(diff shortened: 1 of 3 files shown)


martinpitt:
[there it is again](https://artifacts.dev.testing-farm.io/3d854b72-34ee-4ac8-9d96-f60313bd5d0f/), this inexplicable "where the eff comes this old node_modules/ from??". Adjusted.

marusak (approved):
Danke!

PR #1062, opened 2022-08-02
Show volume permission in container integration tab

tomasmatus:
Read-write volumes use left-right arrow,
Read-only volumes use right arrow.

image

Both arrows have a tooltip:
image
image

### DIFF
--- a/src/ContainerIntegration.jsx
+++ b/src/ContainerIntegration.jsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import cockpit from 'cockpit';
 
-import { Button, DescriptionList, DescriptionListTerm, DescriptionListDescription, DescriptionListGroup, List, ListItem } from "@patternfly/react-core";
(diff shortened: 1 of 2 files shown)


src/ContainerIntegration.jsx
```diff
@@ -35,7 +35,11 @@ const renderContainerVolumes = (volumes) => {
     const result = volumes.map(volume => {
         return (
             <ListItem key={volume.Source + volume.Destination}>
-                {volume.Source} &rarr; {volume.Destination}
+                {volume.Source}
+                {volume.RW
+                    ? <Tooltip content="Read-write access"><span> &harr; </span></Tooltip>
+                    : <Tooltip content="Read-only access"><span> &rarr; </span></Tooltip>}
```
marusak, line 41: You need to localize these strings

garrett (approved):
:+1: Looks good!

---

Here's a screenshot of my system that shows a normal podman container for cockpit-website and a toolbox for Fedora 36, where you can see volume mapping in action.

image

PR #1066, opened 2022-08-08
test: Move to flake8

martinpitt:
This is easier to use, more flexible as pycodestyle errors can be
overridden on specific lines, and IDEs/vim-ale can read .flake8 and stop
complaining about the line length.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 195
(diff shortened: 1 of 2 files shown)


jelly:
Should this not use test/static-code? Like I think we did in cockpit-machines. Switching to flake8 is already a good improvement however!

jelly (approved):

martinpitt:
> Should this not use test/static-code?

I actually tried quickly, but it stumbled over eslint. I didn't investigate this yet. But eventually yes, as that also validates correct i18n quotes and such.

PR #1069, opened 2022-08-12
test: Robustify executed commands

martinpitt:
Running `a && b` is wrong for our intent here, as it will silently
succeed if `a` fails, and never run `b`. This can often cause follow-up
failures which are harder to investigate.

This was solved in in https://github.com/cockpit-project/bots/pull/3706
by running all Machine.execute() commands with `set -e`.

Replace `&&` with `;`, use Machine.write() to write files, and drop the
redundant `set -e` call (we don't care about the `-u`, that shell script does
not have any variables).


allisonkarlitskaya (approved):
:+1:

PR #1074, opened 2022-08-17
Show image history

jelly:
A new History tab has been added to the images view which fetches the
history on demand.

The comment field always seems to be empty.. and therefore cut off.

@garrett please take a look:

image

image

On Mobile

image

On tablet

image

### DIFF
--- a/src/ImageHistory.jsx
+++ b/src/ImageHistory.jsx
@@ -1,0 +1,63 @@
+import React, { useState, useEffect } from 'react';
+import cockpit from 'cockpit';
+import * as utils from './util.js';
+import * as client from './client.js';
+
+import { ListingTable } from "cockpit-components-table.jsx";
+
+const _ = cockpit.gettext;
+
+const IdColumn = Id => {
+    Id = utils.truncate_id(Id);
+    // Not an id but <missing> or something else
+    if (/<[a-z]+>/.test(Id)) {
+        return <div className="pf-u-disabled-color-100">{Id}</div>;
+    }
+    return Id;
+};
+
+const ImageDetails = ({ image }) => {
+    const [history, setHistory] = useState([]);
+    const [error, setError] = useState(null);
+
+    useEffect(() => {
+        client.imageHistory(image.isSystem, image.Id).then(setHistory)
+                .catch(ex => {
+                    console.error("Cannot get image history", ex);
+                    setError(true);
+                });
+    }, []);
+
+    const columns = ["ID", _("Created"), _("Created by"), _("Size"), _("Comments")];
+    let showComments = false;
+    const rows = history.map(layer => {
+        const row = {
+            columns: [
+                { title: IdColumn(layer.Id), props: { className: "ignore-pixels" } },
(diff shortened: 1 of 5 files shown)


garrett:
> The comment field always seems to be empty.. and therefore cut off.

Are we able to see if there are any comments and hide the column if there aren't any?

And can we not show anything if it's just `<missing>` in the ID?

jelly:
Found one with comments:


image

garrett:
> Padding comes from PF, it's not really double padding:

I understand that... but it is literally double padding. There's padding, a frame, and padding again. I'm not talking about individual HTML elements, but from a layout / design standpoint. There shouldn't be. We already have padding and a frame, we don't need two borders with padding with each doubled up for the same element. It's just way too much in mobile mode and the data is constrained by all the redundant space and borders.

It's because we're nesting multiple elements within each other.

This is one of the (many) reasons we made the Machines page has sub-pages, not expandable tables with tabs and stuff like in Cockpit-Podman.

(But we had and still have some CSS hacks to strip away the redundant padding and borders around Cockpit.)

jelly:
What do you think of:

image

src/ImageHistory.jsx
```diff
@@ -0,0 +1,63 @@
...
+import * as client from './client.js';
+
+import { ListingTable } from "cockpit-components-table.jsx";
+
+const _ = cockpit.gettext;
+
+const IdColumn = Id => {
+    Id = utils.truncate_id(Id);
+    // Not an id but <missing> or something else
+    if (Id.length != 12) {
```
marusak, line 13: Not the most robust check, but I don't have any better idea either (except maybe additionally checking if it is also hexadecimal)

src/ImageHistory.jsx
```diff
@@ -0,0 +1,63 @@
...
+    }
+    return Id;
+};
+
+const ImageDetails = ({ image }) => {
+    const [history, setHistory] = useState([]);
+    const [error, setError] = useState(null);
+
+    useEffect(() => {
+        client.imageHistory(image.isSystem, image.Id).then(history => setHistory(history))
```
marusak, line 24: you can just pass `setHistory`, no need for the function wrapper.
`client.imageHistory(image.isSystem, image.Id).then(setHistory)`

marusak (changes requested):
Nice, thanks!

Tests are unhappy, two small comments and please add release note, thanks!

src/ImageHistory.jsx
jelly: An idea is to use regex, like `<[a-z]+>`

src/ImageHistory.jsx
jelly: Good point!

src/ImageHistory.jsx
jelly: Applied, seems to work nicely

jelly:
> Nice, thanks!
> 
> Tests are unhappy, two small comments and please add release note, thanks!

Test should be fixed, React keeps a tbody around from the History tab.. which is kinda annoying. Also added a pixel test for the history table

marusak (approved):
Nice, thank you!

PR #1076, opened 2022-08-18
Translate integration tab loading details

jelly:


tomasmatus:
Thanks for noticing! It must have slipped through...

jelly:
> 



> Thanks for noticing! It must have slipped through...

Yeah, I was wondering if we didn't have a linter for it.

tomasmatus (approved):
LGTM!

PR #1081, opened 2022-08-22
Reload pod state after container removal

### LINKED ISSUE #1075: Pod does not know it's container is deleted
When deleting a container in a pod, and then deleting the pod, it lists that it will also delete the already removed container inside the pod.

Steps to reproduce

1. Create a pod
2. Add a container to the pod
3. Delete the container in the pod
4. Delete the pod and notice it lists the already removed container

image

jelly:
Podman does not send a pod event when a container in a pod is removed,
so the only we have to force reload the pod state after a containe
removal.

Fixes #1075

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -413,6 +413,9 @@ class Application extends React.Component {
             break;
         case 'remove':
         case 'cleanup':
+            // HACK: we don't get a pod event when a container in a pod is removed.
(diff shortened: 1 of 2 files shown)


marusak (approved):
Thanks! Any chance for a test?

jelly:
> Thanks! Any chance for a test?

Test added!

marusak (approved):
Thanks!

PR #1082, opened 2022-08-22
Drop the already fixed pod event

jelly:


marusak (approved):
Thanks!  We have test for it, right?

marusak:
https://github.com/KKoukiou/cockpit-podman/commit/db3159232fbd97740707283460943fd2aac3a620

Also the other line should be dropped?

marusak:
If you would add check that the pod is running here after one container is added to it then it would be tested.
https://github.com/cockpit-project/cockpit-podman/blob/main/test/check-application#L215

jelly:
> 



> [KKoukiou@db31592](https://github.com/KKoukiou/cockpit-podman/commit/db3159232fbd97740707283460943fd2aac3a620)
> 
> Also the other line should be dropped?

Hmm need to investigate.

jelly:
> If you would add check that the pod is running here after one container is added to it then it would be tested. https://github.com/cockpit-project/cockpit-podman/blob/main/test/check-application#L215

Seems it is still broken so my testing was wrong... Test on Fedora-36


image

Just added `b.wait_visible("#table-pod-1 .ct-badge-pod-running")`

PR #1091, opened 2022-09-08
test: obtain image count in setUp

jelly:
To make it easier to assert the amount of images which should be shown in the UI, so adding a new container image to the test image does not require us to change all our tests.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -97,13 +97,11 @@ class TestApplication(testlib.MachineCase):
         # HACK: older c-ws versions always log an assertion, fixed in PR cockpit#16765
         self.allow_journal_messages("json_object_get_string_member: assertion 'node != NULL' failed")
 
-    def waitNumImages(self, expected, auth):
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -755,13 +758,13 @@ class TestApplication(testlib.MachineCase):
...
-                expected += 3
+                expected += self.system_images_count
                 if self.machine.ostree_image:
                     expected += 1  # cockpit/ws
 
             b.wait_in_text("#containers-images", f"{expected} images")
 
-        waitImageCount(4)
+        waitImageCount(self.user_images_count)
         image_id = self.execute(auth, "podman images --sort created --format '{{.Id}}' | head -n 1").strip()
```
jelly, line 768: So.. turns out this is racy and selects the wrong image.

jelly:
Retried the last two flakes.

martinpitt (approved):
This is nice, thank you! Retrying the flake.

PR #1092, opened 2022-09-08
Show all containers by default

### LINKED ISSUE #980: Show all containers by default
Only showing running containers by default was originally made for Cockpit-Docker way back in the day. I don't know the reasoning for it, but I think it was because the old Cockpit-Docker page had containers on top and images below and when there were too many containers, you wouldn't be able to see the images, so only showing the running containers was probably a workaround back then.  Cockpit-Podman originally inherited the Cockpit-Docker UI.

The old layout was especially a problem with how things used to work, as Cockpit-Docker and the previous incarnation of Cockpit-Podman had an image-based flow instead of our current container-based flow. You'd have to first download an image first, then launch a new container from the image. When we redesigned the page, we streamlined the flow so that images are an implementation detail instead, and you build a container now using a remote image (that is then downloaded in the background) or a local image — but this two step process is all mostly hidden away and included in the same step now. Images are now collapsed down by default, with the main interaction being pruning old, unused images (and sometimes viewing the images by expanding the group).

So, I'd like to suggest showing containers by default (just like the rest of Cockpit, where we show the states, even for things that aren't currently running or active).

We also want to improve filtering, but in doing so, we should make sure filter states are still filterable, so someone can show only running containers, for example.

tomasmatus:
Fixes #980

---
## Show all containers by default

Cockpit-podman now shows all containers by default.

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -577,8 +577,8 @@ class Containers extends React.Component {
                     </ToolbarItem>
                     <ToolbarItem>
                         <FormSelect id="containers-containers-filter" value={this.props.filter} onChange={this.props.handleFilterChange}>
-                            <FormSelectOption value='running' label={_("Only running")} />
                             <FormSelectOption value='all' label={_("All")} />
+                            <FormSelectOption value='running' label={_("Only running")} />
                         </FormSelect>
                     </ToolbarItem>
                     <Divider isVertical />
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -48,7 +48,7 @@ class Application extends React.Component {
             userImagesLoaded: false,
             systemImagesLoaded: false,
             containers: null,
-            containersFilter: "running",
+            containersFilter: "all",
             containersStats: {},
             containersDetails: {},
             userContainersLoaded: null,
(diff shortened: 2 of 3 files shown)


tomasmatus:
I had an idea that maybe I could change the filtering of containers and first show running containers and then exited/stopped bellow. Of course this brings up a problem with restarting a container as it would jump around in the UI.

garrett:
This is important, as restarting container and failed containers "disappear" from the list. It's unexpected. Plus: Cockpit doesn't hide elements like this anywhere else.

I don't think we should sort by status by default, as it would shift too much (like you mentioned). We do have labels which indicate the status of each container, which helps by default.

Related, but probably separate: We should also improve filtering & add sorting.

tomasmatus:
> I don't think we should sort by status by default, as it would shift too much (like you mentioned). We do have labels which indicate the status of each container, which helps by default.

Ok! Yes, the blue label makes it quite clear which container is running.

marusak:
Can you add `Fixes #980`? Also should we make this release-note? Not a feature, but significant change I would say.

tomasmatus:
>Can you add Fixes #980?

Done

>Also should we make this release-note? Not a feature, but significant change I would say.

Yes, I guess we should.

marusak (approved):
Thanks!
(I still am not fully on board with showing all container by default, but I will leave this decision to others)

PR #1093, opened 2022-09-08
Reconnect console after container restart

### LINKED ISSUE #1089: Container console doesn't reconnect after restarting a container
Cockpit version: 275
Cockpit-podman version: 52
Podman version: 4.2.0
OS: Fedora 36

Steps to reproduce

1. Open a container console
2. Restart the container
3. Console stays disconnected after container is restarted

skazi0:
During restart container's State is still "running" and console was getting disconnected but not reconnected. Additional flag was added to keep information about restart and connect channels again.

Closes #1089

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -294,7 +294,7 @@ class Application extends React.Component {
                 });
     }
 
-    updateContainerAfterEvent(id, system) {
+    updateContainerAfterEvent(id, system, event) {
         client.getContainers(system, id)
                 .then(reply => Promise.all(
                     (reply || []).map(container =>
@@ -311,6 +311,10 @@ class Application extends React.Component {
                         reply = reply[0];
 
                         reply.isSystem = system;
+                        // HACK: during restart State never changes from "running"
+                        //       override it to reconnect console after restart
+                        if (event && event.Action === "restart")
+                            reply.State = "restarting";
                         this.updateState("containers", reply.Id + system.toString(), reply);
                         if (reply.State == "running") {
                             this.inspectContainerDetail(reply.Id, system);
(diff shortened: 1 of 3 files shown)


src/app.jsx
```diff
@@ -311,6 +311,9 @@ class Application extends React.Component {
                         reply = reply[0];
 
                         reply.isSystem = system;
+                        // HACK: during restart State never changes from "running"
```
marusak, line 314: if we would change this to:
```
+                        if (event.Action === "restart")
+                            reply.State = "restarting";
```
we don't need any additional changes and it will also update the label. We should then only define it in `src/utils.js` for it being localized.

WDYT?

skazi0: That was my first idea but it has some unwanted side effects. See: https://github.com/cockpit-project/cockpit-podman/issues/1089#issuecomment-1240676131

marusak: Ah right. So we switched by default to showing all containers and not only running ones, so this should not be a big deal. Also we could show restarting as running

skazi0: Hmmm. I think changing the filter to include "restarting" even when "only running" is selected might be a good/better solution.

skazi0: done

jelly: Please add the `restarting` translation to `src/utils.js` with a comment why it's explicitly added there.

kubat-david: Hi, the `restarting` translation has been added to `src/utils.js` by @skazi0

marusak (approved):
Thank you!

marusak (approved):
Thanks!

PR #1096, opened 2022-09-14
test: Also compute number of images that root sees

mvollmer:
based on how many got deleted.

This matters with a recent rhel-9-1 refresh: https://cockpit-logs.us-east-1.linodeobjects.com/pull-3861-20220914-064245-a4a883db-rhel-9-1-cockpit-project-cockpit-podman/log.html


jelly:
We have flakes again, retying..

jelly (approved):

mvollmer:
The flakey restore_dir behavior on debian-testing needs some investigation.  Looks like containers are started/stopped while we try to snapshot /var/lib/containers...

mvollmer:
Uff, it went green after all.

PR #1102, opened 2022-09-15
Fix caculation of images for commit of running container with pause

skobyda:


skobyda:
Triggered `fedora-coreos@bots#3865`, let's see if it passes

jelly (approved):

jelly:
> Triggered `fedora-coreos@bots#3865`, let's see if it passes

Passed! Thanks, now you should be able to retrigger the coreos refresh test and it should succeed.

PR #1108, opened 2022-09-26
Make "Create container in pod" secondary

jelly:
There should only be one primary action button per page ideally.

image


garrett (commented):
:+1: LGTM! Thanks!

garrett (approved):
:+1: LGTM! Thanks!

PR #1112, opened 2022-09-29
Handle podId in container events

jelly:
Podman now has the podId in the remove and start event, so we can only reload data for a specific pod only. This feature is still in podman main and 4.3.0-rc1.

https://github.com/containers/podman/issues/15408

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -392,7 +392,11 @@ class Application extends React.Component {
         case 'start':
             // HACK: We don't get 'started' event for pods got started by the first container which was added to them
             // https://github.com/containers/podman/issues/7213
-            this.updatePodsAfterEvent(system);
+            if (event.Actor.Attributes.podId) {
+                this.updatePodAfterEvent(event.Actor.Attributes.podId, system);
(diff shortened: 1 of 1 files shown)


martinpitt:
@jelly , @marusak : funny [pixel failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1112-20220929-152107-6ceca6eb-fedora-36/pixeldiff.html#TestApplication-testBasicSystem-overview) here. test-sh4 container apparently lost its CPU usage meter due to this? or is that an unrelated flake/race condition? Then the test should wait until the metric appears before it takes the screenshot.

martinpitt:
The change looks good to me, thanks! Just not sure about the pixel failure.

marusak (approved):
Thanks! The pixel test is flaky, we should make it stable in the code and never show 0% for stopped containers.

jelly:
> Thanks! The pixel test is flaky, we should make it stable in the code and never show 0% for stopped containers.

Yes, let's discuss this during standup.

PR #1114, opened 2022-10-04
Allow equals sign to be entered as env variable

### LINKED ISSUE #1090: cannot enter equals sign (=) into the value field of an environment variable
cockpit-ws @ 275
cockpit-podman @ 52

while attempting to follow these instructions: https://github.com/dgtlmoon/changedetection.io/wiki/Playwright-content-fetcher#fetching-content-using-playwright

i copied the value to the clipboard and typed in PLAYWRIGHT_DRIVER_URL, when i pasted the rest into the value field i was surprised to find it had changed the env to "ws://127.0.0.1:3000/?stealth" and the value to "1&--disable-web-security"

every attempt through developer tools and stuff to allow me to type in just a naked = failed

the solution to the convenience bug #816 seems to have created the problem

jelly:
An equal sign can be part of an env variable.

Closes #1090

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1567,13 +1567,13 @@ class TestApplication(testlib.MachineCase):
         b.click('#run-image-dialog-env-2-btn-close')
         b.click('.env-form .btn-add')
         # Test inputting an key=var entry
-        b.set_val('#run-image-dialog-env-2-value', "RHUBARB=STRAWBERRY DURIAN=LEMON")
+        b.set_val('#run-image-dialog-env-2-value', "RHUBARB=STRAWBERRY DURIAN=LEMON TEST_URL=wss://cockpit/?start=1&stop=0")
         # set_val does not trigger onChange so append a space.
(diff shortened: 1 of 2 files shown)


test/check-application
```diff
@@ -1567,13 +1567,16 @@ class TestApplication(testlib.MachineCase):
         b.click('#run-image-dialog-env-2-btn-close')
         b.click('.env-form .btn-add')
         # Test inputting an key=var entry
-        b.set_val('#run-image-dialog-env-2-value', "RHUBARB=STRAWBERRY DURIAN=LEMON")
+        # testlib.sit()
```
tomasmatus, line 1570: Is this (and the other two) something you forgot to remove?

tomasmatus (commented):
Looks good except for the 3 leftover `testlib.sit()` comments

test/check-application
jelly: Woops, I guess that's what you get when you want to quickly push before going for dinner ;-)

tomasmatus (approved):
Thanks!

PR #1115, opened 2022-10-06
test: Update pixel tests to include 'pause' image

marusak:


jelly:
This should be, `pause` image is no longer included as it's only needed for podman < 4. Looks good however!

jelly (approved):

PR #1116, opened 2022-10-07
Don't show CPU stats for exited containers

jelly:
Podman reports 0 CPU and ram back for non-running containers, due to our code we show 0.00% CPU and no ram. Make this experience consistent and just only show CPU/RAM for running containers, this also fixes a pixel flake where one container did show CPU and the other not.

Note: probably requires a pixel test update...


jelly:
Ok, rawhide broke itself unrelated to the PR.

marusak (approved):
Thank you!

PR #1118, opened 2022-10-11
Drop mentions of Fedora 35

marusak:


jelly (approved):
Looks good, but it's somehow flaky!

PR #1128, opened 2022-11-03
Bump cockpit lib to dark theme support

jelly:
Include the dark theme patternfly improvements for a consistent experience.


jelly:
This now does not run the dark mode pixel tests as these cannot work yet. We test with an older `cockpit.js` which does not switch dark/light mode.

We do not set `pf-m-dark` yet to the `html` tag as this would default to dark theme for users. This would mean that users could get a white flash if they load cockpit-podman with an up to date cockpit.

Or we decide that the dark theme is so important to be consistent that we hard depend on the new cockpit 279 in cockpit-podman/machines and everything would work out of the box.

Dark theme for podman seems to have one interesting issue:

image

jelly:
To be reviewed pixel tests https://cockpit-logs.us-east-1.linodeobjects.com/pull-1128-20221104-092344-f3250d9b-fedora-36/log.html

marusak:
New pixel tests
https://github.com/cockpit-project/pixel-test-reference/compare/9cf1e0f2f5564a4ff5f5e8520d12752a9f8c9ec4..6ce99a95b0268de501aa016fede7421d613a84a3#diff-696d8727345543e6a5af258821ce22dc2b956fe5f843facc06d4224351c6cc52

marusak (approved):
nice, thank you!

PR #1138, opened 2022-11-17
Upgrade to react 18 and enable the new concurrent renderer

KKoukiou:
https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,8 @@
     "date-fns": "2.28.0",
     "docker-names": "1.2.1",
     "prop-types": "15.8.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
     "throttle-debounce": "2.3.0",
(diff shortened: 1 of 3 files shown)


jelly:
So the workflow fails as the workflow tries to re-create the node_modules commit you made by running `npm install` and comparing the created hash which fails.

As the bots also cannot build cockpit, it seems you somehow created a broken `node_modules` locally?

jelly:
Seems the functionality which keeps sets the `window.location` is broken in the React upgrade.

KKoukiou:
> Seems the functionality which keeps sets the `window.location` is broken in the React upgrade.

setState Batching changed behavior in this release. I am sure this is that. I am investigating right now.

jelly (approved):

jelly:
Rawhide fails due to the new cockpit disallowing root login, fix has been submitted https://github.com/cockpit-project/cockpit-podman/pull/1145

PR #1145, opened 2022-11-22
test: enable root login for tests

jelly:
Cockpit now disallows root login by default, to allow it in tests bump testlib and pass `enable_root_login` so the `disallowed-users` file is emptied.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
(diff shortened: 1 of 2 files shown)


jelly:
I created a mess for myself:

```
    self.sed_file('/root/d', '/etc/cockpit/disallowed-users')
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 1955, in sed_file
    m.execute("sed -i.cockpittest '{0}' {1}".format(expr, path))
  File "/work/bots/make-checkout-workdir/bots/machine/machine_core/ssh_connection.py", line 306, in execute
    res = subprocess.run(command_line,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

As the file does not exists, the test fails...

marusak (approved):
Thank you!

PR #1149, opened 2022-11-29
Checkpoint restore fixes

jelly:
Ideally we would also test the TCP connection / MAC and IP related checkboxes on restore, but from a small test:

```
cat /sys/class/net/*/address
```

The MAC address does not really change, and we can't change it with the alpineimage as it has no.

Turns out while making this PR I realised we have a busybox image which has `ip`

But reading the man page.. we need to create a container with `--mac-address` and `--ip` to test this functionality. This can be done but from my testing it seems to be broken?!

```
podman run -dit --mac-address 92:d0:c6:0a:29:38 --name fo busybox:latest sh
```

```
podman container checkpoint fo
```

```
podman container restore --ignore-static-mac fo
```

```
podman inspect --format '{{.NetworkSettings.MacAddress}}' fo
```

```
[root@fedora-37-127-0-0-2-2201 ~]# podman inspect --format '{{.NetworkSettings.MacAddress}}' fo
92:d0:c6:0a:29:38
```

So this sounds like a podman bug, and I don't see any tests for it in their repository. Reported it here https://github.com/containers/podman/issues/16666

### DIFF
--- a/src/ContainerCheckpointModal.jsx
+++ b/src/ContainerCheckpointModal.jsx
@@ -13,7 +13,6 @@ const ContainerCheckpointModal = ({ containerWillCheckpoint, onAddNotification }
     const [keep, setKeep] = useState(false);
     const [leaveRunning, setLeaveRunning] = useState(false);
     const [tcpEstablished, setTcpEstablished] = useState(false);
-    const [ignoreRootFS, setIgnoreRootFS] = useState(false);
 
     const handleCheckpointContainer = () => {
         setProgress(true);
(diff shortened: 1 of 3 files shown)


KKoukiou (dismissed):
One relevant test failure: wait_js_cond(ph_is_present(".pf-c-modal-box input#checkpoint-dialog-ignoreRootFS:not([disabled]):not([aria-disabled=true])")): Uncaught (in promise) Error: condition did not become true

Otherwise looks great. :)

KKoukiou (approved):
Met dank :)

jelly:
More testing farm issues...

```
> warning: Unhandled event type  volume
Error relocating /lib/ld-musl-x86_64.so.1: RELRO protection failed: Permission denied
Error relocating /bin/sh: RELRO protection failed: Permission denied
```

jelly:
* Rawhide issue is due to CRIU, should be fixed in bodhi / updates
* Centos-stream-9 issue is separate and needs more investigation.

As both are unrelated to the PR, merging.

PR #1155, opened 2022-12-07
Prefill command from image

### LINKED ISSUE #1126: Create Container In Pod does not respect the "command" of the image.
Cockpit version: 278
Cockpit-podman version: 55
Podman version: 4.2.1
OS: Fedora Server 36

When I click "Create container", the command defaults to `sh`.  I think it should default to the "command" option of the image you're using.

Demonstration: if you have the postgres default image installed from docker.io/library/postgres:latest, when you click Create container in the Containers section, the Command is `sh` not `postgres`.

marusak:
We prefilled always `sh` as command for new containers but this is wrong. We should always use default command from image. There are multiple changes:
1. Don't offer 'sh', just leave the command field empty
2. If user selects local image prefill the command from image
3. If user selects remote image leave the command input empty
4. If we preffiled the input in 2. and user unselects the image with (x) remove the command if it is still the same
5. If user touches the command all of the above are not applied

Fixes #1126

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -103,7 +103,7 @@ export class ImageRunModal extends React.Component {
 
     constructor(props) {
         super(props);
-        let command = "sh";
+        let command = "";
         if (this.props.image && this.props.image.Command) {
             command = utils.quote_cmdline(this.props.image.Command);
         }
@@ -463,6 +463,11 @@ export class ImageRunModal extends React.Component {
     }
 
     clearImageSelection = () => {
+        // Reset command if it was prefilled
+        let command = this.state.command;
+        if (this.state.selectedImage && this.state.selectedImage.Command && this.state.command === utils.quote_cmdline(this.state.selectedImage.Command))
+            command = "";
+
         this.setState({
             selectedImage: "",
             image: "",
@@ -470,6 +475,7 @@ export class ImageRunModal extends React.Component {
             imageResults: {},
             searchText: "",
             searchFinished: false,
+            command: command,
         });
     }
 
@@ -483,9 +489,14 @@ export class ImageRunModal extends React.Component {
         if (event === undefined)
             return;
 
+        let command = this.state.command;
+        if (value.Command && !command)
+            command = utils.quote_cmdline(value.Command);
+
         this.setState({
             selectedImage: value,
             isImageSelectOpen: false,
+            command: command,
         });
     }
 
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -8,7 +8,7 @@ import {
     NumberInput, InputGroupTextVariant,
     InputGroup, InputGroupText,
     SelectOption, SelectGroup,
-    TextInput, Tabs, Tab, TabTitleText,
+    TextInput, Tabs, Tab, TabTitleText, Text,
     ToggleGroup, ToggleGroupItem,
     Flex, FlexItem,
     Popover,
@@ -103,10 +103,14 @@ export class ImageRunModal extends React.Component {
 
     constructor(props) {
         super(props);
+
         let command = "";
         if (this.props.image && this.props.image.Command) {
             command = utils.quote_cmdline(this.props.image.Command);
         }
+
+        const entrypoint = utils.quote_cmdline(this.props.image?.Entrypoint);
+
         let selectedImage = "";
         if (this.props.image) {
             selectedImage = utils.image_name(this.props.image);
@@ -119,6 +123,7 @@ export class ImageRunModal extends React.Component {
         this.state = {
             command,
             containerName: dockerNames.getRandomName(),
+            entrypoint,
             env: [],
             hasTTY: true,
             publish: [],
@@ -465,7 +470,7 @@ export class ImageRunModal extends React.Component {
     clearImageSelection = () => {
         // Reset command if it was prefilled
         let command = this.state.command;
-        if (this.state.selectedImage && this.state.selectedImage.Command && this.state.command === utils.quote_cmdline(this.state.selectedImage.Command))
+        if (this.state.command === utils.quote_cmdline(this.state.selectedImage?.Command))
             command = "";
 
         this.setState({
@@ -476,6 +481,7 @@ export class ImageRunModal extends React.Component {
             searchText: "",
             searchFinished: false,
             command: command,
+            entrypoint: "",
         });
     }
 
@@ -493,10 +499,13 @@ export class ImageRunModal extends React.Component {
         if (value.Command && !command)
             command = utils.quote_cmdline(value.Command);
 
+        const entrypoint = utils.quote_cmdline(value?.Entrypoint);
+
         this.setState({
             selectedImage: value,
             isImageSelectOpen: false,
             command: command,
+            entrypoint: entrypoint,
         });
     }
 
@@ -739,6 +748,12 @@ export class ImageRunModal extends React.Component {
                             <Checkbox isChecked={this.state.pullLatestImage} id="run-image-dialog-pull-latest-image"
                                       onChange={value => this.onValueChanged('pullLatestImage', value)} label={_("Pull latest image")}
                             />
+                        </FormGroup>
+                        }
+
+                        {dialogValues.entrypoint &&
+                        <FormGroup fieldId='run-image-dialog-entrypoint' hasNoPaddingTop label={_("Entrypoint")}>
+                            <Text id="run-image-dialog-entrypoint">{dialogValues.entrypoint}</Text>
                         </FormGroup>
                         }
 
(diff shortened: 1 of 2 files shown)


marusak:
(running tests now to see what is broken and if there are not enough broken cases I will write some more tests)

src/ImageRunModal.jsx
```diff
@@ -463,13 +463,19 @@ export class ImageRunModal extends React.Component {
     }
 
     clearImageSelection = () => {
+        // Reset command if it was prefilled
+        let command = this.state.command;
+        if (this.state.selectedImage && this.state.selectedImage.Command && this.state.command === utils.quote_cmdline(this.state.selectedImage.Command))
+            command = "";
```
cockpituous, line 469: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20221207-145112-86329344-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/ImageRunModal.jsx
```diff
@@ -463,13 +463,19 @@ export class ImageRunModal extends React.Component {
...
+
         this.setState({
             selectedImage: "",
             image: "",
             isImageSelectOpen: false,
             imageResults: {},
             searchText: "",
             searchFinished: false,
+            command: command,
```
cockpituous, line 478: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20221207-145112-86329344-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

jelly:
PR looks simple enough, but this doesn't look correct.

Screenshot from 2022-12-08 09-40-00

Is this really the command of the image?

podman image inspect shows:
```
               "Entrypoint": [
                    "/entrypoint.sh"
               ],
               "Cmd": [
                    "/etc/docker/registry/config.yml"
               ],
```

So it's "technically" correct but hmm kinda weird.

src/ImageRunModal.jsx
```diff
@@ -103,10 +103,17 @@ export class ImageRunModal extends React.Component {
...
+
+        let command = "";
         if (this.props.image && this.props.image.Command) {
             command = utils.quote_cmdline(this.props.image.Command);
         }
+
+        let entrypoint = "";
+        if (this.props.image && this.props.image.Entrypoint) {
+            entrypoint = utils.quote_cmdline(this.props.image.Entrypoint);
+        }
```
cockpituous, line 115: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20230104-111228-5fe35f89-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/ImageRunModal.jsx
```diff
@@ -463,13 +471,20 @@ export class ImageRunModal extends React.Component {
     }
 
     clearImageSelection = () => {
+        // Reset command if it was prefilled
+        let command = this.state.command;
+        if (this.state.selectedImage && this.state.selectedImage.Command && this.state.command === utils.quote_cmdline(this.state.selectedImage.Command))
+            command = "";
```
cockpituous, line 477: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20230104-111228-5fe35f89-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/ImageRunModal.jsx
```diff
@@ -463,13 +471,20 @@ export class ImageRunModal extends React.Component {
...
+
         this.setState({
             selectedImage: "",
             image: "",
             isImageSelectOpen: false,
             imageResults: {},
             searchText: "",
             searchFinished: false,
+            command: command,
+            entrypoint: "",
```
cockpituous, line 487: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20230104-111228-5fe35f89-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/ImageRunModal.jsx
```diff
@@ -483,9 +498,19 @@ export class ImageRunModal extends React.Component {
         if (event === undefined)
             return;
 
+        let command = this.state.command;
+        if (value.Command && !command)
+            command = utils.quote_cmdline(value.Command);
+
+        let entrypoint = "";
+        if (value.Entrypoint)
+            entrypoint = utils.quote_cmdline(value.Entrypoint);
```
cockpituous, line 507: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20230104-111228-5fe35f89-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/ImageRunModal.jsx
```diff
@@ -731,6 +756,12 @@ export class ImageRunModal extends React.Component {
                         </FormGroup>
                         }
 
+                        {dialogValues.entrypoint &&
+                        <FormGroup fieldId='run-image-dialog-entrypoint' hasNoPaddingTop label={_("Entrypoint")}>
+                            {dialogValues.entrypoint}
```
cockpituous, line 761: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20230104-111228-5fe35f89-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

marusak:
@jelly I have added showing of entrypoint when present. Looks like this now:

Screenshot from 2023-01-04 12-11-00

jelly:
> @jelly I have added showing of entrypoint when present. Looks like this now:
> 
> Screenshot from 2023-01-04 12-11-00

That feels a bit strange to me, but I am not sure what we can do better here.

jelly (dismissed):
Provided solution seems like the best compromise to me, maybe @garrett still wants to have a look before merging?

garrett:
> Provided solution seems like the best compromise to me, maybe @garrett still wants to have a look before merging?

What's the provided solution? "Entrypoint"? (Using "argument" like this could sound pretty bad, like shouting in a fight between people.)

Checking out the PR now.

garrett:
> If user touches the command all of the above are not applied

Does this mean focus/blur?
Or empty versus having values?
Or original versus different value?

(Original value versus different value makes the most sense, I think, especially if it would be prefilled in some cases.)

marusak:
> Does this mean focus/blur?
> Or empty versus having values?
> Or original versus different value?

Original vs different value. If we prefill "foo" and someone edits it to "bar" we will leave it there. If we prefill "foo" and the value is still "foo" when user changes image then we will change this as well.

marusak:
> What's the provided solution? "Entrypoint"? (Using "argument" like this could sound pretty bad, like shouting in a fight between people.)

Yes, when image has entrypoint we show it. We don't enable editing it as it is not something that normally is done. You need to do `podman run --entrypoint` to overwrite the default one but normally one does not do it. What is expected is that users overwrite command.

src/ImageRunModal.jsx
martinpitt: Can you please add a test for this?

Also, this is an excellent case for using [coalescing operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), i. e. 

    if (this.state.selectedImage?.Command ..)

src/ImageRunModal.jsx
martinpitt: Same thing -- please add a test, and use `?.` if you want to.

src/ImageRunModal.jsx
martinpitt: Style -- please make this a `const` and use `? .. :`

martinpitt (changes requested):
Thanks, this is a nice fix! I'm quite surprised that we didn't do this a long time ago already, in retrospect this feels like quite a big bug..

martinpitt:
@marusak : Hmm, git failure? Your recent [two](https://github.com/cockpit-project/cockpit-podman/compare/0f64fa41b4f073724ec1f65f6b7bdbf05be29a3b..f53584b6798bb0a1d6b801eea0cde019a0853cb7) [force-pushes](https://github.com/cockpit-project/cockpit-podman/compare/b1df780acdaa4290ce9acc755808a66d942cc5da..0f64fa41b4f073724ec1f65f6b7bdbf05be29a3b) are just a bit of noise, and they don't address any of my review comments from above.

martinpitt:
@marusak : I actually did look there, and the "prefill command" does not have tests, and doesn't simplify the code (although that is optional)

marusak:
There is something weird, I just opened `Files changed (2)` and this is what I see:
Screenshot from 2023-01-16 20-53-23

marusak:
Oh I actually have this as two commits. I think they can be squashed or I have to clean them up as I pushed those fixes into second commit. The second commit also has tests for the "prefill command" part.

martinpitt:
Ah, I see -- sorry, I previously only saw the entrypoint tests. This is a bit unclean, but good enough. Thanks!

martinpitt (approved):

PR #1159, opened 2022-12-13
Keep 'Hide images' toggle even when filtering out all images

marusak:
This has two advantages - it is less jumpy when filter hides all images as well as it fixes highlighting on newly added images.


KKoukiou (approved):

KKoukiou:
Testing farm failures should be fixed here https://github.com/cockpit-project/cockpit-podman/pull/1157

PR #1161, opened 2022-12-14
Bump cockpit commit for fetching lib

KKoukiou:


jelly (dismissed):

src/index.html
```diff
@@ -15,7 +15,7 @@
 You should have received a copy of the GNU Lesser General Public License
 along with this package; If not, see <http://www.gnu.org/licenses/>.
 -->
-<html class="pf-m-dark" lang="en">
+<html class="pf-theme-dark" lang="en">
```
jelly, line 18: Hmm this does auto enable dark theme however, so that doesn't seem to do what we want.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-1161-20221214-170725-18688f11-fedora-36/log.html

I think the issue we see here is that by bumping `cockpit.js` we now dropped the dark theme switcher controller by `cockpit-podman` itself but our test image Cockpit is still the old one which does not have the new behaviour so this kinda break.... :|

KKoukiou: Right this needs pkg/shell/base_index.js to set the new theme. This file comes from cockpit-system, and we could theoretically have a dependency for a specific version of cockpit-system on cockpit-podman spec.

KKoukiou: @jelly the only approach I can think , it to block this till cockpit-282 is available in all supported distros we test this, and after we update the images make this update as well. This is pretty unfortunate though. 

Alternatively we can mimic inside this package the work that is done for changing the themes, inside this page.

jelly: I think we should maybe have kept the cockpit.dark_mode behaviour for one more release, because this is a pretty annoying situation where things half work and I keep forgetting our tests have an old cockpit :(

mvollmer (approved):

tomasmatus (approved):

PR #1162, opened 2022-12-15
fix: Unsupported style property row-gap. Did you mean rowGap?

KKoukiou:


KKoukiou:
Actually is that even needed?

garrett (approved):
I don't see any visible difference between this and main. If it's being ignored and isn't (apparently) needed, then this is perfect. :+1:

jelly:
> I don't see any visible difference between this and main. If it's being ignored and isn't (apparently) needed, then this is perfect. +1

Did you see https://cockpit-logs.us-east-1.linodeobjects.com/pull-1162-20221215-141136-7390a520-fedora-36/pixeldiff.html#TestApplication-testBasicSystem-overview-mobile one pixel test changed

garrett (dismissed):
OH! If that pixel test is to be believed (and I'm sure it is), then this is a mobile thing, I guess, and it was being applied! Whoops.

Thanks for spotting that, @jelly!

Ideally, we wouldn't use a style. But if we are, then I guess it needs to be the following?

```jsx
style={{ rowGap: "var(--pf-global--spacer--xs)" }}>
```

tomasmatus (approved):
Thanks!

PR #1165, opened 2022-12-19
Deletion dialog improvements

jelly:


src/Containers.jsx
```diff
@@ -67,7 +67,7 @@ const ContainerActions = ({ container, healthcheck, onAddNotification, version,
 
             Dialogs.show(<ForceRemoveModal name={container.Names}
                                            handleForceRemove={handleForceRemoveContainer}
-                                           reason={_("Container is currently running.")} />);
```
KKoukiou, line 70: I don't like this but its a preexisting 'string' so ok.

KKoukiou (approved):

PR #1170, opened 2023-01-04
test: Increase wait timeout for Services page

martinpitt:
The page is utterly slow, and sometimes takes longer than 15s to load on a busy machine.

---

See https://cockpit-logs.us-east-1.linodeobjects.com/pull-1169-20230103-181159-45c8d761-arch/log.html#20-2


martinpitt:
The same

    (src/bridge/cockpithttpstream.c:942):cockpit_http_stream_close: code should not be reached

as in #1169, so this is an important issue.

KKoukiou (approved):

PR #1173, opened 2023-01-06
Update pixel tests due to new fonts

marusak:
For https://github.com/cockpit-project/bots/pull/4233


martinpitt:
Meh, more uglification.. but hopefully we can revert this all after fixing https://github.com/cockpit-project/cockpit/issues/18125

PR #1175, opened 2023-01-09
Bump pkg/lib checkout to fix 'maximum depth exceeded' issue

marusak:


KKoukiou (approved):

marusak:
>it's 69aac32ebe3e23384a00321b9570f64f4ac07583 the actual commit ID, but current main also works :)

yes, but wanted to just use main to pull in as many changes as we currently have :) And it pulled in dark theme changes so needed to update pixel tests

marusak:
```
$ podman run -d -p 5000:5000 --name registry registry:2
Resolved "registry" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/registry:2...
```

That is not correct. Seems that there is a new alias and we need to better define where the registry is.

martinpitt:
podman-images.setup pulls `quay.io/cockpit/registry:2`, so supposedly we should use that.

marusak:
Looking at [screenshot](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1175-20230109-164729-1552a4db-rhel-9-1/TestApplication-testCreatePodUser-rhel-9-1-127.0.0.2-3001-FAIL.png) the pod is 'exited' but the container in it is 'running'. This is odd.

We do
```py
# Force stop, so tearDown does not hang                                 
self.execute(auth, f"podman pod stop -t0 {pod_name}")   
```
before the hanging cleanup (`podman pod rm --force -t0 --all || true`). That was introduced in 5d98049c71791b5a91be0ba2ce9b8877b24ee904 so no more context why we do this.

marusak:
Ubuntu has old podman and does not know `--timeout` option.

martinpitt (approved):
Thank you!

PR #1184, opened 2023-01-16
tests.speed++

marusak:
We see in tests often messages like:
`WARNING: Waiting for !ph_is_present("#table-pod-1") took 10.6 seconds, which is 70% of the timeout.`

In current run there are 18 of those, which means we wait for 3 minutes longer than we need to.

[test run before these changes](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20230116-102954-0f64fa41-fedora-37/log)

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -2003,6 +2003,7 @@ class TestApplication(testlib.MachineCase):
         b.click("#image-actions-dropdown")
         b.wait_visible("button:contains(Prune unused images).pf-m-disabled")
 
+    @testlib.nondestructive
     def testPruneUnusedImagesSystemSelections(self):
         ''' Test the prune unused images selection options'''
         b = self.browser
@@ -2042,6 +2043,7 @@ class TestApplication(testlib.MachineCase):
         b.click("#image-actions-dropdown")
         b.wait_visible("button:contains(Prune unused images).pf-m-disabled")
 
+    @testlib.nondestructive
     def testCreateContainerValidation(self):
         ''' Test the validation errors'''
         b = self.browser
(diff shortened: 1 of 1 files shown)


marusak:
Still happens when we stop container we created from UI. We only check there terminal, I can move this part somewhere else. [log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1184-20230116-141625-a5a11c59-fedora-37/log)

marusak:
Yesterday [fedora-37](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1155-20230116-134656-f53584b6-fedora-37/log.html) took 27 minutes, with these changes [fedora-37](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1184-20230117-084156-efed8e3e-fedora-37/log.html) takes 11 minutes! Same amount of retries (of affected tests) and skips.

marusak:
All of testing farm failed on
```
        fail: Failed to pull workdir from the guest. This usually means that login as 'root' to the guest does not work.
Failed to run command: yum install -y rsync Reason: Command returned '255'.
```

marusak:
aha `testCheckpointRestoreCGroupsV1` changes cgroups version and then reboots, so this one is not nond. And it also breaks `testLifecycleOperationsUser` as if this test runs on the same machine after `testCheckpointRestoreCGroupsV1` then we check for wrong cgroups.

marusak:
Ok, happy with this. All went green and then I force trigger all status and they again went all green. Now [fedora-37](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1184-20230117-102525-45213315-fedora-37/log.html) took only 10 minutes!

martinpitt (approved):
:racing_car: dakujem!

PR #1185, opened 2023-01-17
test: drop test extended timeout for testCreatePodUser

jelly:
In 3779bb5fc32fb80f4c1bb3 we now start containers with --stop-timeout 0, so this can be dropped.

But let's let the bot check!

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -2339,9 +2339,6 @@ class TestApplication(testlib.MachineCase):
     def testCreatePodSystem(self):
         self._createPod(True)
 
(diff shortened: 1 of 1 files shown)


marusak (approved):
We still don't have 4.0 in ubuntu but it seems to work just fine, thanks!

jelly:
> We still don't have 4.0 in ubuntu but it seems to work just fine, thanks!

Right, that's because you added `--stop-timeout 0` I think it's mostly the same behaviour as `podman rm -t 0`.

PR #1187, opened 2023-01-18
test: test ignore MAC Address on restore feature

### LINKED ISSUE #1179: Add test for container restoring with ignoring MAC address
We don't test this feature yet as it was recently broken https://github.com/containers/podman/issues/16666 (Might work on podman < 4)

Basically we need a test that:

* starts a container with a predictable MAC address using `--mac-address`
* checkpoint the container
* restore the container with the "ignore MAC address" option
* test that the MAC != previous MAC

jelly:
This was broken in 4.0 and now fixed in 4.4 which is still an release candidate. Locally this has been tested with podman-4.4.0~rc1.

Closes #1179

Locally tested with https://koji.fedoraproject.org/koji/buildinfo?buildID=2111508 and Fedora-37

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -32,6 +32,12 @@ DISTROS_WITHOUT_PODMAN_USER_RESTART = ['ubuntu-stable', 'ubuntu-2204']
 DISTROS_WITHOUT_PODMAN_POD_VOLUMES = ['ubuntu-stable', 'ubuntu-2204', 'debian-testing']
 
 
+def podman_version(cls):
+    version = cls.execute(False, "podman -v").strip().split(' ')[-1]
+    # HACK: handle possible rc versions such as 4.4.0-rc2
+    return tuple(int(v.split('-')[0]) for v in version.split('.'))
+
+
 def showImages(browser):
     if browser.attr("#containers-images button.pf-c-expandable-section__toggle", "aria-expanded") == 'false':
         browser.click("#containers-images button.pf-c-expandable-section__toggle")
@@ -1162,7 +1168,8 @@ class TestApplication(testlib.MachineCase):
             return
 
         # Run a container
-        self.execute(True, "podman run -dit --name swamped-crate --stop-timeout 0 busybox:latest sh; podman stop swamped-crate")
+        mac_address = '92:d0:c6:0a:29:38'
+        self.execute(True, f"podman run -dit --mac-address {mac_address} --name swamped-crate --stop-timeout 0 busybox:latest sh; podman stop swamped-crate")
         b.wait(lambda: self.execute(True, "podman ps --all | grep -e swamped-crate -e Exited"))
 
         # Check that the restore option is not present (i.e. start is a regular button)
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -1257,6 +1258,12 @@ class TestApplication(testlib.MachineCase):
         b.click('.pf-c-modal-box button:contains(Restore)')
         b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in 'Running')
 
+        # A new MAC address should have been generated
+        _, _, version = self.execute(False, "podman -v").split(' ')
+        # Fixed in podman 4.4.0 https://github.com/containers/podman/issues/16666
+        if int(version[0]) > 3 and int(version[2]) > 3:
```
marusak, line 1264: This is confusing syntax :) That `.` is `version[1]` maybe you can split it on `.`.

marusak:
I would wait until we actually test this somewhere so if it does not work then we would need to adjust it. Should not take too long

jelly:
> I would wait until we actually test this somewhere so if it does not work then we would need to adjust it. Should not take too long

A new podman was released last Friday, so this should be testable in rawhide today.. maybe?!

test/check-application
jelly: Looked at it again and re-worked it so it's more logical.

jelly:
rawhide has:
```
crun-1.8-1.fc38.x86_64
podman-4.4.0~rc2-2.fc38.x86_64
```

test/check-application
```diff
@@ -1257,6 +1258,12 @@ class TestApplication(testlib.MachineCase):
         b.click('.pf-c-modal-box button:contains(Restore)')
         b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in 'Running')
 
+        # A new MAC address should have been generated
+        version = self.execute(False, "podman -v").strip().split(' ')[-1]
+        # Fixed in podman 4.4.0 https://github.com/containers/podman/issues/16666
+        if int(version.split('.')[0]) >= 4 and int(version.split('.')[1]) > 3:
```
martinpitt, line 1264: This would fail for e. g. podman 5.0 or 5.2. I agree that right now a hardcoded `m.image` list is still too early, as 4.4 will most probably enter Fedora, Debian testing, arch etc. soon. I recommend to convert `version` to a tuple of ints and compare it as tuple, which works as intended. See e.g. `test/verify/check-storage-resize`:

    if self.storaged_version < [2, 7, 6]

martinpitt (changes requested):
Thanks!

jelly:
```
ValueError: invalid literal for int() with base 10: '0-rc2'
```
Arggh ofcourse!

test/check-application
```diff
@@ -32,6 +32,12 @@ DISTROS_WITHOUT_PODMAN_USER_RESTART = ['ubuntu-stable', 'ubuntu-2204']
 DISTROS_WITHOUT_PODMAN_POD_VOLUMES = ['ubuntu-stable', 'ubuntu-2204', 'debian-testing']
 
 
+def podman_version(cls):
+    version = cls.execute(False, "podman -v").strip().split(' ')[-1]
+    # HACK: handle possible rc versions such as 4.4.0-rc2
+    return tuple(int(v.split('-')[0]) for v in version.split('.'))
```
martinpitt, line 38: If only we had a more machine readable API for determining the version. One should invent an RPC protocol over an Unix socket, and call it "data bus" or something such :thinking: 

(snarky comment aside -- this is ugly, but effective, thanks!)

test/check-application
```diff
@@ -1201,6 +1208,13 @@ class TestApplication(testlib.MachineCase):
         b.click('.pf-c-modal-box button:contains(Restore)')
         b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in 'Running')
 
+        # A new MAC address should have been generated
+        # Fixed in podman 4.4.0 https://github.com/containers/podman/issues/16666
+        import json
```
martinpitt, line 1213: Let's please keep imports at the top.

test/check-application
```diff
@@ -1201,6 +1208,13 @@ class TestApplication(testlib.MachineCase):
...
+        print(json.dumps(podman_version(self)))
```
martinpitt, line 1214: Debugging leftover. But to be sure that the version comparison works, you could do an `else:` below and check that the MAC address is equal.

martinpitt (changes requested):
Thanks, almost there!

test/check-application
jelly: I actually think that's a neat idea! Let's test both cases.

martinpitt (approved):
Danjkewel!

PR #1188, opened 2023-01-20
test: bump reference-image to fedora-37

jelly:
The tasks container is Fedora 37 already.

### DIFF
--- a/test/reference-image
+++ b/test/reference-image
@@ -1,1 +1,1 @@
-fedora-36
+fedora-37


martinpitt:
> The tasks container is Fedora 37 already.

This is completely irrelevant. This just controls which test context actually does pixel tests, i.e. from which OS we actually took the reference images. This could technically just as well be Arch or Debian.

martinpitt (approved):
Thanks! I'll land this with editing the confusing commit message.

PR #1194, opened 2023-01-30
build: add support for /pybridge scenario

allisonkarlitskaya:
Build the Python bridge from the cockpit project and install it into the VM so we can run tests against it.

 - [x] cockpit-project/cockpit#18265


martinpitt (approved):
Nice!

PR #1198, opened 2023-02-02
test: Drop explicit cgroupsv1/v2 tests, change default TEST_OS to fedora-37

martinpitt:
That transition has happened several years ago. Let's stop supporting
switching to cgroupsv1 on Fedora, nobody else tests this any more
anyway. Let's just test whatever the OS defaults to. 

This was the only destructive test, and didn't even do anything useful
on all OSes except Fedora. Move `@nondestructive` to the class.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PACKAGE_NAME := $(shell awk '/"name":/ {gsub(/[",]/, "", $$2); print $$2}' packa
 RPM_NAME := cockpit-$(PACKAGE_NAME)
 VERSION := $(shell T=$$(git describe 2>/dev/null) || T=1; echo $$T | tr '-' '.')
 ifeq ($(TEST_OS),)
-TEST_OS = fedora-36
+TEST_OS = fedora-37
 endif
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -3,7 +3,7 @@ PACKAGE_NAME := $(shell awk '/"name":/ {gsub(/[",]/, "", $$2); print $$2}' packa
 RPM_NAME := cockpit-$(PACKAGE_NAME)
 VERSION := $(shell T=$$(git describe 2>/dev/null) || T=1; echo $$T | tr '-' '.')
 ifeq ($(TEST_OS),)
-TEST_OS = fedora-36
+TEST_OS = fedora-37
```
jelly, line 6: `HACKING.md` still mentions `fedora-36` btw.

jelly (dismissed):

Makefile
martinpitt: Thanks for spotting! Fixed.

jelly (approved):

PR #1203, opened 2023-02-14
packit: add Fedora 38 testing and release

jelly:


packit.yaml
```diff
@@ -21,6 +21,7 @@ jobs:
     targets:
       - fedora-36
```
martinpitt, line 22: Let's please drop fedora-36. We have traditionally only supported two parallel releases, there's too much overhead otherwise.

martinpitt (changes requested):
I'll trigger f-38 to see how it works out.

packit.yaml
jelly: Ok, I checked and Fedora 36 is supported till F 35. So do we stop releasing and testing or only testing?

jelly:
> I'll trigger f-38 to see how it works out.

But we have no Fedora 38 image yet? I can't build one yet as the cloud image is not yet out. 

https://download.fedoraproject.org/pub/fedora/linux/releases/38/Cloud/x86_64/images/

packit.yaml
martinpitt: We stop both. (Logically, stop releasing first and then testing, but we can stop both at the same time in packit)

martinpitt:
> But we have no Fedora 38 image yet? I

Ah yes, of course :see_no_evil:

jelly:
F38 succeeded here, so re-pushed with the dropping of F36 https://artifacts.dev.testing-farm.io/3eeaf539-248b-457b-9e6c-62b8c4654e43/

martinpitt (approved):
Nice!

PR #1205, opened 2023-02-15
test fedora-38

jelly:


martinpitt:
See https://github.com/cockpit-project/bots/pull/4411 -- already works, doesn't need changes. So I think this can be closed?

PR #1207, opened 2023-02-16
Add Github code scanning workflow using codeql

jelly:
There are still two issues left, but those require more work to be fixed so let's do that in a follow up. For now this provides some nice code issues so it's worth to already have it.

### DIFF
--- a/src/ContainerRenameModal.jsx
+++ b/src/ContainerRenameModal.jsx
@@ -58,7 +58,7 @@ const ContainerRenameModal = ({ container, version, updateContainerAfterEvent })
     const handleKeyPress = (event) => {
         if (event.key === "Enter") {
             event.preventDefault();
-            handleRename(name);
+            handleRename();
         }
     };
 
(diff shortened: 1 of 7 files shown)


src/ContainerRenameModal.jsx
```diff
@@ -58,7 +58,7 @@ const ContainerRenameModal = ({ container, version, updateContainerAfterEvent })
     const handleKeyPress = (event) => {
         if (event.key === "Enter") {
             event.preventDefault();
-            handleRename(name);
+            handleRename();
```
cockpituous, line 61: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1207-20230216-111025-7f74f495-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt (approved):
Nice, thank you! So the CodeQL workflow won't always be red now, it just complains about the delta, right?

jelly:
> Nice, thank you! So the CodeQL workflow won't always be red now, it just complains about the delta, right?

Hm that is a good question, we'll found that out soon enough I suppose.

PR #1209, opened 2023-02-17
build: Move to a webpack module, generalize bundler language/variable names

martinpitt:
Cockpit recently changed to an ESM build system [1]. Bump COCKPIT_REPO_COMMIT to that and follow suit.

This does not work with old node.js 10 any more which is still the default in RHEL 8. Install the newer version 16 instead.

[1] https://github.com/cockpit-project/cockpit/pull/18366

----

Pretty much identical to https://github.com/cockpit-project/starter-kit/pull/625 except for the node_modules submodule refresh.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
-COCKPIT_REPO_COMMIT = 173bb62c36044f23092fbe1d30f6b21177a74980 # 285 + 173bb62c36044f23092fbe1d30f6b21177a74980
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -13,8 +13,8 @@ APPSTREAMFILE=org.cockpit-project.$(PACKAGE_NAME).metainfo.xml
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 # stamp file to check for node_modules/
 NODE_MODULES_TEST=package-lock.json
-# one example file in dist/ from webpack to check if that already ran
-WEBPACK_TEST=dist/manifest.json
+# one example file in dist/ from bundler to check if that already ran
```
KKoukiou, line 16: I think the whole docs need an article where you refer to 'the bundler', but that's not really important.

KKoukiou (approved):

PR #1211, opened 2023-02-20
Update cockpit repo to hybrid webpack/esbuild plugins

martinpitt:
The API of the plugins changed slightly to also support esbuild.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ COCKPIT_REPO_FILES = \
 	$(NULL)
 
 COCKPIT_REPO_URL = https://github.com/cockpit-project/cockpit.git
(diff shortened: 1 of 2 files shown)


jelly (approved):

jelly:
We don't test translations in podman, so that approval was too soon :)

KKoukiou (approved):

martinpitt:
@jelly: I diffed the whole dist/ between main and this branch, and the only difference is
```diff
--- dist.main/po.cs.js
+++ dist/po.cs.js
@@ -1048,4 +1048,4 @@
   null,
   "uživatel:"
  ]
-});
+})
```
I interactively tested podman in German, and translations work fine.

PR #1212, opened 2023-02-21
Use whole path in patternfly imports

KKoukiou:
Otherwise we rely on treeshaking mechanisms to remove unused code, thus making our bundle size more vulnerable to bundler's ability to drop dead code.

### DIFF
--- a/src/ContainerCheckpointModal.jsx
+++ b/src/ContainerCheckpointModal.jsx
@@ -1,5 +1,8 @@
 import React, { useState } from 'react';
-import { Button, Checkbox, Form, Modal } from '@patternfly/react-core';
+import { Button } from "@patternfly/react-core/dist/esm/components/Button";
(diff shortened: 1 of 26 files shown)


martinpitt:
Is this preparation, i.e. necessary, but not sufficient? And we need a newer cockpit-components-table as well? Then I'm happy to +1

KKoukiou:
> Is this preparation, i.e. necessary, but not sufficient? And we need a newer cockpit-components-table as well? Then I'm happy to +1

Correct. This is just preparation. we need https://github.com/patternfly/patternfly-react/pull/8708 to get merged and update to the new to be released react-table.

martinpitt (approved):
Thanks. Well, then let's get used to that pattern.

Incidentally, it's how [PF elements](https://patternflyelements.org/get-started/) works as well, each component is a separate import.

PR #1218, opened 2023-03-01
Update README.md

GOUKOU007:
Intorduce the way to handle missing msgfmt in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,10 @@ make
 `sudo make install` installs the package in `/usr/share/cockpit/`. This depends
 on the `dist` target, which generates the distribution tarball.
 
+> If a `msgfmt` not found error occurs during `make install`, you may need to
+> install the `gettext` package then `make install` again. e.g. `apt install 
+> gettext` or `yum install gettext`.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,6 +26,10 @@ make
 `sudo make install` installs the package in `/usr/share/cockpit/`. This depends
 on the `dist` target, which generates the distribution tarball.
 
+> If a `msgfmt` not found error occurs during `make install`, you may need to
+> install the `gettext` package then `make install` again. e.g. `apt install 
+> gettext` or `yum install gettext`.
```
jelly, line 31: We indeed do not document the build requirements like we do in cockpit's `HACKING.md`, it's probably better to to link to the Cockpit docs instead which is kept up to date.

For our documentation we strive to write it towards for settings a development setup install XYZ instead of documenting every possible error case.

jelly (changes requested):

jelly:
This [PR](https://github.com/cockpit-project/cockpit-podman/pull/1224) superseedes this one.

PR #1219, opened 2023-03-02
packit: Run PR tests on CentOS 8 stream

martinpitt:


jelly (approved):
\# TESTS PASSED [1357s on ip-172-31-29-219, 0 parallel tests, 32 serial tests: ]

Looks good!

PR #1226, opened 2023-03-15
Assorted fixes

jelly:


src/PodActions.jsx
```diff
@@ -120,7 +120,7 @@ export const PodActions = ({ onAddNotification, pod }) => {
             </DropdownItem>,
         );
     }
-    if (pod.Status == "Created" || pod.Status == "Exited" || pod.Status == "Stopped") {
+    if (pod.Containers.length !== 0 && (pod.Status == "Created" || pod.Status == "Exited" || pod.Status == "Stopped")) {
```
martinpitt, line 123: Hmm .. I almost prefer the status quo, as that explains *why*  the container can't be started. Now it's a bit of a mystery. Could the explanation be somewhere on the page?

jelly: Hmmmm good point it's rather opaque if there is no infra container

Issue #1067, opened 2022-08-10
Explain "system" vs. "user session"


garrett:
We should probably explain the difference between a system container versus running in a user session, in a popover connected to a (?) icon.

It should probably simply state the benefits and drawbacks of each approach too.

(This is the same as https://github.com/cockpit-project/cockpit-machines/issues/792, but for Podman containers.)

PR #1227, opened 2023-03-15
Explain the difference between a user and system session

jelly:
When creating a container with the system session offer a popover with the differences between the owner types.

Closes: #1067

---

image

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -11,7 +11,7 @@ import { NumberInput } from "@patternfly/react-core/dist/esm/components/NumberIn
 import { InputGroup, InputGroupText, InputGroupTextVariant } from "@patternfly/react-core/dist/esm/components/InputGroup";
 import { TextInput } from "@patternfly/react-core/dist/esm/components/TextInput";
 import { Tab, TabTitleText, Tabs } from "@patternfly/react-core/dist/esm/components/Tabs";
-import { Text } from "@patternfly/react-core/dist/esm/components/Text";
+import { Text, TextContent, TextList, TextListItem, TextVariants } from "@patternfly/react-core/dist/esm/components/Text";
 import { ToggleGroup, ToggleGroupItem } from "@patternfly/react-core/dist/esm/components/ToggleGroup";
 import { Flex, FlexItem } from "@patternfly/react-core/dist/esm/layouts/Flex";
 import { Popover } from "@patternfly/react-core/dist/esm/components/Popover";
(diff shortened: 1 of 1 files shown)


src/ImageRunModal.jsx
```diff
@@ -687,7 +687,44 @@ export class ImageRunModal extends React.Component {
...
+                                                              {_("Ports under 1024 can be forwarded")}
+                                                          </TextListItem>
+                                                      </TextList>
+                                                  </TextContent>
+                                                  <TextContent>
+                                                      <Text component={TextVariants.h4}>{cockpit.format("$0 $1", _("User:"), this.props.user)}</Text>
+                                                      <TextList>
+                                                          <TextListItem>
+                                                              {_("Ideal for development")}
+                                                          </TextListItem>
```
jelly, line 718: @garrett I can't think of more points for the user session, can you?

src/ImageRunModal.jsx
```diff
@@ -687,7 +687,44 @@ export class ImageRunModal extends React.Component {
...
+                                                          <TextListItem>
+                                                              {_("Resource limits can be set")}
+                                                          </TextListItem>
+                                                          <TextListItem>
+                                                              {_("Checkpoint/restore support")}
+                                                          </TextListItem>
+                                                          <TextListItem>
+                                                              {_("Ports under 1024 can be forwarded")}
+                                                          </TextListItem>
```
jelly, line 710: IIRC there are more network related differences but I couldn't find them? IIRC host networking but we don't offer that.

src/ImageRunModal.jsx
```diff
@@ -687,7 +687,44 @@ export class ImageRunModal extends React.Component {
...
+                                                              {_("Ideal for running services")}
+                                                          </TextListItem>
+                                                          <TextListItem>
+                                                              {_("Resource limits can be set")}
+                                                          </TextListItem>
+                                                          <TextListItem>
+                                                              {_("Checkpoint/restore support")}
+                                                          </TextListItem>
+                                                          <TextListItem>
+                                                              {_("Ports under 1024 can be forwarded")}
```
garrett, line 709: Can't you use ports under 1024 inside the container and forward to ports 1024 outside? This sounds like the port _inside_ the container must be >= 1024.

Perhaps:

```suggestion
                                                              {_("Ports under 1024 can be mapped")}
```

src/ImageRunModal.jsx
garrett: Looks good. I think these points work well.

Are there any network considerations besides the ports?

Out of scope, but interesting and distantly related: 

(The low ports is such is an absurd default policy in this day and age, but that's a different topic and way out of scope — and I think the policy can even be changed manually, but we might even hardcode that somewhere, either in Podman or Cockpit-Podman).

And, someone could remap the port using firewall rules too. (This would actually be a neat feature for Cockpit to handle, probably in the Firewall page, and it would be awesome if we had integration for both Cockpit-Podman and Cockpit-Machines.) But, again, this is out of scope from this PR.

src/ImageRunModal.jsx
jelly: Yes, it's only on the outside so podman run -p 22:22 only works as root.

jelly:
Updated the screenshot to reflect the review comments.

garrett:
For rootless containers, we may want to also include something from https://developers.redhat.com/blog/2020/09/25/rootless-containers-with-podman-the-basics#why_rootless_containers_:

Something like this?

"Uses a standard built-in user-based security layer"

(That might be too wordy.)

jelly:
> For rootless containers, we may want to also include something from https://developers.redhat.com/blog/2020/09/25/rootless-containers-with-podman-the-basics#why_rootless_containers_:
> 
> Something like this?
> 
> "Uses the standard user-based security layer"

That sounds a bit vague to me, I changed it to "Running containers as root as an unprivileged user"

garrett:
How about "Restricted by user account permissions"? That's both a pro and con.

garrett (dismissed):

src/ImageRunModal.jsx
martinpitt: They are also ideal (or sometimes even required, with e.g. encrypted home dirs) for working with your $HOME data, like handing over your ~/Photos dir as a volume to your image gallery container.

garrett (approved):

src/ImageRunModal.jsx
garrett: It's fine. We have a lot of the most common differences shown already. :+1:

PR #1229, opened 2023-03-16
Bump sass to fix performance regression

KKoukiou:
https://github.com/sass/dart-sass/releases/tag/1.59.3


martinpitt (approved):
Yay! :racehorse:

martinpitt:
That's more like it!
```
❱❱❱ time node build.js 

real	0m6.074s
```

It was around 90 seconds before..

PR #1230, opened 2023-03-16
fix: show dialog error at the top

### LINKED ISSUE #1221: Show dialog errors on top
This is a follow up from the cockpit PR but for podman https://github.com/cockpit-project/cockpit/pull/17221

When we have erorrs in a dialog which can't be showed next to a specific Form element using PF's input validation, we use an `ErrorNotification` as shown below.

image

These errors should however been shown on top.

subhoghoshX:
Fixes #1221

Moved dialog errors at the top in these components.
1. ImageRunModal
2. PodCreateModal
3. ImageSearchModal
4. ContainerCommitModal

### DIFF
--- a/src/ContainerCommitModal.jsx
+++ b/src/ContainerCommitModal.jsx
@@ -3,7 +3,6 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { Checkbox } from "@patternfly/react-core/dist/esm/components/Checkbox";
 import { Form, FormGroup } from "@patternfly/react-core/dist/esm/components/Form";
 import { Modal } from "@patternfly/react-core/dist/esm/components/Modal";
-import { Stack, StackItem } from "@patternfly/react-core/dist/esm/layouts/Stack";
 import { TextInput } from "@patternfly/react-core/dist/esm/components/TextInput";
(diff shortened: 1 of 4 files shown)


src/ContainerCommitModal.jsx
```diff
@@ -158,7 +158,12 @@ const ContainerCommitModal = ({ container, localImages }) => {
                      </Button>
                  </>}
         >
-            {commitContent}
+            <Stack hasGutter>
```
KKoukiou, line 161: You can achieve the same visual result, by putting the dialogError inside the Form; I prefer it as it does not require this extra component. 

```diff
         <Form isHorizontal>
+            {dialogError && <ErrorNotification errorMessage={dialogError} errorDetail={dialogErrorDetail} onDismiss={() => setDialogError("")} />}
```

KKoukiou (changes requested):
This works perfect. I prefer if we follow the error in Form approach for code better readability though.

src/ContainerCommitModal.jsx
subhoghoshX: Thanks. Moved it inside Form.

KKoukiou (approved):

PR #1232, opened 2023-03-17
Make containers table sortable

jelly:
Allow the sorting of the container name, owner, cpu, memory and state. This makes containers sortable and leaves pods unsorted but containers in pods can be sorted. Allowing pods to be sorted with containers would be undesired.

---

# Sort containers

The podman container section now allows a user to sort on name, owner, CPU, memory and state.

image

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -11,7 +11,7 @@ import { Text, TextVariants } from "@patternfly/react-core/dist/esm/components/T
 import { FormSelect, FormSelectOption } from "@patternfly/react-core/dist/esm/components/FormSelect";
 import { Tooltip } from "@patternfly/react-core/dist/esm/components/Tooltip";
 import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core/dist/esm/components/Toolbar";
-import { cellWidth } from '@patternfly/react-table';
+import { cellWidth, SortByDirection } from '@patternfly/react-table';
 import { MicrochipIcon, MemoryIcon, PortIcon, VolumeIcon, } from '@patternfly/react-icons';
 
 import cockpit from 'cockpit';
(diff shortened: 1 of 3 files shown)


garrett (dismissed):
This is straightforward, is what I'd expect, uses PF effectively, and looks great! :+1:

garrett:
Oh, I did realize one thing:

- When you click on CPU and memory, it's sorted by smallest to largest, when people would almost always want to see largest (ones using resources the most) sorted first.
- Same for state; that should be sorted where running is first and stopped comes later.

garrett (changes requested):
Mentioned above: While this is great, please change the sorting so it works as expected with one click, instead of always having to click most of the headings twice. Thanks!

garrett:
Are the upper bounds limits or reserved (for both memory and disk)? It's limits, right? If so, then we should use the actual used memory. (If they're reserved, then that's the actual use on the system, so then that should be sorted first and then the used inside of reserved second. But I don't think that's the case. Not sure though. :shrug:)

jelly:
> Are the upper bounds limits or reserved (for both memory and disk)? It's limits, right? If so, then we should use the actual used memory. (If they're reserved, then that's the actual use on the system, so then that should be sorted first and then the used inside of reserved second. But I don't think that's the case. Not sure though. shrug)

We sort based on the metrics podman provides and that is a the MemoryUsage:

```
ID            NAME              CPU %       MEM USAGE / LIMIT  MEM %       NET IO             BLOCK IO           PIDS        CPU TIME    AVG CPU %
0090559af02b  lucid_maxwell     0.00%       0B / 0B            0.00%       0B / 0B            0B / 0B            0           0s          0.00%
21a0460c06b3  hopeful_payne     0.00%       53.25kB / 1.141GB  0.00%       5.244kB / 1.076kB  0B / 0B            1           2.99ms      0.00%
5269e9f5e848  practical_bohr    0.00%       57.34kB / 512MB    0.01%       1.556kB / 1.006kB  28.67kB / 0B       1           3.556ms     0.00%
a6935eba61ff  suspicious_kalam  0.00%       0B / 0B            0.00%       0B / 0B            0B / 0B            0           0s          0.00%
b41c2c8945a2  focused_feistel   0.00%       0B / 0B            0.00%       0B / 0B            0B / 0B            0           0s          0.00%
d61b109a774c  epic_matsumoto    0.05%       31.24MB / 1.141GB  2.74%       2.912kB / 1.006kB  24.58kB / 114.9MB  13          1.060304s   0.11%
```

So Grafana has the highest, then the memory limited one and then `hopeful_payne`. How we show it is, is a bit confusing as we take the memory limit into account it seems.

image

garrett:
So a problem is that they're not the same units. It's hard to compare across different units.

jelly:
> So a problem is that they're not the same units. It's hard to compare across different units.

Yes, so I don't think we can do much better then it is now.

garrett:
Can we make them all GB or all MB for now? (What's more realistic for containers on a server really?) I know servers usually have lots of RAM — like 16, 32, 64, 128 GB... but the services running inside containers: how much do they typically have?

tomasmatus (approved):

jelly:
> Mentioned above: While this is great, please change the sorting so it works as expected with one click, instead of always having to click most of the headings twice. Thanks!

I can't reproduce this, can you recheck?

garrett (approved):
:+1: It looks good now! Thanks!

PR #1233, opened 2023-03-17
Allow custom healthcheck actions to be configured

jelly:
When a containers healthcheck fails a user can now configure what specific action needs to happen for example, kill or restart. Podman's rest api uses an int for the health check options so we carry our own object with the options

* [ ] This needs a version check as this feature has been added in 4.3
* [ ] Also fix the helpertext... :)


src/ImageRunModal.jsx
```diff
@@ -1054,6 +1064,27 @@ export class ImageRunModal extends React.Component {
                                     onPlus={() => this.onPlusOne('healthcheck_retries')}
                                     onChange={ev => this.onValueChanged('healthcheck_retries', parseInt(ev.target.value) < 0 ? 0 : ev.target.value)} />
                         </FormGroup>
+                        {this.props.version.localeCompare("4.3", undefined, { numeric: true, sensitivity: 'base' }) >= 0 &&
```
martinpitt, line 1067: Locale specific version comparison sounds weird/dangerous. It seems better to split on `.` and compare them as lists, which seems to work very nicely. (e.g. `[4,2] > [4,3]`). But if `numeric` DTRT here, fine.

test/check-application
```diff
@@ -2056,6 +2056,9 @@ class TestApplication(testlib.MachineCase):
         b.wait_val("#run-image-healthcheck-retries input", 2)
         if auth:
             b.assert_pixels('.pf-c-modal-box', "healthcheck-modal", skip_layouts=["rtl"])
+        # Test that the healtcheck option is not available before podman 4.3
```
martinpitt, line 2059: healt"h"

test/check-application
```diff
@@ -2111,6 +2114,39 @@ class TestApplication(testlib.MachineCase):
...
+        b.set_input_text('#run-image-dialog-healthcheck-command', '/bin/test -f /healthy')
+        b.set_input_text('#run-image-healthcheck-interval input', '1')
+        b.set_input_text('#run-image-healthcheck-timeout input', '1')
+        b.select_from_dropdown("#run-image-healthcheck-action", 2)  # Kil
+        b.click('.pf-c-modal-box__footer #create-image-create-run-btn')
+
+        self.waitContainerRow(containername)
+        self.toggleExpandedContainer(containername)
+        b.wait(lambda: self.getContainerAttr(containername, "State") == "RunningHealthy")
+        # Removing the file should kill the container
```
martinpitt, line 2144: kil"l"

martinpitt (commented):
Thanks! This looks good enough to me, I just have some questions and suggestions.

@garrett should also look at this for the string changes, but he already has a review request.

src/ImageRunModal.jsx
```diff
@@ -52,6 +52,14 @@ const units = {
     },
 };
 
+// healthchecks.go HealthCheckOnFailureAction
+const HealthCheckOnFailureAction = {
+    0: _("None"),
```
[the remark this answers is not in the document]
jelly: This is the default option for a full reference from `man podman-run`:

              • none: Take no action.

              • kill: Kill the container.

              • restart: Restart the container.  Do not combine the restart action with the --restart flag.  When running inside of a systemd unit, consider using  the
                kill or stop action instead to make use of systemd's restart policy.

              • stop: Stop the container.

src/ImageRunModal.jsx
jelly: Right, that would require fixing of.. multiple occurrences:

```
[jelle@t14s][~/projects/cockpit-podman]%git grep localeCompare | wc -l
6
```

I'd prefer to do this in a follow up.

test/check-application
jelly: I don't see a typo here :eyes:

src/ImageRunModal.jsx
```diff
@@ -52,6 +52,14 @@ const units = {
...
+    2: _("Kill"),
+    3: _("Restart"),
+    4: _("Stop"),
```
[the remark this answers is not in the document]
jelly: Maybe we should show the text above ^?

src/ImageRunModal.jsx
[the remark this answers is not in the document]
martinpitt: Doesn't explain the use case, but hmmkay

test/check-application
```diff
@@ -2111,6 +2114,39 @@ class TestApplication(testlib.MachineCase):
...
+
+        containername = "healthaction"
+        b.set_input_text("#run-image-dialog-name", containername)
+        b.set_input_text("#run-image-dialog-command", "/bin/sh -c 'echo 1 > /healthy && sleep infinity'")
+
+        b.click("#pf-tab-2-create-image-dialog-tab-healthcheck")
+        b.set_input_text('#run-image-dialog-healthcheck-command', '/bin/test -f /healthy')
+        b.set_input_text('#run-image-healthcheck-interval input', '1')
+        b.set_input_text('#run-image-healthcheck-timeout input', '1')
+        b.select_from_dropdown("#run-image-healthcheck-action", 2)  # Kil
```
martinpitt, line 2138: *This* is the kil"l" typo :grin:

src/ImageRunModal.jsx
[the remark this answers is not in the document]
martinpitt: That doesn't explain anything else, just appends "the container" (which is implied)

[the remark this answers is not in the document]
martinpitt: I honestly don't know what the difference is -- my hunch would be "stop == SIGTERM pid 1" and "kill == SIGKILL all processes in the container". Relying on an orderly shutdown if a health check fails (and processes are likely broken) seems a bit useless. My point is, we should explain when to use a situation. Same for "None" -- if that's the default, it seems totally pointless.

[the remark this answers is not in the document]
jelly: Seems my comment got lost now:

              • none: Take no action.

              • kill: Kill the container.

              • restart: Restart the container.  Do not combine the restart action with the --restart flag.  When running inside of a systemd unit, consider using  the
                kill or stop action instead to make use of systemd's restart policy.

              • stop: Stop the container.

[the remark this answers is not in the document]
martinpitt: Not lost, just hidden behind resolved". :grin: But that's what my [reply above](https://github.com/cockpit-project/cockpit-podman/pull/1233#discussion_r1150439031) refered to.

martinpitt:
> Additionally, the order seems wrong. Shouldn't it be "None", "Restart", "Stop", "Kill"... going from likely and softer actions to less likely and more severe? 

I don't have that much experience with it, but IMHO with broken containers where the watchdog fires, the more "likely" (sensible) action are the "more severe" ones. For example, "None" is not even an action, and watchdogs are a no-op with it.

src/ImageRunModal.jsx
jelly: So, what about we also stop using passing version around everywhere because that is super annoying. Maybe we should use a context provider or useContext? Because we pass a *ton* of properties around.

garrett (approved):
:+1: LGTM! Thanks for making the changes!

src/ImageRunModal.jsx
```diff
@@ -157,6 +165,7 @@ export class ImageRunModal extends React.Component {
             healthcheck_timeout: 30,
             healthcheck_start_period: 0,
             healthcheck_retries: 3,
+            healthcheck_action: 0,
```
martinpitt, line 168: I still object to this default -- "No action" is completely pointless for a watchdog. "Force stop" may be a better default. What is the default on main where users can't configure the action type?

src/ImageRunModal.jsx
```diff
@@ -52,6 +52,14 @@ const units = {
     },
 };
 
+// healthchecks.go HealthCheckOnFailureAction
+const HealthCheckOnFailureActionOrder = [
+    { value: 0, label: _("No action") },
+    { value: 3, label: _("Restart") },
+    { value: 4, label: _("Stop") },
+    { value: 2, label: _("Force stop") },
```
martinpitt, line 60: This looks weird -- what' is value 1? And can we keep it sorted numerically? It looks like something went wrong here.

martinpitt (changes requested):

src/ImageRunModal.jsx
jelly: This is the default of podman itself, should we really override that behaviour. It doesn't sound correct and it would be a breaking change.

https://docs.podman.io/en/latest/markdown/podman-run.1.html#health-on-failure-action

src/ImageRunModal.jsx
jelly: This is just wrong now, we can keep it numerically sorted but then need something else to sort it correctly for the UI.

martinpitt: Oh, that's the order in which we show it in the UI, I see. Keep it then, yes.

src/ImageRunModal.jsx
martinpitt: Hm, this really leaves us between a rock and a hard place. It makes sense to keep the UI compatible with the CLI. But it does tell me that our current feature in `main` is broken: you can configure a health check command, and then it is a complete no-op? Likewise with the CLI, if you do a `--health-cmd`, it goes entirely ignored? That is at least worth a podman bug report.

I have no strong opinion between defaulting to "kill" (and configuring a container to auto-spawn, like on k8s), or "restart" (which may make more sense without wrapping in systemd).

martinpitt: So in the interest of not blocking this PR, we can do this as a follow-up. But I am really not happy about this..

PR #1239, opened 2023-03-22
Make inserting a value with an equals not always split in to key/value

### LINKED ISSUE #1090: cannot enter equals sign (=) into the value field of an environment variable

jelly:
After the introduction of the multi insert feature, users could no longer insert a value with an equals sign.

Closes #1090


jelly:
> Looks good, but lots of test failures. Some might be flakes?

Yes, podman has been flaking a lot in my PR's lately. I'll book some time tomorrow or Friday to investigate the top flakes.

jelly:
Reqruires a pixel test update

martinpitt (approved):
Dankjewel!

PR #1240, opened 2023-03-23
Fix dd flake and improve 

jelly:
This should fix one of the common flakes in the recent PR's.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -15,8 +15,6 @@ sys.path.append(os.path.join(os.path.dirname(TEST_DIR), "bots/machine"))
 from machine_core import ssh_connection  # noqa
 import testlib  # noqa
 
-from lib.constants import OSTREE_IMAGES # noqa
-
 REGISTRIES_CONF = """
 [registries.search]
 registries = ['localhost:5000', 'localhost:6000']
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -1084,8 +1082,9 @@ class TestApplication(testlib.MachineCase):
             self.assertIn(full[1], ["GB", "MB"])
 
             # Test that the value is updated dynamically
-            self.execute(auth, "podman exec -i swamped-crate sh -c 'dd bs=1024 count=1000000 < /dev/urandom > /dev/null'")
+            self.execute(auth, "podman exec -di swamped-crate sh -c 'dd bs=1024 < /dev/urandom > /dev/null'")
```
martinpitt, line 1085: Hmm, that means the `podman exec` command itself will stay around for that long? Does that block the `restart`, or is that fast? I'd feel a bit easier if this would use `&` instead of `-d`, and `pkill dd` instead of `podman restart`.

There are lots of flakes, perhaps that would help a bit?

jelly: There are a ton of unrelated flakes, I don't see how they relate to `podman exec`. But maybe a kill is better as a restart is slow.

martinpitt:
At first sight, the failures are unrelated and random, but there are some patterns. But this only touches testPods, and that now succeeded everywhere except [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1240-20230323-123049-cf8a13e1-rhel-9-2/log.html#25-1). But that's again a different cause.

So I'd say, let's land this (after some retries to not land red), and look at the other flakes separately? Thanks for hunting these down!

PR #1245, opened 2023-03-28
Reuse esbuild configuration with main cockpit project

KKoukiou:
This introduces the override for breakpoints, therefore pixel test updates are required.


martinpitt (approved):
Nice!

PR #1247, opened 2023-03-28
fmf: Split into three plans

martinpitt:
A single run already takes half an hour if all tests pass, and a full hour if there are many failures. That's massively longer than the < 10 mins that they take on our upstream CI.

Split the tests into three parts and run them in parallel plans.

----

Comparison with recent single-plan runs: [c8s](https://artifacts.dev.testing-farm.io/7590103b-29d0-4deb-98c3-69f3f2a5010e/) took 29 min ; [F38](https://artifacts.dev.testing-farm.io/9584be6a-c8be-4063-ac93-85675fefe8c4/) took 27 min (pipeline.log) The tests here take between 16 and 21 mins, so about 33% to 50% less.

### DIFF
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -1,6 +1,19 @@
-summary:
-    Run all tests
 discover:
     how: fmf
 execute:
     how: tmt
+
+/system:
+    summary: Run tests on system podman
+    discover+:
+        test: /test/browser/system
+
+/user:
+    summary: Run tests on user podman
+    discover+:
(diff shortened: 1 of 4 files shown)


martinpitt:
[failure to pull from quay](https://artifacts.dev.testing-farm.io/54c6d825-6142-418b-9f1f-99a0bfbd85ef/), perhaps we should add a little retry loop. But unrelated → followup

KKoukiou (approved):

PR #1250, opened 2023-03-29
test: correctly restore root login on ostree

jelly:
The third parameter of write_file is append but instead we want to invoke a post restore action.


martinpitt (approved):
Oops! Are the failures all flakes?

jelly:
> Oops! Are the failures all flakes?

Yes, we have a ton of flakes... which I can't seem to reproduce locally.

PR #1241, opened 2023-03-23
test: debug check if pods are really not started yet

jelly:
We should probably do:

```
             if not isinstance(state, list):
                 state = [state]
-            b.wait(lambda: b.text(sel + " td[data-label=State]") in state)
+            with b.wait_timeout(10):
+                b.wait(lambda: b.text(sel + " td[data-label=State]") in state)
```

But let's verify.

PR #1251, opened 2023-03-29
test: don't lower test timeouts to 5

jelly:
Our normal test timeout is 15, for restarting containers the timeout of 5 is not enough and it flakes occasionally.

context: https://github.com/cockpit-project/cockpit-podman/pull/1241#issuecomment-1488363794

Be aware, we have more flakes, but let's tackle them off one by one


jelly:
Alright this still flakes, but not for testLifeCycleOperations!

martinpitt (approved):
Thanks! Retried the flakes.

PR #1254, opened 2023-03-30
Use orientation instead of deprecated isVertical

jelly:
PatternFly deprecated isVertical in favour of the orientation property.

https://www.patternfly.org/v4/components/divider/#divider


martinpitt (approved):
Thanks!

PR #1256, opened 2023-03-30
Show HealthCheck Failure action in container details

tomasmatus:
image


src/ContainerHealthLogs.jsx
```diff
@@ -35,12 +35,20 @@ const format_nanoseconds = (ns) => {
     return cockpit.format(cockpit.ngettext("$0 second", "$0 seconds", seconds), seconds);
 };
 
+const HealthcheckOnFailureActionText = {
+    none: _("No action"),
+    restart: _("Restart"),
+    stop: _("Stop"),
+    kill: _("Force stop"),
+};
```
jelly, line 43: Ah right, the API wants integers for creation, and returns strings when we request it...

tomasmatus: when do you think API will change the returned string?  :)

jelly:
Please update the pixel tests https://cockpit-logs.us-east-1.linodeobjects.com/pull-1256-20230331-155645-4bd0fc64-fedora-37/log.html

jelly:
Sorry @tomasmatus as I merged the sort containers PR you'll need to regenerate the pixel tests again :(

jelly (dismissed):

KKoukiou:
The new testHealthsystem test is very flaky. Checking this locally now

KKoukiou (dismissed):

garrett:
I was just made aware of this; I didn't realize we'd say "failure action" in the UI when it isn't failing. That's a problem. Had I seen this earlier, I would've suggested calling it something else here and in the create container dialog. :disappointed:

garrett (approved):
Looks good now! Thanks for the string update.

Please also update the same label in the "Create container" modal dialog as well, either as a separate commit here or in a follow-up PR that would land at the same time. Thanks!

garrett:
(Oh ... and we should probably update the release notes and merge the two halves into one item on the notes... or at least update the create modal screenshot with the new text there too!)

tomasmatus:
> Please also update the same label in the "Create container" modal dialog as well, either as a separate commit here or in a follow-up PR that would land at the same time. Thanks!

Done in separate commit on this branch

> (Oh ... and we should probably update the release notes and merge the two halves into one item on the notes... or at least update the create modal screenshot with the new text there too!)

@KKoukiou pinging you because this is probably something you should do when finishing the release. Thank you!

KKoukiou (approved):

PR #1257, opened 2023-03-30
Apply regressed pixel test update

KKoukiou:


martinpitt (approved):
Sounds like a collision between me updating container images, and someone else updating PF breakpoints in parallel? Odd, I didn't see that in e.g. #1253. But *shrug*, if it works..

PR #1259, opened 2023-04-05
packaging: Declare bundled NPM dependencies in spec file

martinpitt:
See https://bugzilla.redhat.com/show_bug.cgi?id=2180520

Turn the pattern make rule into an explicit one, as pattern rules don't support dependencies. We only need it for our single spec file anyway.

Cherry-picked from starter-kit commit 8817969d1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,9 +81,9 @@ po/LINGUAS:
 #
 # Build/Install/dist
 #
-
-%.spec: packaging/%.spec.in
-	sed -e 's/%{VERSION}/$(VERSION)/g' $< > $@
+$(SPEC): packaging/$(SPEC).in $(NODE_MODULES_TEST)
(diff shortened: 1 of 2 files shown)


martinpitt:
[This pixel failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1259-20230405-083526-983f3009-fedora-37/log.html) is the same (unrelated) one as in #1258 -- it exists on main, and is a time bomb. Let's discuss it there?

martinpitt:
[This pixel failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1259-20230405-083526-983f3009-fedora-37/log.html) is the same (unrelated) one as in #1258 -- it exists on main, and is a time bomb. Let's discuss it there?

marusak (approved):

PR #1260, opened 2023-04-05
Skip mobile layout where pixel testing overview as it's flaky

KKoukiou:
The image creation dates vary between test runs, and their wrapping is neither reliably broken into one nor two lines. Since this is not possible to ignore with a selector as it shifts the UI let's just skip the mobile layout which makes these dates wrap.

### DIFF
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -178,7 +178,7 @@ class Images extends React.Component {
         const columnTitles = [
             { title: _("Image"), transforms: [cellWidth(20)] },
             { title: _("Owner"), props: { className: "ignore-pixels" } },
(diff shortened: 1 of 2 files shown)


martinpitt:
Thanks! This needs a pixel ref update, though:

> Unused reference image TestApplication-testBasicSystem-overview-mobile-pixels.png

KKoukiou:
iff same problem other test - https://cockpit-logs.us-east-1.linodeobjects.com/pull-1260-20230406-081303-fe467274-fedora-37/pixeldiff.html#TestApplication-testBasicSystem-overview

skobyda (approved):

PR #1262, opened 2023-04-11
XXX test TF

martinpitt:
Checking whether we now get the [same failure](https://artifacts.dev.testing-farm.io/89293494-3014-4425-9897-7ca92214b766/) that [broke the release](https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/103) in upstream tests:

```
package runc is not installed
crun-1.8.3-2.fc38.x86_64
podman-4.4.2-2.fc38.x86_64
[...]
+ for img in localhost/test-alpine localhost/test-busybox localhost/test-registry
+ podman save localhost/test-alpine
+ sudo -i -u admin podman load
Error: default OCI runtime "crun" not found: invalid argument
```


martinpitt:
OK, this doesn't reproduce here. Why oh why?

PR #1263, opened 2023-04-11
test: Switch back to offline VM preparation

martinpitt:
Now that all our test VM images picked up the new container images from <https://github.com/cockpit-project/bots/pull/4585>, we can go back to preparing the test VMs offline, and drop the hack again.

----

This still requires some image refreshes:
 - [ ] https://github.com/cockpit-project/bots/pull/4620


jelly (approved):
Danke!

PR #1271, opened 2023-04-17
Fix management of containersDetails

tomasmatus:
Currently containerDetails are deleted when container reaches any other state than "running". Because of this when new container is created or is stopped the Integration tab shows infinite loading until the page is refreshed or the container is started again.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -317,7 +317,7 @@ class Application extends React.Component {
                         if (event && event.Action === "restart")
                             reply.State = "restarting";
                         this.updateState("containers", reply.Id + system.toString(), reply);
-                        if (reply.State == "running") {
+                        if (["running", "created", "exited", "paused", "stopped"].find(containerState => containerState === reply.State)) {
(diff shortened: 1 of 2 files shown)


martinpitt:
testHealthcheck{System,User} [fail everywhere](https://artifacts.dev.testing-farm.io/3916ad58-4bf2-470b-801e-ed7281042a1b/) in TF. We've landed a few PRs while package builds in TF were broken, but the most recent plausible cause # 1256 still did pass tests. Here the test has an ambiguous selector as the container has two badges: "Exited" and "Unhealthy". I think that actually is fallout from this PR?

martinpitt:
Yes it is, same failures on [our CI](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1271-20230417-124744-c740826a-debian-testing/log.html).

KKoukiou (approved):
@tomasmatus does not it worth to add a test for it? Could be follow up if you prefer.

PR #1272, opened 2023-04-18
.github: add Cockpit library update action

jelly:
This action works similar to the npm-update workflow, it creates a PR every Thursday (usually after the Cockpit release) to update the COCKPIT_REPO_COMMIT to the latest version.


martinpitt (approved):
Thanks!

PR #1277, opened 2023-04-24
test: Check image history textually instead of with pixels

martinpitt:
The real container images on our VM images change all the time. In particular their creation date may be very recent ("last Monday at 11:00 AM") or older ("2023/03/17"). That changes the table column width and breaks the "history" pixel test. Also, the differing widths of the image IDs and dates cause small column width variation. Finally, all of the actual text was ignored by the pixel test, leaving only the table frame (which isn't very interesting, as it's a standard component).

So all of this makes this a bad use case for pixel tests. Instead, create a dynamic dummy image with predictable layers, date, and content, and check it textually.

---

This blocks https://github.com/cockpit-project/bots/pull/4683 , and we've had this kind of noise for each image refresh many times. I tried for about two hours to salvage this pixel test, running into countless annoying deltas where the column widths wiggled just a bit. But IMHO in the end it's just not worth it.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -498,6 +498,16 @@ class TestApplication(testlib.MachineCase):
         # images are sorted alphabetically
         self.assertRegex(text, ".*/test-alpine.*/test-busybox.*/test-registry")
 
+        # build a dummy image so that the timestamp is "today" (for predictable pixel tests)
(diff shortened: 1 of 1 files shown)


marusak (approved):
Thanks! Not a fan of dropping pixel tests but if everything was ignored anyway this can go.

PR #1285, opened 2023-05-08
package.json: Bump patternfly and pkg/lib

KKoukiou:
FormSelect onChange callback parameters order changed.

This allows/forces us also to also bump pkg/lib and get the latests changes there. Most notable is the removal of scrollbars when pixel testing


martinpitt (approved):
Nice! I really like the scrollbar removal trick. I retried the aarch64 test, which failed on a `podman pull` network glitch.

PR #1292, opened 2023-05-18
Fix:issue#1216 allow the search filters to be cleared or reset to the default

### LINKED ISSUE #1216: Allows the search filters to be cleared or reset to the default
Add the same clear filter functionality as the logs page in Cockpit.

image

This probably means we have to change the `TextInput` into a `SearchInput` and `FormSelect` into a `Select` which supports onClear=()

shubhit7:
Fixes #1216 

- Change owner filter, from `FormSelect` to `Select` next component of patternfly alpha release. Old `Select` has been deprecated.
- Change the search field beside owner filter, from `TextInput` to `SearchInput`.

### DIFF
--- a/src/ContainerHeader.jsx
+++ b/src/ContainerHeader.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import cockpit from 'cockpit';
 import { FormSelect, FormSelectOption } from "@patternfly/react-core/dist/esm/components/FormSelect";
-import { TextInput } from "@patternfly/react-core/dist/esm/components/TextInput";
+import { SearchInput } from "@patternfly/react-core/dist/esm/components/SearchInput";
 import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core/dist/esm/components/Toolbar";
 const _ = cockpit.gettext;
 
@@ -24,10 +24,11 @@ const ContainerHeader = ({ user, twoOwners, ownerFilter, handleOwnerChanged, tex
                     </>
                 }
                 <ToolbarItem>
-                    <TextInput id="containers-filter"
+                    <SearchInput id="containers-filter"
                                    placeholder={_("Type to filter…")}
                                    value={textFilter}
-                                   onChange={handleFilterChanged} />
+                                   onChange={(_event, value) => handleFilterChanged(value)}
+                                   onClear={() => handleFilterChanged('')} />
                 </ToolbarItem>
             </ToolbarContent>
         </Toolbar>


shubhit7:
I figured out how we can use deprecated `Select`.
@KKoukiou Shall we use deprecated `Select` or `Select` Next ?

KKoukiou:
> I figured out how we can use deprecated `Select`. @KKoukiou Shall we use deprecated `Select` or `Select` Next ?

Either of the two works for now. Tending towards the deprecated one, so that we can move all Selects at once  to the new.

src/ContainerHeader.jsx
```diff
@@ -1,11 +1,27 @@
...
+import { Select, SelectOption, SelectVariant } from "@patternfly/react-core/dist/esm/deprecated/components/Select";
 const _ = cockpit.gettext;
 
 const ContainerHeader = ({ user, twoOwners, ownerFilter, handleOwnerChanged, textFilter, handleFilterChanged }) => {
+    const [isOpenOwnersFilter, setIsOpenOwnersFilter] = React.useState(false);
+    const selectOptions = {
+        system: _("System"),
+        all: _("All"),
+        user
+
```
KKoukiou, line 14: Newline inside object should be removed.

KKoukiou (changes requested):
I am happy with the search input adoption - but totally disagree with adopting the typeahead for the system/user filter. Logs page uses the typeahead in a case where there are hunderds of select options, not 3 like here. 
I know it's suggested in https://github.com/cockpit-project/cockpit-podman/issues/1216#issue-1601474369 but it's just not correct design wise..

Please undo the select part, and keep the SearchInput. Let's also add a test for the new functionality and it's good to go.

src/ContainerHeader.jsx
```diff
@@ -24,10 +24,11 @@ const ContainerHeader = ({ user, twoOwners, ownerFilter, handleOwnerChanged, tex
                     </>
                 }
                 <ToolbarItem>
-                    <TextInput id="containers-filter"
+                    <SearchInput id="containers-filter"
                                    placeholder={_("Type to filter…")}
                                    value={textFilter}
-                                   onChange={handleFilterChanged} />
+                                   onChange={(_event, value) => handleFilterChanged(value)}
+                                   onClear={() => handleFilterChanged('')} />
```
cockpituous, line 31: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1292-20230606-085317-ba80941c-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

shubhit7:
> PR looks good from my side but the tests fail everywhere, it looks like you will need to adjust how test for the value of the search input:
> 
> ```
> wait_js_cond(ph_has_val("#containers-filter","pod-1")): Uncaught (in promise) Error: #containers-filter does not have a value
> ```

You are right, using the SearchInput component changes the structure of dom elements compared to using simple TextInput component. Now the input box will be accessible by selector `"#containers-filter input"`.

`wait_js_cond(ph_has_val("#containers-filter input","pod-1")) ` should not throw any exception. I am not able to get hands on test suite setup.

martinpitt:
Continuing in #1968

PR #1301, opened 2023-06-01
Filter out unwanted RegExp characters when searching

jelly:
Instead of oops'ing filter out invalid regexp characters.

Closes: #1300

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -556,7 +556,8 @@ export class ImageRunModal extends React.Component {
             imageRegistries.push(this.state.searchByRegistry);
         }
 
-        let regexString = searchText;
+        // Strip out all non-allowed container image characters when filtering.
+        let regexString = searchText.replace(/[^\w_.:-]/g, "");
         // Strip image registry option if set for comparing results for docker.io searching for docker.io/fedora
         // returns docker.io/$username/fedora for example.
         if (regexString.includes('/')) {
--- a/test/check-application
+++ b/test/check-application
@@ -1550,6 +1550,10 @@ class TestApplication(testlib.MachineCase):
         b.click("#containers-containers button.pf-v5-c-button.pf-m-primary")
         b.set_input_text("#run-image-dialog-name", container_name)
 
+        # Test invalid input
+        b.set_input_text("#create-image-image-select-typeahead", "|alpi*ne?\\")
+        b.wait_text("button.pf-v5-c-select__menu-item:not(.pf-m-disabled)", "localhost/test-alpine:latest")
+
         # No local results found
         b.set_input_text("#create-image-image-select-typeahead", "notfound")
 


src/ImageRunModal.jsx
```diff
@@ -559,7 +559,8 @@ export class ImageRunModal extends React.Component {
             imageRegistries.push(this.state.searchByRegistry);
         }
 
-        let regexString = searchText;
+        // Filter out characters which are not allowed in RegExp.
+        let regexString = searchText.replace(/[^a-zA-Z0-9_\\.-]/g, "").replaceAll('\\');
```
martinpitt, line 563: `/g` and replace*All* are redundant, but I suppose it doesn't hurt.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping has a more restricted set of characters to escape. Is this safe against messing up multi-byte unicode? And is it safe to escape any and all characters that way?

test/check-application
```diff
@@ -1417,6 +1417,11 @@ class TestApplication(testlib.MachineCase):
         b.click("#containers-containers button.pf-c-button.pf-m-primary")
         b.set_input_text("#run-image-dialog-name", container_name)
 
+        # Test invalid input
+        b.set_input_text("#create-image-image-select-typeahead", "*\\|?")
+        b.click('button.pf-c-toggle-group__button:contains("Local")')
+        b.wait_text("button.pf-c-select__menu-item.pf-m-disabled", "No images found")
```
martinpitt, line 1423: This doesn't prove much, though. Could this actually create an image with a ? or * in it, and make sure it gets found?

jelly: Ah, haven't considered that, my main concern was that we wouldn't oops. What I would really like is to show an validation error when entering non-allowed characters. But this is a bit tricky with the select when the dropdown is shown above the validation line of PF.

src/ImageRunModal.jsx
jelly: I don't think it's foolproof, we should probably catch any exception when calling `RegExp()` but I wasn't sure how to handle the exception? Do we just not filter case insensitive?

martinpitt: I'm not sure how that is related to case sensitivity. An exception in RegExp() after escaping is an internal programming error. So an Oops isn't actually too bad. Of course you could recover by console.error()'ing the exception and resetting the search form or so, but it's not a much better UX.

test/check-application
martinpitt: Is there some spec what kind of characters can appear in container names? If so, that would help indeed. 

Indeed docker seems to be rather  picky about it:
```
❱❱❱ podman tag docker.io/library/debian:unstable 'localhost:foo?bar'    
Error: normalizing name "localhost:foo?bar": normalizing name "localhost:foo?bar": invalid reference format    
❱❱❱ podman tag docker.io/library/debian:unstable 'localhost:foo*bar'    
Error: normalizing name "localhost:foo*bar": normalizing name "localhost:foo*bar": invalid reference format    
❱❱❱ podman tag docker.io/library/debian:unstable 'localhost:foo|bar'    
Error: normalizing name "localhost:foo|bar": normalizing name "localhost:foo|bar": invalid reference format    
❱❱❱ podman tag docker.io/library/debian:unstable 'localhost:foobär'    
Error: normalizing name "localhost:foobär": normalizing name "localhost:foobär": invalid reference format    
```

jelly: The restriction seems to be this regex https://github.com/moby/moby/blob/be97c66708c24727836a22247319ff2943d91a03/daemon/names/names.go#L6

martinpitt:
@jelly Are you still working on this? This doesn't seem too far from landing, and would make a nice stabilization fix for RHEL (today's release is the last chance for current y-stream releases)

src/ImageRunModal.jsx
jelly: I think we are approaching this in a very wrong way, we are adding regex to late imo. This is a <Select> sadly but let's see if we can just do form validation.

Also unrelated, we could handle:

image

test/check-application
jelly: This is about filtering containers, not about creating. So you can't actually select an invalid container image as those should not be returned from the registry (that would be a bug imo)

jelly:
This needs work, it actually crashes in the test:

```
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 598, in wait_js_cond
    raise e
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 582, in wait_js_cond
    result = self.cdp.invoke("Runtime.evaluate",
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/bots/make-checkout-workdir/test/common/cdp.py", line 205, in invoke
    res = self.command(cmd)
          ^^^^^^^^^^^^^^^^^
  File "/work/bots/make-checkout-workdir/test/common/cdp.py", line 232, in command
    raise RuntimeError(res["error"])
RuntimeError: SyntaxError: Invalid regular expression: /\/i: \ at end of pattern
```

src/ImageRunModal.jsx
```diff
@@ -556,7 +556,8 @@
             imageRegistries.push(this.state.searchByRegistry);
         }
 
-        let regexString = searchText;
+        // Strip out all non-allowed container image characters when filtering.
+        let regexString = searchText.replace(/[^\w0-9_.:-]/g, "");
```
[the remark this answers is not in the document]
martinpitt: Yes, \w includes 0-9

test/check-application
martinpitt: Can this also test a string with some valid and some invalid characters? In this check, *all* characters are dropped.

martinpitt (approved):
Nice and clean!

PR #1304, opened 2023-06-14
test: add compatibility for Cockpit versions

jelly:
A newer Cockpit has different patternfly classnames prefixed with pf-v5, as updating PF brought up some bigger issues.


martinpitt:
Good enough! Thanks!

martinpitt (approved):

PR #1307, opened 2023-06-16
fix: Popover broken issue

### LINKED ISSUE #1302: Bug - Popover in Create container dialog is broken
Cockpit version: 292
Cockpit-podman version: 70
Podman version: 4.5.0
OS: Fedora Workstation 38

image

shubhit7:
Closes #1302 

Giving the owner div z-index and removing the `min-width: "revert"` property from popover component should fix this issue.

Although, I am not really able to remove the `min-width` property, feels like there is some issue at patternfly upstream.

https://github.com/cockpit-project/node-cache/blob/b3aed397c0415407d9b4b4d21e71fcc5f2160d05/%40patternfly/react-core/src/components/Popover/Popover.tsx#L467

Is `min-width: "revert"` hardcoded here?

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -699,9 +699,9 @@ export class ImageRunModal extends React.Component {
                 <Tabs activeKey={activeTabKey} onSelect={this.handleTabClick}>
                     <Tab eventKey={0} title={<TabTitleText>{_("Details")}</TabTitleText>} className="pf-c-form pf-m-horizontal">
                         { this.props.userServiceAvailable && this.props.systemServiceAvailable &&
-                        <FormGroup isInline hasNoPaddingTop fieldId='run-image-dialog-owner' label={_("Owner")}
+                        <FormGroup isInline hasNoPaddingTop fieldId='run-image-dialog-owner' label={_("Owner")} style={{ zIndex: "1" }}
                                    labelIcon={
-                                       <Popover aria-label={_("Owner help")}
+                                       <Popover aria-label={_("Owner help")} minWidth='0'
                                           enableFlip
                                           bodyContent={
                                               <>


martinpitt:
@jelly can you please have a look? Is it legit to put `style=` into jsx? Shouldn't that go into the .scss due to our C-S-P?

jelly:
We have the same issue here, https://github.com/cockpit-project/cockpit-machines/issues/1100

> @jelly can you please have a look? Is it legit to put `style=` into jsx? Shouldn't that go into the .scss due to our C-S-P?

I tried the fix locally and it doesn't seem to be fixed.

The real issue seems to be the `minWidth=revert` which passing `minWidth` does not circumvent. So I'd opt for a `scss` rule instead to override `min-width`.

jelly:
The demo from @KKoukiou uses @patternfly/react-core 50 while we are on 5.0.0-alpha.83

shubhit7:
> We have the same issue here, [cockpit-project/cockpit-machines#1100](https://github.com/cockpit-project/cockpit-machines/issues/1100)
> 
> > @jelly can you please have a look? Is it legit to put `style=` into jsx? Shouldn't that go into the .scss due to our C-S-P?
> 
> I tried the fix locally and it doesn't seem to be fixed.
> 
> The real issue seems to be the `minWidth=revert` which passing `minWidth` does not circumvent. So I'd opt for a `scss` rule instead to override `min-width`.
> 
> We should look if Cockpit has a Popover in a modal which is also broken and maybe ask PF devs what's up? So that would be making a Modal with a Popover component.

yes, it would not fix it locally as minWidth is always getting set to revert.
If you try to apply the fix on dev tools in browser, fix works fine.
But again, is this a bug at PF ? I had already raised an issue on Patternfly-react, no response yet :/

shubhit7:
> > We have the same issue here, [cockpit-project/cockpit-machines#1100](https://github.com/cockpit-project/cockpit-machines/issues/1100)
> > > @jelly can you please have a look? Is it legit to put `style=` into jsx? Shouldn't that go into the .scss due to our C-S-P?
> > 
> > 
> > I tried the fix locally and it doesn't seem to be fixed.
> > The real issue seems to be the `minWidth=revert` which passing `minWidth` does not circumvent. So I'd opt for a `scss` rule instead to override `min-width`.
> > We should look if Cockpit has a Popover in a modal which is also broken and maybe ask PF devs what's up? So that would be making a Modal with a Popover component.
> 
> yes, it would not fix it locally as minWidth is always getting set to revert. If you try to apply the fix on dev tools in browser, fix works fine. But again, is this a bug at PF ? I had already raised an issue on Patternfly-react, no response yet :/
> 
> Meanwhile, I will try to override it by scss

I don't think it is really possible to override the minWidth=revert, I have tried scss and also inline styling (has the highest precedence), but the revert property doesn't get overridden. I guess PF popover component needs a fix.

shubhit7:
It seems, minWidth issue has been fixed upstream.
https://github.com/patternfly/patternfly-react/issues/9279.

If we update the patternfly deps, it should work.

jelly:
> It seems, minWidth issue has been fixed upstream. [patternfly/patternfly-react#9279](https://github.com/patternfly/patternfly-react/issues/9279).
> 
> If we update the patternfly deps, it should work.

Thanks for looking into this! We updated to 53 but I suppose that is not new enough.

shubhit7:
> > It seems, minWidth issue has been fixed upstream. [patternfly/patternfly-react#9279](https://github.com/patternfly/patternfly-react/issues/9279).
> > If we update the patternfly deps, it should work.
> 
> Thanks for looking into this! We updated to 53 but I suppose that is not new enough.

Yeah, the fix has been released in @patternfly/react-core@5.0.0-prerelease.14

jelly:
This has been fixed by PatternFly and now looks good:

image

PR #1310, opened 2023-06-22
test: Update pixel references for Chromium 114

martinpitt:
The recent cockpit/tasks refresh got a new Chromium version, which causes subtle pixel changes.


martinpitt:
Nice, no updates actually needed.

PR #1325, opened 2023-07-03
test: Print container logs on failure

martinpitt:
That may help to debug failures where containers unexpectedly stop.

---

[example log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1324-20230703-082205-84171830-rhel-9-3/log.html)

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -175,6 +175,16 @@ class TestApplication(testlib.MachineCase):
         self.system_images_count = int(self.execute(True, "podman images -n | wc -l").strip())
         self.user_images_count = int(self.execute(False, "podman images -n | wc -l").strip())
 
+    def tearDown(self):
+        if self.getError():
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -175,6 +175,16 @@ class TestApplication(testlib.MachineCase):
         self.system_images_count = int(self.execute(True, "podman images -n | wc -l").strip())
         self.user_images_count = int(self.execute(False, "podman images -n | wc -l").strip())
 
+    def tearDown(self):
```
jelly, line 178: addcleanup is called after tearDown so this sounds sensible!

martinpitt: Yep, I learned the same in https://github.com/cockpit-project/cockpit/commit/163555f90dc2872be715da073ae468c9b77867cf

PR #1326, opened 2023-07-03
Update font class for PF5

jelly:
In the last PF5, we forgot to update the tabular number class.

Pixel test updates https://cockpit-logs.us-east-1.linodeobjects.com/pull-1326-20230703-093916-959ebbfc-fedora-37/log.html


jelly:
Only triggered the required tests as this we are a bit flaky right now.

martinpitt (approved):
Ah, fun -- that essentially reverts the bits that I commented on in the PF5 update PR. Thanks!

PR #1324, opened 2023-07-03
Debug flakes

martinpitt:
I got [this testCreatePodUser error](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1322-20230703-042644-64a79ae1-fedora-38-devel/log.html#12-1) locally once when I was investigating something else. But of course since then I could never reproduce it again. The pod container is in state "Created", while it's supposed to be "Running". That isn't just a timeout question, it stays in "Created" forever. 

[another example](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1322-20230703-063401-7cc9e4bf-fedora-38/log.html#12)

PR #1329, opened 2023-07-04
Don't re-fetch all containers for "cleanup" event

martinpitt:
This does not mean that the container goes away, that will only happen for the "remove" event. "cleanup" will also happen (twice!) between "died" and "init" for restarting a container.

This avoids two parallel getContainer() calls with possibly inconsistent information, and generally reduces the number of getContainer() calls.

---

This fixes testLifecycleOperations failures like [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1324-20230703-141527-37092ff0-fedora-38/log.html#26) and [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1324-20230703-141527-37092ff0-fedora-38-devel/log.html#24). [debugging notes](https://github.com/cockpit-project/cockpit-podman/pull/1324#issuecomment-1619493649), stress-tested in #1324

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -406,6 +406,7 @@ class Application extends React.Component {
         case 'died':
         case 'exec_died':
         case 'kill':
+        case 'cleanup':
         case 'mount':
         case 'pause':
         case 'prune':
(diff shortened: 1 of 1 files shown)


jelly (approved):
Sounds logical, sadly the events are not documented https://docs.podman.io/en/latest/_static/api.html#tag/system/operation/SystemEventsLibpod

PR #1334, opened 2023-07-05
Serialize start and remove event handling

martinpitt:
Don't start pod and container state updates at the same time. They do asynchronous state updates which sometimes step on each other's toes and reverse the real event order.

This mitigates the problem with getting podman API call results in the wrong order (https://github.com/containers/podman/issues/19124); that causes containers to sometimes appear in a non-current state.

-----

See the above podman issue, this can't be perfect. I stress-tested this in #1324, and at least it got a lot better. E.g. in commit 5a28f0d8503bfc6e4845d87733373eed0ac13bc0 I just got one failure out of 5x3x13 runs (amplification x affected retries x TEST_OS). But it makes runs a lot easier to get green.

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -199,7 +199,7 @@ class Application extends React.Component {
     }
 
     updateContainersAfterEvent(system, init) {
-        client.getContainers(system)
+        return client.getContainers(system)
                 .then(reply => Promise.all(
                     (reply || []).map(container =>
                         this.isContainerCheckpointPresent(container.Id, system)
@@ -267,7 +267,7 @@ class Application extends React.Component {
     }
 
     updatePodsAfterEvent(system) {
-        client.getPods(system)
+        return client.getPods(system)
                 .then(reply => {
                     this.setState(prevState => {
                         // Copy only pods that could not be deleted with this event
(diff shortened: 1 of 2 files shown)


src/app.jsx
```diff
@@ -420,14 +418,14 @@ class Application extends React.Component {
...
-                this.updatePodAfterEvent(event.Actor.Attributes.podId, system);
-            } else {
-                this.updatePodsAfterEvent(system);
-            }
-            this.updateContainersAfterEvent(system);
+            this.updateContainersAfterEvent(system).then(() => {
+                // HACK: we don't get a pod event when a container in a pod is removed.
+                // https://github.com/containers/podman/issues/15408
+                if (event.Actor.Attributes.podId)
+                    this.updatePodAfterEvent(event.Actor.Attributes.podId, system);
```
cockpituous, line 425: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1334-20230705-071328-f502185e-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

jelly (approved):
Good find, would have been impossible without the debug logging.

PR #1341, opened 2023-07-07
onKeyPress is replaced with onKeyDown

themukeshmahiya:
trying to solve this good first issue

https://github.com/cockpit-project/cockpit/issues/19061


jelly (approved):
Thanks!

PR #1343, opened 2023-07-07
test: adjust pod delete error

jelly:
In podman 4.6.0-rc1 the returned exception message is changed and no longer mentions paused/running containers.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -371,7 +371,8 @@ class TestApplication(testlib.MachineCase):
 
         self.performPodAction("pod-1", "system", "Delete")
         b.click(".pf-v5-c-modal-box button:contains(Delete)")
-        b.wait_in_text(".pf-v5-c-modal-box__body", "running or paused containers cannot be removed without force")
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -371,7 +371,11 @@ class TestApplication(testlib.MachineCase):
 
         self.performPodAction("pod-1", "system", "Delete")
         b.click(".pf-v5-c-modal-box button:contains(Delete)")
-        b.wait_in_text(".pf-v5-c-modal-box__body", "running or paused containers cannot be removed without force")
+        # https://github.com/containers/podman/issues/19159
+        if podman_version(self) >= (4, 6, 0):
```
mvollmer, line 375: So if the podman issues is fixed, this will change again?  What about just waiting for any kind of error to appear?  (Likely a .pf-v5-c-alert thing, no?)

jelly: Oh, actually might be a good idea.

mvollmer (approved):
LGTM!

Issue #1287, opened 2023-05-10
Show if container has checkpoint


marusak:
Right now it is possible to checkpoint container. If container has checkpoint then we show 'Restore' option but there is no other indicator of this fact. We should visibly show that there is available version ideally with some information (at least time of the checkpoint)

PR #1345, opened 2023-07-10
Show time of container's checkpoint, if any is present

skobyda:
Fixes https://github.com/cockpit-project/cockpit-podman/issues/1287

## Show time of container's latest checkpoint

Screenshot from 2023-07-11 13-47-14

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1329,6 +1329,10 @@ class TestApplication(testlib.MachineCase):
         self.performContainerAction("swamped-crate", "Start")
         b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in 'Running')
 
+        self.toggleExpandedContainer("swamped-crate")
+        b.wait_visible(".pf-m-expanded button:contains('Details')")
+        b.wait_not_present(f'#containers-containers tr:contains("{IMG_BUSYBOX}") dt:contains("Latest checkpoint")')
+
         # Checkpoint the container
         self.performContainerAction("swamped-crate", "Checkpoint")
         b.set_checked('.pf-v5-c-modal-box input#checkpoint-dialog-keep', True)
@@ -1340,6 +1344,10 @@ class TestApplication(testlib.MachineCase):
 
         if self.has_criu:
             b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in NOT_RUNNING)
+            b.wait_in_text(
+                f'#containers-containers tr:contains("{IMG_BUSYBOX}") dt:contains("Latest checkpoint") + dd',
+                'today at'
+            )
         else:
             # expect proper error message
             b.wait_in_text(".pf-v5-c-alert.pf-m-danger", "Failed to checkpoint container swamped-crate")
--- a/src/ContainerDetails.jsx
+++ b/src/ContainerDetails.jsx
@@ -67,6 +67,10 @@ const ContainerDetails = ({ container }) => {
                         <DescriptionListTerm>{_("State")}</DescriptionListTerm>
                         <DescriptionListDescription>{render_container_state(container)}</DescriptionListDescription>
                     </DescriptionListGroup>
+                    {container?.State.Checkpointed && <DescriptionListGroup>
+                        <DescriptionListTerm>{_("Latest checkpoint")}</DescriptionListTerm>
+                        <DescriptionListDescription>{utils.localize_time(Date.parse(container.State.CheckpointedAt) / 1000)}</DescriptionListDescription>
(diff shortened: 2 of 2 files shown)


garrett (changes requested):
"Checkpointed" sounds odd to me, especially without an "at". But even "Checkpointed at" is odd.

**We should say "Latest checkpoint" instead.** (And latest instead of last, as "Last checkpoint" would imply that there will be no more checkpoints.) "Recent checkpoint" or "Current checkpoint" all also work, but they each mean something slightly different.

skobyda:
> "Checkpointed" sounds odd to me, especially without an "at". But even "Checkpointed at" is odd.
> 
> **We should say "Latest checkpoint" instead.** (And latest instead of last, as "Last checkpoint" would imply that there will be no more checkpoints.) "Recent checkpoint" or "Current checkpoint" all also work, but they each mean something slightly different.

Updated to "Latest checkpoint":

Screenshot from 2023-07-11 13-46-25

garrett (approved):
:+1: Thanks! LGTM

src/app.jsx
```diff
@@ -314,6 +314,7 @@ class Application extends React.Component {
                                 delete copyStats[reply.Id + system.toString()];
                                 return { containersDetails: copyDetails, containersStats: copyStats };
                             });
+                            this.inspectContainerDetail(reply.Id, system);
```
martinpitt, line 317: I'd like to put a hold on this. I am currently completely reworking the state/API handling to not call inspectContainerDetail() and getContainers() a zillion times (it causes flakes and makes everything slow), see #1324. I should be done today, then we can rebase this PR and avoid this call.

martinpitt (changes requested):
Thanks! See below, I'd like to queue this.

src/app.jsx
skobyda: Sure no problem

martinpitt:
@skobyda Can you please update the screenshot? The string changed, and we also shouldn't have both images in the release note. Thanks!

martinpitt:
@skobyda : The API rework in PR #1350 landed. Can you please rebase? You'll also need to fix the overlong line length which made the tests fail everywhere. Thanks!

src/app.jsx
martinpitt: Unblocked. This call can be dropped now.

test/check-application
```diff
@@ -1340,6 +1340,11 @@ class TestApplication(testlib.MachineCase):
 
         if self.has_criu:
             b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in NOT_RUNNING)
+            self.toggleExpandedContainer("swamped-crate")
+            b.wait_in_text(
+                f'#containers-containers tr:contains("{IMG_BUSYBOX}") dt:contains("Latest checkpoint") + dd',
+                'today at'
```
martinpitt, line 1346: Can you please also add a check that this description item does *not* appear for an unchecked container? In my local runs I see some rather inconsistent values. For example, I never saw the `State.Checkpointed` value (it may not exist at all for containers without a checkpoint), and the default value of `CheckpointedAt` is "0001-01-01T00:00:00Z", i. e. not quite a null-ish value as podman interprets and formats the zero as time string.

martinpitt:
Ignore [this healthcheck failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1345-20230713-072534-d2cf1d09-arch/log.html#15-1) - I'm handling it in https://github.com/cockpit-project/cockpit-podman/pull/1324#issuecomment-1633697209 and will soon post a PR.

test/check-application
skobyda: Fixed

test/check-application
```diff
@@ -1329,6 +1329,9 @@ class TestApplication(testlib.MachineCase):
         self.performContainerAction("swamped-crate", "Start")
         b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in 'Running')
 
+        self.toggleExpandedContainer("swamped-crate")
+        b.wait_not_present(f'#containers-containers tr:contains("{IMG_BUSYBOX}") dt:contains("Latest checkpoint")')
```
martinpitt, line 1333: That's a race condition. After toggling, you first need a positive check for something else in the expanded area, as immediately after clicking the toggle icon the "not present" check will always be true for a while.

skobyda: Fixed

jelly (approved):

martinpitt (approved):
Cheers!

PR #1349, opened 2023-07-12
test: Change restart detection from Pid to StartedAt

martinpitt:
The `Pid` property is not exposed in the /containers/ID/json API, and generally not recommended as they can be recycled. We are going to drop `data-pid` in the near future.

Move the restart detection in the test from comparing Pid to comparing StartedAt.

Change the `data-started-at` rendering from Unix time stamp (from /containers/json API) to ISO string (from /containers/ID/json API), as we are going to move to the latter soon, and that also coincides to the CLI output, so it's easier to compare.


marusak (approved):
Thanks!

PR #1353, opened 2023-07-13
Fix race condition in testLifecycleOperations, Serialize updateContainer() calls, Ignore irrelevant container events

martinpitt:


src/app.jsx
```diff
@@ -320,9 +331,16 @@ class Application extends React.Component {
         /* The following events do not need to trigger any state updates */
         case 'attach':
         case 'exec':
+        case 'exec_died':
```
cockpituous, line 334: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1353-20230713-111150-a8183c4d-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/app.jsx
```diff
@@ -320,9 +331,16 @@ class Application extends React.Component {
...
         case 'export':
         case 'import':
         case 'init':
+        case 'kill':
+        case 'mount':
+        case 'prune':
+        case 'restart':
+        case 'sync':
+        case 'unmount':
```
cockpituous, line 343: These 6 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1353-20230713-111150-a8183c4d-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/app.jsx
```diff
@@ -352,6 +352,7 @@ class Application extends React.Component {
         case 'create':
         case 'died':
         case 'exec_died':
+        case 'health_status':
```
marusak, line 355: But does it do actually anything? When does podman send this event? Every time healthcheck is performed? If so, then this would be rather annoying to receive this event every time and reload container.

martinpitt: That is, I'll check when this event is sent more closely tomorrow.

martinpitt: a-ha! yes, it does something. It happens every time when there is a new healthcheck result. This [fails consistently](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1353-20230713-173031-9e7430ad-fedora-38-devel/log.html#14) when I *remove* the "Handle 'health_status' container event" patch, and "*add* the "Ignore irrelevant container events" patch (so that there are no other events which would often, but not always, catch up with health updates). So having both patches is overall a better solution IMHO.

martinpitt:
[testHealthcheck](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1353-20230713-182931-035e7973-ubuntu-2204/log.html) is still very flaky. But that already [happens on main](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1345-20230713-072534-d2cf1d09-fedora-37/log.html#15), investigating in https://github.com/cockpit-project/cockpit-podman/pull/1324#issuecomment-1633697209 . So retrying.

src/app.jsx
```diff
@@ -331,9 +331,16 @@ class Application extends React.Component {
         /* The following events do not need to trigger any state updates */
         case 'attach':
         case 'exec':
+        case 'exec_died':
         case 'export':
         case 'import':
         case 'init':
+        case 'kill':
+        case 'mount':
+        case 'prune':
```
marusak, line 340: prune for example would have effect on the UI but I assume we get different event as well which is enough?

marusak (approved):
Looks good! The last commit with ignoring events may be potentially dangerous, see my comment

src/app.jsx
martinpitt: Do you know what this event actually does/is?

martinpitt: In my [discussion](https://github.com/containers/podman/issues/19124#issuecomment-1621297821) with @vrothberg he recommended to ignore internal transient events, but as I don't know what "prune" does, I can't judge if that's something internal, or an action that the user can do.

martinpitt: It's [podman container prune](https://docs.podman.io/en/latest/markdown/podman-container-prune.1.html). Running these commands:
```
podman create localhost/test-busybox sleep infinity
podman container prune -f
```
does remove the container. `podman events` will show
```
2023-07-14 06:45:10.698039538 +0000 UTC container create a76bcda78923883a1abe5db42efbf0e4492d636888b3fd5b75529399d8844b76 (image=localhost/test-busybox:latest, name=gracious_dirac)
2023-07-14 06:45:10.683002845 +0000 UTC image pull 839899c307141ddc522d4018c208d46f5ac3050909a9a429775bb248ef5e6e42 localhost/test-busybox
2023-07-14 06:45:21.569598655 +0000 UTC container remove a76bcda78923883a1abe5db42efbf0e4492d636888b3fd5b75529399d8844b76 (image=localhost/test-busybox:latest, name=gracious_dirac)
```
i.e. it does not even get a `prune` event, but it does give a `remove` event.

martinpitt:
Argh, this makes the health check tests more flaky, especially on [ubuntu-2204](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1353-20230714-080809-035e7973-ubuntu-2204/log.html) (but not limited to that). I analyzed that in https://github.com/cockpit-project/cockpit-podman/pull/1324#issuecomment-1633697209 , reported it as https://github.com/containers/podman/issues/19237 , and documented our accidental workaround explicitly.

src/app.jsx
martinpitt: I had to revert this, see below. It would be the right thing to do on paper, but it's utterly broken.

marusak (approved):
I am so sorry you have to deal with this :/ Thanks!

PR #1355, opened 2023-07-14
test: Clean up test skips

martinpitt:
Current Debian and ubuntu-stable images support all functionality. Restrict the test skips to ubuntu-2204.


skobyda (approved):
Thanks

PR #1357, opened 2023-07-15
fix: Screen width for podman containers

### LINKED ISSUE #1319: Screen Width for Podman Containers list
Cockpit version: 294.1
Cockpit-podman version: 71
Podman version: podman version 4.5.1
OS: Fedora Server 38

Podman Containers screen is wider than the overall app width. This makes the containers list run off the screen to the right. No scrollbars are present and one must select text and drag the screen to the right to see the container commands list. This happens in Safari and Chrome.

Steps to reproduce

1. Create a container with a very long command.
2. Observe container list that is too wide.
3. Example command is a Cloudflared Tunnel (then has been purposely changed to invalid but same length):
4. tunnel --no-autoupdate run --token eyJhIjoiNsdfsdfsfdsdffmYTk4YzZlNDA3MWIasdfadcdsfsdfsfsddsverlYmZhMjgtZDMyMSsdfsfdZGUxYWNlOGQyIiwicyI6Ik9UTmlfsdfssdfsfsdfTm1NM0xUaGhNelV0WVRKa05XWmxNVsdfsfsfsdfsdfsfsdfsdsdfsdfdfsfsdfsfsdfsdfsdfsdfsdsdsdf

image

shubhit7:
Fixes #1319

### DIFF
--- a/src/Containers.scss
+++ b/src/Containers.scss
@@ -1,6 +1,8 @@
 @import "global-variables";
 
 .container-pod {
+    overflow-x: auto;
+
     .pf-v5-c-card__header {
         border-color: #ddd;
         padding-top: var(--pf-v5-global--spacer--md);
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -27,7 +27,7 @@
 .container-block {
     display: flex;
     flex-direction: column;
-    word-break: break-all
+    word-break: break-all;
 }
 
 .container-block small {
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -27,6 +27,7 @@
 .container-block {
     display: flex;
     flex-direction: column;
+    word-break: break-all
 }
 
 .container-block small {


martinpitt:
Thanks for looking into this! This fails to build because of the

>   ✖  30:25  Expected a trailing semicolon  declaration-block-trailing-semicolon

ESLint failure. But moreover, it does not actually seem to fix the problem, see https://github.com/cockpit-project/cockpit-podman/issues/1319#issuecomment-1637388475 . Does that make any actual difference for you?

shubhit7:
> Thanks for looking into this! This fails to build because of the
> 
> > heavy_multiplication_x  30:25  Expected a trailing semicolon  declaration-block-trailing-semicolon
> 
> ESLint failure. But moreover, it does not actually seem to fix the problem, see [#1319 (comment)](https://github.com/cockpit-project/cockpit-podman/issues/1319#issuecomment-1637388475) . Does that make any actual difference for you?

Yes, it does. But the issue on smaller screen is not related to the long command/ container names. This will require a different approach.
see https://github.com/cockpit-project/cockpit-podman/issues/1319#issuecomment-1637928405

Making PR draft for now.

martinpitt (approved):
Thanks! This still leaves some broken screen widths:

image

But it improves this a lot!

PR #1359, opened 2023-07-19
test: extend rename dialog test coverage

jelly:
Test the error conditions, and submitting the rename modal via an enter key.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -2551,9 +2551,30 @@ class TestApplication(testlib.MachineCase):
 
         # the container name should be in the "Rename container" header
         b.wait_in_text("#pf-modal-part-1", container_name)
+        b.set_input_text("#rename-dialog-container-name", "")
+        b.wait_in_text("#commit-dialog-image-name-helper", "Container name is required")
+        b.set_input_text("#rename-dialog-container-name", "banana???")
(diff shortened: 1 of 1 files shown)


jelly:
Should cover most of https://cockpit-logs.us-east-1.linodeobjects.com/pull-1354-20230713-111536-64696d17-fedora-38-devel/Coverage/src/ContainerRenameModal.jsx.gcov.html

jelly:
91% coverage now :partying_face:

martinpitt (dismissed):
> this is basically "unreachable" code

Right, then I suppose it could just be removed. The test clearly covers this case. I leave it up to you if you want to do that here (I'm happy to re-approve quickly) or in a follow-up. Thanks!

jelly:
> > this is basically "unreachable" code
> 
> Right, then I suppose it could just be removed. The test clearly covers this case. I leave it up to you if you want to do that here (I'm happy to re-approve quickly) or in a follow-up. Thanks!

But it isn't dead code! If you submit via the enter key we should handle that as well.

jelly:
> [Noooice!](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1359-20230720-071540-95dfbc7e-fedora-38-devel/Coverage/src/ContainerRenameModal.jsx.gcov.html)

Need to discuss improving the `not-covered` with @mvollmer as it would be nice if it could recognize blocks.

PR #1363, opened 2023-07-26
build.js: support flags

jelly:
Adds flags to build.js to use rsync, disable linting and use watch mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ $(DIST_TEST): $(COCKPIT_REPO_STAMP) $(shell find src/ -type f) package.json buil
 	$(MAKE) package-lock.json && NODE_ENV=$(NODE_ENV) ./build.js
 
 watch: $(NODE_MODULES_TEST)
-	NODE_ENV=$(NODE_ENV) ESBUILD_WATCH=true ./build.js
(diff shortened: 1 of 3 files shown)


martinpitt (approved):
Danjkewel!

martinpitt (changes requested):
Wait, what about the README bits?

martinpitt (approved):
Ah, fair enough.

martinpitt:
Haha -- I was about to write "let's use build.js everywhere" and then reconsidered that the Make rules also ensure all the dependencies, like test/common and pkg/lib. Well *shrug* either way.

jelly:
> Haha -- I was about to write "let's use build.js everywhere" and then reconsidered that the Make rules also ensure all the dependencies, like test/common and pkg/lib. Well _shrug_ either way.

Indeed it does handle that. It's getting late here, we can decide this tomorrow :)

chabad360:
> "let's use build.js everywhere"

I would like to mention here that in my docker port I added some extra lines to fudge away issues with the package.json not having the same contents as it's equivalent in the node_modules repo. I would hope that functionality doesn't become much more complicated with build.js (TBF, I haven't really looked over build.js at all, maybe it's easy to port).

PR #1365, opened 2023-07-31
packit: Build COPR for main commits

martinpitt:
Build every commit into main into
https://copr.fedorainfracloud.org/coprs/g/cockpit/main-builds/

We will use that for cross-project testing (e.g. running cockpit-podman's tests in podman upstream PRs). Users can also install that to test bug fixes and new features after they landed, but before they got released.

---

See https://github.com/cockpit-project/cockpit/pull/19117#issuecomment-1657895111 for the plan.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -26,7 +26,7 @@ jobs:
     - centos-stream-9-x86_64
     - centos-stream-9-aarch64
     - centos-stream-8-x86_64
-  
+
   - job: tests
     trigger: pull_request
     targets:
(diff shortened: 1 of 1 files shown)


martinpitt:
@lachmanfrantisek,  @lbarcziova : I'm a bit confused here. This already triggered rpm-build:main-build-copr* tasks, but this PR didn't land yet. Shouldn't this only trigger *after* a PR lands? I did set `branch: main` in packit.yaml, did I screw this up somehow? I thought I followed https://packit.dev/docs/configuration/upstream/copr_build#optional-parameters

lbarcziova:
@martinpitt I think the problem here is that the source branch of this PR is named `main-*` and we do `re.match()` when checking the match in the configuration, therefore there are additional checks for the `main-build-copr` branch

martinpitt:
@lbarcziova Erk, right; If that's a regexp, I just make it tighter. TIL!

martinpitt:
[COPR build succeeded](https://copr.fedorainfracloud.org/coprs/g/cockpit/main-builds/build/6224947/), so switching back to the tight and correct `branch:`.

martinpitt:
Cool, I [got a build](https://copr.fedorainfracloud.org/coprs/g/cockpit/main-builds/build/6225065/)! :+1:

PR #1371, opened 2023-08-04
Fix undefined property checks

martinpitt:
In both places we already know that `container` is defined. What we meant in ContainerDetails.jsx was to guard against a missing `.State` object.

Spotted by coverity.


jelly (approved):
Tests agree!

PR #1377, opened 2023-08-18
packaging: Remove po/LINGUAS during package cleaning

martinpitt:
This would normally be done during `make clean`, but that is deliberately skipped. Remove LINGUAS to avoid an unclean source after a binary package build.

https://bugs.debian.org/1043810


skobyda:
centos-stream-8 fails but that's unrelated race: `(src/bridge/cockpithttpstream.c:942):cockpit_http_stream_close: code should not be reached`

skobyda (approved):

PR #1379, opened 2023-08-21
Use DynamicListForm from pkg/lib

skobyda:
DynamicListForm was moved to pkg/lib at https://github.com/cockpit-project/cockpit/pull/19204


jelly (approved):
less code more coverage :)

PR #1392, opened 2023-09-05
workflows: Force nightly trigger

martinpitt:
It may well happen that nothing has landed on main since the last time the nightly run happened. Disable the "isn't in triggerable state (is: success)" check.

---

This [failed last night](https://github.com/cockpit-project/cockpit-podman/actions/runs/6079229511/job/16491455998) as I already manually triggered it yesterday evening.

I [triggered a run from this branch](https://github.com/cockpit-project/cockpit-podman/actions/runs/6080223985), it succeeded and created a status. It's also visible in this PR, as it's the same SHA that I triggered the workflow on.


jelly (approved):
Good idea, totally forgot about that edge case.

PR #1393, opened 2023-09-05
Support the updates-testing scenario in `test/run`

jelly:
Slowly we are getting there, sorry I forgot that we have a custom `test/run` per project while I should have known and noted in my PR original Cockpit PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,24 +154,12 @@ rpm: $(TARFILE)
 	find tmp/rpmbuild -name '*.rpm' -printf '%f\n' -exec mv {} . \;
 	rm -r tmp/rpmbuild
 
-# pybridge scenario: build and install the python bridge from cockpit repo
-ifeq ("$(TEST_SCENARIO)","pybridge")
-COCKPIT_PYBRIDGE_REF = main
-COCKPIT_WHEEL = cockpit-0-py3-none-any.whl
-
-$(COCKPIT_WHEEL):
-	pip wheel git+https://github.com/cockpit-project/cockpit.git@${COCKPIT_PYBRIDGE_REF}
-
-VM_DEPENDS = $(COCKPIT_WHEEL)
-VM_CUSTOMIZE_FLAGS += --install $(COCKPIT_WHEEL)
-endif
-
 ifeq ("$(TEST_SCENARIO)","updates-testing")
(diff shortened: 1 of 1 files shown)


jelly:
Hmm I do see `dnf -y update` but I don't see any updates output https://cockpit-logs.us-east-1.linodeobjects.com/pull-1393-20230905-074549-c169a45a-fedora-38-updates-testing/log

martinpitt:
It feels a little weird to me to do it that way -- When accidentally not dropping --fresh, it would silently succeed. Makefile already handles the pybridge scenario, why not updates-testing as well? That's easier to reproduce locally as a developer as well.

That said, fine to land of course -- so far the running test looks good.

jelly:
> It feels a little weird to me to do it that way -- When accidentally not dropping --fresh, it would silently succeed. Makefile already handles the pybridge scenario, why not updates-testing as well? That's easier to reproduce locally as a developer as well.
> 
> That said, fine to land of course -- so far the running test looks good.

Yeah feels a bit hackish right? I'm happy to improve this, I tried using `VM_CUSTOMIZE_FLAGS but for some reason that seems to clash with the normal `image-customize`. I'm happy to wait till I have better internet so I can debug this better.

martinpitt:
I meant, just do this *entirely* in Makefile, just like it handles /pybridge.

jelly:
> I meant, just do this _entirely_ in Makefile, just like it handles /pybridge.

Sure, but what I am wondering is if this requires us to run `vm-customize` twice or can it be done with one invocation?

martinpitt:
One is enough, you can string arbitrarily many options together. It's a trade-off between being faster, and being able to run parts of your setup with `--no-network`.

Makefile
```diff
@@ -164,6 +166,10 @@ VM_DEPENDS = $(COCKPIT_WHEEL)
 VM_CUSTOMIZE_FLAGS = --install $(COCKPIT_WHEEL)
```
martinpitt, line 166: This now should also get `--no-network`

martinpitt (changes requested):
Looks nice! just one detail

jelly:
We never tested centos-8-stream so this is expected to fail

martinpitt:
@jelly please add a second commit to drop the pybridge scenario. Cheers!

martinpitt (approved):
Cheers!

PR #1406, opened 2023-09-13
Allow deleting intermediate images

### LINKED ISSUE #1320: Cannot delete intermediate images without name (tagged as "<none>:<none>")
Cockpit version: 294.1
Cockpit-podman version: 71
Podman version: 4.5.1

OS: Fedora 38

I cannot delete in podman-cokpit images without name, or displayed as "< none>:< none>". When i click in menu's image on delete, it shows the confirm dialog but no tag appear, and i cannot click on the delete button which is disabled. I can delete these images by command line : podman rmi "ID of the image". I think it will be better to use ID images than tag/name. This issue was already present in the previous versions.

Steps to reproduce

1. Click on the menu right of any images named as "< none>:< none>"
2. Click on delete
3. Cannot click on delete button on the dialog box as if none image is selected.

subhoghoshX:
fixes #1320

### DIFF
--- a/src/ImageDeleteModal.jsx
+++ b/src/ImageDeleteModal.jsx
@@ -24,6 +24,7 @@ function sortTags(a, b) {
 export const ImageDeleteModal = ({ imageWillDelete, onAddNotification }) => {
     const Dialogs = useDialogs();
     const repoTags = imageWillDelete.RepoTags ? imageWillDelete.RepoTags : [];
+    const isIntermediateImage = repoTags.length === 0;
 
     const [tags, setTags] = useState(repoTags.sort(sortTags).reduce((acc, item, i) => {
         acc[item] = (i === 0);
@@ -55,7 +56,7 @@ export const ImageDeleteModal = ({ imageWillDelete, onAddNotification }) => {
         if (all)
             client.delImage(imageWillDelete.isSystem, imageWillDelete.Id, false)
                     .catch(ex => {
-                        Dialogs.show(<ForceRemoveModal name={imageWillDelete.RepoTags[0]}
+                        Dialogs.show(<ForceRemoveModal name={isIntermediateImage ? _("intermediate image") : repoTags[0]}
                                                        handleForceRemove={handleForceRemoveImage}
                                                        reason={ex.message} />);
                     });
@@ -75,7 +76,7 @@ export const ImageDeleteModal = ({ imageWillDelete, onAddNotification }) => {
         }
     };
 
-    const imageName = repoTags[0]?.split(":")[0].split("/").at(-1) ?? "";
+    const imageName = repoTags[0]?.split(":")[0].split("/").at(-1) ?? _("intermediate");
 
     let isAllSelected = null;
     if (checkedTags.length === repoTags.length)
@@ -90,9 +91,9 @@ export const ImageDeleteModal = ({ imageWillDelete, onAddNotification }) => {
                  onClose={Dialogs.close}
                  title={cockpit.format(_("Delete $0 image?"), imageName)}
                  footer={<>
-                     <Button id="btn-img-delete" variant="danger" isDisabled={checkedTags.length === 0}
+                     <Button id="btn-img-delete" variant="danger" isDisabled={!isIntermediateImage && checkedTags.length === 0}
                              onClick={() => handleRemoveImage(checkedTags, checkedTags.length === repoTags.length)}>
-                         {_("Delete tagged images")}
+                         {isIntermediateImage ? _("Delete image") : _("Delete tagged images")}
                      </Button>
                      <Button variant="link" onClick={Dialogs.close}>{_("Cancel")}</Button>
                  </>}
--- a/test/check-application
+++ b/test/check-application
@@ -880,6 +880,29 @@ class TestApplication(testlib.MachineCase):
         b.click(".pf-v5-c-modal-box .btn-cancel")
         b.wait_not_present(".pf-v5-c-modal-box")
 
+        # Delete intermediate images
+        intermediate_image_sel = "#containers-images tbody:last-child:contains('<none>:<none>')"
+        b.click(".listing-action button:contains('Show intermediate images')")
+        b.click(intermediate_image_sel + " .pf-v5-c-dropdown__toggle")
+        b.click(intermediate_image_sel + " button.btn-delete")
+        self.confirm_modal("Delete")
+        b.wait_not_present(intermediate_image_sel)
+        # Delete intermediate image which is in use
+        tmpdir = self.execute(auth, "mktemp -d").strip()
+        self.execute(auth, f"echo 'FROM {IMG_REGISTRY}\nRUN ls' > {tmpdir}/Dockerfile")
+        IMG_INTERMEDIATE = 'localhost/test-intermediate'
+        self.execute(auth, f"podman build -t {IMG_INTERMEDIATE} {tmpdir}")
+        b.click(f'#containers-images tbody tr:contains("{IMG_INTERMEDIATE}") .ct-container-create')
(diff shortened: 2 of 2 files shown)


jelly:
This feels like an anti feature, intermediate images are likely used by something and can't be removed usually like:

```
[jelle@t14s][~]%podman rmi cab04ad2eb6d
Error: image used by 2f0a0f7f5c4ff8f09dd7efcf214b56238fa9bbe35d20e30b291580e6dc66bb61: image is in use by a container: consider listing external containers and force-removing image
```

chabad360:
Is there a way to better indicate that it's an intermediate image (other than `<none>:<none>`)? Perhaps with it being greyed out, or being a thinner item? Or maybe hiding it behind the parent images somehow?

I would think the desire to delete those comes from the fact that it doesn't look like it's there for any reason other than to take up space. So providing a better indicator that they can't/shouldn't be deleted would likely avoid this issue entirely.

jelly:
> Is there a way to better indicate that it's an intermediate image/undelete-able(other than `<none>:<none>`)? Perhaps with it being greyed out, or being a thinner item? Or maybe hiding it behind the parent images somehow?

It is already hidden by default (under `show intermediate images`)

> I would think the desire to delete those comes from the fact that it doesn't look like it's there for any reason other than to take up space. So providing a better indicator that they can't/shouldn't be deleted would likely avoid this issue entirely.

We do show that they are used by running containers? Maybe it can show that it's a child of X.

subhoghoshX:
Fixed the error! It was just trying to get the name for ForceRemoveModal out of a `null` value. I ended up overthinking things for no reason.

The dialog for Remove and Force Remove indeed looks a bit empty, both the title (as there's no tags) and body.

image

image

The format of the title is changed a little in #1451, I should land that first...

martinpitt (commented):
Thanks! LGTM, but I'll leave the last word to @jelly

jelly:
> Thanks! LGTM, but I'll leave the last word to @jelly

I was under the impression that we should first land https://github.com/cockpit-project/cockpit-podman/pull/1451

jelly:
So I thought we should show something more then this for an intermediate image? Should we change it to delete intermediate image?

image

This is the same behavior as normal images so this is acceptable. I wonder if we should warn upfront however as we know if it's un use.
image

jelly (changes requested):

test/check-application
```diff
@@ -877,6 +877,15 @@ class TestApplication(testlib.MachineCase):
...
         b.wait_not_present(".pf-v5-c-select__menu-item:contains('none')")
+        b.click("button:contains('Cancel')")
+
+        # Delete intermediate images
+        intermediate_image_sel = "#containers-images tbody:last-child:contains('<none>:<none>')"
+        b.click(".listing-action button:contains('Show intermediate images')")
+        b.click(intermediate_image_sel + " .pf-v5-c-dropdown__toggle")
+        b.click(intermediate_image_sel + " button.btn-delete")
+        self.confirm_modal("Delete")
+        b.wait_not_in_text("#containers-images", intermediate_image_sel)
```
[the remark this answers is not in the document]
jelly: NVM, we already do this!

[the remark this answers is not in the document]
jelly: We only don't test if one is in use and try deleting that one.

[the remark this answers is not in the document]
subhoghoshX: Yes that's my plan as well to create the image (and also a container for force delete), before I just didn't know how to create one in test :)

subhoghoshX:
@jelly Thanks for looking into it. So you meant something like this?

image

This actually fixes an issue that I probably introduced in #1451, there is two spaces in "Delete  image?", because of the missing tag in between which would've been there otherwise.

There's also a extra space between "Delete ?"

image

jelly:
> @jelly Thanks for looking into it. So you meant something like this?
> 
> image
> 
> This actually fixes an issue that I probably introduced in #1451, there is two spaces in "Delete image?", because of the missing tag in between which would've been there otherwise.
> 
> There's also a extra space between "Delete ?"
> 
> image

Good point, if you can fix that, it would be great! But that should show intermediate image as well right?

src/ForceRemoveModal.jsx
```diff
@@ -6,7 +6,7 @@ import cockpit from 'cockpit';
 
 const _ = cockpit.gettext;
 
-const ForceRemoveModal = ({ name, reason, handleForceRemove }) => {
+const ForceRemoveModal = ({ name = "intermediate image", reason, handleForceRemove }) => {
```
subhoghoshX, line 9: I'm not sure whether I should translate it, it's used in `cockpit.format(_("Delete $0?"), name)` down below. It is automatically translated when used this way, right? I'm confused.

subhoghoshX, line 9: I'm not sure whether I should translate it, it's used in `cockpit.format(_("Delete $0?"), name)` down below. It is automatically translated when used this way, right? I'm confused.

src/ImageDeleteModal.jsx
```diff
@@ -75,7 +76,7 @@ export const ImageDeleteModal = ({ imageWillDelete, onAddNotification }) => {
         }
     };
 
-    const imageName = repoTags[0]?.split(":")[0].split("/").at(-1) ?? "";
+    const imageName = repoTags[0]?.split(":")[0].split("/").at(-1) ?? "intermediate";
```
subhoghoshX, line 79: Same here.

subhoghoshX:
Oh it has conflicts! I'll fix it tomorrow.

src/ImageDeleteModal.jsx
```diff
@@ -55,7 +56,7 @@ export const ImageDeleteModal = ({ imageWillDelete, onAddNotification }) => {
         if (all)
             client.delImage(imageWillDelete.isSystem, imageWillDelete.Id, false)
                     .catch(ex => {
-                        Dialogs.show(<ForceRemoveModal name={imageWillDelete.RepoTags[0]}
+                        Dialogs.show(<ForceRemoveModal name={repoTags[0] ?? _("intermediate image")}
```
cockpituous, line 59: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1406-20231121-080751-114b2988-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

jelly, line 59: Here we should already know about `isIntermediateImage` so I would prefer using that.

jelly (changes requested):
small fixup, then it's good

subhoghoshX:
All green :rocket: for some reason the tests are failing locally with `TEST_SHOW_BROWSER=true`, but I trust the CI.

jelly (approved):

test/check-application
```diff
@@ -880,6 +880,27 @@ class TestApplication(testlib.MachineCase):
...
+        IMG_INTERMEDIATE = 'localhost/test-intermediate'
+        self.execute(auth, f"podman build -t {IMG_INTERMEDIATE} {tmpdir}")
+        b.click(f'#containers-images tbody tr:contains("{IMG_INTERMEDIATE}") .ct-container-create')
+        b.wait_visible('div.pf-v5-c-modal-box header:contains("Create container")')
+        b.click("#create-image-create-btn")
+        self.execute(auth, f"podman untag {IMG_INTERMEDIATE}")
+        b.click(intermediate_image_sel + " .pf-v5-c-dropdown__toggle")
+        b.click(intermediate_image_sel + " button.btn-delete")
+        self.confirm_modal("Delete")
+        self.confirm_modal("Force delete")
```
jelly, line 902: We don't confirm here if the image was gone and container was gone.

jelly (changes requested):

jelly (approved):
Thanks!

PR #1409, opened 2023-09-14
Show toolbox & distrobox containers different in containers listing

### LINKED ISSUE #865: Show toolbox containers different in the containers listing
For power users who use toolbox and cockpit-podman on their machine, it might be nice to show the toolbox container different in our containers listing before accidentally removing them.

Official toolbox containers have specific labels to identify them as toolbox containers:
```
[jelle@t14s][~/projects/cockpit]%podman inspect d28e01ad2c37 | jq '.[].Config.Labels'
{
  "architecture": "x86_64",
  "authoritative-source-url": "registry.fedoraproject.org",
  "build-date": "2021-11-02T03:13:49.622981",
  "com.github.containers.toolbox": "true",
  "com.github.debarshiray.toolbox": "true",
  "com.redhat.build-host": "osbs-node02.iad2.fedoraproject.org",
  "com.redhat.component": "fedora-toolbox",
  "distribution-scope": "public",
  "license": "MIT",
  "maintainer": "Debarshi Ray <rishi@fedoraproject.org>",
  "name": "fedora-toolbox",
  "release": "12",
  "summary": "Base image for creating Fedora toolbox containers",
  "usage": "This image is meant to be used with the toolbox command",
  "vcs-ref": "c637c4bfa90fb54cee97f7e942174f908857200d",
  "vcs-type": "git",
  "vendor": "Fedora Project",
  "version": "35"
}
```

subhoghoshX:
fixes #865

image

---

# Show toolbox & distrobox containers different in containers listing

image

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -375,6 +375,8 @@ class Containers extends React.Component {
     renderRow(containersStats, container, localImages) {
         const containerStats = containersStats[container.Id + container.isSystem.toString()];
         const image = container.ImageName;
+        const isToolboxContainer = container.Config?.Labels?.["com.github.containers.toolbox"] === "true";
+        const isDistroboxContainer = container.Config?.Labels?.manager === "distrobox";
         let localized_health = null;
 
         // this needs to get along with stub containers from image run dialog, where most properties don't exist yet
@@ -397,7 +399,11 @@ class Containers extends React.Component {
         }
         const info_block = (
             <div className="container-block">
-                <span className="container-name">{container.Name}</span>
+                <Flex alignItems={{ default: 'alignItemsCenter' }}>
+                    <span className="container-name">{container.Name}</span>
+                    {isToolboxContainer && <Badge className='ct-badge-toolbox'>toolbox</Badge>}
+                    {isDistroboxContainer && <Badge className='ct-badge-distrobox'>distrobox</Badge>}
+                </Flex>
                 <small>{image}</small>
                 <small>{utils.quote_cmdline(container.Config?.Cmd)}</small>
             </div>
--- a/src/podman.scss
+++ b/src/podman.scss
@@ -69,6 +69,26 @@
   color: white;
 }
 
+.ct-badge-toolbox {
+    background-color: var(--pf-v5-global--palette--purple-100);
+    color: var(--pf-v5-global--palette--purple-600);
+
+    .pf-v5-theme-dark & {
+        background-color: var(--pf-v5-global--palette--purple-500);
+        color: white;
+    }
+}
+
+.ct-badge-distrobox {
+    background-color: var(--pf-v5-global--palette--gold-100);
+    color: var(--pf-v5-global--palette--gold-600);
+
+    .pf-v5-theme-dark & {
+        background-color: var(--pf-v5-global--palette--gold-500);
+        color: white;
+    }
+}
+
 .green {
     color: var(--pf-v5-global--success-color--100);
 }
--- a/test/check-application
+++ b/test/check-application
@@ -2607,6 +2607,27 @@ class TestApplication(testlib.MachineCase):
         for i in range(31):
             self.execute(True, f"podman rm -f container{i}")
 
+    def testSpecialContainers(self):
+        m = self.machine
+        b = self.browser
+
+        toolbox_label = "com.github.containers.toolbox=true"
+        distrobox_label = "manager=distrobox"
+
+        container_1_id = m.execute(f"podman run -d --name container_1 -l {toolbox_label} {IMG_BUSYBOX}").strip()
+        container_2_id = m.execute(f"podman run -d --name container_2 -l {distrobox_label} {IMG_BUSYBOX}").strip()
+
+        self.login()
+
+        self.waitContainerRow('container_1')
+        self.waitContainerRow('container_2')
+
+        container_1_sel = f"#containers-containers tbody tr[data-row-id=\"{container_1_id}{'true'}\"]"
+        container_2_sel = f"#containers-containers tbody tr[data-row-id=\"{container_2_id}{'true'}\"]"
+
+        b.wait_visible(container_1_sel + " .ct-badge-toolbox:contains('toolbox')")
+        b.wait_visible(container_2_sel + " .ct-badge-distrobox:contains('distrobox')")
+
     def testCreatePodSystem(self):
         self._createPod(True)
 


subhoghoshX:
I wonder whether gray looks better:

image

src/Containers.jsx
```diff
@@ -397,7 +398,10 @@ class Containers extends React.Component {
         }
         const info_block = (
             <div className="container-block">
-                <span className="container-name">{container.Name}</span>
+                <Flex alignItems={{ default: 'alignItemsCenter' }}>
+                    <span className="container-name">{container.Name}</span>
+                    {isToolboxContainer && <Badge>toolbox</Badge>}
```
cockpituous, line 403: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1409-20230914-111755-35d42e9d-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

garrett:
Yeah, blue is probably too in-your-face. Grey works, but might be too understated. I wonder if any other color could work. (Green, orange, red are all probably out, as they often mean "good", "problem", and "error".)

Perhaps a light purple or light cyan?
https://www.patternfly.org/design-foundations/colors#color-palette

Can you also add distrobox support? It should be the same, but say "distrobox" instead of "toolbox". Thanks!

jelly (commented):
LGTM, but needs a design review from @garrett

jelly:
> Yeah, blue is probably too in-your-face. Grey works, but might be too understated. I wonder if any other color could work. (Green, orange, red are all probably out, as they often mean "good", "problem", and "error".)
> 
> Perhaps a light purple or light cyan? https://www.patternfly.org/design-foundations/colors#color-palette
> 
> Can you also add distrobox support? It should be the same, but say "distrobox" instead of "toolbox". Thanks!

Seems it uses a different label? https://github.com/89luca89/distrobox/blob/3099d53a71ea5511eec23e83c33e25c032729e4f/distrobox-create#L579

jelly:
A test would be great, just create a container with a label which matches toolbox's behaviour

src/Containers.jsx
cockpituous, line 403: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1409-20230915-112344-e0e65ca9-fedora-38-devel//Coverage/lcov/github-pr.diff.gcov.html)

subhoghoshX:
Now it looks like this (I went with purple for toolbox and gold for distrobox, which is very loosely based on their brand colors):

image

garrett:
Previously, I was thinking that we'd just use one color, because how many people use both toolbox _and_ distrobox? I mean, I do, and perhaps others on the team as well... but we're probably not that ordinary. :wink: 

But this works for me.

My one concern is the contrast ratio for the colors, primarily the gold color.

If I plug the colors into https://accessible-colors.com/ — I get this:

image

Which means we either need to make the background or foreground darker to have the correct color contrast ratio, so people can more easily read it.

The purple also has the same problem, but not as much:

image

We'll want to stick with the PF color palette (extended), so moving a shade down for the background colors of each would work... Or moving the shade up and flipping the text to dark instead of light.

(At the size and weight do matter, but not at this size or smaller, which this is. For the purple to pass without color changes, it'd have to be 18px and bold... which we definitely don't want to do here.)

https://www.patternfly.org/design-foundations/colors/#color-palette

subhoghoshX:
Fixed the contrast issues @garrett, new look:

image

In dark mode it looks good:

image

PS: Using https://accessible-colors.com/ the image badge doesn't seem to have a good contrast.

subhoghoshX:
umm not sure why the tests failed...

garrett:
Ah! I was working on color choices and used the hex on that website... and made this:

image

subhoghoshX:
> Ah! I was working on color choices and used the hex on that website... and made this:
> 
> image

Thanks @garrett. So which one should we go for? Or use both? second one for light theme and first one for dark theme?

garrett:
I like the first set overall (dark on light), but if we have to do custom CSS anyway for assigning the colors, then we could do the light/dark, with the first set as light and second set as dark.

subhoghoshX:
Thanks @garrett, it indeed looks cool now.

Light:

image

Dark:

image

src/podman.scss
```diff
@@ -69,6 +69,26 @@
...
+
+.pf-v5-theme-dark .ct-badge-toolbox {
+    background-color: var(--pf-v5-global--palette--purple-500);
+    color: white;
+}
+
+.pf-v5-theme-dark .ct-badge-distrobox {
+    background-color: var(--pf-v5-global--palette--gold-500);
+    color: white;
+}
```
garrett, line 90: As this is SCSS instead of CSS, it's also possible to write it with nesting, so the associated CSS values are grouped by type, like this:

```scss
.ct-badge-toolbox {
    background-color: var(--pf-v5-global--palette--purple-100);
    color: var(--pf-v5-global--palette--purple-600);

    .pf-v5-theme-dark & {
        background-color: var(--pf-v5-global--palette--purple-500);
        color: white;
    }
}

.ct-badge-distrobox {
    background-color: var(--pf-v5-global--palette--gold-100);
    color: var(--pf-v5-global--palette--gold-600);

    .pf-v5-theme-dark & {
        background-color: var(--pf-v5-global--palette--gold-500);
        color: white;
    }
}
```

This is also the way you'd write plain CSS with nesting, which is a new feature inspired by SCSS. https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting — but while native nesting is now available in all browsers (except Firefox extended support release), we can't use it yet. But we're using SCSS, so it's fine to use where we know it'll get compiled down. https://caniuse.com/css-nesting

It's not a big deal though; what's here in the PR is totally correct. :grin:

garrett (approved):
:+1: Looks great! Thanks for working on this.

I've added a note about an alternate way to write the SCSS (which is probably how I would've done it), but it's totally fine as-is.

subhoghoshX:
> 👍 Looks great! Thanks for working on this.
> 
> I've added a note about an alternate way to write the SCSS (which is probably how I would've done it), but it's totally fine as-is.

Makes sense. It did cross my mind once to nest it, but I brushed it off for some reason :P... fixed it! thanks again!

jelly (approved):
Thanks

PR #1414, opened 2023-09-20
.github: add dependabot integration

jelly:
Dependabot creates a PR for us so we have to introduce a new action which updates node-modules for us.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,83 @@
+name: Update node_modules
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, labeled]
+
+jobs:
+  dependabot:
+    environment: npm-update
+    permissions:
+      contents: read
+      pull-requests: write
+    timeout-minutes: 5
+    # 22.04's podman has issues with piping and causes tar errors
+    runs-on: ubuntu-20.04
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'node_modules') }}
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          fetch-depth: 0
+
+      - name: Clear node_modules label
+        uses: actions/github-script@v6
+        with:
+          script: |
+            try {
+              await github.rest.issues.removeLabel({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                name: 'node_modules'
+              });
+            } catch (e) {
+              if (e.name == 'HttpError' && e.status == 404) {
+                /* expected: 404 if label is unset */
+              } else {
+                throw e;
+              }
+            }
+
+      - name: Update node_modules for package.json changes
+        run: |
+          make tools/node-modules
+          git config --global user.name "GitHub Workflow"
+          git config --global user.email "cockpituous@cockpit-project.org"
+          eval $(ssh-agent)
+          ssh-add - <<< '${{ secrets.NODE_CACHE_DEPLOY_KEY }}'
+          ./tools/node-modules install
+          ./tools/node-modules push
+          git add node_modules
+          ssh-add -D
+          ssh-agent -k
+
+      - name: Clear no-test label
+        uses: actions/github-script@v6
+        with:
+          script: |
+            try {
+              await github.rest.issues.removeLabel({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                name: 'no-test'
+              });
+            } catch (e) {
+              if (e.name == 'HttpError' && e.status == 404) {
+                /* expected: 404 if label is unset */
+              } else {
+                throw e;
+              }
+            }
+
+      - name: Force push node_modules update
+        run: |
+          git commit --amend --no-edit node_modules
+          eval $(ssh-agent)
+          ssh-add - <<< '${{ secrets.SELF_DEPLOY_KEY }}'
+          git push --force 'git@github.com:${{ github.repository }}' '${{ github.head_ref }}'
+          ssh-add -D
+          ssh-agent -k
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "node_modules"
+      - "no-test"
+    groups:
+      eslint:
+        patterns:
+          - "eslint*"
+      esbuild:
+        patterns:
+          - "esbuild*"
+      stylelint:
+        patterns:
+          - "stylelint*"
+      patternfly:
+        patterns:
+          - "@patternfly*"


jelly:
@martinpitt any idea how I could test this on my own fork. The obvious issue is that I have no `node-modules` cache repository but that sounds easily fixable. The other issues might be more difficult.

martinpitt:
@jelly I think the easier way to test this is to do this PR from an origin branch, and tell dependabot to create PRs against that.

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Run npm-update bot
+        run: |
+          make tools/node-modules
+          git config --global user.name "GitHub Workflow"
+          git config --global user.email "cockpituous@cockpit-project.org"
+          mkdir -p ~/.config/cockpit-dev
+          echo ${{ github.token }} >> ~/.config/cockpit-dev/github-token
+          eval $(ssh-agent)
+          ssh-add - <<< '${{ secrets.NODE_CACHE_DEPLOY_KEY }}'
```
martinpitt, line 24: Redundant, the "force push" step does that.  I'd say, do the `./tools/node-modules push` in the second step (so that you keep all SSH operations in one step), or just merge them both.

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,39 @@
...
+          ./tools/node-modules install
+          ./tools/node-modules push
+          git add node_modules
+          ssh-add -D
+          ssh-agent -k
```
martinpitt, line 29: Dito. Let's not do the ssh setup and teardown twice.

martinpitt (changes requested):
Thanks! Looking forward to getting this for our git-tracked node_modules/ as well!

jelly:
> @jelly I think the easier way to test this is to do this PR from an origin branch, and tell dependabot to create PRs against that.

Hmm, actually it wasn't too hard to set up the `node-cache` but it seems the secret isn't loaded? Let's figure this out tomorrow we should have this documented somewhere as we did that for anaconda?

jelly:
How Cockpit org loads the secrets https://github.com/cockpit-project/bots/blob/main/setup-deploy-keys

.github/workflows/dependabot.yml
jelly: Uhmm there are two different deploy keys at play here, one for the `node-cache` repo and one for the pushing to the existing PR.

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,39 @@
+name: Dependabot update node-modules
+on: pull_request
+
+permissions:
+  pull-requests: write
```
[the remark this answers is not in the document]
jelly: We also need to update the PR with the ammended git submodule commit.

.github/workflows/dependabot.yml
martinpitt: Ah, I see, sorry!

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,39 @@
...
+          git add node_modules
+          ssh-add -D
+          ssh-agent -k
+
+      - name: Force push the change to trigger testing workflows
+        run: |
+          sleep 1 # make sure the timestamp changes
+          git commit --amend --no-edit
+          eval $(ssh-agent)
+          ssh-add - <<< '${{ secrets.COCKPIT_DEPLOY_KEY }}'
```
martinpitt, line 36: Our [npm-update environment](https://github.com/cockpit-project/cockpit-podman/settings/environments/508657671/edit) calls this `SELF_DEPLOY_KEY`

.github/workflows/dependabot.yml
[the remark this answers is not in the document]
martinpitt: But that happens through ssh, not https+token.

.github/workflows/dependabot.yml
jelly: Ah this was copied from Cockpit which is a bit different (annoyingly)

jelly:
> @jelly I think the easier way to test this is to do this PR from an origin branch, and tell dependabot to create PRs against that.

I can make a PR against origin, but no idea how to tell dependabot to create a PR against that?

jelly:
> > @jelly I think the easier way to test this is to do this PR from an origin branch, and tell dependabot to create PRs against that.
> 
> I can make a PR against origin, but no idea how to tell dependabot to create a PR against that?
> 
> Figured that out in #1416 with a WIP commit.

Nevermind, this won't work because of GH restrictions. This whole approach needs more work: https://stackoverflow.com/a/58740879

martinpitt:
You can tell dependabot to use a different target branch, not sure if it helps: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#target-branch

jelly:
@martinpitt after some :facepalm: I got it working, but this requires some thought:

Successful: https://github.com/jelly/cockpit-podman/actions/runs/6274863130/job/17041081841?pr=23
Updated PR: https://github.com/jelly/cockpit-podman/pull/23

So this still makes `schutzbot` unhappy. Do we want to set a `github-changes` label when we amended the commit?
Or do we not run `schutzbot` on these PR's so if author == dependabot?

Another issue is that we might not be able to rebase these PR's now, we can as `dependabot` to re-recreate the PR but I am not sure how awesome that workflow is.

Something else to think about is, should this dependabot set a `no-test` label to prevent the tests from being triggered when the PR is made and this workflow then can remove that label?

martinpitt:
Your test pr [failed](https://github.com/jelly/cockpit-podman/actions/runs/6274882473/job/17041141095?pr=23) with

> fatal: remote error: upload-pack: not our ref dc4ea5cd02749a98911984a310a5ea4c6476503e

Which smells like the workflow didn't actually push the (correct or at all) updated node_modules?

> Another issue is that we might not be able to rebase these PR's now

That ought to work fine? `on: pull_request` workflows run on force-pushes as well. What do you think would go wrong?

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,40 @@
+name: Dependabot update node-modules
+on: pull_request
```
martinpitt, line 2: How about `branches: '**dependabot**'` ? That avoids unnecessary runs for all other PRs, and should replace the `if actor` below.

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,40 @@
...
+          echo ${{ github.token }} >> ~/.config/cockpit-dev/github-token
+          eval $(ssh-agent)
+          ssh-add - <<< '${{ secrets.NODE_CACHE_DEPLOY_KEY }}'
+          ./tools/node-modules install
+          ./tools/node-modules push
+          git add node_modules
+          ssh-add -D
+          ssh-agent -k
+
+      - name: Force push the change to trigger testing workflows
```
martinpitt, line 32: This is a bit misleading: There's no push through the github token (which would skip tests). It needs to force-push to actually commit the `node_modules` change from the `git add` above.

martinpitt (changes requested):
Thanks! There's still some current threads from my earlier review, I closed the resolved ones.

.github/workflows/dependabot.yml
jelly: That could work, and allows us to keep amending and maybe getting rebasing working.

jelly:
FYI: we might want to push as dependabot:


> Looks like this PR has been edited by someone other than Dependabot. That means Dependabot can't rebase it - sorry!
>If you're happy for Dependabot to recreate it from scratch, overwriting any edits, you can request @dependabot recreate.

jelly:
An existing re-created PR: https://github.com/jelly/cockpit-podman/pull/24
A fresh new Pr: https://github.com/jelly/cockpit-podman/pull/25

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,66 @@
...
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          fetch-depth: 0
+
+      - name: Clear .github-changes label
```
jelly, line 25: fixme, `node_modules` cleared

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,85 @@
+name: Update node_modules
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, labeled]
```
martinpitt, line 4: It's not immediately clear to me why you want "opened" **and** labeled? Won't you get two parallel runs then if dependabot opens a PR with labels?

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,85 @@
...
+              });
+            } catch (e) {
+              if (e.name == 'HttpError' && e.status == 404) {
+                /* expected: 404 if label is unset */
+              } else {
+                throw e;
+              }
+            }
+
+      - name: Clear no-test label
```
martinpitt, line 44: Please move that after the node_modules generation (but before the PR force-push). It could otherwise be that tests-scan gets on this between opening the PR and while `tools/node-modules install` is running (which can easily take a minute)

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,85 @@
...
+              }
+            }
+
+      - name: Update node_modules for package.json changes
+        run: |
+          make tools/node-modules
+          git config --global user.name "GitHub Workflow"
+          git config --global user.email "cockpituous@cockpit-project.org"
+          mkdir -p ~/.config/cockpit-dev
+          echo ${{ github.token }} >> ~/.config/cockpit-dev/github-token
```
martinpitt, line 69: What does this do? tools/node-modules doesn't do any GitHub API stuff. (I asked this before, but you resolved it without an answer) AFAIK the only token-related stuff here is the label unsetting above.

martinpitt (changes requested):
Thanks! I feel the workflow is complex and subtle enough that it could do with a high-level comment at the top what it does, especially with the juggling of the two labels.

.github/workflows/dependabot.yml
jelly: Haven't seen that, and we use the same logic for `reposchutz` so this seems to go fine.

.github/workflows/dependabot.yml
jelly: Hmmm good question, let's drop it and re-try.

martinpitt (approved):
Thanks! I previously thought it was tied tightly to dependabot, but it really isn't -- in principle this works for any PR where we set the node_modules label. This is clear enough now -- I'd say, let it lose and we'll see what happens :grin: 

Not sure if @allisonkarlitskaya still wants to review this, though. Although there's nothing there that we can't change later on.

allisonkarlitskaya (approved):
Very very quick look over it and I'm happy with the approach, thanks!

PR #1418, opened 2023-09-22
packit: Enable tests on Fedora ELN

martinpitt:
This will likely require a quick discussion with our naughties, but let's first see how well this works in general.


martinpitt:
This is currently being discussed. TF doesn't offer fedora-eln at the moment, and the mere COPR build is very uninteresting for c-podman.

PR #1429, opened 2023-09-28
.github: set `[no-test]` prefix for dependabot PR's

jelly:
Dependabot sets labels after PR creation so our pull requests tests would run regardless which is a waste of resources.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,4 +1,4 @@
-name: Update node_modules
+name: update node_modules
 on:
   pull_request_target:
     types: [opened, reopened, synchronize, labeled]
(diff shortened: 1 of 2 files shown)


jelly:
> Hmm, that clears the no-test flag from the PR title, but it also needs to be removed from the actual commit when force-pushing. I assume that `commit-message:` will literally put it into the commit, not just the PR title/description?

Yup, should we put in some effort to drop it there too

.github/workflows/dependabot.yml
```diff
@@ -54,28 +54,27 @@ jobs:
...
-                name: 'no-test'
-              });
-            } catch (e) {
-              if (e.name == 'HttpError' && e.status == 404) {
-                /* expected: 404 if label is unset */
-              } else {
-                throw e;
-              }
-            }
+            const title = process.env['TITLE'].replace(/\[no-test\]\W+ /, '')
```
martinpitt, line 64: Why the `env` dance? This could just directly use `github.event.pull_request.title` surely?

martinpitt (changes requested):
Thanks! Let's clean up the title handling a bit.

But I must say, this is super ugly. Originally I thought that our API sets the labels right in the "create PR" call -- but that's [not actually supported](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#create-a-pull-request) (unlike with creating a PR interactively), and our [bots API doesn't do this](https://github.com/cockpit-project/bots/blob/main/task/__init__.py#L335) , it POSTs the new /pull and then labels it. In theory that *should* cause a race condition with our bots, as tests-scan takes the webhook event into account, not the current state. So :thinking: how does that work for us?

jelly:
> Thanks! Let's clean up the title handling a bit.
> 
> But I must say, this is super ugly. Originally I thought that our API sets the labels right in the "create PR" call -- but that's [not actually supported](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#create-a-pull-request) (unlike with creating a PR interactively), and our [bots API doesn't do this](https://github.com/cockpit-project/bots/blob/main/task/__init__.py#L335) , it POSTs the new /pull and then labels it. In theory that _should_ cause a race condition with our bots, as tests-scan takes the webhook event into account, not the current state. So 🤔 how does that work for us? In other words, what could dependabot do differently to work for us?

No no, when you manually create a PR you can set `no-test` as a label in the creation screen.

For bot workflows we set `[no-test]`.

.github/workflows/dependabot.yml
jelly: Uh, because I have no confidence in my scripting of github actions, it's frustrating to test this seems to be the most easy to get to work solution without magic. This example claims your suggestion works however https://github.com/actions/github-script#use-esm-import

martinpitt: Not sure how this is related to imports -- your script already uses the `github` object, it's just about dropping that title env and replacing it with
```js
const title = github.event.pull_request.title.replace(/\[no-test\]\W+ /, '')
```

martinpitt:
> For bot workflows we set [no-test].

Ah, thanks! that clears it up.

.github/workflows/dependabot.yml
jelly: IF only it was that easy https://github.com/jelly/cockpit-podman/actions/runs/6348462088/job/17245272304?pr=32

The example in the github-script page makes it even more of a soup:

```
            const { default: printStuff } = await import('${{ github.workspace }}/src/print-stuff.js')
```
So it seems they do some templating and then run the nodejs script.

jelly: If you don't mind we keep it this way.

martinpitt: Weird.. ok.

martinpitt (approved):
Ack, thanks!

jelly:
Also FYI, I had a discussion with @allisonkarlitskaya in Matrix, maybe we should try to improve this a bit by rewriting it into a `bots/node-modules-update` and `tools/node-modules rebase` so we can remove the `[no-test]` title in Python and more easily handle multiple commits.

But  eh, that's a lot of work. Let's wait for things to settle first.

PR #1430, opened 2023-09-29
.github: drop npm-update workflows

jelly:
Now that we handle our updates via dependabot these workflows are obsolete.


martinpitt (approved):
Dankjewel!

Issue #1408, opened 2023-09-14
Details not showing when creating a container off an existing image (of different user) and crash


subhoghoshX:
Cockpit version: xxx
Cockpit-podman version: 76-7-g149fc4f
Podman version: 4.6.2
OS: fedora-38

[Screencast from 2023-09-14 15-09-08.webm](https://github.com/cockpit-project/cockpit-podman/assets/108616679/350d7b94-f243-4197-a504-13dc467e7495)

jelly:
Can reproduce this with the tasks container, as it's a toolbox image maybe that's somehow special

jelly:
Actually since this image is so big, it's more the Integration and other tabs which don't handle the downloading Status.

PR #1452, opened 2023-10-16
Don't crash when image is being downloaded and integration/console tab is clicked

jelly:
Large images which take some time download already shown the tabs but clicking on it would result in a JavaScript error as the Container object would not yet be available.

Closes:  #1408

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -330,7 +330,7 @@ export class ImageRunModal extends React.Component {
             tempImage.isSystem = isSystem;
             tempImage.State = { Status: _("downloading") };
             tempImage.Created = new Date();
-            tempImage.Names = [tempImage.name];
+            tempImage.Name = tempImage.name;
             tempImage.Image = createConfig.image;
             tempImage.isDownloading = true;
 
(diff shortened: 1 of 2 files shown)


subhoghoshX:
So it stops the expandable. Nice! It's better than expanding and going "Ooops" I guess.

Little nitpick, the presentation of it is a little weird. I guess that's for garrett. Is it possible to have the name of the container shown in the row at least?

image

jelly:
> So it stops the expandable. Nice! It's better than expanding and going "Ooops" I guess.
> 
> I think the presentation of it is a little weird. That may be something for garrett. Is it possible to have the name of the container shown in the row at least?
> 
> image

Yes, that is unfortunate. maybe we can show it as a ListingTable without the expendable section or:

image

src/ImageRunModal.jsx
```diff
@@ -330,7 +330,7 @@ export class ImageRunModal extends React.Component {
             tempImage.isSystem = isSystem;
             tempImage.State = { Status: _("downloading") };
             tempImage.Created = new Date();
-            tempImage.Names = [tempImage.name];
+            tempImage.Name = tempImage.name;
```
subhoghoshX, line 333: I'm not sure what this does. Changing property name/type can potentially break something somewhere, unless it's not used absolutely anywhere :-)

jelly: That is expected, the property is Name in podman's API not Names.

martinpitt (approved):
Thanks!

PR #1478, opened 2023-11-07
test: update reference-image to Fedora-39

jelly:


martinpitt (approved):
That was easy! :+1: (that's not often the case these days..)

PR #1481, opened 2023-11-10
Validate hostPath based on the error callback from FileAutoComplete

jelly:
This depends on this PR in Cockpit: https://github.com/cockpit-project/cockpit/pull/19600

@martinpitt you looked at the validation PR before, mind giving this a look as well?


src/Volume.jsx
```diff
@@ -38,9 +40,11 @@ export const Volume = ({ id, item, onChange, idx, removeitem, additem, options,
             >
                 <FileAutoComplete id={id + "-host-path"}
                     value={item.hostPath || ''}
-                    onChange={value => {
+                    onChange={(value, error) => {
+                        console.log(`value: ${value}, error: ${error}`);
+                        console.log(validationFailed);
                         utils.validationClear(validationFailed, "hostPath", onValidationChange);
-                        utils.validationDebounce(() => onValidationChange({ ...validationFailed, hostPath: validateVolume(value, "hostPath") }));
+                        utils.validationDebounce(() => onValidationChange({ ...validationFailed, hostPath: validateVolume(error, "hostPath") }));
```
martinpitt, line 47: This looks strange.. is `onValidationChange()` *only* being called for errors? the code certainly doesn't look like it. Or I don't understand this change just by looking at the patch.

src/ImageRunModal.jsx
```diff
@@ -659,6 +659,7 @@ export class ImageRunModal extends React.Component {
     *            Index needs to corellate with a row number
     */
     dynamicListOnValidationChange = (value, key) => {
+        console.log('dynamic list validation', value, key);
```
jelly, line 662: debug leftover....

src/Volume.jsx
```diff
@@ -38,9 +40,11 @@ export const Volume = ({ id, item, onChange, idx, removeitem, additem, options,
             >
                 <FileAutoComplete id={id + "-host-path"}
                     value={item.hostPath || ''}
-                    onChange={value => {
+                    onChange={(value, error) => {
+                        console.log(`value: ${value}, error: ${error}`);
```
jelly, line 44: debug leftover

src/Volume.jsx
jelly: No, it is always called onChange?

martinpitt: That's what I thought, but now it doesn't even get the value passed any more?

jelly: So the validation function `validateVolume` usually takes the new value, but the problem is that when an error occurs the value is empty and this is also the initial value.

We could make `validateVolume` check if the file exists, but that is also impossible as a faulty file won't be returned. So instead I pass the `error` to the validation function.

martinpitt: Right, I see. validateVolume() does not actually do anything with it, and just passes it back to the caller. So that's fine.

martinpitt:
Okay, can you please clean up the debugging leftovers, and add an integration test? Thanks!

jelly:
> Okay, can you please clean up the debugging leftovers, and add an integration test? Thanks!

How do we deal with the dependent Cockpit PR?

jelly:
> Okay, can you please clean up the debugging leftovers, and add an integration test? Thanks!

Done, but now I found another issue. When we provide a path without a trailing slash the validation doesn't work.

PR #1490, opened 2023-11-21
test: Fix image deletion check

martinpitt:
Thanks to @subhoghoshX for spotting!

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -804,7 +804,7 @@ class TestApplication(testlib.MachineCase):
         self.assertTrue(b.get_checked(f".pf-v5-c-check__input[aria-label='{IMG_BUSYBOX}:4']"))
         self.confirm_modal("Delete")
         self.confirm_modal("Force delete")
-        b.wait_not_in_text("#containers-images", busybox_sel)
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -840,7 +840,7 @@ class TestApplication(testlib.MachineCase):
         b.click(alpine_sel + " button.btn-delete")
         self.confirm_modal("Delete")
         self.confirm_modal("Force delete")
-        b.wait_not_in_text("#containers-images", alpine_sel)
+        b.wait_not_in_text("#containers-images", IMG_ALPINE_LATEST)
```
subhoghoshX, line 843: Thanks! There another one at https://github.com/cockpit-project/cockpit-podman/blob/main/test/check-application#L807 and  I was expecting it to be more like `b.wait_not_present(alpine_sel)` because the row shouldn't be present at all.

martinpitt: Thanks, fixed that as well. And the original fix was broken too, as we have both system and user images.

subhoghoshX (approved):
Thanks!

PR #1500, opened 2023-12-06
packaging: drop Fedora 36/37 workarounds

jelly:
According to the GitHub issue this was also fixed in the packaging in F37. But as F37 is now EOL it is also fine to drop.


martinpitt:
COPR is kaputt (or clogged) today, ignoring.

PR #1527, opened 2024-01-09
Update to changed DynamicListForm import from latest cockpit

martinpitt:
Adjust code and tests for the `undefined` "holes" that the data model now gets on deletion.

---

I couldn't reproduce the state confusion from https://issues.redhat.com/browse/RHEL-19598 at least with the port mapping. This only happened in machines with the "validated" SSH  keys, where the whole element was replaced. React magic.. But interactive testing with the updated component still seems to work.

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -26,7 +26,7 @@ import rest from './rest.js';
 import cockpit from 'cockpit';
 import { onDownloadContainer, onDownloadContainerFinished } from './Containers.jsx';
 import { PublishPort, validatePublishPort } from './PublishPort.jsx';
-import { DynamicListForm } from 'DynamicListForm.jsx';
+import { DynamicListForm } from 'cockpit-components-dynamic-list.jsx';
 import { validateVolume, Volume } from './Volume.jsx';
 import { EnvVar, validateEnvVar } from './Env.jsx';
 
@@ -166,9 +166,9 @@ export class ImageRunModal extends React.Component {
             createConfig.resource_limits = resourceLimit;
         }
         createConfig.terminal = this.state.hasTTY;
-        if (this.state.publish.length > 0)
+        if (this.state.publish.some(port => port !== undefined))
             createConfig.portmappings = this.state.publish
-                    .filter(port => port.containerPort)
+                    .filter(port => port?.containerPort)
                     .map(port => {
                         const pm = { container_port: parseInt(port.containerPort), protocol: port.protocol };
                         if (port.hostPort !== null)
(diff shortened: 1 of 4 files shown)


martinpitt:
@jelly please apply a critical eye. Yesterday I wasn't aware at all that the internal state of DynamicListForm shines through that much. Perhaps you have a better idea how to rearrange this, or hide it from callers?

src/ImageRunModal.jsx
```diff
@@ -177,14 +177,17 @@ export class ImageRunModal extends React.Component {
                             pm.host_ip = port.IP;
                         return pm;
                     });
-        if (this.state.env.length > 0) {
-            const ports = {};
-            this.state.env.forEach(item => { ports[item.envKey] = item.envValue });
-            createConfig.env = ports;
```
jelly, line 183: This was a super confusing old code! Good to see it cleaned up

jelly (approved):
Thanks!

jelly:
> @jelly please apply a critical eye. Yesterday I wasn't aware at all that the internal state of DynamicListForm shines through that much. Perhaps you have a better idea how to rearrange this, or hide it from callers?

Not at the moment. What is the biggest issue? Because in essence we have three things we want:

* Dynamic List State - this should be kept in DynamicListForm
* A way to "export" values to a Form for submission, so in React land you don't query a component but update your state via a callback. That's what we do, but with an exact copy of the internal state (hence all the silly undefined checks) 
Basically:
```
        }, () => this.props.onChange(this.state.list.concat()));
```
So this could strip out undefined rows.

* A way to "validate" rows, `onValidationChange` this uses a delta. Likely not a big problem?

martinpitt:
> ```
>         }, () => this.props.onChange(this.state.list.concat()));
> ```
> 
> So this could strip out undefined rows when we emit `onChange`?

Hmm, would that actually work? I.e. can we assume that *nothing* in the caller uses the indexes in the array to change stuff? Then that'd simplify the code a lot.

jelly:
> > ```
> >         }, () => this.props.onChange(this.state.list.concat()));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So this could strip out undefined rows when we emit `onChange`?
> 
> Hmm, would that actually work? I.e. can we assume that _nothing_ in the caller uses the indexes in the array to change stuff? Then that'd simplify the code a lot.

Good question, for podman the order likely does not matter? But even if order matters, if we simply filter out `undefined` the order should remain right? (With `this.state.list.filter(x => x !== undefined))`

martinpitt:
@jelly That doesn't work. I tried that, but then the indexes that are visible to the caller again move around and don't match up with the `onValidationChange` handler. This is annoyingly complex..

PR #1540, opened 2024-01-28
Don't assume `State.Health` always exists

martinpitt:
https://github.com/containers/podman/pull/21388 drops the `State.Health` field for containers without health checks. This causes an `undefined` crash in the container sort function.

---

I reproduced the crash with https://dashboard.packit.dev/results/copr-builds/1287492 and validated it against that build.

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -645,17 +645,20 @@ class Containers extends React.Component {
             // Remove infra containers
             filtered = filtered.filter(id => !this.props.containers[id].IsInfra);
 
+            const getHealth = id => {
+                const state = this.props.containers[id]?.State;
+                return state?.Health?.Status || state?.Healthcheck?.Status;
+            };
+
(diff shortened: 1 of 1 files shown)


src/Containers.jsx
```diff
@@ -645,17 +645,20 @@ class Containers extends React.Component {
             // Remove infra containers
             filtered = filtered.filter(id => !this.props.containers[id].IsInfra);
 
+            const getHealth = id => {
+                const state = this.props.containers[id]?.State;
+                return state?.Health?.Status || state?.Healthcheck?.Status;
```
cockpituous, line 650: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1540-20240128-084429-20c044a7-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

PR #1546, opened 2024-02-02
Try out ruff

jelly:


jelly:
The new version is not yet in the tasks container.

PR #1547, opened 2024-02-05
Various test cleanups and robustifications

martinpitt:


martinpitt:
I'm now spoonfeeding the fixes here, with the most trivial ones first. [This crash](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1547-20240205-071508-d9abb32d-fedora-39/log.html#6) is important and we should fix it ASAP, and it's not related at all to the changes here.

martinpitt:
I have some [serious issues with this crash](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1547-20240205-071508-d9abb32d-fedora-39/log.html#6). It doesn't reproduce locally, and I just audited all our usages of `.includes()`, and they should all be fine -- I didn't find any place where the value could be undefined. Let's see whether it reproduces again after the rebase, then I can match the line numbers better.

jelly (approved):
Thanks!

PR #1549, opened 2024-02-07
test: Drop redundant CLI pod port binding check

martinpitt:
The "as we do not show them in the UI yet" is not true any more, the code above does just that.

This avoids assuming that `podman inspect` returns a JSON object. https://github.com/containers/podman/pull/21514 changes it to a list, so the test fails with "list indices must be integers or slices, not str" with that PR.

---

I locally validated that the test passes with [that podman version](https://dashboard.packit.dev/results/copr-builds/1310429).


martinpitt:
The [devel crash](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1549-20240207-064020-c26d8da9-fedora-39-devel/log.html) in lcov.py is known, let's ignore here as this  PR is urgent.

martinpitt:
I've retried arch a few times, but it's pretty nigh impossible to get it to green right now. [These failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1549-20240207-080328-c26d8da9-arch/log.html#8) are unrelated.

martinpitt:
I see and also fix some issues in #1547, let's address it there.

jelly (approved):

PR #1585, opened 2024-02-27
test: drop allowing cockpit-ws assertion

jelly:
This assertion should no longer appear since Cockpit 260.

---

If this is good to go, we should also clean up machines.


martinpitt (approved):
Thanks!

PR #1592, opened 2024-02-29
test: Replace setup/teardown

martinpitt:
Fixes #1591

PR #1593, opened 2024-02-29
test against new cockpit/tasks container

martinpitt:


martinpitt:
This works.

martinpitt:
[This is messy](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1593-20240229-175021-e1d050f9-fedora-40/log.html#35), but at first sight it doesn't look like the same kind of trouble that I see in #1592.

PR #1596, opened 2024-03-01
test: Make system container cleanup faster

martinpitt:
`podman system reset` defaults to waiting for 10s for containers to shut down, which unnecessarily slows down tests (see
https://github.com/containers/podman/issues/21874). To work around this, force-stop all containers first with a zero timeout.

---

Landing some parts of #1592 separately.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -113,15 +113,22 @@ class TestApplication(testlib.MachineCase):
         # backup/restore pristine podman state, so that tests can run on existing testbeds
         self.restore_dir("/var/lib/containers")
 
-        # HACK: sometimes podman leaks mounts
         self.addCleanup(m.execute, """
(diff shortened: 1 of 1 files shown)


martinpitt:
Interesting that *this* is a bit which is broken :exploding_head: [log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1596-20240301-082058-a93fe2af-fedora-39-devel/log.html) (also on other distros).

Let's cut that in half.

jelly (approved):
Neat! Wondering if system reset should have such a hammer option `podman system reset --dont-wait-on-me`

martinpitt:
@jelly I filed https://github.com/containers/podman/issues/21874 yesterday, and they seemed vaguely to agree.

PR #1597, opened 2024-03-01
test: Drop /var/lib/containers/ subdirectory unmounting

martinpitt:
podman creates an overlay mount or btrfs volume in a subdirectory there, which is legit. Let's see how far the tests get without this.


martinpitt:
I'm surprised that this seems to mostly work (of all the things, that seemed to be the most brittle one..) but it [did fall over hard on RHEL 9](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1597-20240301-083051-62a14354-rhel-9-4/log.html#36-1). 

OMG, there doesn't seem to be any way out of this hackery..

PR #1600, opened 2024-03-01
test: HACK: Add fake default route for podman-next scenario

martinpitt:
The introduction of pasta [1] regressed user containers if there is no default route [2]. While that is being sorted out, add a fake interface with a default route for our offline tests, to unbreak upstream podman COPR testing.

[1] https://github.com/containers/podman/pull/21563
[2] https://github.com/containers/podman/issues/21896


martinpitt:
TF is  having a bad day, but this doesn't affect TF runs at all, only our own CI.

jelly (approved):

PR #1601, opened 2024-03-01
test: Move from `restore_dir()` to `podman system reset` for system

martinpitt:
Similar to commit 9e3d0c724e00, avoid the `restore_dir()` and replace it with `podman system reset`. With that we can drop the unmount hack, let podman handle overlay/subvolume mounts and stop rummaging in its guts.


martinpitt:
I haven't seen cleanup related c-podman failures recently, and no time to pursue this.

PR #1605, opened 2024-03-05
Use Cockpit's Kebab dropdown

jelly:


src/Containers.jsx
```diff
@@ -243,7 +222,7 @@ const ContainerActions = ({ container, healthcheck, onAddNotification, localImag
         }
         if (container.isSystem && container.State?.CheckpointPath) {
             actions.push(
-                <DropdownSeparator key="separator-0" />,
+                <Divider key="separator-0" />,
```
cockpituous, line 225: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1605-20240306-082630-f936f937-fedora-39-devel//Coverage/lcov/github-pr.diff.gcov.html)

test/check-application
```diff
@@ -916,7 +916,7 @@ class TestApplication(testlib.MachineCase):
 
         # Delete intermediate image which is in use
         self.execute(auth, f"podman untag {IMG_INTERMEDIATE}")
-        b.click(intermediate_image_sel + " .pf-v5-c-dropdown__toggle")
+        b.click(intermediate_image_sel + " .pf-v5-c-menu-toggle")
```
jelly, line 919: Maybe we should just use a .... helper function

jelly: This becomes more and more of a yakshave as we repeat the same stuff in `deleteImage`... This feels follow up territory.

martinpitt (approved):
Niiice! and no pixel changes any more! ^5s

PR #1609, opened 2024-03-11
lint: Ignore test functions in vulture

martinpitt:
These aren't unused, but discovered/enumerated by `unittest` or `pytest`.

---

Fixes [this static-code failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6048-20240311-121551-77e2f887-ubuntu-stable-cockpit-project-cockpit-podman/log.html) after we added vulture to the tasks container.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,3 +43,8 @@ mark-parentheses = false
 
 [tool.ruff.lint.isort]
 known-first-party = ["cockpit"]
+
+[tool.vulture]
+ignore_names = [
+   "test[A-Z0-9]*",
+]


jelly (approved):

pyproject.toml
```diff
@@ -43,3 +43,8 @@ mark-parentheses = false
 
 [tool.ruff.lint.isort]
 known-first-party = ["cockpit"]
+
+[tool.vulture]
+ignore_names = [
+   "test[A-Z0-9]*",
```
allisonkarlitskaya, line 49: I'd prefer to see `test[A-Z]*` and `test_*` handled separately to avoid pulling in possibly unrelated things.

martinpitt: Then we should do that in cockpit as well, I just copied it from there.

PR #1612, opened 2024-03-12
test: Apply pasta hack to scenarios as well

### LINKED ISSUE #1611: Tests failed on 6c0463d52605079c5d826a7ccaaedf6795f8cae2
The job `fedora-39/podman-next` failed on commit 6c0463d52605079c5d826a7ccaaedf6795f8cae2.

Log: https://cockpit-logs.us-east-1.linodeobjects.com/pull-0-6c0463d5-20240312-014411-fedora-39-podman-next/log.html

martinpitt:
Commit 9279fd46f1c1a73 didn't apply the hack to the "updates-testing" and "podman-next" scenarios, as these *overwrite* `VM_CUSTOMIZE_FLAGS` to get rid of the `--offline`.

Reorganize VM_CUSTOMIZE_FLAGS a bit to deal with `--no-network` in a cleaner way.

Fixes #1611

-----

I additionally triggered the scenarios here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,21 @@ DIST_TEST=dist/manifest.json
 COCKPIT_REPO_STAMP=pkg/lib/cockpit-po-plugin.js
 # common arguments for tar, mostly to make the generated tarballs reproducible
 TAR_ARGS = --sort=name --mtime "@$(shell git show --no-patch --format='%at')" --mode=go=rX,u+rw,a-s --numeric-owner --owner=0 --group=0
-# default customize_flags
-VM_CUSTOMIZE_FLAGS = --no-network
+
+VM_CUSTOMIZE_FLAGS =
 
 # HACK: https://github.com/containers/podman/issues/21896
 VM_CUSTOMIZE_FLAGS += --run-command 'nmcli con add type dummy con-name fake ifname fake0 ip4 1.2.3.4/24 gw4 1.2.3.1 >&2'
+
+# the following scenarios need network access
+ifeq ("$(TEST_SCENARIO)","updates-testing")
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -158,11 +159,11 @@ rpm: $(TARFILE)
 	rm -r tmp/rpmbuild
 
 ifeq ("$(TEST_SCENARIO)","updates-testing")
-VM_CUSTOMIZE_FLAGS = --run-command 'dnf -y update --setopt=install_weak_deps=False --enablerepo=updates-testing >&2'
+VM_CUSTOMIZE_FLAGS = $(PASTA_HACK) --run-command 'dnf -y update --setopt=install_weak_deps=False --enablerepo=updates-testing >&2'
```
allisonkarlitskaya, line 162: Not enjoying the shadowing on this variable.  What's the story there?

martinpitt: Mostly that we don't have a `--yes-network` to override the original `--no-network`

martinpitt: we can probably do this with an `else` too, though

martinpitt: Changed, WDYT?

allisonkarlitskaya (commented):
Hate to do this, but can you split this into two commits, one for the cleanup of the variable handling and one which we can revert later when podman gets fixed?

martinpitt:
@allisonkarlitskaya I'm afraid that's not quite possible any more, as the hack was already introduced in commit cecb2cc6e8f24 and modified in commit 9279fd46f1c1a737a . But I cleaned it up once more so that we only need to drop a single line (plus the comment) when it gets fixed.

allisonkarlitskaya (approved):
This looks a lot nicer, indeed.  Thanks!

PR #1614, opened 2024-03-13
test: avoid a double wait_visible

jelly:
The click and wait_in_text function already call wait_visible so we end up waiting twice.


martinpitt (approved):
Thanks!

Issue #1622, opened 2024-03-15
Environment variable validation


jelly:
Some issues with validating env variables:

* We allow all characters, even invalid ones. The regex should likely be `[a-zA-Z_]{1,}[a-zA-Z0-9_]*`
* Empty value is allowed, but that likely makes no sense?
* Removing an invalid entry shifts the invalidate state to the valid one, see below:

![image](https://github.com/cockpit-project/cockpit-podman/assets/67428/d2a74953-9b5e-4c64-bd3a-e1ae2e8b879a)

Then remove the top one:

![image](https://github.com/cockpit-project/cockpit-podman/assets/67428/18e7758a-18ee-4efb-80d7-08b0eddb3e4c)

jelly:
The last issue is due to the validationState being wrong compared to `this.state.env`

```
[
    null,
    null,
    null,
    null,
    null,
    {
        "envKey": "Key must not be empty"
    },
    {}
]
```

While `this.state.env` is:

```
[
    null,
    null,
    null,
    null,
    null,
    null,
    {
        "key": 6,
        "envKey": null,
        "envValue": null
    },
    {
        "key": 7,
        "envKey": null,
        "envValue": null
    },
    {
        "key": 8,
        "envKey": "jijij",
        "envValue": null
    }
]
```

While the UI shows:

![image](https://github.com/cockpit-project/cockpit-podman/assets/67428/3e18bbb1-1cea-4c48-b111-15064abae3f4)

PR #1623, opened 2024-03-15
Handle undefined state entries

### LINKED ISSUE #1619: Removing Environment variables when creating a container breaks validation
Cockpit version: 312-1
Cockpit-podman version: 84.1-1
Podman version: 84.1-1
OS: Arch Linux

When creating a container and adding and subsequently removing environment variables, the container will not create and will not produce an error.

Steps to reproduce

1. Start container creation
2. Add environment variables
3. Remove said environment variables
4. Click create

An error occurs in the console while validating the variables as they are undefined instead of removed

image

image

I can't tell whether 85 fixes this, it's not in extra yet

jelly:
The dynamiclistform only grows and removes entries by setting them to undefined which we need to account for. Previously this would map() and return [undefined, obj] but that raises an exception when you call Object.keys(undefined).

Fixes #1619

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -626,7 +626,7 @@ export class ImageRunModal extends React.Component {
                 containerPort: validatePublishPort(a.containerPort, "containerPort"),
             };
         });
-        if (publishValidation.some(entry => Object.keys(entry).length > 0))
+        if (publishValidation.some(entry => entry && Object.keys(entry).length > 0))
             validationFailed.publish = publishValidation;
 
         const volumesValidation = volumes.map(a => {
@@ -638,7 +638,7 @@ export class ImageRunModal extends React.Component {
                 containerPath: validateVolume(a.containerPath, "containerPath"),
             };
         });
-        if (volumesValidation.some(entry => Object.keys(entry).length > 0))
+        if (volumesValidation.some(entry => entry && Object.keys(entry).length > 0))
             validationFailed.volumes = volumesValidation;
 
         const envValidation = env.map(a => {
@@ -650,7 +650,7 @@ export class ImageRunModal extends React.Component {
                 envValue: validateEnvVar(a.envValue, "envValue"),
             };
         });
-        if (envValidation.some(entry => Object.keys(entry).length > 0))
+        if (envValidation.some(entry => entry && Object.keys(entry).length > 0))
             validationFailed.env = envValidation;
 
         const containerNameValidation = await this.validateContainerName(containerName);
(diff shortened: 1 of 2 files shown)


jelly:
There are some things I want to fix in a follow up:

https://github.com/cockpit-project/cockpit-podman/issues/1622#issue-2188051305

martinpitt (approved):
Ah, I remember that pain in machines. Thanks!

fedora/rawhide is known broken (https://github.com/containers/podman/issues/21896), but the others should work, retrying.

jelly:
> Ah, I remember that pain in machines. Thanks!
> 
> fedora/rawhide is known broken ([containers/podman#21896](https://github.com/containers/podman/issues/21896)), but the others should work, retrying.

There is some serious nonsense going on here, I had to add that podman inspect as waiting on the container to appear breaks the test because it opens the "Create container modal" too fast. This is very annoying imo, especially because waiting for a container to be "Running" requires doing an inspect with `waitContainer`

```
Killing browser (pid 665418)
  File "/home/jelle/projects/cockpit-podman/./test/check-application", line 2393, in testCreateContainerValidation
    b.click(f'#containers-images tbody tr:contains("{IMG_BUSYBOX}") .ct-container-create')
  File "/home/jelle/projects/cockpit-podman/test/common/testlib.py", line 412, in click
    self.mouse(selector + ":not([disabled]):not([aria-disabled=true])", "click", 0, 0, 0)
  File "/home/jelle/projects/cockpit-podman/test/common/testlib.py", line 405, in mouse
    self.call_js_func('ph_mouse', selector, event, x, y, btn, ctrlKey, shiftKey, altKey, metaKey)
  File "/home/jelle/projects/cockpit-podman/test/common/testlib.py", line 358, in call_js_func
    return self.eval_js("%s(%s)" % (func, ','.join(map(jsquote, args))))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jelle/projects/cockpit-podman/test/common/testlib.py", line 336, in eval_js
    result = self.cdp.invoke("Runtime.evaluate", expression=code, trace=code,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jelle/projects/cockpit-podman/test/common/cdp.py", line 204, in invoke
    res = self.command(cmd)
          ^^^^^^^^^^^^^^^^^
  File "/home/jelle/projects/cockpit-podman/test/common/cdp.py", line 231, in command
    raise RuntimeError(res["error"])
RuntimeError: TypeError: Converting circular structure to JSON
```

PR #1624, opened 2024-03-15
test: wait on the dialog to close 

jelly:
podman inspect seems to fail randomly when we just created a container, likely because it is still "setting up". As we try to work around the issue where we click too fast on the "create container" button while the dialog is still tearing down, we aren't too interested in much of podman inspect more in if the container is running.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -2333,9 +2333,9 @@ class TestApplication(testlib.MachineCase):
         # Changing the port should allow creation of container
         b.set_input_text('#run-image-dialog-publish-0-host-port', '5001')
         b.click('.pf-v5-c-modal-box__footer #create-image-create-run-btn')
+        b.wait_not_present("#run-image-dialog-name")
+
         self.waitContainerRow(container_name)
-        container_sha = self.execute(False, f"podman inspect --format '{{{{.Id}}}}' {container_name}").strip()
-        self.waitContainer(container_sha, False, name=container_name, image=IMG_BUSYBOX, state='Running')
 
         # Test validation JavaScript errors when removing invalid environment entries
         container_name = 'env-var-validation'


martinpitt:
TBH I don't understand  the commit message. How is that related to a dialog and tearing down? In that case we should wait for the dialog to go away -- that's sorely missing and should be done in any case. Then the container should be guaranteed to run.

jelly:
> TBH I don't understand the commit message. How is that related to a dialog and tearing down? In that case we should wait for the dialog to go away -- that's sorely missing and should be done in any case. Then the container should be guaranteed to run.

Right, that is true. I think it is also a good thing to check if the container is actually running

test/check-application
```diff
@@ -2384,9 +2384,9 @@ class TestApplication(testlib.MachineCase):
         # Changing the port should allow creation of container
         b.set_input_text('#run-image-dialog-publish-0-host-port', '5001')
         b.click('.pf-v5-c-modal-box__footer #create-image-create-run-btn')
+        b.wait_not_present("#run-image-dialog-name")
+
         self.waitContainerRow(container_name)
-        container_sha = self.execute(False, f"podman inspect --format '{{{{.Id}}}}' {container_name}").strip()
-        self.waitContainer(container_sha, False, name=container_name, image=IMG_BUSYBOX, state='Running')
```
martinpitt, line 2390: Is that a rebase error, or did you deliberately drop this? Then the commit message should explain.

martinpitt (changes requested):

test/check-application
jelly: Deliberate, this was there to make sure the dialog was closed but there is imo a better way :)

martinpitt:
Meh, we need the pinning from https://github.com/cockpit-project/cockpit-podman/pull/1628/commits/f64b901885868deb364ac1e82059bea2d47c096e first -- can you please push that here, and we land this first?

martinpitt (approved):

PR #1628, opened 2024-03-15
test/browser: run tests inside the tasks container

allisonkarlitskaya:
This simplifies the "outside" setup quite a bit and gives us the same tasks container that test runs on the Cockpit CI run under.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -1,27 +1,7 @@
 #!/bin/sh
+
 set -eux
-
-# test plan name, passed on to run-test.sh
-PLAN="$1"
-
-export TEST_BROWSER=${TEST_BROWSER:-firefox}
-
-TESTS="$(realpath $(dirname "$0"))"
-export SOURCE="$(realpath $TESTS/../..)"
-
-# https://tmt.readthedocs.io/en/stable/overview.html#variables
-export LOGS="${TMT_TEST_DATA:-$(pwd)/logs}"
-mkdir -p "$LOGS"
-chmod a+w "$LOGS"
-
-# install firefox (available everywhere in Fedora and RHEL)
-# we don't need the H.264 codec, and it is sometimes not available (rhbz#2005760)
-dnf install --disablerepo=fedora-cisco-openh264 -y --setopt=install_weak_deps=False firefox
-
-# nodejs 10 is too old for current Cockpit test API
-if grep -q platform:el8 /etc/os-release; then
-    dnf module switch-to -y nodejs:16
-fi
+cd "${0%/*}/../.."
 
 # HACK: ensure that critical components are up to date: https://github.com/psss/tmt/issues/682
 dnf update -y podman crun conmon criu
@@ -37,6 +17,11 @@ fi
 # Show critical package versions
 rpm -q runc crun podman criu kernel-core selinux-policy cockpit-podman cockpit-bridge || true
 
+# allow test to set up things on the machine
+mkdir -p /root/.ssh
+curl https://raw.githubusercontent.com/cockpit-project/bots/main/machine/identity.pub >> /root/.ssh/authorized_keys
+chmod 600 /root/.ssh/authorized_keys
+
 # create user account for logging in
 if ! id admin 2>/dev/null; then
     useradd -c Administrator -G wheel admin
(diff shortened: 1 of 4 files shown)


allisonkarlitskaya:
This is going to need a fairly extensive treatment of finding all of the places we make exceptions for the -ws container on ostree images and extending them to include exceptions for testing farm.  Probably we want to add a new API for that...

martinpitt:
@allisonkarlitskaya As a data point: RHEL8 has systemd-nspawn (in the `systemd-container` package). However, nspawn can't directly run {oci,docker}-{archive,directory}. It [is possible](https://seanmcgary.com/posts/run-docker-containers-with-systemd-nspawn/) to let podman construct the file system into a flat tree and run that, but it's a lot of overhead especially for the large tasks container:
```
mkdir /var/tmp/nspawntest
podman export "$(podman create --name nspawntest localhost/test-busybox true)" | tar -x -C /var/tmp/nspawntest
systemd-nspawn -D /var/tmp/nspawntest/
```
(I didn't specify any of the network options for this)

allisonkarlitskaya:
> Treating the -ws container is already awkward, but we can get away with it because it can be killed and restarted for each test. We don't have that option for the tasks container. Given how brittle the clean up is already, I have a pretty strong feeling that we can't do robust _podman_ tests and at the same time rely on that _podman_ container -- so using docker or possibly nspawn or even just chroot feels more promising here.

:100:

Same basic justification as what I wrote in the commit message:


> We have to deviate from the starter-kit template here: the tests have setup/teardown logic which wipes the podman slate completely clean.  We can't possibly survive that, and trying to carve out an exemption would be difficult and would cut against the spirit of the deep clean that we're trying for.  This isn't a problem for the cockpit-ws container, since it gets started after the tests start running.
>
> We can avoid all that by just running the tasks container in docker.

martinpitt:
Ah right -- I had blindly assumed the "all tests red" was still due to the podman conflict, but it's "just" testCreateContainerValidation now

jelly:
> Ah right -- I had blindly assumed the "all tests red" was still due to the podman conflict, but it's "just" testCreateContainerValidation now

See https://github.com/cockpit-project/cockpit-podman/pull/1624 I suspected I created a flake here.

martinpitt:
hmm, "Failed to start docker.service: Unit docker.service not found." in c8s/c9s [log](https://artifacts.dev.testing-farm.io/6da1b8d7-9ed2-437b-a6d0-b12bb051477f/)

martinpitt:
I updated the docker approach comment above a little bit, and that looks like a dead end. So I had another look at nspawn:
```
mkdir /var/tmp/tasks
podman export "$(podman create --name tasks-import ghcr.io/cockpit-project/tasks)" | tar -x -C /var/tmp/tasks
podman rm tasks-import
podman rmi ghcr.io/cockpit-project/tasks

# disable seccomp, see https://bugzilla.redhat.com/show_bug.cgi?id=2040247
SYSTEMD_SECCOMP=0 systemd-nspawn -D /var/tmp/tasks/ --ephemeral --user user
```

I'll add a fixup commit here to play around with this a little bit.

test/browser/browser.sh
```diff
@@ -49,25 +49,28 @@ for retry in $(seq 5); do
     sleep $((5 * retry * retry))
 done
 
+CONTAINER="$(cat .cockpit-ci/container)"
+
+# import the test CONTAINER image as a directory tree for nspawn
+mkdir /var/tmp/tasks
+podman export "$(podman create --name tasks-import $CONTAINER)" | tar -x -C /var/tmp/tasks
```
allisonkarlitskaya, line 56: You might want to look into `podman mount`...

martinpitt: Well no -- *not* having created podman containers and storage files was the whole point of that exercise..

martinpitt:
The [rawhide errors](https://artifacts.dev.testing-farm.io/70c6c217-f118-48c9-b432-3136efafbe44/) are unrelated:
```
Error: pasta failed with exit code 1:
netlink: Unexpected sequence number (6 != 10)
```
At least that's now slightly different than the previous complete FUBAR in https://github.com/containers/podman/issues/22052 , but we can continue to ignore this on our side. I'll report it to podman/pasta ASAP.

allisonkarlitskaya (commented):
I mean, I'm not thrilled about it, but it seems to work.  I'm happy to land it in its current state.

martinpitt (approved):
Ack, let's go with that then. It's at least closer to the new paradigm than keeping the old setup.

PR #1634, opened 2024-03-19
workflows: add tasks-container-update workflow

jelly:
This workflow tries to update our cockpit CI container on a weekly interval.


martinpitt (approved):
Thanks! This comes from a fork, so we can't easily test it with running the workflow. But it looks good, and it just looks like a copy from -machines.

martinpitt:
I manually triggered the workflow and it produced #1635 . Nice, thanks!

PR #1644, opened 2024-03-26
spec: Support SUSE specific BuildRequires

Nykseli:
In order to support testing on opensuse's tumbleweed image, the spec file needs to be updated to support it.

### DIFF
--- a/packaging/cockpit-podman.spec.in
+++ b/packaging/cockpit-podman.spec.in
@@ -24,7 +24,12 @@ URL:            https://github.com/cockpit-project/cockpit-podman
 
 Source0:        https://github.com/cockpit-project/%{name}/releases/download/%{version}/%{name}-%{version}.tar.xz
 BuildArch:      noarch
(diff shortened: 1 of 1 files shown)


martinpitt:
Until https://github.com/cockpit-project/bots/pull/6136 lands, I applied a #5 hammer and [forced the tests to run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1644-f646bc98-20240326-115723-opensuse-tumbleweed/log.html). "Failed to connect to bus: No medium found" sounds like there's no D-Bus for the user session of "admin", i.e. something wrong with the user bus?

(Ignore the sssd.service failure, that's not fatal)

Nykseli:
Created a PR fixing known issues with the testing image: https://github.com/cockpit-project/bots/pull/6137

martinpitt (approved):
You know, this doesn't break anything, and is clearly a step into the right direction, so let's just land it.

PR #1652, opened 2024-04-03
tmt: Pass TEST_AUDIT_NO_SELINUX into test container

mvollmer:


martinpitt (dismissed):

mvollmer:
centos-8 seems to be too old to understand `--setenv=VAR` without the `=VALUE` part.

test/browser/browser.sh
```diff
@@ -75,6 +75,7 @@ SYSTEMD_SECCOMP=0 systemd-nspawn \
     -D /var/tmp/tasks/ \
     --ephemeral \
     --user user \
+    --setenv=TEST_AUDIT_NO_SELINUX="$TEST_AUDIT_NO_SELINUX" \
```
martinpitt, line 78: Doesn't this need `${TEST_AUDIT_NO_SELINUX:-}` ? This script is `set -u`.

mvollmer: Yes, let's do that.

PR #1660, opened 2024-04-09
workflows: Run tasks-container-update in "self" environment

martinpitt:
Just like cockpit-lib-update, this workflow produces PRs which need to run reposchutz, so they need to be pushed via SSH, not the GitHub token.

----

I landed yesterday's #1658 with my admin hammer, but let's fix this properly.

### DIFF
--- a/.github/workflows/tasks-container-update.yml
+++ b/.github/workflows/tasks-container-update.yml
@@ -6,6 +6,7 @@ on:
   workflow_dispatch:
 jobs:
   tasks-container-update:
+    environment: self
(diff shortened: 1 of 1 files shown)


martinpitt:
[Ran the action against this branch](https://github.com/cockpit-project/cockpit-podman/actions/runs/8611698317), it produced #1661 and [reposchutz ran successfully](https://github.com/cockpit-project/cockpit-podman/actions/runs/8611708130/job/23599421707?pr=1661).

jelly (approved):

PR #1672, opened 2024-04-18
cockpit.format_bytes() cleanups

martinpitt:
See discussion here: https://github.com/cockpit-project/cockpit/pull/20322/files#r1567471946

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -49,16 +49,16 @@ export function format_memory_and_limit(usage, limit) {
         return "";
 
     let mtext = "";
-    let units = 1000;
+    let unit;
     let parts;
     if (limit) {
-        parts = cockpit.format_bytes(limit, units, true);
+        parts = cockpit.format_bytes(limit, undefined, { separate: true });
         mtext = " / " + parts.join(" ");
-        units = parts[1];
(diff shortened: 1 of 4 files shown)


src/util.js
```diff
@@ -49,16 +49,16 @@ export function format_memory_and_limit(usage, limit) {
         return "";
 
     let mtext = "";
-    let units = 1000;
+    let unit;
     let parts;
     if (limit) {
-        parts = cockpit.format_bytes(limit, units, true);
+        parts = cockpit.format_bytes(limit, undefined, { separate: true });
```
allisonkarlitskaya, line 55: imho, this `separate` business is the ugliest part of the existing API.  No cleanup would be complete without taking an axe to this...

martinpitt: I :100: agree. That's why I just sent #1673, but I realize that it may be more controversial. This PR should hopefully be straightforward, so I sent it separately.

jelly: It is interesting how it went from `bool => object`

jelly: Note that we have a lot of separate still left in storage as common pattern.

PR #1673, opened 2024-04-18
Rework display of current container memory usage

martinpitt:
The previous view like "0.001/16 GB" was not very helpful. For containers
without an explicit memory limit, the denominator `containerStats.MemLimit` is
the *host* memory, not something specific to the container. This makes it hard
to see which containers actually have a configured limit. For containers which
*do* have a memory limit, the numerator applied to that, but that makes it
difficult to compare to other containers and its relative size on the host
(from the "see the forest, not the tree" perspective).

Rework this to show Memory in three parts:

- The first line shows the container's memory usage as a percentage of
the host's memory. That makes them comparable and gives a good
feeling about host resource utilization, similar to the CPU usage
view.

- The second line shows the container's absolute memory usage. This
uses proper SI prefixes and thus uses normalized numbers, not like "0.001".

- *If* the container has an explicit memory limit, show its value and used
percentage, in lower/gray font (similar to the image name and process).
Don't show it for the common case of no limit.

This also gets rid of calling `cockpit.format_bytes()` with the `separate` API,
which we'd like to get rid of (as it encourages these questionable string
trickeries).

----

See [this discussion](https://github.com/cockpit-project/cockpit/pull/20322/files#r1570020217).

On main it looks like this:
podman-mem-main

image

 - [x] Builds on top of PR #1672 

[pixel diff](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1673-bee7fa0f-20240425-151140-fedora-39/log.html)

## Podman: Improve display of memory usage

The Podman Containers page now shows the memory used by a container as percentage of host memory and also the absolute usage with an appropriate unit. It also does not show a limit any more for containers which don't actually have a configured limit. For those which do, it shows both the absolute limit and which percentage of this is currently used.

This makes it easier to get an overview of the host's RAM utilization and relative sizes of the containers, similar to how it already presents the CPU usage. It also makes it clear which containers have a memory limit.

Screen Shot 2024-04-25 at 16 55 07

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -44,30 +44,6 @@ export function localize_time(unix_timestamp) {
     return formatRelative(unix_timestamp * 1000, Date.now(), { locale });
 }
 
-export function format_memory_and_limit(usage, limit) {
-    if (usage === undefined || isNaN(usage))
-        return "";
-
-    let mtext = "";
-    let unit;
-    let parts;
-    if (limit) {
-        parts = cockpit.format_bytes(limit, undefined, { separate: true });
-        mtext = " / " + parts.join(" ");
-        unit = parts[1];
-    }
-
-    if (usage) {
-        parts = cockpit.format_bytes(usage, unit, { separate: true });
-        if (mtext)
-            return _(parts[0] + mtext);
-        else
-            return _(parts.join(" "));
-    } else {
-        return "";
-    }
-}
-
 /*
  * The functions quote_cmdline and unquote_cmdline implement
  * a simple shell-like quoting syntax.  They are used when letting the
--- a/test/check-application
+++ b/test/check-application
@@ -4,6 +4,7 @@
 # "class Browser" and "class MachineCase" for the available API.
 
 import os
+import re
 import sys
 import time
 
@@ -1225,6 +1226,14 @@ class TestApplication(testlib.MachineCase):
         self.waitContainer(container_sha, auth, name='swamped-crate', image=IMG_BUSYBOX,
                            state='Running', owner="system" if auth else "admin")
 
+        # run another container with explicit memory 400 MiB limit, with a process that uses 300 MB memory
+        self.execute(
+            auth,
+            f"podman run -d --name picket-fence --memory 400m --stop-timeout 0 {IMG_ALPINE} "
+            """awk 'BEGIN { y = sprintf("%300000000s",""); system("sleep infinity") }'""")
+        picket_fence_sha = self.execute(auth, "podman inspect --format '{{.Id}}' picket-fence").strip()
+        self.waitContainer(picket_fence_sha, auth, name="picket-fence", state="Running")
+
         def get_cpu_usage(sel):
             cpu = self.getContainerAttr(sel, "CPU")
             self.assertIn('%', cpu)
@@ -1238,17 +1247,40 @@ class TestApplication(testlib.MachineCase):
         if auth or self.has_cgroupsV2:
             cpu = get_cpu_usage(IMG_BUSYBOX)
 
-            self.assertIn('/', memory)
-            numbers = memory.split('/')
-            self.assertTrue(numbers[0].strip().replace('.', '', 1).isdigit())
-            full = numbers[1].strip().split()
-            self.assertTrue(full[0].replace('.', '', 1).isdigit())
-            self.assertIn(full[1], ["GB", "MB"])
+            # swamped-crate has no explicit --memory limit
+            # FIXME: current format_bytes() does "KB"; fixed in https://github.com/cockpit-project/cockpit/pull/20356
+            m = re.match(r'([0-9]+)%([0-9.]+) [kK]B$', memory)
+            self.assertTrue(m, memory)
+            mem_pct = int(m.group(1))
+            mem_abs = float(m.group(2))
+            self.assertGreaterEqual(mem_pct, 0)
+            self.assertGreaterEqual(mem_abs, 1.0)
+            # this container doesn't do anything, should have small memory usage
+            self.assertLessEqual(mem_pct, 5)
+            self.assertLessEqual(mem_abs, 300.0)
 
-            # Test that the value is updated dynamically
+            # Test that the CPU value is updated dynamically
             self.execute(auth, "podman exec -i swamped-crate sh -c 'dd bs=1024 < /dev/urandom > /dev/null &'")
             b.wait(lambda: get_cpu_usage(IMG_BUSYBOX) > cpu)
             self.execute(auth, "podman exec swamped-crate sh -c 'pkill dd'")
+
+            # picket-fence has a memory limit
+            memory = self.getContainerAttr(IMG_ALPINE, "Memory")
+            m = re.match(r'([0-9]+)%([0-9.]+) MB([0-9]+)% of ([0-9.]+) MB limit$', memory)
+            self.assertTrue(m, memory)
+            # wide range here, to work on custom testbeds
+            host_mem_pct = int(m.group(1))
+            self.assertGreater(host_mem_pct, 2)
+            self.assertLess(host_mem_pct, 90)
+            # this is fairly well predictable
+            mem_abs = float(m.group(2))
+            self.assertGreaterEqual(mem_abs, 300.0)
+            self.assertLess(mem_abs, 305.0)
+            limit_mem_pct = int(m.group(3))
+            self.assertGreaterEqual(limit_mem_pct, 70)
+            self.assertLessEqual(limit_mem_pct, 75)
+            # 400 MiB limit
+            self.assertEqual(m.group(4), "419")
         else:
             # No support for CGroupsV2
             self.assertEqual(self.getContainerAttr(IMG_BUSYBOX, "CPU"), "n/a")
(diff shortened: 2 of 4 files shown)


martinpitt:
@garrett I'll let you have the first word here -- if you don't like the percentage, there's an alternative to say e.g. "10 MB / 1.13 GB", or replacing the '/' with an 'of', or keeping the slash: "10% / 1.13 GB". Any opinions?

martinpitt:
@allisonkarlitskaya @jelly FYI for the typing discussion.

src/util.js
```diff
@@ -44,30 +44,6 @@ export function localize_time(unix_timestamp) {
...
-    let unit;
-    let parts;
-    if (limit) {
-        parts = cockpit.format_bytes(limit, undefined, { separate: true });
-        mtext = " / " + parts.join(" ");
-        unit = parts[1];
-    }
-
-    if (usage) {
-        parts = cockpit.format_bytes(usage, unit, { separate: true });
```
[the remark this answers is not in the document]
allisonkarlitskaya: https://github.com/cockpit-project/cockpit/pull/20334 kills this API.

allisonkarlitskaya (commented):
I like it.  I actually find it easier to read/understand.

src/util.js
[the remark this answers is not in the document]
martinpitt: Does it really? It just marks it as deprecated, but it still has the compat code and the unit tests.

But, just FTR: Neither of these PRs really blocks on the other. c-podman bundles cockpit.js, it's not a runtime dependency. **If** that cockpit PR gets modified to actually drop the API, existing c-podman will run fine. Of course tests will blow up if/when we land the cockpit PR first and then get a  "bump cockpit lib import" PR, but that's totally fine.

martinpitt:
Rebased/deconflicted/unblocked.

garrett:
I'm not sure if the amount of memory per container is as important as the amount of memory as a percentage of the system.

Additionally, if you go with the per-container method, the amount of memory allocated for the containers is probably going to vary, so seeing the percentage relative to that means you cannot compare between containers.

Without that context, it's fine. But I think these above two points likely matter, so the change isn't really good overall when you consider all containers and the amount of memory usage overall.

martinpitt:
@garrett Good point -- I suppose that's a matter of perspective, whether users are more interested in absolute memory usage from the host's POV (→ absolute value), or how much of its memory allocation a container uses (→ relative value). If you think the absolute values matter more, WDYT about "10 MB / 1.13 GB" then? Or even some combination, like "45 MB / 1.13 GB (4%)" (that isn't mathematically correct, but you get the idea)

garrett:
We'd still need to use the same units across containers to compare containers. So everything would need to be MB, GB, etc. Mixing and matching makes comparison difficult to impossible. (As results will update over time, it could jump from 15 KB to 15 MB to 15 GB in the matter of seconds, which is why I said "impossible".)

martinpitt:
I had a look at the reported "max memory" in podman. When `podman run`, `podman inspect` shows `HostConfig.Memory: 0`. When I run it with `--memory=500M`, inspect shows "Memory: 5024......" (it's the right number, I just didn't copy it). But our tests also doesn't start the container with `--memory`.

Our UI shows a maximum because it doesn't look at `HostConfig.Memory`, but the runtime stats (`/libpod/containers/stats` API, `containerStats.MemLimit`. But that's just the default cgroup limit of the host memory:

```
podman container stats -l --no-stream|cat
ID            NAME           CPU %       MEM USAGE / LIMIT  MEM %       NET IO           BLOCK IO    PIDS        CPU TIME    AVG CPU %
2026c9e1f408  swamped-crate  0.00%       36.86kB / 1.134GB  0.00%       2.6kB / 1.048kB  0B / 0B     1           3.312ms     0.00%

# free -h
               total        used        free      shared  buff/cache   available
Mem:           1.1Gi       492Mi       349Mi       5.3Mi       385Mi       588Mi
```

So, quite independently of this I propose to not actually show that "implied" maximum in cockpit-podman if it wasn't actually configured. We can do this quite easily, but I think before I change the code, I'd like to get some agreement on the new behaviour/design.

So for the common case of *not* configuring a limit, the memory usage could just be shown as "12 MB" (or whatever a sensible unit is, `cockpit.format_bytes()` figures that out nicely). And *if* there is a limit, we could show it as e.g. "14.2 MB (1% of 1.5 GB)"?

garrett:
What about something like this?

Container memory percentage and limit

1. You can compare the containers to one another, as they're the same unit (%)
2. The % is relative to the system usage (which is how the CPU is measured also, right?)
3. Limits are not shown, unless they are set.
4. Limits are not tied to the unit of measurement.
5. Limits can be set to the best size relative to the container (hopefully even what was specified in while setting up the container).
6. Limits are using a subdued color and size, which makes it easier to scan the percentages.
7. Limits are _under_ percentages as secondary text, mirroring the container details under the container name.

What do you think?

garrett:
Here's the percentage of limit:

Container memory percentage and limit with percentage

Here's the same, but with limit integrated at the end (instead of at the start with a colon — it's cleaner looking, I think):

Container memory percentage and limit with percentage, limit at end

martinpitt:
@garrett I implemented your second proposal -- I agree that keeping the numbers at the start is a bit cleaner:

image

The first container has no limit, the second does.

(not pushed yet, still needs tests)

martinpitt:
@garrett ah, that still has my "n% (absolute usage)" proposal. I can be talked into taking it out.

garrett:
The problem with the () and the arbitrary wide number and whatever unit is that:
1. it's cluttered looking
2. completely breaks the scalability
3. makes it harder to compare
4. shifts widths every time anything in the table is updated
5. likely shifts the width of *everything* as it's all interconnected in layout calculations

martinpitt:
> arbitrary wide number

Note that it's not arbitrary wide -- `format_bytes()` limits them to at most three digits, larger numbers get the next-higher suffix (kB, MB, TB, PB, etc.)

martinpitt:
In other words, it's not going to be wider than the more verbose "limit" line below

garrett:
The sizing is completely up to change based on several factors:

It's determined by:
- digits, which can change
- unit, which can change
- proportional characters within the units (the numbers _should_ be monospace via tabular numbers)

Which exactly proves my point that it can and will vary in width, which means it breaks the scan-ability of the text, which is critical when you're comparing things in a table.

garrett:
For reference, here's what it'd look like on another line:

Container memory percentage and limit with percentage, limit at end(1)

Alternatively, I guess we could move some of the information to tooltips, if it's not really necessary to show by default.

martinpitt:
@garrett I implemented the above functionality, including tests (I want to see how they fare on all OSes/images, but that is for me).

But I really struggled with the right-aligned columns. I pushed a second WIP commit here with what I have so far, this is how it looks like:

image

It needs some more fine-tuning, like perhaps aligning the sorting arrows with the right side of the content text, but at least it works in principle.

garrett:
I have this sort of opposite problem:

image

It's "Me..." but I think that isn't related to this PR. The header labels, at least in this case, probably should never be truncated.

martinpitt:
@garrett Right, that's independent, but also needs fixing. There's not enough minwidth on the other columns, so a long command line (such as toolbox) squeezes them too far. I see that too on my laptop, I did the screenshot in a test VM without toolbox.

garrett:
> @garrett Right, that's independent, but also needs fixing. 

Yes, but, until there's a fix for that, I can't see the problem here to fix it, which is my point.

martinpitt:
I added a commit to fix the minimum width. Before:

image

now:

image

This could also affect the "Owner" column:

image

now:

image

martinpitt:
OK, I think I figured it out. This now works for very wide widths:

image

and very narrow ones:

image

and I also tested everything in between. The arrow isn't precisely aligned to the right side of the text, but at this point I'd call it "good enough" and "wait for the design adults from PF to do it perfectly"..

garrett (dismissed):
Thanks! Much better now! I think we're all happier with this.

I can see a couple of very nitpicky things, but they're PF-related and would need some minor adjusting. (Mainly: The spacing of the arrows relative to the headings is very odd, especially at small sizes.) But that's not for this PR. This PR looks great from my point of view! Thanks!

martinpitt:
[erk](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1673-ed6ab945-20240425-145202-fedora-39-devel/log.html#16), that's a bug, looking:

> error: Warning: Each child in a list should have a unique "key" prop.%s%s See https://reactjs.org/link/warning-keys for more information.%s

test/check-application
```diff
@@ -1238,17 +1247,40 @@ class TestApplication(testlib.MachineCase):
...
-            self.assertIn('/', memory)
-            numbers = memory.split('/')
-            self.assertTrue(numbers[0].strip().replace('.', '', 1).isdigit())
-            full = numbers[1].strip().split()
-            self.assertTrue(full[0].replace('.', '', 1).isdigit())
-            self.assertIn(full[1], ["GB", "MB"])
-
-            # Test that the value is updated dynamically
+            # swamped-crate has no explicit --memory limit
+            # FIXME: current format_bytes() does "KB"; fixed in https://github.com/cockpit-project/cockpit/pull/20356
```
jelly, line 1251: I believe this should now already be in `pkg/lib`, as I pulled and merged an update yesterday.

But feel free to do this in a follow up PR.

jelly (approved):

test/check-application
martinpitt: Ah right -- I was concerned that this has to work against older cockpit-bridge, but podman bundles cockpit.js now. I'll send a follow-up!

martinpitt: → PR #1692

PR #1692, opened 2024-04-26
test: Tighten expected unit for kilobyte

martinpitt:
We pulled in https://github.com/cockpit-project/cockpit/pull/20356 now, so we don't expect "KB" any more.


jelly (approved):
Thanks for the quick follow up!

PR #1699, opened 2024-04-29
test: update reference-image to Fedora-40

jelly:


jelly:
Flake is persistent, or expected?

PR #1701, opened 2024-04-30
Makefile: workaround pasta user containers being broken

jelly:
When a system container runs a user container cannot be created as it exits with a pasta error.

Reported as https://bugzilla.redhat.com/show_bug.cgi?id=2277954

### DIFF
--- a/test/run
+++ b/test/run
@@ -12,6 +12,11 @@ if [ "$TEST_SCENARIO" == "devel" ]; then
     export TEST_COVERAGE=yes
 fi
 
+# Workaround broken podman pasta on Fedora-CoreOS https://bugzilla.redhat.com/show_bug.cgi?id=2277954
+if [ "$TEST_OS" == "fedora-coreos" ]; then
+    export RUN_TESTS_OPTIONS="${RUN_TESTS_OPTIONS} --enable-network"
+fi
+
 make codecheck
 make check
 make po/podman.pot


Makefile
```diff
@@ -41,6 +41,11 @@ RUN_TESTS_OPTIONS+=--coverage
 NODE_ENV=development
 endif
 
+# Workaround broken podman pasta on Fedora-CoreOS https://bugzilla.redhat.com/show_bug.cgi?id=2277954
+ifeq ("$(TEST_OS)","fedora-coreos")
+RUN_TESTS_OPTIONS+=--enable-network
+endif
```
martinpitt, line 47: This would fit a lot better into test/run

martinpitt (approved):
Thanks! If you are in a hurry, ok, but I promise a quick re-review when you clean this up.

jelly:
> Thanks! If you are in a hurry, ok, but I promise a quick re-review when you clean this up.

No hurry, we can handle your review comment.

test/run
```diff
@@ -12,6 +12,11 @@ if [ "$TEST_SCENARIO" == "devel" ]; then
     export TEST_COVERAGE=yes
 fi
 
+# Workaround broken podman pasta on Fedora-CoreOS https://bugzilla.redhat.com/show_bug.cgi?id=2277954
+if [ "$TEST_OS" == "fedora-coreos" ]; then
+    export RUN_TESTS_OPTIONS+=" --enable-network"
```
martinpitt, line 17: That "+" syntax doesn't work here. Please do the `="$RUN_TESTS_OPTIONS --enable-network"` shell approach.

jelly: Curiously it does? 

```
TEST_AUDIT_NO_SELINUX=1 test/common/run-tests --track-naughties --enable-network
```

martinpitt:
TF doesn't use test/run anyway, no need to wait.

PR #1730, opened 2024-05-27
Implement pull option for existing images

StefanNienhuis:
This PR implements a 'Pull' option for existing images. This action is added in the dropdown menu for an image. In addition to this, it includes a 'Pull all images' option at the top level to pull all images in storage. The goal is to easily allow updating containers when using tags like 'latest' or a just major version number.

The implementation makes use of the existing downloadImage function that is used when downloading a new image. It was slightly modified to support multiple images being downloaded simultaneously.

I'm not familiar with the localization mechanism used in Cockpit, but in line with other strings I wrapped the labels in `_("Label")`. Upon changing the display language this (somewhat obviously) didn't automatically work, so translations will need to be added.

---
# Podman: Add image pull action

The new `Pull` option in the image context menu updates the existing image to the latest version on its registry.

image

### DIFF
--- a/src/Images.jsx
+++ b/src/Images.jsx
@@ -35,6 +35,8 @@ class Images extends React.Component {
         this.state = {
             intermediateOpened: false,
             isExpanded: false,
+            // List of container image names which are being downloaded
+            imageDownloadInProgress: [],
         };
 
         this.downloadImage = this.downloadImage.bind(this);
@@ -46,10 +48,10 @@ class Images extends React.Component {
         if (imageTag)
             pullImageId += ":" + imageTag;
 
-        this.setState({ imageDownloadInProgress: imageName });
+        this.setState(previous => ({ imageDownloadInProgress: [...previous.imageDownloadInProgress, imageName] }));
         client.pullImage(system, pullImageId)
                 .then(() => {
-                    this.setState({ imageDownloadInProgress: undefined });
+                    this.setState(previous => ({ imageDownloadInProgress: previous.imageDownloadInProgress.filter((image) => image != imageName) }));
                 })
                 .catch(ex => {
                     const error = cockpit.format(_("Failed to download image $0:$1"), imageName, imageTag || "latest");
@@ -58,7 +60,7 @@ class Images extends React.Component {
                             <samp>{cockpit.format("$0 $1", ex.message, ex.reason)}</samp>
                         </p>
                     );
-                    this.setState({ imageDownloadInProgress: undefined });
+                    this.setState(previous => ({ imageDownloadInProgress: previous.imageDownloadInProgress.filter((image) => image != imageName) }));
                     this.props.onAddNotification({ type: 'danger', error, errorDetail });
                 });
     }
@@ -71,6 +73,10 @@ class Images extends React.Component {
                               userServiceAvailable={this.props.userServiceAvailable}
                               systemServiceAvailable={this.props.systemServiceAvailable} />
         );
+    };
+
+    onPullAllImages = () => {
+        Object.values(this.props.images).forEach((image) => this.downloadImage(utils.image_name(image), null, image.isSystem));
     };
 
     onOpenPruneUnusedImagesDialog = () => {
@@ -138,7 +144,8 @@ class Images extends React.Component {
                 title: <ImageActions image={image} onAddNotification={this.props.onAddNotification}
                                      user={this.props.user}
                                      userServiceAvailable={this.props.userServiceAvailable}
-                                     systemServiceAvailable={this.props.systemServiceAvailable} />,
+                                     systemServiceAvailable={this.props.systemServiceAvailable}
+                                     downloadImage={this.downloadImage} />,
                 props: { className: 'pf-v5-c-table__action content-action' }
             },
         ];
@@ -290,6 +297,7 @@ class Images extends React.Component {
                         </FlexItem>
                         <FlexItem>
                             <ImageOverActions handleDownloadNewImage={this.onOpenNewImagesDialog}
+                                              handlePullAllImages={this.onPullAllImages}
                                               handlePruneUsedImages={this.onOpenPruneUnusedImagesDialog}
                                               unusedImages={unusedImages} />
                         </FlexItem>
@@ -319,15 +327,15 @@ class Images extends React.Component {
                   onAddNotification={this.props.onAddNotification}
                   userServiceAvailable={this.props.userServiceAvailable}
                   systemServiceAvailable={this.props.systemServiceAvailable} /> }
-                {this.state.imageDownloadInProgress && <CardFooter>
-                    <div className='download-in-progress'> {_("Pulling")} {this.state.imageDownloadInProgress}... </div>
+                {this.state.imageDownloadInProgress.length > 0 && <CardFooter>
+                    <div className='download-in-progress'> {_("Pulling")} {this.state.imageDownloadInProgress.join(', ')}... </div>
                 </CardFooter>}
             </Card>
         );
     }
 }
 
-const ImageOverActions = ({ handleDownloadNewImage, handlePruneUsedImages, unusedImages }) => {
+const ImageOverActions = ({ handleDownloadNewImage, handlePullAllImages, handlePruneUsedImages, unusedImages }) => {
     const actions = [
         <DropdownItem
             key="download-new-image"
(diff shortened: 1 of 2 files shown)


jelly (commented):
Hi! Changes look alright, note that pulling won't update existing containers and we also have a `always pull latest` image option when creating a container.

Would you be interested in adding tests? It is a bit complicated as you need to push an updated image to our test and then click on `pull`.

https://github.com/cockpit-project/cockpit-podman/blob/main/test/check-application#L444

StefanNienhuis:
Yeah, I noticed that too while testing. I thought about also introducing a 'recreate' option for containers, but with the various different ways containers can be created I'm not sure that's feasible. Especially when the containers are created using Quadlets like I am doing most of the time. My current solution is to just do a system reboot, so the containers are recreated with the latest image (only works with Quadlets though I believe).

I haven't contributed to anything Cockpit-related before so I'm not familiar with the testing software, but I can make an attempt. Is there already a test case where an image is being pushed or would I have to out the pushing logic?

jelly:
I'll help out and add some tests, sorry for long delay.

jelly:
One thing I realized, is that the UI might be a bit deceptive as an image as grafana:10.0 is unlikely to get updates but of course a `:latest` isn't. Then again if you use `grafana:10` you might get 10.x updates.

StefanNienhuis:
Thanks. I checked it out, but I wasn't really able to do anything meaningful.

Yeah, my intended usage is to use a tag like `grafana:10` and be able to know that I won't be getting any breaking changes. If a version 11 is released you'd need to update the container config manually, but you kind of want that to avoid breaking changes happening without knowing it. Do you have any suggestions for a better button label? This way of pulling existing images was inspired by Portainer btw, which has a similar UI and terminology.

StefanNienhuis:
Not sure what these failed test cases are about. Do you need anything from me to get this merged?

src/Images.jsx
```diff
@@ -35,6 +35,7 @@ class Images extends React.Component {
         this.state = {
             intermediateOpened: false,
             isExpanded: false,
+            imageDownloadInProgress: [],
```
martinpitt, line 38: This could use a comment what the structure of that state is -- it's not obvious from an empty list. I think it's a list of strings with image names?

src/Images.jsx
```diff
@@ -46,10 +47,10 @@ class Images extends React.Component {
         if (imageTag)
             pullImageId += ":" + imageTag;
 
-        this.setState({ imageDownloadInProgress: imageName });
```
martinpitt, line 49: Yay, it seems that previously wasn't initialized in the constructor?

src/Images.jsx
```diff
@@ -46,10 +47,10 @@ class Images extends React.Component {
...
+        this.setState((previous) => ({ imageDownloadInProgress: [...previous.imageDownloadInProgress, imageName] }));
```
martinpitt, line 50: Style nitpick throughout the patch: Please avoid 1-tuples. `previous`, not `(previous)`.

test/check-application
```diff
@@ -1184,6 +1185,86 @@ class TestApplication(testlib.MachineCase):
...
+        # Add local insecure registry into registries conf
+        m.write("/etc/containers/registries.conf", REGISTRIES_CONF)
+        self.execute(True, "systemctl stop podman.service")
+
+        self.execute(True, f"""
+            podman tag {IMG_BUSYBOX} localhost:5000/my-busybox
+            podman push localhost:5000/my-busybox
+            # Remove image with the same sha id
+            podman rmi localhost/test-busybox:latest
+        """)
```
martinpitt, line 1206: This copies the same code from testDownloadImage() which are of course also related. Perhaps either split this out in to a `self.setupLocalRegistries()` helper, or move the new tests into testDownloadImage()?

test/check-application
```diff
@@ -1184,6 +1185,86 @@ class TestApplication(testlib.MachineCase):
...
+        b.click(busybox_sel + " button.pf-v5-c-menu__item:contains('Pull')")
+        b.wait_in_text("div.download-in-progress", "Pulling")
+        b.wait_not_present(busybox_sel)
+
+        # Updated image has a different sha
+        self.assertNotEqual(busybox_id,
+                self.execute(True, "podman inspect --format '{{.Id}}'  localhost:5000/my-busybox").strip())
+
+        # Try to pull multiple images, drop to superuser as the registry runs as superuser
+        b.drop_superuser()
```
martinpitt, line 1240: I don't understand this justification?

test/check-application
```diff
@@ -1184,6 +1185,86 @@ class TestApplication(testlib.MachineCase):
...
+
+        # Ubuntu has an old podman which ships with a k8s image
+        if m.image == "ubuntu-2204":
+            self.execute(False, "podman rmi k8s.gcr.io/pause:3.5")
+
+        b.wait_in_text("#containers-images", "2 images")
+        b.click("#image-actions-dropdown")
+        b.click("button:contains(Pull all images)")
+        b.wait_in_text("div.download-in-progress", "Pulling")
+        b.wait_not_present('h4.pf-v5-c-alert__title:contains("Failed to download image")')
```
martinpitt, line 1266: That should wait until the pulling is complete, and ensure that it succeeded.

martinpitt (changes requested):
Thanks! I mostly have some style nitpicks which are optional. But the test looks like it got cut off?

test/check-application
jelly: The reason I run the tests as "normal" user is that you don't see superuser images which get in the way. (Basically the counting of `#container-images` becomes easier.)

martinpitt: Ah, that makes it clear. The current comment totally doesn't explain that, can you please update it? Thanks!

jelly: Done, to explain it better, only having images around which can be pulled makes the test not show any error dialogs as pulling `localhost/test-alpine` will return in an error.

martinpitt:
The new test [fails on rawhide](https://artifacts.dev.testing-farm.io/aa46898a-7323-4105-ba59-0b8cbe555b78/). I `--allow`ed our tests to get more data, but probably an unstable test?

src/Images.jsx
martinpitt: This is still current, note the "throughout the patch"

test/check-application
martinpitt: That's also a nice helper, if you need it multiple times.

martinpitt (commented):
Thanks! LGTM now (other than the style nit). But the test needs fixing.

src/Images.jsx
```diff
@@ -73,6 +75,10 @@ class Images extends React.Component {
         );
     };
 
+    onPullAllImages = () => {
+        Object.values(this.props.images).forEach((image) => this.downloadImage(utils.image_name(image), null, image.isSystem));
```
cockpituous, line 79: These 2 added lines are not executed by any test.

src/Images.jsx
```diff
@@ -336,6 +344,13 @@ const ImageOverActions = ({ handleDownloadNewImage, handlePruneUsedImages, unuse
         >
             {_("Download new image")}
         </DropdownItem>,
+        <DropdownItem
+            key="pull-all-images"
+            component="button"
+            onClick={() => handlePullAllImages()}
```
cockpituous, line 350: This added line is not executed by any test.

martinpitt:
Sorry, testPullImage still fails both on [our CI](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1730-9e96a71a-20240621-151725-fedora-40-devel/log.html#33) and  on [TF](https://artifacts.dev.testing-farm.io/e1a534da-272b-437b-bd10-0c42290e593e/)

test/check-application
```diff
@@ -1184,6 +1196,95 @@ class TestApplication(testlib.MachineCase):
...
+
+        self.login()
+
+        image1_sel = f"#containers-images tbody tr[data-row-id=\"{image1_sha}true\"]".lower()
+
+        # Wait new image sha now visible
+        b.click("#containers-images button.pf-v5-c-expandable-section__toggle")
+        b.click(f'{image1_sel} .pf-v5-c-menu-toggle')
+        b.click(image1_sel + " button.pf-v5-c-menu__item:contains('Pull')")
+        b.wait_in_text("div.download-in-progress", "Pulling")
```
martinpitt, line 1237: Pulling a tiny image from a local registry just takes a blink of an eye.. that feels like inviting a flake? (This isn't what the current failures were about, though)

martinpitt:
Retrying unrelated test failures, oh my :sob:

martinpitt (approved):
Thanks!

martinpitt:
I updated the release note a bit, please review. Thanks!

jelly:
> I updated the release note a bit, please review. Thanks!

LGTM!

PR #1735, opened 2024-05-31
Fix tests for opensuse tumbleweed

Nykseli:
Fix tests to support testing on tumbleweed. Sorry that this took a while. Got quite stuck on the cgroups issue for while :-)

This depends on: https://github.com/cockpit-project/bots/pull/6450

### DIFF
--- a/packaging/cockpit-podman.spec.in
+++ b/packaging/cockpit-podman.spec.in
@@ -41,6 +41,8 @@ Requires:       podman >= 2.0.4
 # HACK https://github.com/containers/crun/issues/1091
 %if 0%{?centos} == 9
 Requires:       criu-libs
+%elif 0%{?suse_version}
+Requires:       libcriu2
 %endif
 
 %{NPM_PROVIDES}
--- a/test/check-application
+++ b/test/check-application
@@ -135,7 +135,7 @@ class TestApplication(testlib.MachineCase):
 
         # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008249
         self.has_criu = "debian" not in m.image and "ubuntu" not in m.image
-        self.has_selinux = "arch" not in m.image and "debian" not in m.image and "ubuntu" not in m.image
+        self.has_selinux = not any(img in m.image for img in ["arch", "debian", "ubuntu", "suse"])
         self.has_cgroupsV2 = not m.image.startswith('rhel-8')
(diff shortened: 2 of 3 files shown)


jelly:
Triggered the tests with the new image, but linting fails so we can't check if the new image makes the tests green.

Nykseli:
Hmm, weird. Looks like `TestApplication.testCommitUser` is failing but I'm not able to get it to fail locally

jelly:
> Hmm, weird. Looks like `TestApplication.testCommitUser` is failing but I'm not able to get it to fail locally

I've seen this fail on podman 5.1 but opensuse has:

```
opensuse-tumbleweed-127-0-0-2-22:~ # rpm -qa podman
podman-5.0.3-1.1.x86_64
```

martinpitt:
That test "only" failed on the second retry, so it's a flake. But also, the refreshed image landed. I'll rebase this and retry -- there might still be a flake, but as long as it doesn't regress anywhere else (and I don't think it will, this looks fine), it's good to land. Thanks!

martinpitt:
[happened again](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1735-462a5e58-20240603-055447-opensuse-tumbleweed/log.html). The test expects the "force commit" dialog, but the [screenshot](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1735-462a5e58-20240603-055447-opensuse-tumbleweed/TestApplication-testCommitUser-opensuse-tumbleweed-127.0.0.2-2701-FAIL.png) shows that there's no dialog. That may be a general race in the test, or some bug in podman's current tumbleweed version?

I tried this locally:
```
bots/vm-run --no-network opensuse-tumbleweed
for i in `seq 10`; do TEST_OS=opensuse-tumbleweed test/check-application TestApplication.testCommitUser $RUNC -stv || break; done
```

and it passed. I retried it on CI once more.

This is annoying enough to block adding this to the automatic test map (i.e. gating for c-podman PRs), but good enough to land as it doesn't cause regressions.

martinpitt (approved):
Happened again.. But anyway, let's land this. Thanks!

Nykseli:
Thanks! 

That test failing/being flaky is indeed a bit annoying. I'll have to try and see if there's something odd going on with tumbleweed's podman.

PR #1744, opened 2024-06-03
packit: Enable centos-stream-10

martinpitt:
https://issues.redhat.com/browse/RHEL-34650 finally got fixed, so podman works in CentOS/RHEL 10 now.

### DIFF
--- a/test/browser/main.fmf
+++ b/test/browser/main.fmf
@@ -4,8 +4,6 @@ require:
   - cockpit-ws
   - cockpit-system
   - criu
-  # HACK: https://bugzilla.redhat.com/show_bug.cgi?id=2269485
-  - slirp4netns
 duration: 30m
 
 /system:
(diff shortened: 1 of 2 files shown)


martinpitt:
Note: centos-10 passed here, but I'm fine with not adding that to the testmap for now. But following TF tests in a non-blocking way is interesting and important for being able to update our package there.

test/browser/main.fmf
```diff
@@ -4,8 +4,6 @@ require:
   - cockpit-ws
   - cockpit-system
   - criu
-  # HACK: https://bugzilla.redhat.com/show_bug.cgi?id=2269485
-  - slirp4netns
```
martinpitt, line 8: This still worked in c10s, but broke rhel-10 gating tests. So after this lands, I'll do a new upstream release.

jelly (approved):

PR #1748, opened 2024-06-05
Use binary http channel for podman socket

### LINKED ISSUE #1733: testCommitUser fails on updates-testing
The job `fedora-40/updates-testing` failed on commit 19ffacf6e31bbcab58e02eb651b849a2bd210270.

Log: https://cockpit-logs.us-east-1.linodeobjects.com/pull-0-19ffacf6-20240531-015714-fedora-40-updates-testing/log.html

https://bugzilla.redhat.com/show_bug.cgi?id=2284131

martinpitt:
Cockpit's (default) text channels ignore frames with split UTF-8 multi-byte characters. This causes broken JSON. This is ultimately an issue with cockpit's channel API design, and Cockpit should hide that by buffering and partially decoding text channel frames, see https://github.com/cockpit-project/cockpit/issues/19235

Until that happens, convert the REST API to use a binary channel.

Fixes #1733

---

podman 5.1.0 has also reached c9s and c10s, so e.g. in #1746 these failed on the exact same issue in TF. So this PR should go all green, and I'll trigger an /updates-testing run as well.


src/rest.js
```diff
@@ -17,28 +17,39 @@ function manage_error(reject, error, content) {
 // calls are async, so keep track of a call counter to associate a result with a call
 let call_id = 0;
 
+const NL = '\n'.charCodeAt(0); // always 10, but avoid magic constant
```
jelly, line 20: :netherlands:

jelly (approved):

src/rest.js
martinpitt: Ik spreek een beetje Nederlands! :wink:

PR #1749, opened 2024-06-05
Replace cockpit.utf8_{de,en}coder with standard JS API

martinpitt:
There's no reason to use that Cockpit API, and we want to deprecate it.


jelly (approved):
:clap: yay!

martinpitt:
The TF centos failures are #1733, see fix in #1748.

PR #1754, opened 2024-06-12
build.js: support outputting esbuild metadata file

jelly:
ESbuild supports outputting a metadata file which can be used to view bundle size information using https://esbuild.github.io/analyze/


martinpitt (approved):
Cheers!

PR #1763, opened 2024-06-21
po: Don't mark JavaScript strings as c-format

### LINKED ISSUE #1712: cockpit.format() strings are incorrectly marked as c-format
https://translate.fedoraproject.org/translate/cockpit-podman/main/sv/?q=state:%3Ctranslated

this string looks a bit wierd guess the o should not be a part of $0% o

martinpitt:
We don't use printf-style C format macros like "%s" in the JavaScript code, at least not for translations. But xgettext interprets the "$0% Free" in pkg/kubernetes/scripts/nodes.js as C format string, which confuses translation tools and blocks the proper translation of this string.

As there doesn't seem to be a way to change the `--keyword=ngettext` argument to do that, just filter out the `c-format` tag with sed.

Taken from https://github.com/cockpit-project/cockpit/commit/449d76c0e6d9

Fixes #1712

----

Tested with `make po/podman.pot`. diff:

```diff
--- /tmp/podman.pot.orig	2024-06-21 14:39:15.548457329 +0200
+++ po/podman.pot	2024-06-21 14:39:35.140548117 +0200
@@ -73,8 +73,8 @@
 msgstr[0] ""
 msgstr[1] ""
 
+#
 #: src/Containers.jsx:379
-#, c-format
 msgid "$0% of $1 limit"
 msgstr ""
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,13 +70,14 @@ $(COCKPIT_REPO_STAMP): Makefile
 LINGUAS=$(basename $(notdir $(wildcard po/*.po)))
 
 po/$(PACKAGE_NAME).js.pot:
-	xgettext --default-domain=$(PACKAGE_NAME) --output=$@ --language=C --keyword= \
+	xgettext --default-domain=$(PACKAGE_NAME) --output=- --language=C --keyword= \
 		--keyword=_:1,1t --keyword=_:1c,2,2t --keyword=C_:1c,2 \
 		--keyword=N_ --keyword=NC_:1c,2 \
 		--keyword=gettext:1,1t --keyword=gettext:1c,2,2t \
 		--keyword=ngettext:1,2,3t --keyword=ngettext:1c,2,3,4t \
 		--keyword=gettextCatalog.getString:1,3c --keyword=gettextCatalog.getPlural:2,3,4c \
-		--from-code=UTF-8 $$(find src/ -name '*.[jt]s' -o -name '*.[jt]sx')
+		--from-code=UTF-8 $$(find src/ -name '*.[jt]s' -o -name '*.[jt]sx') | \
+		sed '/^#/ s/, c-format//' > $@
 
 po/$(PACKAGE_NAME).html.pot: $(NODE_MODULES_TEST) $(COCKPIT_REPO_STAMP)
 	pkg/lib/html2po.js -o $@ $$(find src -name '*.html')


martinpitt:
From @allisonkarlitskaya : Check `--keyword=_:1,1t,no-c-format`, see https://www.gnu.org/software/gettext/manual/html_node/c_002dformat-Flag.html

allisonkarlitskaya:
> From @allisonkarlitskaya : Check `--keyword=_:1,1t,no-c-format`, see https://www.gnu.org/software/gettext/manual/html_node/c_002dformat-Flag.html

See also this wall of text: https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html#index-_002d_002dflag_002c-xgettext-option

martinpitt:
@allisonkarlitskaya These don't work. Your first reference applies to comments *in the code*, but I'm not willing to do that. `--keyword=_:1,1t:no-c-format` or `--keyword=_:1,1t,no-c-format` don't work and are also not documented to exist. The `--flag` documentation is hard to understand, but something like `-flag=_:1:no-c-format` just causes *every* message id to receive a "no-c-format" commit which is stupid as well.

I suppose I already pulled my hair out over that six years ago, so I'd like to keep y seddery. It has served us  well enough.

martinpitt:
I tried to change `--language=C` to `--language=JavaScript`, and this is bad: It ignores two strings, and has this weird `javascript-format` marker -- not sure what weblate and other tools will do about that.

```diff
--- /tmp/podman.pot	2024-06-24 14:40:19.848671108 +0200
+++ po/podman.pot	2024-06-24 14:40:37.571759516 +0200
@@ -74,7 +74,7 @@
 msgstr[1] ""
 
 #: src/Containers.jsx:379
-#, c-format
+#, javascript-format
 msgid "$0% of $1 limit"
 msgstr ""
 
@@ -143,12 +143,11 @@
 msgstr ""
 
 #: src/PruneUnusedContainersModal.jsx:96 src/ContainerRenameModal.jsx:97
-#: src/ImageRunModal.jsx:1175 src/ContainerCheckpointModal.jsx:50
-#: src/ForceRemoveModal.jsx:25 src/ContainerDeleteModal.jsx:34
-#: src/ContainerRestoreModal.jsx:53 src/ImageSearchModal.jsx:152
-#: src/ImageDeleteModal.jsx:98 src/ContainerCommitModal.jsx:157
-#: src/PodActions.jsx:52 src/PruneUnusedImagesModal.jsx:97
-#: src/PodCreateModal.jsx:213
+#: src/ContainerCheckpointModal.jsx:50 src/ForceRemoveModal.jsx:25
+#: src/ContainerDeleteModal.jsx:34 src/ContainerRestoreModal.jsx:53
+#: src/ImageSearchModal.jsx:152 src/ImageDeleteModal.jsx:98
+#: src/ContainerCommitModal.jsx:157 src/PodActions.jsx:52
+#: src/PruneUnusedImagesModal.jsx:97 src/PodCreateModal.jsx:213
 msgid "Cancel"
 msgstr ""
 
@@ -250,7 +249,7 @@
 msgid "Containers"
 msgstr ""
 
-#: src/ImageRunModal.jsx:1172 src/PodCreateModal.jsx:210
+#: src/PodCreateModal.jsx:210
 msgid "Create"
 msgstr ""
 
@@ -258,19 +257,10 @@
 msgid "Create a new image based on the current state of the $0 container."
 msgstr ""
 
-#: src/ImageRunModal.jsx:1169
-msgid "Create and run"
-msgstr ""
-
-#: src/Containers.jsx:773 src/ImageRunModal.jsx:1166 src/Images.jsx:400
-#: src/Images.jsx:409
+#: src/Containers.jsx:773 src/Images.jsx:400 src/Images.jsx:409
 msgid "Create container"
 msgstr ""
 
-#: src/ImageRunModal.jsx:1166
-msgid "Create container in $0"
-msgstr ""
-
 #: src/Containers.jsx:856
 msgid "Create container in pod"
 msgstr ""

```

allisonkarlitskaya (approved):
Don't like it, but trust you that it's the best way.

PR #1772, opened 2024-06-28
Settle containers lifecycle

jelly:
Played around in this PR https://github.com/cockpit-project/cockpit-podman/pull/1771

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1379,6 +1379,10 @@ class TestApplication(testlib.MachineCase):
             self.execute(auth, "podman exec swamped-crate sh -c 'pkill dd'")
 
             # picket-fence has a memory limit
+
+            # Wait for CPU usage to settle before checking the memory usage, on
+            # a busy host the awk script takes longer to run.
+            b.wait(lambda: get_cpu_usage(IMG_ALPINE) < 10)
             memory = self.getContainerAttr(IMG_ALPINE, "Memory")
             m = re.match(r'([0-9]+)%([0-9.]+) MB([0-9]+)% of ([0-9.]+) MB limit$', memory)
             self.assertTrue(m, memory)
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -1358,6 +1358,8 @@ class TestApplication(testlib.MachineCase):
         b.wait(lambda: self.getContainerAttr(IMG_BUSYBOX, "Memory") != "")
         memory = self.getContainerAttr(IMG_BUSYBOX, "Memory")
         if auth or self.has_cgroupsV2:
+            # Wait for CPU usage to settle before saving CPU usage for comparison later
+            b.wait(lambda: get_cpu_usage(IMG_BUSYBOX) < 25)
```
martinpitt, line 1362: Good catch. Might need a `b.wait_timeout(60)` even, but let's see!

martinpitt (approved):
If this goes green, it'll at least improve things. Thanks!

test/check-application
jelly: Doesn't seem to be needed, I tested this in the test PR and it didn't fail on that so far.

martinpitt: :+1:

martinpitt:
[This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1772-e74a5c7c-20240628-100949-fedora-coreos/log.html#17) is related. Not sure if it's a missed spot or the patch here needs to try/wait harder?

jelly:
And it goes red of course. The [screenshot](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1772-e74a5c7c-20240628-100949-fedora-coreos/TestApplication-testLifecycleOperationsUser-fedora-coreos-127.0.0.2-2801-FAIL.png) clearly shows that not all memory was allocated yet. 

I could change this to wait on memory to be 300, that would be better, because CPU usage is not a real indicator anyway. or alternatively wait for a smaller cpu usage, locally it is `0%`. Lowered it to `< 10`, lets see now it might need a longer timeout...

jelly:
There is one failure left, where it flakes in a different way. I'd say this is progress :)

https://cockpit-logs.us-east-1.linodeobjects.com/pull-1772-338bd188-20240628-103015-rhel-9-4/log.html#38

martinpitt (approved):
Thanks!

PR #1777, opened 2024-07-01
Show all dates/times as relative

martinpitt:
We want to drop the date-fns runtime dependency, and already did so in all other projects. For consistency, show all dates (such as container creation, health checks, or image history) as relative times via `timeformat.distanceToNow()`, with a tooltip that shows the precise absolute time. This is similar to what we did in machines [1].

Introduce a `RelativeTime` helper component for that, and teach it to accept both ISO-formatted strings as well as raw timestamps to get rid of a few repeated `Date.parse()` calls. Also, most time stamps that we get are in ms (just like `Date` itself), so make it accept that instead of seconds.

[1] https://github.com/cockpit-project/cockpit-machines/commit/b63a54e6721

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -1,10 +1,11 @@
 import React, { useContext } from "react";
 
-import cockpit from 'cockpit';
-
 import { debounce } from 'throttle-debounce';
-import * as dfnlocales from 'date-fns/locale';
-import { formatRelative } from 'date-fns';
+import { Tooltip } from "@patternfly/react-core/dist/esm/components/Tooltip";
+
+import cockpit from 'cockpit';
+import * as timeformat from 'timeformat';
+
 const _ = cockpit.gettext;
 
 export const PodmanInfoContext = React.createContext();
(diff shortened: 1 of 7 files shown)


src/util.js
```diff
@@ -39,10 +40,15 @@ export function truncate_id(id) {
...
 }
 
-export function localize_time(unix_timestamp) {
-    const locale = (cockpit.language == "en") ? dfnlocales.enUS : dfnlocales[cockpit.language.replace('_', '')];
-    return formatRelative(unix_timestamp * 1000, Date.now(), { locale });
-}
+// this supports formatted strings (via Date.parse) or raw timestamps
+export const RelativeTime = ({ time }) => {
+    if (!time)
+        return null;
```
cockpituous, line 46: This added line is not executed by any test.

cockpituous, line 46: This added line is not executed by any test.

martinpitt: Yeah, this is more like belt-and-suspenders. We directly plumb the podman API to this component, and at least in the past these fields could be null or undefined.

martinpitt:
I pushed another commit which goes all the way to dropping date-fns. Rest in peace!

src/util.js
cockpituous, line 46: This added line is not executed by any test.

martinpitt:
[uh, what now](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1777-c948d570-20240701-154314-rhel-8-10/log.html)? That doesn't feel related. But I'll save this and go back to my original proposal, and just save the last step as a follow-up.

src/util.js
cockpituous, line 46: This added line is not executed by any test.

jelly (approved):
Thanks!

PR #1781, opened 2024-07-03
eslint: configure import sorting rules

jelly:
This more or less enforces existing practice, but cleans up many places where we've deviated from that.  We sort into these groups:

- 'react', if present, always first
- external (`node_modules/`) imports
- cockpitlib (`pkg/lib/`) imports
- parent (`../`) imports
- sibling (`./`) imports

---

This is taken from cockpit-files except the restriction on extensions, as we don't have Typescript here, this likely will flake a bunch but lets hope it won't :crossed_fingers:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,6 +17,16 @@
                 "MemberExpression": 2,
                 "ignoredNodes": [ "JSXAttribute" ]
             }],
+        "import/order": ["error",
+            {
+                "alphabetize": { "order": "asc" },
+                "groups": ["builtin", "external", "internal", "parent", "sibling"],
+                "newlines-between": "always",
+                "pathGroupsExcludedImportTypes": ["react"],
+                "pathGroups": [
+                    { "pattern": "react", "group": "builtin", "position": "before" }
+                ]
+            }],
         "newline-per-chained-call": ["error", { "ignoreChainWithDepth": 2 }],
         "no-var": "error",
         "lines-between-class-members": ["error", "always", { "exceptAfterSingleLine": true }],
@@ -51,5 +61,10 @@
     "globals": {
         "require": "readonly",
         "module": "readonly"
+    },
+    "settings": {
+        "import/resolver": {
+            "node": { "moduleDirectory": [ "pkg/lib" ], "extensions": [ ".js", ".jsx", ".ts", ".tsx" ] }
+        }
     }
 }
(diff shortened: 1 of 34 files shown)


src/index.js
```diff
@@ -19,7 +19,9 @@
 
 import "cockpit-dark-theme";
 import React from 'react';
+
```
martinpitt, line 22: This newline is a bit weird, as react and react-dom really belong together. Possibly that needs to go into the config, too?

.eslintrc.json
```diff
@@ -17,6 +17,16 @@
...
                 "ignoredNodes": [ "JSXAttribute" ]
             }],
+        "import/order": ["error",
+            {
+                "alphabetize": { "order": "asc" },
+                "groups": ["builtin", "external", "internal", "parent", "sibling"],
+                "newlines-between": "always",
+                "pathGroupsExcludedImportTypes": ["react"],
+                "pathGroups": [
+                    { "pattern": "react", "group": "builtin", "position": "before" }
```
martinpitt, line 27: Is that substring match? It either doesn't seem to treat "react-dom/client" as matching here, or it's treated as a separate group.

martinpitt (commented):
Yes please! just a question/fine-tuning.

.eslintrc.json
jelly: This is super complicated https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/order.md

https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/order.md#pathgroups-array-of-objects


This uses `minimatch` under the hood, I just allowed `react` and `react-dom` as `react.*` also matches patternfly imports @patternfly/react-core etc.

jelly: So with `debug`
```
matchOne {
  this: Minimatch {
    options: { debug: true },
    set: [ [Array] ],
    pattern: 'react(-dom)?',
    regexp: null,
    negate: false,
    comment: false,
    empty: false,
    partial: false,
    globSet: [ 'react(-dom)?' ],
    debug: [Function: debug],
    globParts: [ [Array] ]
  },
  file: [ 'timeformat' ],
  pattern: [
    /^(?=.)react\(-dom\)[^/]$/ {
      _glob: 'react(-dom)?',
      _src: '(?=.)react\\(-dom\\)[^/]'
    }
  ]
}
```

martinpitt: Ah, don't sweat it -- I'd rather enforce this to be consistent, but "slightly weird", instead of the current wild mess. We can tweak this in a follow-up.

martinpitt (approved):
Cheers!

PR #1807, opened 2024-07-30
Fix searching containers with full URL

tomasmatus:
Fixes bug that makes it impossible to search containers with full URL such as `docker.io/{user}/{container_name}`.

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -516,7 +516,7 @@ export class ImageRunModal extends React.Component {
         }
 
         // Strip out all non-allowed container image characters when filtering.
-        let regexString = searchText.replace(/[^\w_.:-]/g, "");
+        let regexString = searchText.replace(/[^/\w_.:-]/g, "");
         // Strip image registry option if set for comparing results for docker.io searching for docker.io/fedora
         // returns docker.io/$username/fedora for example.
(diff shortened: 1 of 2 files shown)


tomasmatus:
Funnily enough tests didn't catch this likely because of debounce in the search input + delay to resolve all promises to search repositories so UI didn't re-render fast enough to cause test failure.

src/ImageRunModal.jsx
```diff
@@ -516,7 +516,7 @@ export class ImageRunModal extends React.Component {
         }
 
         // Strip out all non-allowed container image characters when filtering.
-        let regexString = searchText.replace(/[^\w_.:-]/g, "");
+        let regexString = searchText.replace(/[^/\w_.:-]/g, "");
```
jelly, line 519: So this strips out the `/` right? But that is also handled below?

tomasmatus: eg: without this fix when you search for `docker.io/library/fedora` the final input regex string ends up being `/docker.iolibraryfedora/i` which when filtering podman search results matches nothing:
image

with this fix the final regex is `/\/library\/fedora/i` so it can match viable images: 
image

jelly: Aha, silly me :)

jelly (approved):

PR #1817, opened 2024-08-09
build: Switch to @xterm/xterm and @xterm/addon-canvas

mvollmer:
These are the new names, and they have more recent releases.


martinpitt:
Wohoo, thanks! .github/dependabot.yml needs adjusting too -- I leave it up to you if you want to do this right here (cleaner) or as an immediate follow-up.

martinpitt (approved):
Danke!

Issue #1220, opened 2023-03-02
The create container dialog should accept a full "container url"


jelly:
Some registries do not support searching as this is not a mandatory feature, as we build our "create container" dialog around searching we should allow users to input a full container url for example `ghcr.io/github/pages-gem`.

This now fails and shows an ugly error:

![image](https://user-images.githubusercontent.com/67428/222410502-cbb2c49f-ebbb-47e3-a0a9-448edb7820ba.png)

## Reproducer:

* Click create new container
* As `Image` input: `ghcr.io/github/pages-gem`
* Try to create the container without the dropdown clearing the input field

## Proposed solution

When a user provides a full container url, try to not search or if that's too complicated and search fails show a warning not an error. Also try to not show the dropdown with "no images found" as that is rather confusing.

Alternatively we need to make our `download new image dialog` prominent but to me that feels like adding some ducktape and working around the issue.

Related to: https://github.com/cockpit-project/cockpit-podman/issues/819

https://issues.redhat.com/browse/RHEL-13022

jelly:
I believe we talked about it during the re-design of the "Create image dialog" but never to implementing it.

jelly:
Seems we can use something similiar to `skopeo manifest inspect docker://fo` or `podman manifest inspect ghcr.io/github/pages-gem`.

```
[jelle@t14s][~]%curl -v -X GET -s -g --no-buffer --unix-socket /run/user/1000/podman/podman.sock 'http://localhost/v1.12/libpod/manifests/ghcr.io/github/pages-gem/json'
*   Trying /run/user/1000/podman/podman.sock:0...
* Connected to localhost (/run/user/1000/podman/podman.sock) port 80 (#0)
> GET /v1.12/libpod/manifests/ghcr.io/github/pages-gem/json HTTP/1.1
> Host: localhost
> User-Agent: curl/8.0.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Api-Version: 1.41
< Content-Type: application/json
< Libpod-Api-Version: 4.4.3
< Server: Libpod/4.4.3 (linux)
< X-Reference-Id: 0xc000694210
< Date: Thu, 30 Mar 2023 10:25:01 GMT
< Content-Length: 104
<
{"schemaVersion":2,"mediaType":"application/vnd.docker.distribution.manifest.v2+json","manifests":null}
* Connection #0 to host localhost left intact
[...]

PR #1821, opened 2024-08-14
fix searching on non-searchable image registries

tomasmatus:
fixes: #1220

Add support to pull images from registries which do not support search API.

https://issues.redhat.com/browse/COCKPIT-1148

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -62,6 +62,8 @@ const HealthCheckOnFailureActionOrder = [
     { value: 2, label: _("Force stop") },
 ];
 
+const RE_CONTAINER_TAG = /:[\w\-\d]+$/;
+
 export class ImageRunModal extends React.Component {
     constructor(props) {
         super(props);
@@ -351,68 +353,119 @@ export class ImageRunModal extends React.Component {
         });
     };
 
+    buildFilterRegex(searchText, includeRegistry) {
+        // Strip out all non-allowed container image characters when filtering.
+        let regexString = searchText.replace(/[^/\w_.:-]/g, "");
+        // drop registry from regex to allow filtering only by container names
+        if (!includeRegistry && regexString.includes('/')) {
+            regexString = '/' + searchText.split('/')
+                    .slice(1)
+                    .join('/');
+        }
+
+        return new RegExp(regexString, 'i');
+    }
+
     onSearchTriggered = value => {
+        let dialogError = "";
+        let dialogErrorDetail = "";
+
+        const changeDialogError = (reason) => {
+            // Only report first encountered error
+            if (dialogError === "" && dialogErrorDetail === "") {
+                dialogError = _("Failed to search for new images");
+                // TODO: add registry context, podman does not include it in the reply.
+                dialogErrorDetail = reason ? cockpit.format(_("Failed to search for images: $0"), reason.message) : _("Failed to search for images.");
+            }
+        };
+
+        const imageExistsLocally = (searchText, localImages) => {
+            const regex = this.buildFilterRegex(searchText, true);
+            return localImages.some(localImage => localImage.Name.search(regex) !== -1);
+        };
+
+        const handleManifestsQuery = (result) => {
+            if (result.status === "fulfilled") {
+                return JSON.parse(result.value);
+            } else if (result.reason.status !== 404) {
+                changeDialogError(result.reason);
+            }
+
+            return null;
+        };
+
         // Do not call the SearchImage API if the input string  is not at least 2 chars,
         // The comparison was done considering the fact that we miss always one letter due to delayed setState
         if (value.length < 2)
             return;
-
-        // Don't search for a value with a tag specified
-        const patt = /:[\w|\d]+$/;
-        if (patt.test(value)) {
-            return;
-        }
 
         if (this.activeConnection)
             this.activeConnection.close();
 
         this.setState({ searchFinished: false, searchInProgress: true });
         this.activeConnection = rest.connect(client.getAddress(this.isSystem()), this.isSystem());
-        let searches = [];
+        const searches = [];
 
-        // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
-        // podman search in the user specified registry.
-        if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
-            searches.push(this.activeConnection.call({
-                method: "GET",
-                path: client.VERSION + "libpod/images/search",
-                body: "",
-                params: {
-                    term: value,
-                }
-            }));
-        } else {
-            searches = searches.concat(utils.fallbackRegistries.map(registry =>
-                this.activeConnection.call({
+        // Try to get specified image manifest
+        searches.push(this.activeConnection.call({
+            method: "GET",
+            path: client.VERSION + "libpod/manifests/" + value + "/json",
+            body: "",
+        }));
+
+        // Don't start search queries when tag is specified as search API doesn't support
+        // searching for specific image tags
+        // instead only rely on manifests query (requires image:tag name)
+        if (!RE_CONTAINER_TAG.test(value)) {
+            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // podman search in the user specified registry.
+            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
+                searches.push(this.activeConnection.call({
                     method: "GET",
                     path: client.VERSION + "libpod/images/search",
                     body: "",
                     params: {
-                        term: registry + "/" + value
+                        term: value,
                     }
-                })));
+                }));
+            } else {
+                searches.push(...utils.fallbackRegistries.map(registry =>
+                    this.activeConnection.call({
+                        method: "GET",
+                        path: client.VERSION + "libpod/images/search",
+                        body: "",
+                        params: {
+                            term: registry + "/" + value
+                        }
+                    })));
+            }
         }
 
         Promise.allSettled(searches)
                 .then(reply => {
                     if (reply && this._isMounted) {
                         let imageResults = [];
-                        let dialogError = "";
-                        let dialogErrorDetail = "";
+                        const manifestResult = handleManifestsQuery(reply[0]);
 
-                        for (const result of reply) {
+                        for (let i = 1; i < reply.length; i++) {
+                            const result = reply[i];
                             if (result.status === "fulfilled") {
                                 imageResults = imageResults.concat(JSON.parse(result.value));
-                            } else {
-                                dialogError = _("Failed to search for new images");
-                                // TODO: add registry context, podman does not include it in the reply.
-                                dialogErrorDetail = result.reason ? cockpit.format(_("Failed to search for images: $0"), result.reason.message) : _("Failed to search for images.");
+                            } else if (!manifestResult && !imageExistsLocally(value, this.props.localImages)) {
+                                changeDialogError(result.reason);
                             }
                         }
+
+                        // Add manifest query result if search query did not find the same image
+                        if (manifestResult && !imageResults.find(image => image.Name === value)) {
+                            manifestResult.Name = value;
+                            imageResults.push(manifestResult);
+                        }
+
                         // Group images on registry
                         const images = {};
                         imageResults.forEach(image => {
-                            // Add Tag is it's there
+                            // Add Tag if it's there
                             image.toString = function imageToString() {
                                 if (this.Tag) {
                                     return this.Name + ':' + this.Tag;
@@ -434,6 +487,7 @@ export class ImageRunModal extends React.Component {
                                 images[index] = [image];
                             }
                         });
+
                         this.setState({
                             imageResults: images || {},
                             searchFinished: true,
@@ -484,17 +538,20 @@ export class ImageRunModal extends React.Component {
             isImageSelectOpen: false,
             command,
             entrypoint,
+            dialogError: "",
+            dialogErrorDetail: "",
         });
     };
 
     handleImageSelectInput = value => {
+        const trimmedValue = value.trim();
         this.setState({
-            searchText: value,
+            searchText: trimmedValue,
             // Reset searchFinished status when text input changes
             searchFinished: false,
             selectedImage: "",
         });
-        this.onSearchTriggered(value);
+        this.onSearchTriggered(trimmedValue);
     };
 
     debouncedInputChanged = debounce(300, this.handleImageSelectInput);
@@ -524,14 +581,7 @@ export class ImageRunModal extends React.Component {
             imageRegistries.push(this.state.searchByRegistry);
         }
 
-        // Strip out all non-allowed container image characters when filtering.
-        let regexString = searchText.replace(/[^/\w_.:-]/g, "");
-        // Strip image registry option if set for comparing results for docker.io searching for docker.io/fedora
-        // returns docker.io/$username/fedora for example.
-        if (regexString.includes('/')) {
-            regexString = searchText.replace(searchText.split('/')[0], '');
-        }
-        const input = new RegExp(regexString, 'i');
+        const input = this.buildFilterRegex(searchText, false);
 
         const results = imageRegistries
                 .map((reg, index) => {
--- a/test/check-application
+++ b/test/check-application
@@ -15,7 +15,7 @@ REGISTRIES_CONF = """
 registries = ['localhost:5000', 'localhost:6000']
 
 [registries.insecure]
-registries = ['localhost:5000', 'localhost:6000']
+registries = ['localhost:80', 'localhost:5000', 'localhost:6000']
 """
 
 NOT_RUNNING = ["Exited", "Stopped"]
@@ -27,6 +27,24 @@ IMG_BUSYBOX = "localhost/test-busybox"
 IMG_BUSYBOX_LATEST = IMG_BUSYBOX + ":latest"
 IMG_REGISTRY = "localhost/test-registry"
 IMG_REGISTRY_LATEST = IMG_REGISTRY + ":latest"
+
+# nginx configuration to simulate a registry without search API (like ghcr.io)
+NGINX_DEFAULT_CONF = """
+server {
+  listen 80;
+  listen [::]:80;
+  server_name localhost;
+
+  location / {
+    proxy_pass http://localhost:5000/;
+  }
+
+  # Simulate a registry without search API, like ghcr.io
+  location ^~ /v2/_catalog {
+    return 403;
+  }
+}
+"""
 
 
 def podman_version(cls):
@@ -3048,6 +3066,92 @@ class TestApplication(testlib.MachineCase):
             self.assertIn("Podman", b.text("#host-apps"))
         else:
             self.assertNotIn("Podman", b.text("#host-apps"))
+
+    @testlib.skipImage("nginx not installed", "centos-*", "rhel-*", "debian-*", "ubuntu-*", "arch")
+    @testlib.skipOstree("nginx not installed")
+    def testNonSearchableRegistry(self):
+        b = self.browser
+        m = self.machine
+
+        self.setupRegistry()
+
+        # Nginx config simulates behavior of ghcr.io in terms of `podman search` and `podman manifest inspect`
+        self.write_file("/etc/nginx/conf.d/default.conf", NGINX_DEFAULT_CONF,
+            post_restore_action="systemctl stop nginx.service; setsebool -P httpd_can_network_connect 0")
+        self.execute(True, """
+            setsebool -P httpd_can_network_connect 1
+            systemctl start nginx.service
+        """)
+
+        container_name = "localhost:80/my-busybox"
+        m.execute(f"""
+            podman manifest create my-busybox
+            podman manifest add my-busybox {IMG_BUSYBOX}
+            podman manifest push my-busybox {container_name}
+            podman manifest push my-busybox {container_name + ":nosearch-tag"}
+            podman manifest push my-busybox "localhost:5000/my-busybox:search-tag"
+        """)
+        # Untag busybox image which duplicates the image we are about to download
+        m.execute(f"""
+            podman manifest rm localhost/my-busybox
+            podman rmi -f {IMG_BUSYBOX}
+        """)
+
+        self.login()
+
+        b.click("#containers-containers button.pf-v5-c-button.pf-m-primary")
+        b.set_input_text("#run-image-dialog-name", "new-busybox")
+        # Searching for container by prefix fails
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:80/my-busy")
+        b.wait_text("button.pf-v5-c-select__menu-item.pf-m-disabled", "No images found")
+        b.wait_in_text(".pf-v5-c-alert.pf-m-danger", "couldn't search registry")
+
+        # Giving full image name finds valid manifest therefore image is pullable
+        b.set_input_text("#create-image-image-select-typeahead", container_name)
+        b.wait_text("button.pf-v5-c-select__menu-item:not(.pf-m-disabled)", container_name)
+        # Select image and create a container
+        b.click(f"button.pf-v5-c-select__menu-item:contains({container_name})")
+        b.click("#create-image-create-btn")
+
+        # Wait for download to finish
+        sel = "span:not(.downloading)"
+        b.wait(lambda: self.getContainerAttr(container_name, "State", sel) in "Created")
+
+        checkImage(b, f"{container_name}:latest", "system")
+
+        b.click("#containers-containers button.pf-v5-c-button.pf-m-primary")
+        # "couldn't search registry" error is hidden when some local image is found
+        b.set_input_text("#create-image-image-select-typeahead", "")
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:80/my-busy")
+        b.wait_text("button.pf-v5-c-select__menu-item:not(.pf-m-disabled)", f"{container_name}:latest")
+        b.wait_not_present(".pf-v5-c-alert.pf-m-danger")
+        b.click('button.pf-v5-c-toggle-group__button:contains("Local")')
+
+        # Error is shown again when no image was found locally
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:80/their-busy")
+        b.wait_text("button.pf-v5-c-select__menu-item.pf-m-disabled", "No images found")
+        b.wait_in_text(".pf-v5-c-alert.pf-m-danger", "couldn't search registry")
+
+        # Searching for full container name with tag finds the image
+        b.set_input_text("#run-image-dialog-name", "tagged-busybox")
+        # Search should still work with spaces around image name
+        b.set_input_text("#create-image-image-select-typeahead", " localhost:80/my-busybox:nosearch-tag ")
+        b.click('button.pf-v5-c-toggle-group__button:contains("All")')
+        b.wait_text("button.pf-v5-c-select__menu-item:not(.pf-m-disabled)", container_name + ":nosearch-tag")
+        b.wait_js_cond("document.getElementsByClassName('pf-v5-c-select__menu-item').length === 1")
+        b.click(f"button.pf-v5-c-select__menu-item:contains({container_name + ":nosearch-tag"})")
+        b.click("#create-image-create-btn")
+
+        # Wait for download to finish
+        sel = "span:not(.downloading)"
+        b.wait(lambda: self.getContainerAttr("tagged-busybox", "State", sel) in "Created")
+        checkImage(b, f"{container_name}:nosearch-tag", "system")
+
+        # Check that manifest search with tag also works on searchable repository
+        b.click("#containers-containers button.pf-v5-c-button.pf-m-primary")
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:5000/my-busybox:search-tag")
+        b.wait_text("button.pf-v5-c-select__menu-item:not(.pf-m-disabled)", "localhost:5000/my-busybox:search-tag")
+        b.wait_js_cond("document.getElementsByClassName('pf-v5-c-select__menu-item').length === 1")
 
 
 if __name__ == '__main__':
--- a/test/browser/main.fmf
+++ b/test/browser/main.fmf
@@ -4,6 +4,7 @@ require:
   - cockpit-ws
   - cockpit-system
   - criu
+  - nginx
 duration: 30m
 
 /system:


src/ImageRunModal.jsx
```diff
@@ -384,6 +401,15 @@ export class ImageRunModal extends React.Component {
                 })));
         }
 
+        // Query if used url is a valid container that can be pulled
+        if ((value.includes('.') || value.startsWith("localhost")) && value.includes('/')) {
```
tomasmatus, line 405: oh yeah this was only for dev testing, needs to be improved

src/ImageRunModal.jsx
```diff
@@ -393,17 +419,43 @@ export class ImageRunModal extends React.Component {
 
                         for (const result of reply) {
                             if (result.status === "fulfilled") {
-                                imageResults = imageResults.concat(JSON.parse(result.value));
+                                try {
+                                    imageResults = imageResults.concat(JSON.parse(result.value));
+                                } catch (err) {
+                                    console.error(err);
```
cockpituous, line 425: These 2 added lines are not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -393,17 +419,43 @@ export class ImageRunModal extends React.Component {
...
+                                }
                             } else {
+                                // manifests query returns 404 Not Found in case used URL is not an image
+                                // supress this error as it will either result in "No images found" or prioritize other erros from search queries
+                                if (result.reason.status === 404) {
+                                    continue;
+                                }
+                                // Supress "could't search registry" error when local image is found
+                                if (result.reason.message.includes("couldn't search registry") && imageExistsLocally(value, this.props.localImages)) {
+                                    continue;
```
cockpituous, line 435: These 2 added lines are not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -393,17 +419,43 @@ export class ImageRunModal extends React.Component {
...
-                            // Add Tag is it's there
+                            // Manifests query
+                            if (image.manifests !== undefined) {
+                                // Only use manifests result if nothing else was found (non-searchable image repository)
+                                if (imageResults.length === 1) {
+                                    image.Name = value;
+                                    dialogError = "";
+                                    dialogErrorDetail = "";
+                                } else {
+                                    return;
```
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 425: These 2 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

garrett:
I already had this container on my system:

image

I tried a few example ones, but perhaps they're not actually there...

image

Then I tried a container that GitHub talked about called "Super Linter", which has this as a container URI:
`ghcr.io/github/super-linter@sha256:b723c817982b85f6f98c9c24a653f9f760bf5941577f5ea9532d31bd89aa8823`
...as seen @ https://github.com/orgs/github/packages/container/super-linter/207523715
image

If you add the URI without the @..., it seems like it'd work:

image

But you have to specifically click on the item in the dropdown. (Is that intended?) If you change the focus, the entry goes blank. Also, if you try to edit the text (like if you have a typo, like copying the URI incorrectly), the entry also goes blank.

However, it does seem to work if you click on the entry, so that's great!

I see a related discussion at https://github.com/orgs/community/discussions/26279 which uses tags/list but it seems things need a token, so I guess that won't work for us.

src/ImageRunModal.jsx
cockpituous, line 425: These 2 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 425: These 2 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 425: These 2 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 425: These 2 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -342,7 +342,24 @@ export class ImageRunModal extends React.Component {
...
     };
 
+    buildFilterRegex(searchText) {
+        // Strip out all non-allowed container image characters when filtering.
+        let regexString = searchText.replace(/[^/\w_.:-]/g, "");
+        // Strip image registry option if set for comparing results for docker.io searching for docker.io/fedora
+        // returns docker.io/$username/fedora for example.
+        if (regexString.includes('/')) {
+            regexString = searchText.replace(searchText.split('/')[0], '');
+        }
```
martinpitt, line 352: I know this is moved code, but the comment doesn't match up what the code does:
```
> searchText = 'docker.io/user/image:tag'
'docker.io/user/image:tag'
> searchText.replace(searchText.split('/')[0], '')
'/user/image:tag'
```

I.e. it does split off the registry, but the "returns docker.io" comment is wrong. Moreover, that's a weird way of doing that. It relies on `.replace()` changing only the first match by default, but that's by no means intuitive.

I'd find it more readable/explicit to do

    searchText.split('/').slice(1).join('/')

(and possibly prepend another `/` if the search API actually cares for that).

This refactoring could be split out into a separate commit.

Also, I don't understand the comment: "searching for docker.io/fedora" may give you some user/fedora results via the dockerhub search API, but how is that related to `buildFilterRegex()`?

Finally, what's the rationale for stripping off the registry in the first place? Is that a requirement for the search API?

src/ImageRunModal.jsx
```diff
@@ -342,7 +342,24 @@ export class ImageRunModal extends React.Component {
...
+
+        return new RegExp(regexString, 'i');
+    }
+
     onSearchTriggered = value => {
+        const imageExistsLocally = (searchText, localImages) => {
+            const regex = this.buildFilterRegex(searchText);
+            return localImages.some(localImage => localImage.Name.search(regex) !== -1);
```
martinpitt, line 360: `buildFilterRegex` strips off the first `/` component (the registry name). This *may* be needed for the docker.io search API (see my question above), but it's wrong here. If I search for "docker.io/fedora" and have a local "quay.io/mpitt-fedora/test:img" image, this should *not* match. This definitively needs to include the registry here. So maybe make that an option of that function?

src/ImageRunModal.jsx
martinpitt: I ignore this for now. What's that supposed to do? But the "localhost" special case must go, otherwise the integration test is meaningless.

src/ImageRunModal.jsx
martinpitt, line 425: How can this fail? If the promise succeeds, it should not throw random garbage at you, but properly formatted JSON. If this failed because the user typed some garbage, then it should be reflected in the UI somehow (but then *not* through a `JSON.parse()` exception, but a semantic error)

src/ImageRunModal.jsx
```diff
@@ -393,17 +419,43 @@ export class ImageRunModal extends React.Component {
...
                             } else {
+                                // manifests query returns 404 Not Found in case used URL is not an image
+                                // supress this error as it will either result in "No images found" or prioritize other erros from search queries
+                                if (result.reason.status === 404) {
+                                    continue;
+                                }
+                                // Supress "could't search registry" error when local image is found
+                                if (result.reason.message.includes("couldn't search registry") && imageExistsLocally(value, this.props.localImages)) {
+                                    continue;
+                                }
```
martinpitt, line 436: I get it that an image which is locally present needs some special treatment. But the main use case here is that someone copy&pastes a full image name like "ghcr.io/cockpit-project/tasks" into the input box and expects our page to download and run that exact image. I suppose that triggers the "couldn't search registry" case (as ghcr.io doesn't have that  search API), but it would't be present locally already. How/where is that handled?

src/ImageRunModal.jsx
cockpituous, line 425: These 2 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -393,17 +419,43 @@ export class ImageRunModal extends React.Component {
...
                         }
+
                         // Group images on registry
                         const images = {};
                         imageResults.forEach(image => {
-                            // Add Tag is it's there
+                            // Manifests query
+                            if (image.manifests !== undefined) {
+                                // Only use manifests result if nothing else was found (non-searchable image repository)
+                                if (imageResults.length === 1) {
```
martinpitt, line 449: This needs some further explanation in a comment. What is that single result which we are ignoring? Can this be asserted/checked?

test/check-application
```diff
@@ -28,6 +28,22 @@ IMG_BUSYBOX_LATEST = IMG_BUSYBOX + ":latest"
 IMG_REGISTRY = "localhost/test-registry"
 IMG_REGISTRY_LATEST = IMG_REGISTRY + ":latest"
 
+NGINX_DEFAULT_CONF = """
+server {
+  listen 80;
+  listen  [::]:80;
```
martinpitt, line 34: Do we need  both?

test/check-application
```diff
@@ -28,6 +28,22 @@ IMG_BUSYBOX_LATEST = IMG_BUSYBOX + ":latest"
...
+  server_name localhost;
+
+  location / {
+    proxy_pass http://localhost:5000/;
+  }
+
+  location ^~ /v2/_catalog {
```
martinpitt, line 41: Please add a comment for our future selves: Something like "Simulate a registry without search API, like ghcr.io"

test/check-application
```diff
@@ -3005,6 +3026,65 @@ class TestApplication(testlib.MachineCase):
...
+    @testlib.skipOstree("nginx not installed")
+    def testNonSearchableRegistry(self):
+        b = self.browser
+        m = self.machine
+
+        self.setupRegistry()
+
+        # Nginx config simulates behavior of ghcr.io in terms of `podman search` and `podman manifest inspect`
+        self.execute(True, f"""
+            echo '{NGINX_DEFAULT_CONF}' > /etc/nginx/conf.d/default.conf
```
martinpitt, line 3034: Note: `self.write_file()` is more convenient and also cleans up after itself.

test/check-application
```diff
@@ -3005,6 +3026,65 @@ class TestApplication(testlib.MachineCase):
...
+            setsebool -P httpd_can_network_connect 1
+            systemctl start nginx.service
+        """)
+        self.addCleanup(m.execute, """
+            systemctl stop nginx.service
+            setsebool -P httpd_can_network_connect 0
+            rm -f /etc/nginx/default.conf
```
martinpitt, line 3041: Asymmetric -- this isn't the file that the code above created. Please use `self.write_file()` :grin:

test/check-application
```diff
@@ -3005,6 +3026,65 @@ class TestApplication(testlib.MachineCase):
...
+        b.click("#create-image-create-btn")
+
+        # Wait for download to finish
+        sel = "span:not(.downloading)"
+        b.wait(lambda: self.getContainerAttr(container_name, "State", sel) in "Created")
+
+        b.click("#containers-containers button.pf-v5-c-button.pf-m-primary")
+        # "couldn't search registry" error is hidden when some local image is found
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:80/my-busy")
+        b.wait_not_present(".pf-v5-c-alert.pf-m-danger")
```
martinpitt, line 3075: You need a positive check before a `wait_not_present()` call, as this is always immediately true. I.e. wait for the found local container to be displayed first.

test/check-application
```diff
@@ -3005,6 +3026,65 @@ class TestApplication(testlib.MachineCase):
...
+        b.click('button.pf-v5-c-toggle-group__button:contains("Local")')
+        b.wait_text("button.pf-v5-c-select__menu-item:not(.pf-m-disabled)", f"{container_name}:latest")
+
+        # Error is shown again when no image was found locally
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:80/their-busy")
+        b.wait_text("button.pf-v5-c-select__menu-item.pf-m-disabled", "No images found")
+        b.wait_in_text(".pf-v5-c-alert.pf-m-danger", "couldn't search registry")
+
+        testlib.sit()
```
martinpitt: :grin:

martinpitt (changes requested):
Cheers! Please update the commit message to explain the change more verbosely (use case, reference to https://issues.redhat.com/browse/COCKPIT-1148 and the correspondig upstream and RHEL issues).

src/ImageRunModal.jsx
tomasmatus: This regex is only used for filtering search results and not for the searching itself. When you specify `docker.io/fedora` it will only search docker.io as expected.

What the comment was probably trying to say is when you search for `docker.io/fedora` the results looks like `docker.io/organization/fedora` and using regex `docker.io/fedora` would not match. So either drop the registry from regex or change regex to work with any organization

src/ImageRunModal.jsx
tomasmatus: True, thanks for catching this

src/ImageRunModal.jsx
tomasmatus: In the case where you'd start typing to search for `ghcr.io/cockpit/ta`. This will start two searches. The standard search query and the manifests query. Standard search will return status 500 and the manifest search will return 404 status.
This is here to cover the standard search which returns 500 but only supresses the warning when regex in `imageExistsLocally` can match an image.

> but it would't be present locally already. How/where is that handled?

the case where manifests search is successful, image does not exist locally and other search queries fail is handled in the code bellow (also the following comment under this)

src/ImageRunModal.jsx
tomasmatus: As mentioned in the comment above this is here to handle case where manifest search is successful but other searches fail and image isn't available locally.

test/check-application
tomasmatus: Can't say for sure but I think it's safer to keep both. `localhost` can resolve to ipv4 or ipv6

src/ImageRunModal.jsx
tomasmatus: It's extremely ugly but this is meant to only fire up the manifests query when user input looks like container image url. It could be just `value.includes('/')`  but I feel like I should come up with something better

src/ImageRunModal.jsx
```diff
@@ -395,15 +423,37 @@ export class ImageRunModal extends React.Component {
...
+                                    image.Name = value;
+                                    dialogError = "";
+                                    dialogErrorDetail = "";
+                                } else {
+                                    return;
```
cockpituous, line 452: This added line is not executed by any test.

tomasmatus:
fedora 41 fails because of this:
```

[root@fedora-41-127-0-0-2-2201 ~]# podman manifest inspect localhost:80/my-busybox
Error: parsing manifest blob "{\"schemaVersion\":2,\"mediaType\":\"application/vnd.oci.image.manifest.v1+json\",\"config\":{\"mediaType\":\"application/vnd.oci.image.config.v1+json\",\"digest\":\"sha256:02154912ea3e0a26dd3d9878cc35592995a559c1894ed9f1f3ad2e67c601e8a3\",\"size\":1051},\"layers\":[{\"mediaType\":\"application/vnd.oci.image.layer.v1.tar+zstd\",\"digest\":\"sha256:979ef0eafe970832b8df3c3233c50a1ceb863501afd0fdd0819f4e74459a269b\",\"size\":783654,\"annotations\":{\"io.github.containers.zstd-chunked.manifest-checksum\":\"sha256:1818e53836a49be9f33ab634b8cecc242402366faff167cc29c346acba6c786c\",\"io.github.containers.zstd-chunked.manifest-position\":\"769759:3878:48899:1\",\"io.github.containers.zstd-chunked.tarsplit-position\":\"773645:9937:339414\"}},{\"mediaType\":\"application/vnd.oci.image.layer.v1.tar+zstd\",\"digest\":\"sha256:5126ce266a4b0023a57d535a42dfdedee252be565ab10967140a8cf5c2babfc6\",\"size\":973884,\"annotations\":{\"io.github.containers.zstd-chunked.manifest-checksum\":\"sha256:9a91d82526737b46f2f6e5d9478f597fe989943445cda912aa8af5e28b7351c9\",\"io.github.containers.zstd-chunked.manifest-position\":\"859954:57329:277072:1\",\"io.github.containers.zstd-chunked.tarsplit-position\":\"917291:56521:1414602\"}}]}" as a "application/vnd.oci.image.manifest.v1+json": Treating single images as manifest lists is not implemented
```

I checked f40 and it gives a warning rather than "not implemented" error. Both are on version 5.2.0 idk what this means/how to fix for now, will look into it

src/ImageRunModal.jsx
cockpituous, line 452: This added line is not executed by any test.

test/check-application
martinpitt: Right, but `[::]` should listen on all interfaces already. Anyway, not important.

src/ImageRunModal.jsx
martinpitt: with "image URL" you mean something like "ghcr.io/owner/name:tag"? (that's not an URL). I'm not sure if that can be captured with a regex -- registries can be pretty much arbitrary and don't *have* to contain a `.` (e.g. `localhost` or a custom defined one), and "localhost" by itself isn't special. Frankly, your update to just check `/` seems to be the best that you can do -- try to query it if it's a valid image/manifest, and if not then use the term as a search query. Or do both in parallel?

src/ImageRunModal.jsx
martinpitt: OK -- this is higher-order hacking and I don't know the background of this, so I'll let it slide :grin:

martinpitt (commented):
Thanks for the fixes! Next round.

src/ImageRunModal.jsx
```diff
@@ -395,15 +423,37 @@ export class ImageRunModal extends React.Component {
                             if (result.status === "fulfilled") {
                                 imageResults = imageResults.concat(JSON.parse(result.value));
                             } else {
+                                // manifests query returns 404 Not Found in case used URL is not an image
+                                // supress this error as it will either result in "No images found" or prioritize other erros from search queries
+                                if (result.reason.status === 404) {
```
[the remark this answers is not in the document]
tomasmatus: I am not sure I can drop the error message string matching as it is the specific search query error I want to ignore.

src/ImageRunModal.jsx
tomasmatus: Another option that comes to mind is to filter out duplicate entries in `filterImages` rather than this seemingly confusing "ignoring manifest result" in specific scenarios

src/ImageRunModal.jsx
```diff
@@ -342,46 +342,71 @@ export class ImageRunModal extends React.Component {
...
+        }));
+
+        // Don't start search queries when tag is specified as search API doesn't support
+        // searching for specific image tags
+        // instead only rely on manifests query (requires full image url)
+        const patt = /:[\w-|\d]+$/;
+        if (!patt.test(value)) {
+            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // podman search in the user specified registry.
+            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
```
cockpituous, line 390: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -342,46 +342,71 @@ export class ImageRunModal extends React.Component {
...
-                })));
+                }));
+            } else {
+                searches = searches.concat(utils.fallbackRegistries.map(registry =>
+                    this.activeConnection.call({
+                        method: "GET",
+                        path: client.VERSION + "libpod/images/search",
+                        body: "",
+                        params: {
+                            term: registry + "/" + value
```
cockpituous, line 406: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 390: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 406: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 390: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 406: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -402,30 +409,22 @@ export class ImageRunModal extends React.Component {
...
         Promise.allSettled(searches)
                 .then(reply => {
                     if (reply && this._isMounted) {
                         let imageResults = [];
                         let dialogError = "";
                         let dialogErrorDetail = "";
 
-                        for (const result of reply) {
+                        for (let i = 0; i < reply.length; i++) {
+                            const result = reply[i];
```
martinpitt, line 420: You can do this more elegantly with `reply.forEach((result, i) => { ...})`

src/ImageRunModal.jsx
```diff
@@ -402,7 +403,8 @@ export class ImageRunModal extends React.Component {
         }
 
         // Query if used url is a valid container that can be pulled
-        if ((value.includes('.') || value.startsWith("localhost")) && value.includes('/')) {
+        // TODO
```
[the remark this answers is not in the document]
martinpitt: Always checking the manifest sounds good, thanks!

src/ImageRunModal.jsx
martinpitt: I feel like it should be the *other* way around: If the user typed in something which is a valid manifest -- i.e. the string uniquely identifies a valid container -- then this should count as the one and only search result IMHO. Search should then be discarded (and most likely won't bring up anything else anyway?). However, that's only true if there is no tag: if I type "docker.io/ubuntu" it should show me tags like ":stable", ":devel" and so on.

So, tricky. 

But that code is still horribly confusing. First of all the length check happens *inside* of an iteration of that very list. And second, at that point the information is lost where it came from -- from the manifest or the search query. I think you should handle these two in completely separate code paths, they are just too different.

src/ImageRunModal.jsx
```diff
@@ -366,12 +366,6 @@ export class ImageRunModal extends React.Component {
         if (value.length < 2)
             return;
 
-        // Don't search for a value with a tag specified
-        const patt = /:[\w|\d]+$/;
```
martinpitt, line 352: This means "the tag name can contain a literal `|`" which I believe is not true. A regexp character group is just a simple concatenation, i.e. `[\w\d]`.

src/ImageRunModal.jsx
```diff
@@ -386,27 +380,33 @@ export class ImageRunModal extends React.Component {
...
-                }
-            }));
-        } else {
-            searches = searches.concat(utils.fallbackRegistries.map(registry =>
-                this.activeConnection.call({
+        // Don't start search queries when tag is specified as search API doesn't support
+        // searching for specific image tags
+        // instead only rely on manifests query (requires full image url)
+        const patt = /:[\w-|\d]+$/;
+        if (!patt.test(value)) {
```
martinpitt, line 387: Either define `patt` at the top level as `RE_TAG` or so and recycle it  (preferable IMHO), or write it shorthand as

```js
if (!patt.matches(/:[\w-\d]+$/))
```

src/ImageRunModal.jsx
```diff
@@ -386,27 +380,33 @@ export class ImageRunModal extends React.Component {
...
                     path: client.VERSION + "libpod/images/search",
                     body: "",
                     params: {
-                        term: registry + "/" + value
+                        term: value,
                     }
-                })));
+                }));
+            } else {
+                searches = searches.concat(utils.fallbackRegistries.map(registry =>
```
martinpitt, line 400: Please use `searches.push` consistently, like above. Then you can also make it const.

test/check-application
```diff
@@ -3086,6 +3087,19 @@ class TestApplication(testlib.MachineCase):
         b.wait_text("button.pf-v5-c-select__menu-item.pf-m-disabled", "No images found")
         b.wait_in_text(".pf-v5-c-alert.pf-m-danger", "couldn't search registry")
 
+        # Searching for full url with tag finds the image
+        b.set_input_text("#run-image-dialog-name", "tagged-busybox")
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:80/my-busybox:my-tag")
+        b.click('button.pf-v5-c-toggle-group__button:contains("All")')
+        b.wait_text("button.pf-v5-c-select__menu-item:not(.pf-m-disabled)", container_name + ":my-tag")
```
martinpitt, line 3094: Could this also ensure that this is the only displayed image, i.e. count the number of result rows? To make sure it's not duplicated.

test/check-application
```diff
@@ -3086,6 +3087,19 @@ class TestApplication(testlib.MachineCase):
...
+        b.click(f"button.pf-v5-c-select__menu-item:contains({container_name + ":my-tag"})")
+        b.click("#create-image-create-btn")
+
```
martinpitt, line 3097: Do we already have a test somewhere which checks the behaviour on a registry *with* search? I.e. the localhost:3000 one (or whatever the port was)? There is a lot of if/else for this, and the missing coverage also hints that there isn't -- can you please add one?

martinpitt:
I believe [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1821-93fb631f-20240828-133741-rhel-10-0/log.html#6) is our well-known flake #1836, so ignore here.

The [TF failures](https://artifacts.dev.testing-farm.io/af20b1e9-0f94-4381-8720-538f171ce350/) are due to

    Failed to start nginx.service: Unit nginx.service not found.

You can fix that by adding `nginx` to the dependency list in test/browser/main.fmf

test/check-application
tomasmatus: The missing coverage is for a fallback when there are no configured search registries. And since we always set these it doesn't go to the `else` branch. However I added a check that the `manifest` query works on searchable registry.

src/ImageRunModal.jsx
```diff
@@ -342,68 +344,118 @@ export class ImageRunModal extends React.Component {
...
+            body: "",
+        }));
+
+        // Don't start search queries when tag is specified as search API doesn't support
+        // searching for specific image tags
+        // instead only rely on manifests query (requires full image url)
+        if (!RE_CONTAINER_TAG.test(value)) {
+            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // podman search in the user specified registry.
+            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
```
cockpituous, line 391: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -342,68 +344,118 @@ export class ImageRunModal extends React.Component {
...
-                })));
+                }));
+            } else {
+                searches.push(...utils.fallbackRegistries.map(registry =>
+                    this.activeConnection.call({
+                        method: "GET",
+                        path: client.VERSION + "libpod/images/search",
+                        body: "",
+                        params: {
+                            term: registry + "/" + value
```
cockpituous, line 407: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 391: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 407: These 7 added lines are not executed by any test.

garrett:
It's getting there!

I found two interaction problems so far:

1. It doesn't strip spaces at the end (or start), so if I copy a URI from somewhere and it has a space around it, it fails.
2. Can't use the keyboard: After pasting the URI, hitting enter fails; it empties the line. I would expect it to accept the URL, so I could continue configuring it.

I was testing with `ghcr.io/cockpit-project/tasks-tmp:10449236342.1-arm64 ` (which had a space when I copied it), which I found at https://github.com/cockpit-project/cockpituous/pkgs/container/tasks-tmp (as it's on ghcr and isn't our standard tasks container, which I'm using as my dev environment in toolbox already).

garrett (changes requested):
(fix the issues with the spaces and hitting enter, as pointed out above; thanks!)

test/check-application
```diff
@@ -3005,6 +3023,91 @@ class TestApplication(testlib.MachineCase):
...
+        """)
+
+        self.login()
+
+        b.click("#containers-containers button.pf-v5-c-button.pf-m-primary")
+        b.set_input_text("#run-image-dialog-name", "new-busybox")
+        # Searching for container by prefix fails
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:80/my-busy")
+        b.wait_text("button.pf-v5-c-select__menu-item.pf-m-disabled", "No images found")
+        b.wait_in_text(".pf-v5-c-alert.pf-m-danger", "couldn't search registry")
```
martinpitt, line 3063: This shouldn't be "danger", but "info" or "default". (It needs to be changed in the code, but it's visible here)

src/ImageRunModal.jsx
```diff
@@ -342,68 +344,118 @@ export class ImageRunModal extends React.Component {
...
+        // Try to get specified image manifest
+        searches.push(this.activeConnection.call({
+            method: "GET",
+            path: client.VERSION + "libpod/manifests/" + value + "/json",
+            body: "",
+        }));
+
+        // Don't start search queries when tag is specified as search API doesn't support
+        // searching for specific image tags
+        // instead only rely on manifests query (requires full image url)
```
martinpitt, line 387: It's really not an image "URL", it's an image:tag name

src/ImageRunModal.jsx
```diff
@@ -342,68 +344,118 @@ export class ImageRunModal extends React.Component {
...
                             if (result.status === "fulfilled") {
+                                // manifest query is always first
+                                if (i === 0) {
+                                    manifestResult = JSON.parse(result.value);
+                                    return;
+                                }
+
                                 imageResults = imageResults.concat(JSON.parse(result.value));
                             } else {
+                                // manifests query returns 404 Not Found in case used URL is not an image
```
martinpitt, line 431: s/URL/value/

test/check-application
```diff
@@ -3005,6 +3023,91 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_text("button.pf-v5-c-select__menu-item:not(.pf-m-disabled)", f"{container_name}:latest")
+        b.wait_not_present(".pf-v5-c-alert.pf-m-danger")
+        b.click('button.pf-v5-c-toggle-group__button:contains("Local")')
+
+        # Error is shown again when no image was found locally
+        b.set_input_text("#create-image-image-select-typeahead", "localhost:80/their-busy")
+        b.wait_text("button.pf-v5-c-select__menu-item.pf-m-disabled", "No images found")
+        b.wait_in_text(".pf-v5-c-alert.pf-m-danger", "couldn't search registry")
+
+        # Searching for full url with tag finds the image
```
martinpitt, line 3091: s/url with //

src/ImageRunModal.jsx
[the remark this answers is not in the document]
martinpitt: Included in pending review for easier tracking.

src/ImageRunModal.jsx
```diff
@@ -342,68 +344,118 @@ export class ImageRunModal extends React.Component {
...
+                                }
                                 dialogError = _("Failed to search for new images");
                                 // TODO: add registry context, podman does not include it in the reply.
                                 dialogErrorDetail = result.reason ? cockpit.format(_("Failed to search for images: $0"), result.reason.message) : _("Failed to search for images.");
                             }
+                        });
+
+                        // Skip manifests query result when search query found the same image
+                        if (manifestResult && !imageResults.find(image => image.Name === value)) {
+                            manifestResult.Name = value;
```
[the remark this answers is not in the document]
tomasmatus: It doesn't https://docs.podman.io/en/latest/_static/api.html?version=v5.0#tag/manifests/operation/ManifestInspectLibpod

martinpitt:
@tomasmatus thanks! I cleaned up outdated review threads and comments to make the remaining stuff easier to see. There's a few small things to fix; if you really don't want to rewrite/simplify the search result loop (which is so hard to read and so easy to get wrong!), perhaps @jelly can help with that, or also me when I'm back from PTO. I also wouldn't veto landing it as-is and cleaning up later, if @jelly is ok with it. But in three weeks nobody of us will be able to follow that logic..

src/ImageRunModal.jsx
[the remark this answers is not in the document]
martinpitt: ok!

tomasmatus:
> I also wouldn't veto landing it as-is and cleaning up later.

yes, I agree with this, let's not

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,116 @@ export class ImageRunModal extends React.Component {
...
+    }
+
     onSearchTriggered = value => {
+        let dialogError = "";
+        let dialogErrorDetail = "";
+
+        const changeDialogError = (reason) => {
+            dialogError = _("Failed to search for new images");
+            // TODO: add registry context, podman does not include it in the reply.
+            dialogErrorDetail = reason ? cockpit.format(_("Failed to search for images: $0"), reason.message) : _("Failed to search for images.");
```
cockpituous, line 376: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,116 @@ export class ImageRunModal extends React.Component {
...
+            body: "",
+        }));
+
+        // Don't start search queries when tag is specified as search API doesn't support
+        // searching for specific image tags
+        // instead only rely on manifests query (requires image:tag name)
+        if (!RE_CONTAINER_TAG.test(value)) {
+            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // podman search in the user specified registry.
+            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
```
cockpituous, line 419: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,116 @@ export class ImageRunModal extends React.Component {
...
-                })));
+                }));
+            } else {
+                searches.push(...utils.fallbackRegistries.map(registry =>
+                    this.activeConnection.call({
+                        method: "GET",
+                        path: client.VERSION + "libpod/images/search",
+                        body: "",
+                        params: {
+                            term: registry + "/" + value
```
cockpituous, line 435: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,116 @@ export class ImageRunModal extends React.Component {
...
-                        for (const result of reply) {
+                        for (let i = 1; i < reply.length; i++) {
+                            const result = reply[i];
                             if (result.status === "fulfilled") {
                                 imageResults = imageResults.concat(JSON.parse(result.value));
-                            } else {
-                                dialogError = _("Failed to search for new images");
-                                // TODO: add registry context, podman does not include it in the reply.
-                                dialogErrorDetail = result.reason ? cockpit.format(_("Failed to search for images: $0"), result.reason.message) : _("Failed to search for images.");
+                            } else if (!manifestResult && !imageExistsLocally(value, this.props.localImages)) {
```
cockpituous, line 451: This added line is not executed by any test.

tomasmatus:
@martinpitt if you'd like to have a look it is this [Select](https://github.com/cockpit-project/cockpit-podman/pull/1821/files#diff-73dca0639248224a73bef57d4a17f8731b58e805ef10da2453fb0a90a87a84a9R888) that is misbehaving.

What happens is when you type something in and press enter the input bar is cleared rather than preserve the value. It is also cleared once the input loses focus. This shouldn't be happening as the `isInputValuePersisted` prop is used. I haven't figured out if this is a bug in the now deprecated PF4 Select or if it is somehow cleared in our code.

martinpitt:
I can reproduce the issue on main. AFAICS the only sensible thing is to ignore the Enter key completely then -- if you type a random string into the image bar, it cannot be "accepted" as a valid value (which is usually the meaning of "Enter"). You must actually select a valid image for the dialog to make sense. Clearing an invalid value on focus change also makes sense IMHO.

However, I do see how this would collide with the workflow of this PR -- it'd still force you to actually click on the search result instead of just copy&pasting an image name and doing nothing else. Did I get this right?

src/ImageRunModal.jsx
cockpituous, line 376: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 419: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 435: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 451: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 376: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 419: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 435: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 451: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,116 @@ export class ImageRunModal extends React.Component {
...
+        return new RegExp(regexString, 'i');
+    }
+
     onSearchTriggered = value => {
+        let dialogError = "";
+        let dialogErrorDetail = "";
+
+        const changeDialogError = (reason) => {
+            dialogError = _("Failed to search for new images");
+            // TODO: add registry context, podman does not include it in the reply.
```
martinpitt, line 375: Follow-up, or is that easy to do and you just forgot about it?

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,116 @@ export class ImageRunModal extends React.Component {
...
+            }
         }
 
         Promise.allSettled(searches)
                 .then(reply => {
                     if (reply && this._isMounted) {
                         let imageResults = [];
-                        let dialogError = "";
-                        let dialogErrorDetail = "";
+                        const manifestResult = handleManifestsQuery(reply[0]);
```
martinpitt, line 445: So if this fails, but the other search replies succeed, this quickly sets an error in the dialog, but the code below will then reset/update the error, right? I.e. that may be intended (even though it's a bit ugly). The error handling strategy should at least be commented here, or better made explicit.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,116 @@ export class ImageRunModal extends React.Component {
...
+                        for (let i = 1; i < reply.length; i++) {
+                            const result = reply[i];
                             if (result.status === "fulfilled") {
                                 imageResults = imageResults.concat(JSON.parse(result.value));
-                            } else {
-                                dialogError = _("Failed to search for new images");
-                                // TODO: add registry context, podman does not include it in the reply.
-                                dialogErrorDetail = result.reason ? cockpit.format(_("Failed to search for images: $0"), result.reason.message) : _("Failed to search for images.");
+                            } else if (!manifestResult && !imageExistsLocally(value, this.props.localImages)) {
+                                changeDialogError(result.reason);
```
martinpitt, line 452: So this overwrites the error set by `handleManifestQuery()`, which seems fine -- but it also overwrites earlier error from reply[i-1]. It probably doesn't matter that much if we show the error from the first or last registry, it just feels a bit random.

martinpitt (approved):
Cheers @tomasmatus ! The [previous review](https://github.com/cockpit-project/cockpit-podman/pull/1821#pullrequestreview-2276732644) still has three current threads, and I have some more questions. This is coming together!

src/ImageRunModal.jsx
tomasmatus: I'm not sure if manifests errors should be reset by successful search queries? Since it looks for a specific image in a specific location and search queries are a bit more lax. Error 404 is handled separately and other errors from the 500s are unexpected so I believe that should be visible in this case.

src/ImageRunModal.jsx
tomasmatus: yes, it's going to show the last error encountered. It's the same behavior as before

src/ImageRunModal.jsx
tomasmatus: ah, that's a moved code and that todo has probably been there for a while 🤔

martinpitt: I see, nevermind then.

src/ImageRunModal.jsx
martinpitt: I re-read this again: changeDialogError() is only called by the search results if `!manifestResult`. So indeed if the manifest query fails, then `manifestResult === null`, and the error message *does* get overwritten by search failures. I agree that the manifest query error is worth showing, but that's not what the code is doing?

tomasmatus: So I suppose the way to go is to show the first encountered error

test/check-application
tomasmatus: I can't think of a way how to set it to "info" for this specific error message. I've learnt from the previous reviews that `msg.includes("couldn't search registry")` is a no go 😅

src/ImageRunModal.jsx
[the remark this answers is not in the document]
tomasmatus: (moved code) I think this wasn't tested before since `fallbackRegistries` is hardcoded as 
`export const fallbackRegistries = ["docker.io", "quay.io"];`

src/ImageRunModal.jsx
martinpitt: Yeah, at this point all promises are settled, so you can consider all failures and know where they come from.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,119 @@ export class ImageRunModal extends React.Component {
...
     onSearchTriggered = value => {
+        let dialogError = "";
+        let dialogErrorDetail = "";
+
+        const changeDialogError = (reason) => {
+            // Only report first encountered error
+            if (dialogError === "" && dialogErrorDetail === "") {
+                dialogError = _("Failed to search for new images");
+                // TODO: add registry context, podman does not include it in the reply.
+                dialogErrorDetail = reason ? cockpit.format(_("Failed to search for images: $0"), reason.message) : _("Failed to search for images.");
```
cockpituous, line 378: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,119 @@ export class ImageRunModal extends React.Component {
...
+            body: "",
+        }));
+
+        // Don't start search queries when tag is specified as search API doesn't support
+        // searching for specific image tags
+        // instead only rely on manifests query (requires image:tag name)
+        if (!RE_CONTAINER_TAG.test(value)) {
+            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // podman search in the user specified registry.
+            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
```
cockpituous, line 422: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,119 @@ export class ImageRunModal extends React.Component {
...
-                })));
+                }));
+            } else {
+                searches.push(...utils.fallbackRegistries.map(registry =>
+                    this.activeConnection.call({
+                        method: "GET",
+                        path: client.VERSION + "libpod/images/search",
+                        body: "",
+                        params: {
+                            term: registry + "/" + value
```
cockpituous, line 438: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -351,68 +353,119 @@ export class ImageRunModal extends React.Component {
...
-                        for (const result of reply) {
+                        for (let i = 1; i < reply.length; i++) {
+                            const result = reply[i];
                             if (result.status === "fulfilled") {
                                 imageResults = imageResults.concat(JSON.parse(result.value));
-                            } else {
-                                dialogError = _("Failed to search for new images");
-                                // TODO: add registry context, podman does not include it in the reply.
-                                dialogErrorDetail = result.reason ? cockpit.format(_("Failed to search for images: $0"), result.reason.message) : _("Failed to search for images.");
+                            } else if (!manifestResult && !imageExistsLocally(value, this.props.localImages)) {
```
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
martinpitt: Thanks, I'll stop picking on the error handling -- this is good enough now. Thanks!

src/ImageRunModal.jsx
cockpituous, line 378: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 422: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 438: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 378: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 422: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 438: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 378: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 422: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 438: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 378: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 422: This added line is not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 438: These 7 added lines are not executed by any test.

src/ImageRunModal.jsx
cockpituous, line 454: This added line is not executed by any test.

martinpitt (approved):
Wohoo! Thanks!

martinpitt:
@tomasmatus Can you please add an initial release note to the description, and an up to date screenshot? (Please use the inspector to cut out the dialog precisely, and add a `border: 2px solid`)

PR #1827, opened 2024-08-19
Allow running image with privileged option

xuhuashan:
Some applications running in containers require additional permissions to run, such as accessing the host's hardware devices. The most convenient way is to run in privileged mode.
This mode is disabled by default. Users can choose to enable it by checking the Privileged option when running an image.

image

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -88,6 +88,7 @@ export class ImageRunModal extends React.Component {
             entrypoint,
             env: [],
             hasTTY: true,
+            privileged: false,
             publish: [],
             image: props.image,
             memory: 512,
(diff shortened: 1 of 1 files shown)


jelly (commented):
I was wondering if privileged was supported for a rootless container and it seems to work.

When I create a privileged container now there is no indication in the UI that it is privileged.

Some tests need to be added to ensure this keeps working in the future, I am happy to assist here.

jelly:
Checked podman desktop for inspiration, this doesn't show if a container is privileged or not.

xuhuashan:
> I was wondering if privileged was supported for a rootless container and it seems to work.

In most cases, rootless containers cannot access files in the /dev directory, even when running privileged. Some device files can be accessed via supplementary group memberships (e.g. the video group). This requires adding the "keep-groups" flag to the "groups" parameter, and possibly mapping the groups with idmappings. But this flag seems to be supported only when running with "crun".

However, the privileged option adds all capabilities to the container, which makes commands like ping work right away. It also disables Seccomp, SELinux, and Apparmor confinement.

Full support for rootless container access to devices seems to complicate things, or privileged should be disabled for rootless containers first.

> 
> When I create a privileged container now there is no indication in the UI that it is privileged.

I really hadn't considered that. it should be indicated in the right position.

> 
> Some tests need to be added to ensure this keeps working in the future, I am happy to assist here.

PR #1831, opened 2024-08-19
fix disabled button check

tomasmatus:
Test checks if a warning is displayed therefore create button should be disabled.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-1821-bfa093ee-20240819-082934-debian-testing/log.html#10-2

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -2433,10 +2433,10 @@ class TestApplication(testlib.MachineCase):
 
     def testCreateContainerValidation(self):
         def validateField(groupSelector, value, errorMessage, resetValue=""):
-            b.set_input_text(f"{groupSelector} input", value)
             b.wait_visible(".pf-v5-c-modal-box__footer #create-image-create-run-btn:not(:disabled)")
+            b.set_input_text(f"{groupSelector} input", value)
             b.wait_in_text(f"{groupSelector} .pf-v5-c-helper-text__item-text", errorMessage)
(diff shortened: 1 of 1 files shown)


tomasmatus:
I am honestly confused... how does tagging work?

https://cockpit-logs.us-east-1.linodeobjects.com/pull-1831-5b29b738-20240819-122636-debian-stable/log.html#13

test/check-application
```diff
@@ -1174,7 +1174,7 @@ class TestApplication(testlib.MachineCase):
         second_tag = "localhost/copybox:latest"
         self.execute(False, f"podman tag localhost:5000/my-busybox {second_tag}")
         # expand details
-        b.click("#containers-images tr:contains('my-busybox') td.pf-v5-c-table__toggle button")
+        b.click(f"#containers-images tr:contains('{second_tag}') td.pf-v5-c-table__toggle button")
```
martinpitt, line 1177: This isn't a "fix" in the sense that the old behaviour was wrong. The original code checks that adding a second tag to an existing image correctly gets picked up by the UI and shown as "additional tag" names in the *original* image info.

You are welcome to *add* code which checks that the new tag also appears as an image by itself (and then also shows `my-busybox` as alternative tag name), but let's please keep that check.

tomasmatus: I see, my bad. I'll revert this then

martinpitt (approved):
Cheers!

PR #1834, opened 2024-08-21
Some test fixes

martinpitt:
Split out from #1832.

I'll trigger an extra /firefox run.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1732,10 +1732,9 @@ class TestApplication(testlib.MachineCase):
 
         # Local results found
         b.set_input_text("#create-image-image-select-typeahead", "registry")
+        b.wait_text("button.pf-v5-c-select__menu-item", IMG_REGISTRY_LATEST)
         if auth:
             b.assert_pixels(".pf-v5-c-modal-box", "image-select", skip_layouts=["rtl"])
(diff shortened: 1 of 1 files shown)


martinpitt:
Ah, [this run](https://artifacts.dev.testing-farm.io/3a44f391-ea98-46b4-b63f-9bb8d565c2ea/) found a similar bug as the testCreateContainerInPodUser() failure, and locally I then found yet another one. Also, the "create button eternally disabled in pod dialog" fails a *lot* here. @jelly is investigating that right now, for this PR I'll just lean on the retry button for these.

martinpitt:
Hm, we've had [this failure](https://artifacts.dev.testing-farm.io/6fe9744b-14d2-491f-b7cb-914cec22d48b/) twice now:
> Error: copying system image from manifest list: writing blob: adding layer with blob "sha256:d2ecaa8d83f5e0ee1a77f22f3ade6f096f71b58290b47ce0fb4e892aee694d44"/""/"sha256:3d0f0bc504b3bf6b24a765edb1764270f1c6d94b1760aa030698316e3441ec42": unpacking failed (error: exit status 1; output: write /usr/lib64/libatspi.so.0.0.1: no space left on device)

But well, all the other TF tests passed, so good enough.

Issue #1836, opened 2024-08-22
Pod creation dialog state handling is broken


martinpitt:
So I conclude that we are doing all the state interaction wrong with validation..

The PodCreateModal has a state `validationFailed` and `setValidationFailed` setter.

`validationFailed` is pushed to the DynamicListForm as a property:

```
            <DynamicListForm id='create-pod-dialog-publish'
                        emptyStateString={_("No ports exposed")}
                        formclass='publish-port-form'
                        label={_("Port mapping")}
                        actionLabel={_("Add port mapping")}
                        validationFailed={validationFailed.publish}
                        onValidationChange={value => dynamicListOnValidationChange('publish', value)}
                        onChange={value => setPublish(value)}
                        default={{ IP: null, containerPort: null, hostPort: null, protocol: 'tcp' }}
                        itemcomponent={ <PublishPort />} />
```

This component has `dynamicListOnValidationChange` which is this function that modifies the state:

```
    const dynamicListOnValidationChange = (key, value) => {
        console.log("validationchange", key, value);
        setValidationFailed(prevState => {
            prevState[key] = value;
            if (prevState[key].every(a => a === undefined))
                delete prevState[key];
            return prevState;
        });
    };
```

The value passed comes from PublishPort:

```
                <TextInput id={id + "-container-port"}
                            type='number'
                            step={1}
                            min={1}
                            max={MAX_PORT}
                            validated={validationFailed?.containerPort ? "error" : "default"}
                            value={item.containerPort || ''}
                            onChange={(_event, value) => {
                                console.log("onChange", validationFailed, _event, value);
                                utils.validationClear(validationFailed, "containerPort", onValidationChange);
                                console.log({ ...validationFailed, containerPort: validatePublishPort(value, "containerPort") });
                                utils.validationDebounce(() => onValidationChange({ ...validationFailed, containerPort: validatePublishPort(value, "containerPort") }));
                                onChange(idx, 'containerPort', value);
                            }} />
```

So this modifies React state and then passes it to onValidationChange which is also a same named function in DynamicListForm:

```
                                return React.cloneElement(this.props.itemcomponent, {
                                    idx,
                                    item,
                                    id: id + "-" + idx,
                                    key: idx,
                                    onChange: this.onItemChange,
                                    removeitem: this.removeItem,
                                    additem: this.addItem,
                                    options: this.props.options,
                                    validationFailed: validationFailed && validationFailed[idx],
                                    onValidationChange: value => {
[...]

PR #1843, opened 2024-09-05
Test how bad our flakes are when retrying

jelly:


jelly:
So both issues are related to https://github.com/cockpit-project/cockpit-podman/issues/1836

TestApplication.testCreateContainerSystem is something different.

PR #1850, opened 2024-09-10
test: workaround missing iptables-nft required for criu

jelly:
Without iptables-nft `podman container checkpoint --tcp-established` fails as `iptables-restore` is missing.


mvollmer (approved):
Thanks!

PR #1861, opened 2024-09-26
Fix irritating Enter key behaviour in container creation dialog

martinpitt:
Pressing "Enter" in the dialog on any active form element (such as the search input or the checkboxes) previously bubbled up to the Form, which then activated the first popover help. This is unexpected and very irritating especially when using the search input.

Just ignore the Enter key to fix that, similar to what the image search modal already does.

See https://github.com/cockpit-project/cockpit-podman/pull/1821#issuecomment-2374302608

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -743,8 +743,10 @@ export class ImageRunModal extends React.Component {
             </ToggleGroup>
         );
 
+        /* ignore Enter key, it otherwise opens the first popover help; this clears
+         * the search input and is still irritating from other elements like check boxes */
         const defaultBody = (
-            <Form>
+            <Form onKeyDown={e => e.key === 'Enter' && e.preventDefault()}>
                 {this.state.dialogError && <ErrorNotification errorMessage={this.state.dialogError} errorDetail={this.state.dialogErrorDetail} />}
                 <FormGroup id="image-name-group" fieldId='run-image-dialog-name' label={_("Name")} className="ct-m-horizontal">
                     <TextInput id='run-image-dialog-name'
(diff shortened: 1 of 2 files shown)


src/ImageRunModal.jsx
```diff
@@ -744,7 +744,7 @@ export class ImageRunModal extends React.Component {
         );
 
         const defaultBody = (
-            <Form>
+            <Form onKeyDown={e => e.key !== 'Enter' || e.preventDefault()}>
```
garrett, line 747: This was the kind of fix I was hoping would be possible when we talked about it on the call! Nice!

I'm curious why you did this as a negative check with `!` `||` instead of a positive check like this:

```
<Form onKeyDown={e => e.key === 'Enter' && e.preventDefault()}>
```

martinpitt: Ah, I suppose old shell script writing habit (where `&&` has a strange semantics). But indeed it looks better with `&&`. I wanted to keep it as an expression (not a full statement) to keep the Arrow simple.

martinpitt: There, changed to `&&`. Thanks for spotting!

garrett: Yeah, it felt like a "double negative" to me, which is usually harder to read (both in written language and in code). It's true that shell scripting definitely has its issues. (I'm glad things like [shellcheck](https://www.shellcheck.net/) exist!)

src/ImageRunModal.jsx
```diff
@@ -744,7 +744,7 @@ export class ImageRunModal extends React.Component {
         );
 
         const defaultBody = (
-            <Form>
+            <Form onKeyDown={e => e.key === 'Enter' && e.preventDefault()}>
```
mvollmer, line 747: I think this needs a comment.

mvollmer (dismissed):

PR #1865, opened 2024-10-02
Validate environment variables

ashley-cui:
Validate env var keys with regex. Remove Required asterisk for value. This might sound weird but Podman and shells in general allow setting env vars without values.

Fixes: https://github.com/cockpit-project/cockpit-podman/issues/1622

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -2535,7 +2535,8 @@ class TestApplication(testlib.MachineCase):
 
         # Test validation of environment variables
         b.click('.env-form .btn-add')
-        b.set_input_text("#run-image-dialog-env-0-key-group input", "sometext")
+        validateField("#run-image-dialog-env-0-key-group", "[]]]", "invalid characters")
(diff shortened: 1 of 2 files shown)


test/check-application
```diff
@@ -2517,7 +2517,7 @@ class TestApplication(testlib.MachineCase):
 
         # Test validation of environment variables
         b.click('.env-form .btn-add')
-        b.set_input_text("#run-image-dialog-env-0-key-group input", "sometext")
+        validateField("#run-image-dialog-env-0-key-group", "[]]]", "invalid characters")
```
jelly, line 2520: Maybe a test for `1A` is also worth it?

jelly (commented):
Looks good, this is likely to cause a lot of retries to get green due to https://github.com/cockpit-project/cockpit-podman/issues/1836

jelly (approved):

martinpitt:
Whoops, this slipped through the cracks.. This needs a pixel update due to the dropped "isRequired".

martinpitt (approved):

PR #1879, opened 2024-10-14
tmt: Disable fedora repo for C10 tests

martinpitt:
https://issues.redhat.com/browse/TFT-2564 roars its ugly head again and completely breaks installation due to installing Fedora packages on CentOS 10.

See https://github.com/containers/podman/pull/24255

Same solution as https://github.com/cockpit-project/cockpit-machines/commit/41f337fd85b85

### DIFF
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -3,6 +3,14 @@ discover:
 execute:
     how: tmt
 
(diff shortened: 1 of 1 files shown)


martinpitt:
It's not sufficient, but it helps a little bit and doesn't  hurt, so let's land?

jelly (approved):

PR #1884, opened 2024-10-17
pods: Fix PodCreateModal validation issues

mvollmer:
I have found four issues:

 - Validation results would not always make it into the UI while typing into the fields. This was confusing, but hitting "Create" would always straighten things out.
 - Removing port rows and then hitting "Create" while something was invalid would shift the (genuine) validation errors into the places of the previous removed ports. There they would invisbly keep the "Create" button from ever getting enabled again.
 - `<TextInput type="number" />` does not usefully distinguish between empty strings and strings that are not numbers, and it sends fewer change notifications than expected. For example, starting to type random letters into a previously empty field will produce no change events and we get no opportunity to run our validation code.
 - The Volumes are not validated at the top-level, but do have their own validation rules. (Just "container path can not be empty".) Instead of showing validation errors, invalid volumes are simply omitted when creating the pod.

### DIFF
--- a/src/PodCreateModal.jsx
+++ b/src/PodCreateModal.jsx
@@ -77,10 +77,10 @@ export const PodCreateModal = ({ user, systemServiceAvailable, userServiceAvaila
     */
     const dynamicListOnValidationChange = (key, value) => {
         setValidationFailed(prevState => {
-            prevState[key] = value;
-            if (prevState[key].every(a => a === undefined))
-                delete prevState[key];
-            return prevState;
+            const newState = Object.assign({}, prevState, { [key]: value });
+            if (newState[key].every(a => a === undefined))
+                delete newState[key];
+            return newState;
         });
     };
 
--- a/src/PodCreateModal.jsx
+++ b/src/PodCreateModal.jsx
@@ -129,7 +129,7 @@ export const PodCreateModal = ({ user, systemServiceAvailable, userServiceAvaila
             };
         });
         if (publishValidation.some(entry => entry && Object.keys(entry).length > 0))
-            newValidationFailed.publish = publishValidation.filter(entry => entry !== undefined);
+            newValidationFailed.publish = publishValidation;
 
         const podNameValidation = validatePodName(podName);
 
--- a/src/PodCreateModal.jsx
+++ b/src/PodCreateModal.jsx
@@ -101,12 +101,20 @@ export const PodCreateModal = ({ user, systemServiceAvailable, userServiceAvaila
     };
 
     const isFormInvalid = validationFailed => {
-        const groupHasError = row => row && Object.values(row)
-                .filter(val => val) // Filter out empty/undefined properties
-                .length > 0; // If one field has error, the whole group (dynamicList) is invalid
+        function publishGroupHasError(row, idx) {
+            // We always ignore errors for empty slots in
+            // publish. Errors for these slots might show up when the
+            // debounced validation runs after a row has been removed.
+            if (!row || !publish[idx])
+                return false;
+
+            return Object.values(row)
+                    .filter(val => val) // Filter out empty/undefined properties
+                    .length > 0; // If one field has error, the whole group (dynamicList) is invalid
+        }
 
         // If at least one group is invalid, then the whole form is invalid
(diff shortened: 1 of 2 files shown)


src/PodCreateModal.jsx
```diff
@@ -77,10 +77,10 @@ export const PodCreateModal = ({ user, systemServiceAvailable, userServiceAvaila
     */
     const dynamicListOnValidationChange = (key, value) => {
         setValidationFailed(prevState => {
-            prevState[key] = value;
-            if (prevState[key].every(a => a === undefined))
-                delete prevState[key];
-            return prevState;
+            const newState = Object.assign({}, prevState, { [key]: value });
+            if (newState[key].every(a => a === undefined))
```
cockpituous, line 81: This added line is not executed by any test.

cockpituous, line 81: This added line is not executed by any test.

mvollmer:
Here is a step-by-step play-through of the bug re keeping publish and validationFailed aligned. The value shown come from JSON.stringify, so "undefined" and "empty slot" show as "null".

```
 - "Create pod"

   publish: []
   valididationFailed: {}

 - "Add port mapping"

    publish: [{"key":0,"IP":null,"containerPort":null,"hostPort":null,"protocol":"tcp"}]
    valididationFailed: {}

 - Trashcan

   publish: [null]
   valididationFailed: {}

 - "Add port mapping"

   publish: [null,{"key":1,"IP":null,"containerPort":null,"hostPort":null,"protocol":"tcp"}]
   valididationFailed: {}

 - Type "-3" into "Container port"

   publish: [null,{"key":1,"IP":null,"containerPort":"-3","hostPort":null,"protocol":"tcp"}]
   valididationFailed: {"publish":[null,{"containerPort":"1 to 65535"}]}

   This is correct, there is a "null" entry in validationFailed to
   correspond to the "null" entry in publish. The error shows up in
   the UI in the correct spot.

 - Trashcan

   publish: [null,null]
   valididationFailed: {}

 - "Add port mapping"

   publish: [null,null,{"key":2,"IP":null,"containerPort":null,"hostPort":null,"protocol":"tcp"}]
   valididationFailed: {}

 - "Create"

    publish: [null,null,{"key":2,"IP":null,"containerPort":null,"hostPort":null,"protocol":"tcp"}]
    valididationFailed: {"publish":[{"containerPort":"Container port must not be empty"}]}

    ^- This is the bug: validateForm() has removed the undefined
    entries from validationFailed and shifted the validation error
    into index 0. The error doesn not appear in the UI, and from now
    on, isFormInvalid will always say "true" and the "Create" button
    will stay disabled.
```

mvollmer:
Since the tests don't actually remove ports in the dialog, I think all the flakiness was caused by the bug in `dynamicListOnValidationchange`.

mvollmer:
> Since the tests don't actually remove ports in the dialog, I think all the flakiness was caused by the bug in `dynamicListOnValidationChange`.

Oh, they do, sorry.

src/PodCreateModal.jsx
cockpituous, line 81: This added line is not executed by any test.

cockpituous, line 81: This added line is not executed by any test.

mvollmer:
I guess I look at the TestApplication.testCreateContainerInPodUser flakes next.

src/PodCreateModal.jsx
```diff
@@ -101,12 +101,20 @@ export const PodCreateModal = ({ user, systemServiceAvailable, userServiceAvaila
...
+            // We always ignore errors for empty slots in
+            // publish. Errors for these slots might show up when the
+            // debounced validation runs after a row has been removed.
+            if (!row || !publish[idx])
+                return false;
+
+            return Object.values(row)
+                    .filter(val => val) // Filter out empty/undefined properties
+                    .length > 0; // If one field has error, the whole group (dynamicList) is invalid
+        }
```
jelly, line 114: This might also need to be fixed in CreateContainerModal.

jelly (approved):

martinpitt (approved):
Awesome, thanks! Even if there are still other races left, these are all good and important fixes, so let's get them in.

PR #1885, opened 2024-10-18
test: Wait for image search to be over in CreateContainer tests

mvollmer:


test/check-application
```diff
@@ -2795,13 +2795,15 @@ class TestApplication(testlib.MachineCase):
...
+        b.wait_visible(f'button.pf-v5-c-select__menu-item:contains("{IMG_ALPINE_LATEST}")')
+        b.wait_visible(f'button.pf-v5-c-select__menu-item:contains("{IMG_BUSYBOX_LATEST}")')
+        # Filter it down to just Busybox
         b.set_input_text("#create-image-image-select-typeahead", IMG_BUSYBOX_LATEST)
-        # we want to switch to "Local", but to make waiting for IMG_BUSYBOX race-free, we first need to
-        # switch to another view; different images have different registries, so don't assume their name, just
-        # that at least one more exists
-        b.click('.image-search-footer .pf-v5-c-toggle-group__item:nth-of-type(3) button')
-        b.wait_in_text(".pf-v5-c-select__menu-list", "No images found")
-        b.click('button.pf-v5-c-toggle-group__button:contains("Local")')
```
martinpitt, line 2804: I don't remember much of the history here, but switching to the "Local" filter  is also useful coverage -- that seems to have gone now?

Could you do both, first wait for the "All registries" view to update, and then switch to Local and ensure that this works as well?

Unless of course we cover this somewhere else.

mvollmer: > Could you do both, first wait for the "All registries" view to update, and then switch to Local and ensure that this works as well?

Yes, let's do this.

jelly: > I don't remember much of the history here, but switching to the "Local" filter is also useful coverage -- that seems to have gone now?

I remember you made these changes when we introduced BiDi, but could only find https://github.com/cockpit-project/cockpit-podman/pull/1834/files#diff-e86755442bc7d21f0414014b3c356b54d6650e8ed9bffcc4f5858d1154df80e1R1801

mvollmer: Ah, there are two instances of this pattern. I'll change the second one as well.

mvollmer:
Looks pretty good now. TestApplication.testHealthcheckUser on ubuntu-2204 flakes, but that's for another day/PR, I'd say.

martinpitt (approved):
Yay, danke! And yes, the ubuntu healthcheck flake has been around forever, and is unrelated.

PR #1887, opened 2024-10-21
test: Wait a bit on ubuntu-2204 before running the second health check

mvollmer:


mvollmer:
The second health check is triggered, but doesn't seem to start. Systemd seems to be involved on ubuntu-2204 when running a health check, and maybe it throttles things...

martinpitt:
Oh dear, so many flakes -- but given the change here they must be unrelated. As this only affects ubuntu-2204, I'll look the other way :grin:

martinpitt (approved):

mvollmer:
> Oh dear, so many flakes -- but given the change here they must be unrelated. As this only affects ubuntu-2204, I'll look the other way 😁

Yeah, the amplified version of testHealthcheckUser was solidly green.

Issue #1336, opened 2023-07-05
CPU utilisation figure for Pod is not cumulative


poVoq:
The large CPU utilisation figure for a Pod seems to display what looks like the highest current use by one of the included containers and not the cumulative CPU utilisation like it does for the RAM use of the Pod.

As CPU use by individual containers is already displayed below, it makes the Pod figure quite redundant, but a cumulative figure would be nice to have if the Pod includes many containers.

jelly:
Showing the CPU / Memory usage was a bit tricky for pods so we use the following solution:

https://github.com/cockpit-project/cockpit-podman/blob/main/src/Containers.jsx#L488

So we just show the highest CPU usage of the shown containers as adding up CPU usage might show > 100% which might be weird/unexpected.

martinpitt:
Please don't use line refs on main (as these change all the time). This is stable: https://github.com/cockpit-project/cockpit-podman/blob/3c5c63ec1a11613e525b5ff6b3656b425002ec7e/src/Containers.jsx#L486

I agree to  this request --  the CPU usage should just be added up. It's *correct* to be > 100% on multi-CPU systems -- `top` shows "%CPUs" as > 100% all the time.

If that is a bother, podman could also normalize the number by dividing by `nproc`. But then this would need to happen for the per-container metrics as well. The latter *already* show > 100% numbers:

![image](https://github.com/cockpit-project/cockpit-podman/assets/200109/a6b16ca1-b2a7-4b79-bc92-70ea28f6c46b)

but that is an independent request.

PR #1889, opened 2024-10-21
Show cumulative CPU use in pod

tomasmatus:
fixes: #1336


martinpitt (approved):
Makes sense, thanks! And agreed that this is hard (i.e. too fragile) to test automatically

PR #1897, opened 2024-10-29
test: Remove cockpit/ws container on non-OSTree images

martinpitt:
In https://github.com/cockpit-project/bots/pull/7046 we want to pre-install the cockpit/ws container to the fedora-40 image (and in the future, whichever Fedora version the container is based on). This "breaks" pixel tests due to the additional container image, so remove it for anything except OSTree images.

We don't just want to include it in the pixel tests, as the container's base image and pixel reference OS move independently.


jelly (approved):

jelly:
Retrying the rawhide tests for science.

PR #1907, opened 2024-11-07
test: wait on new image tag

jelly:
testDownloadImage flaked a very often as the test succeeds when runs fast enough. Tagging an image to a new name in this case changes the image list name to `copybox` possibly because it is sorted before `my-busybox`.

---

Alternatively we can wait on either?

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -3,6 +3,7 @@
 # See https://github.com/cockpit-project/cockpit/blob/main/test/common/testlib.py
 # "class Browser" and "class MachineCase" for the available API.
 
+import json
 import re
 import sys
 import time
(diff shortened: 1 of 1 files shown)


jelly:
Ding ding ding, fails in some cases still so this needs to be conditional

martinpitt (changes requested):
Ah, you are saying the "contains my-busybox" isn't specific enough? But the parent image shouldn't be shown in the "tr", just in the info below. :thinking: This smells funny, and also fails on some OSes. Let's figure out Monday.

jelly:
> Ah, you are saying the "contains my-busybox" isn't specific enough? But the parent image shouldn't be shown in the "tr", just in the info below. 🤔 This smells funny, and also fails on some OSes. Let's figure out Monday.

The image has two tags, but which one is shown is not predictable which is the issue shown in the failing tests. Either the UI does not update or it is unpredicable.

jelly:
Cool this works now

martinpitt:
[this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1907-ee681156-20241111-145135-rhel-9-6/log.html#39) looks like flake, /me leans on retry button

martinpitt (approved):
Ah, makes sense -- UI has the same order as .RepoTags

jelly:
> Ah, makes sense -- UI has the same order as .RepoTags

Yup, apparently that is not deterministic :D

PR #1913, opened 2024-11-21
Convert substr() usage to substring()

jelly:
String.prototype.substr() is a deprecated API.

---

Couldn't find an eslint rule for this, so we'll have to pay attention.


martinpitt (approved):
Dankjewel!

PR #1918, opened 2024-11-29
test: Move reference-image to Fedora 41

martinpitt:


martinpitt:
Oh nice, that Just Works.. (other than the package mirror SNAFU in F41 on TF, but irrelevant here..)

jelly (approved):

PR #1920, opened 2024-12-03
fmf: Add CentOS image mode testing

martinpitt:
See https://gitlab.com/testing-farm/general/-/blob/main/rfd/rfd5-testing-image-mode.adoc

https://issues.redhat.com/browse/COCKPIT-1213

---

This doesn't work locally:
```
tmt run --until report provision --how virtual --image https://artifacts.dev.testing-farm.io/images/Fedora-40-image-mode-x86_64.qcow2
```
fails with

```
       Command 'rpm -q --whatprovides cockpit-ws cockpit-podman cockpit-system nginx systemd-container || rpm-ostree install --apply-live --idempotent --allow-inactive  cockpit-ws cockpit-podman cockpit-system nginx systemd-container' returned 1.
[...]
        error: Could not depsolve transaction; 3 problems detected:
         Problem 1: package nginx-2:1.26.2-1.fc40.x86_64 from updates requires nginx-core = 2:1.26.2-1.fc40, but none of the providers can be installed
          - conflicting requests
          - nothing provides nginx-mimetypes needed by nginx-core-2:1.26.2-1.fc40.x86_64 from updates
         Problem 2: package cockpit-ws-329.1-1.fc40.x86_64 from updates requires (selinux-policy >= 40.29 if selinux-policy-targeted), but none of the providers can be installed
          - cannot install both selinux-policy-40.29-2.fc40.noarch from updates and selinux-policy-40.28-1.fc40.noarch from @System
          - conflicting requests
         Problem 3: package systemd-container-255.15-1.fc40.x86_64 from updates requires libsystemd-shared-255.15-1.fc40.so()(64bit), but none of the providers can be installed
          - package systemd-container-255.15-1.fc40.x86_64 from updates requires libsystemd-shared-255.15-1.fc40.so(SD_SHARED)(64bit), but none of the providers can be installed
          - package systemd-container-255.15-1.fc40.x86_64 from updates requires systemd(x86-64) = 255.15-1.fc40, but none of the providers can be installed
          - cannot install both systemd-255.15-1.fc40.x86_64 from updates and systemd-255.13-1.fc40.x86_64 from @System
          - conflicting requests
          - nothing provides libbz2.so.1 needed by systemd-container-255.15-1.fc40.i686 from updates
          - nothing provides liblzma.so.5 needed by systemd-container-255.15-1.fc40.i686 from updates
          - nothing provides liblzma.so.5(XZ_5.0) needed by systemd-container-255.15-1.fc40.i686 from updates
          - nothing provides libacl.so.1 needed by systemd-container-255.15-1.fc40.i686 from updates
          - nothing provides libacl.so.1(ACL_1.0) needed by systemd-container-255.15-1.fc40.i686 from updates
```

This feels like an rpm-ostree issue together with a dnf depsolver. Reported to https://issues.redhat.com/browse/TFT-3020 . Let's see how that works in packit/TF. Also, this isn't quite the correct way, we should really build a derived container and boot into that.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -19,13 +19,15 @@ jobs:
   - job: tests
     trigger: pull_request
     targets: &test_targets
-      - fedora-41
-      - fedora-42
-      - fedora-latest-stable-aarch64
-      - fedora-rawhide
-      - centos-stream-9-x86_64
-      - centos-stream-9-aarch64
-      - centos-stream-10
+      fedora-41: {}
+      fedora-42: {}
+      fedora-latest-stable-aarch64: {}
+      fedora-rawhide: {}
+      centos-stream-9-x86_64:
+        distros: ["centos-stream-9", "CentOS-Stream-9-image-mode"]
+      centos-stream-9-aarch64: {}
+      centos-stream-10:
+          distros: ["centos-stream-10", "CentOS-Stream-10-image-mode"]
 
   - job: copr_build
     trigger: pull_request
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -4,7 +4,9 @@ set -eux
 cd "${0%/*}/../.."
 
 # HACK: ensure that critical components are up to date: https://github.com/psss/tmt/issues/682
-dnf update -y podman crun conmon criu
+if [ ! -e /sysroot/ostree ]; then
+    dnf update -y podman crun conmon criu
+fi
 
 # if we run during cross-project testing against our main-builds COPR, then let that win
 # even if Fedora has a newer revision
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -20,7 +20,9 @@ fi
 . /run/host/usr/lib/os-release
 export TEST_OS="${ID}-${VERSION_ID/./-}"
 
-if [ "$TEST_OS" = "centos-9" ]; then
+if [ -e /sysroot/ostree ]; then
+    TEST_OS="${TEST_OS}-ostree"
+elif [ "$TEST_OS" = "centos-9" ]; then
     TEST_OS="${TEST_OS}-stream"
 fi
 


martinpitt:
Hm, no TF test got triggered here. @lbarcziova @LachmanFrantisek Is packit supposed to know about the "CentOS-Stream-9-image-mode" test target already? I took that name from https://gitlab.com/testing-farm/general/-/blob/main/rfd/rfd5-testing-image-mode.adoc?ref_type=heads#user-content-composes .

martinpitt:
@lbarcziova @lachmanfrantisek  I filed https://issues.redhat.com/browse/PACKIT-3975 about this, probably better to discuss it there than some random PR.

lbarcziova:
Thanks @martinpitt , going to have a look and comment there.

packit.yaml
```diff
@@ -19,24 +19,13 @@ jobs:
...
     trigger: pull_request
     targets:
-      - fedora-40
-      - fedora-41
-      - fedora-latest-aarch64
-      - fedora-development
-      - centos-stream-9-x86_64
-      - centos-stream-9-aarch64
-      - centos-stream-10
+      centos-stream-9-image-mode:
```
lbarcziova, line 27:
```suggestion
      centos-stream-9-x86_64:
```
could you please try this? The key here should be a valid build target

martinpitt: @lbarcziova Thanks for the hint! That indeed queued a test. I expect this to fail due to reasons not related to packit (https://issues.redhat.com/browse/TFT-3020). But it should at least boot the correct image.

I suppose I could do

     distros: ["CentOS-Stream-9-image-mode", "centos-stream-9-x86_64"]

to actually test on both classic rpm and bootc?

lbarcziova: yes, the lists are supported, but if I remember correctly it should be without the architecture in the distros list, so 

```
distros: ["CentOS-Stream-9-image-mode", "centos-stream-9"]
```

packit.yaml
```diff
@@ -19,24 +19,13 @@ jobs:
...
     targets:
-      - fedora-40
-      - fedora-41
-      - fedora-latest-aarch64
-      - fedora-development
-      - centos-stream-9-x86_64
-      - centos-stream-9-aarch64
-      - centos-stream-10
+      centos-stream-9-x86_64:
+        distros: ["centos-stream-9", "CentOS-Stream-9-image-mode"]
```
thrix, line 29:
```suggestion
    tf_extra_params:
        environments:
          - artifacts:
              - type: repository
                id: https://my.repo/repository
```

thrix: @martinpitt let's try this, this should disable the artifacts installation I hope :)

martinpitt: @thrix Thanks for the hint! I'm a bit sceptical, as in [cockpit](https://github.com/cockpit-project/cockpit/blob/main/packit.yaml#L46) that felt additive, not replacing. But pushed, let's try!

martinpitt: @thrix No, didn't help unfortunately: https://artifacts.dev.testing-farm.io/5ba18f34-5551-42e1-867f-16d24984ee20/

martinpitt:
Installed build deps programmatically, so a local `tmt` run actually gets pretty far :star: But it now runs straight into https://github.com/cockpit-project/cockpit/issues/21201 , so we need cockpit 330 for that.

packit.yaml
martinpitt: Reported as https://issues.redhat.com/browse/TFT-3028

thrix: No need to report that, we have other ways to try it out, let me open a new MR so i can experiment and help

thrix: continuing myself in !1924 for a prototype

martinpitt:
I'm revisiting that, as we'll need TF image mode testing in the context of system-roles. CentOS-Stream-9-image-mode-x86_64 TF job is not getting picked up. https://api.dev.testing-farm.io/v0.1/composes now has it without the -x86 prefix, trying that.

martinpitt:
@lbarcziova Interesting, thanks! So that (1) needs to be conditionalized for *-image-mode targets only, and (2) we somehow need to get passed a reference/URL/repo to the artifacts so that the test plan can build its own bootc derivative OS with the artifact rpms. This is all rather non-discoverable and complicated, and ideally TF would do that just by itself (see https://issues.redhat.com/browse/TFT-3172, in particular https://issues.redhat.com/browse/TFT-3177). In fact it was already [announced to work that way](https://docs.testing-farm.io/Testing%20Farm/0.1/releases/2025-01.6.html)

Laura, do you happen to know some example packit config/plan where this has been worked out? Thanks!

test/browser/browser.sh
```diff
@@ -4,7 +4,9 @@ set -eux
 cd "${0%/*}/../.."
 
 # HACK: ensure that critical components are up to date: https://github.com/psss/tmt/issues/682
-dnf update -y podman crun conmon criu
+if [ ! -e /sysroot/ostree ]; then
+    dnf update -y podman crun conmon criu
+fi
```
jelly, line 9: Wondering if `bootcl status` isn't a better fit if bootc ever moves out of ostree

jelly (approved):
Small comment, otherwise LGTM!

test/browser/browser.sh
martinpitt: I indeed started with this, but:we have to check for bootc in run-test.sh as well, but it fails as non-root; and then I didn't want to have two different checks in these files, so I moved both to the `-e /sysroot` check. At least if the dir stops existing in ostree/bootc, the test will fail, so we'll know.

PR #1932, opened 2024-12-09
debug tasks-container-update failure

martinpitt:
The last runs of this workflow failed in [machines](https://github.com/cockpit-project/cockpit-machines/actions/runs/12228090535/job/34105778704) , [podman](https://github.com/cockpit-project/cockpit-podman/actions/runs/12230669044/job/34112363978), and cockpit. It worked in starter-kit and ostree, and curiously also in files. The latter doesn't use `environment: self`, so supposedly that's the crucial bit.

Not sure what changed since October, when they were still working. Locally this works:

```
PYTHONPATH=bots python3 -c 'import task; print(task.default_branch())'
main
```

But of course I'm also not using a deploy key locally nor the GitHub token. This needs to be debugged in the workflow infra.

### DIFF
--- a/.github/workflows/tasks-container-update.yml
+++ b/.github/workflows/tasks-container-update.yml
@@ -31,6 +31,12 @@ jobs:
       - name: Pacify git's permission check
         run: git config --global --add safe.directory /__w/cockpit-podman/cockpit-podman
 
+      - run: dnf install -y tmate curl
+
+      - uses: mxschmitt/action-tmate@v3
+        with:
+          install-dependencies: false
+
       - name: Run tasks-container-update
         run: |
           make bots
--- a/packit.yaml
+++ b/packit.yaml
@@ -17,28 +17,6 @@ srpm_build_deps:
 
 jobs:
   - job: copr_build
-    trigger: pull_request
-    targets:
-    - fedora-40
-    - fedora-41
-    - fedora-latest-aarch64
-    - fedora-development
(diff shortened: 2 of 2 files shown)


martinpitt:
In the workflow shell this reproduces effortlessly:
```
bash-5.2# PYTHONPATH=bots python3 -c 'import task; print(task.default_branch())'
  File "<string>", line 1, in <module>
  File "/__w/cockpit-podman/cockpit-podman/bots/task/__init__.py", line 441, in default_branch
    return api.get()["default_branch"]
           ~~~~~~~~~^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

GITHUB_REF seems alright:
```
bash-5.2# echo $GITHUB_REF
refs/heads/debug-workflow
```

but the way the code is written, `default_branch()` still gets called anyway. We could fix that, but let's find out what broke -- getting the API is important.

```
# PYTHONPATH=bots python3 -c 'import task; print(task.github.GitHub().get())'
None
```

This is the problem:

```
# PYTHONPATH=bots python3 -c 'import task; print(task.github.GitHub().repo)'
cockpit-project/cockpit-podman.git
```

The `.git` extension needs to go. This needs fixing in bots/task/github.py get_origin_repo().

martinpitt:
Fixed in https://github.com/cockpit-project/bots/pull/7198

PR #1936, opened 2024-12-12
test: show catatonit version for testing farm runs

jelly:
CentOS 10 is currently broken with:

Error: building local pause image: finding pause binary: exec: "catatonit": executable file not found in $PATH

So check if the package is installed.

---

Can already locally reproduce the issue but this seems useful regardless

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -21,7 +21,7 @@ if [ -n "$main_builds_repo" ]; then
 fi
 
 # Show critical package versions
-rpm -q runc crun podman criu passt kernel-core selinux-policy cockpit-podman cockpit-bridge || true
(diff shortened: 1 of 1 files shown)


martinpitt:
ugh @ https://artifacts.dev.testing-farm.io/5045dea9-2e01-4852-82f1-4678146dffb6/ -- I pretend I didn't see that :see_no_evil:

martinpitt (approved):

martinpitt:
Works as intended, [c10s](https://artifacts.dev.testing-farm.io/c30f6082-7604-4feb-bfe5-05957276f6e7/) run shows

> catatonit-0.2.0-1.el10.x86_64

and the run still fails on the catatonit error.

PR #1938, opened 2024-12-16
test: Drop redundant cleanup

martinpitt:
We don't need to do the same commands on test setup *and* teardown. Let's do it on teardown only.


martinpitt:
Meh, [seems we really do need](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1938-7e4f7197-20241216-071003-ubuntu-2404/log.html) the redundant cleanup.

PR #1951, opened 2025-01-03
Use TypeaheadSelect in ImageRunModal

mvollmer:
- [x] https://github.com/cockpit-project/cockpit/pull/21482


mvollmer:
Registry searching is too slow for our tests on Arch...

martinpitt (approved):
Cheers! That is the last one in all our projects :star2:

PR #1953, opened 2025-01-07
test: Factorize registry setup in testDownloadImage()

martinpitt:
Use our setupRegistry() helper.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -1036,12 +1036,7 @@ class TestApplication(testlib.MachineCase):
         execute = self.execute
 
         def prepare():
-            # Create and start registry containers
-            self.execute(True, f"podman run -d -p 5000:5000 --name registry --stop-timeout 0 {IMG_REGISTRY}")
-            self.execute(True, f"podman run -d -p 6000:5000 --name registry_alt --stop-timeout 0 {IMG_REGISTRY}")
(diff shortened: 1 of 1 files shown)


test/check-application
```diff
@@ -1036,11 +1036,7 @@ class TestApplication(testlib.MachineCase):
         execute = self.execute
 
         def prepare():
-            # Create and start registry containers
-            self.execute(True, f"podman run -d -p 5000:5000 --name registry --stop-timeout 0 {IMG_REGISTRY}")
-            self.execute(True, f"podman run -d -p 6000:5000 --name registry_alt --stop-timeout 0 {IMG_REGISTRY}")
-            # Add local insecure registry into registries conf
-            self.machine.write("/etc/containers/registries.conf", REGISTRIES_CONF)
+            self.setupRegistry()
             self.execute(True, "systemctl stop podman.service")
```
mvollmer, line 1040: This is already included in `setupRegistry` and doesn't need to be repeated here.

mvollmer (changes requested):
Thanks! Let's remove the redundant podman stopping. It's harmless now but can only cause trouble in the future.

test/check-application
martinpitt: Thanks for spotting! I meant to do that, but just forgot.

PR #1955, opened 2025-01-07
test: Disable firewalld StrictForwardPorts on RHEL 10

martinpitt:
RHEL 10 started to break `podman -p` port forwarding to localhost by default [1]. This was a deliberate change [2][3]. We really don't care about firewalling localhost ports from containers (does anybody really?), so revert back to the previous behaviour (which is still the default on Fedora) of having StrictForwardPorts=no.

This obsoletes our naughty https://github.com/cockpit-project/bots/issues/7291

[1] https://issues.redhat.com/browse/RHEL-72937
[2] https://firewalld.org/2024/11/strict-forward-ports
[3] https://github.com/firewalld/firewalld/issues/1380

---

This should make the [four skipped tests](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1951-68a71ea4-20250107-150223-rhel-10-0/log.html) green again. After this lands, I'll send a bots PR to drop the naughty.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -23,6 +23,13 @@ Delegate=cpu cpuset io memory pids
 EOF
 fi
 
+if grep -q platform:el10 /usr/lib/os-release && [ -e /etc/firewalld/firewalld.conf ]; then
+    # HACK: unbreak container port forwarding to localhost
+    # https://firewalld.org/2024/11/strict-forward-ports and https://github.com/firewalld/firewalld/issues/1380
+    # TF runs have no firewalld
+    sed -i 's/StrictForwardPorts=yes/StrictForwardPorts=no/' /etc/firewalld/firewalld.conf
+fi
+
 # don't force https:// (self-signed cert)
 mkdir -p /etc/cockpit
 printf "[WebService]\\nAllowUnencrypted=true\\n" > /etc/cockpit/cockpit.conf


martinpitt:
[works](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1955-90a66b47-20250107-154642-rhel-10-0/log.html)!

martinpitt:
Ah, [no firewalld on c10s TF](https://artifacts.dev.testing-farm.io/9eb0c92b-551b-48aa-ab86-d6aaaa8ea843/)

test/vm.install
```diff
@@ -23,6 +23,13 @@ Delegate=cpu cpuset io memory pids
 EOF
 fi
 
+if grep -q platform:el10 /usr/lib/os-release && [ -e /etc/firewalld/firewalld.conf ]; then
```
mvollmer, line 26: Do we need to restrict this to rhel-10? We could just erase StrictForwardPort settings everywhere.

mvollmer: Also, could we install a config snippet that overrides StrictPortForward, instead of having to know which file has the changed OS default?

mvollmer (approved):
Thanks! How annoying!

test/vm.install
martinpitt: > Do we need to restrict this to rhel-10

We don't strictly have to, but as the default is explicitly "no" on Fedora and everywhere else, and this is currently a RHEL 10-ism, I'd rather restrict the hack to that, and learn if it creeps in anywhere else. TBH this is so ridiculously broken that I don't expect this to stay around for long -- this needs to go back to the drawing board.

> could we install a config snippet

I'd love to, and that's usually the first thing I look for. Changing big distro-provided conffiles is baaad. But firewalld.conf doesn't seem to have anything that resembles an "include conf.d", nor is there any directory like that. But again, I expect this to be a temporary hack.

PR #1959, opened 2025-01-10
Support listing volumes

jelly:
Rought mockup with basic volume listing

image

Let's add a basic listing and then delete support.

Tests needed

* [x] Delete volume
* [ ] Force delete volume
* [ ] Long volume names
* [ ] Unused volume detection
* [ ] Filter on volume names
* [ ] Quadlet volume listing?

As follow up:

* Show sizes, there is a /system/df endpoint but it needs investigation on why its slow with btrfs graphDriver and other performance concerns
* Prune unused volumes like images
* Create volume support in the container page
* Export volume (download), which begs the question about import support.

Related to: #2342

### DIFF
--- a/src/VolumeDeleteModal.jsx
+++ b/src/VolumeDeleteModal.jsx
@@ -1,0 +1,52 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+import React, { useState } from 'react';
+
+import { Button } from "@patternfly/react-core/dist/esm/components/Button";
+import {
+    Modal, ModalBody, ModalFooter, ModalHeader
+} from '@patternfly/react-core/dist/esm/components/Modal';
+import { useDialogs } from "dialogs.jsx";
+
+import cockpit from 'cockpit';
+
+import * as client from './client.js';
+
+const _ = cockpit.gettext;
+
+export const VolumeDeleteModal = ({ con, volume }) => {
+    const [force, setForce] = useState(false);
+    const [reason, setReason] = useState(null);
+    const Dialogs = useDialogs();
+
+    const handleRemoveVolume = async () => {
+        setReason(null);
+        try {
+            await client.deleteVolume(con, volume.Name, force);
+            Dialogs.close();
+        } catch (exc) {
+            setReason(exc.message);
+            setForce(true);
+        }
+    };
+
+    return (
+        <Modal isOpen
+                 position="top" variant="medium"
+                 onClose={Dialogs.close}
+        >
+            <ModalHeader title={cockpit.format(_("Delete $0 volume?"), volume.Name)}
+                titleIconVariant="warning"
+            />
+            <ModalBody>
+                {reason}
+            </ModalBody>
+            <ModalFooter>
+                <Button id="btn-volume-delete" variant="danger"
+                        onClick={() => handleRemoveVolume()}>
+                    {force ? _("Force delete volume") : _("Delete volume")}
+                </Button>
+                <Button variant="link" onClick={Dialogs.close}>{_("Cancel")}</Button>
+            </ModalFooter>
+        </Modal>
+    );
+};
--- a/src/Volumes.jsx
+++ b/src/Volumes.jsx
@@ -1,0 +1,188 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+import React from 'react';
+
+import { Card, CardBody, CardHeader, CardTitle } from "@patternfly/react-core/dist/esm/components/Card";
+import { DropdownItem } from '@patternfly/react-core/dist/esm/components/Dropdown/index.js';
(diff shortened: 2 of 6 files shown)


tomasmatus:
I like this, thanks! Another idea I have is to show which containers use the volume. Maybe a popover for the text in the "Used by" column?

jelly:
> I like this, thanks! Another idea I have is to show which containers use the volume. Maybe a popover for the text in the "Used by" column?

Right we have this for images!

image

So maybe an expandable makes sense here.

src/app.jsx
```diff
@@ -196,6 +198,30 @@ class Application extends React.Component {
                 .catch(e => console.warn("initContainers uid", con.uid, "getContainers failed:", e.toString()));
     }
 
+    initVolumes(con) {
+        return client.getVolumes(con)
+                .then(volumesList => {
+                    console.log(volumesList);
```
Venefilyn: Should be removed from the commit

src/app.jsx
```diff
@@ -22,6 +22,7 @@ import { superuser } from "superuser";
 import ContainerHeader from './ContainerHeader.jsx';
 import Containers from './Containers.jsx';
 import Images from './Images.jsx';
+import Volumes from './Volumes.jsx';
```
Venefilyn, line 25: Hasn't been added on or before this commit

src/VolumeDeleteModal.jsx
```diff
@@ -0,0 +1,52 @@
...
+            <ModalHeader title={cockpit.format(_("Delete $0 volume?"), volume.Name)}
+                titleIconVariant="warning"
+            />
+            <ModalBody>
+                {reason}
+            </ModalBody>
+            <ModalFooter>
+                <Button id="btn-volme-delete" variant="danger"
+                        onClick={() => handleRemoveVolume()}>
+                    {force ? _("Force delete volume") : _("Delete volume")}
```
Venefilyn, line 46: Force is never used atm, state is never modified

src/Volumes.jsx
```diff
@@ -0,0 +1,186 @@
...
+    );
+
+    return (
+        <Card id="containers-volumes" className="containers-volumes">
+            <CardHeader>
+                <Flex flexWrap={{ default: 'nowrap' }} className="pf-v6-u-w-100">
+                    <FlexItem grow={{ default: 'grow' }}>
+                        <Flex>
+                            <CardTitle>
+                                <h2 className="containers-images-title">{_("Volumes")}</h2>
```
Venefilyn: Should be
```suggestion
                                <Content component={ContentVariants.h1} className="containers-volumes-title">{_("Volumes")}</Content>
```
Currently the heading differs from images and containers. The class name also needs a change

src/Volumes.jsx
```diff
@@ -0,0 +1,186 @@
...
+                        volume={volume}
+
+        />);
+    };
+
+    const dropdownActions = [
+        <DropdownItem key={volume.Name + "delete"}
+                    component="button"
+                    className="pf-m-danger btn-delete"
+                    onClick={removeImage}>
```
Venefilyn:
```suggestion
                    onClick={removeVolume}>
```

src/Volumes.jsx
```diff
@@ -0,0 +1,186 @@
...
+                    : cardBody}
+            </CardBody>
+        </Card>
+    );
+};
+
+const VolumeActions = ({ con, volume }) => {
+    const Dialogs = useDialogs();
+
+    const removeImage = () => {
```
Venefilyn:
```suggestion
    const removeVolume = () => {
```

src/app.jsx
```diff
@@ -196,6 +198,29 @@ class Application extends React.Component {
...
 
+    initVolumes(con) {
+        return client.getVolumes(con)
+                .then(volumesList => {
+                    this.setState(prevState => {
+                        const copyVolumes = {};
+                        Object.entries(prevState.volumes || {}).forEach(([id, volume]) => {
+                            if (volume.uid !== con.uid)
+                                copyVolumes[id] = volume;
+                        });
```
Venefilyn, line 209: This seems expensive. Can we keep an index of everything we found so far or otherwise not do a heavy `Object.entries` on existing state?

Venefilyn (changes requested):
Overall looks good! Couple of comments going through the commits

src/app.jsx
jelly: This is how our state works for containers, images and now also volumes. However I don't understand your concern here iterating over an array is not *that* slow in JavaScript and we do need new state (ie  an object).

If we want to improve this I am not sure how as I don't understand your proposal, getting the volumes basically resets the state for the current logged in user (uid).

src/Volumes.jsx
jelly: Fixed.

PR #1963, opened 2025-01-13
Hide container/image Command details for quadlets

jelly:
Quadlets have an `EntryPoint` and no `Command` so for quadlets we would have shown an empty `Command` which isn't terribly useful or informative.

---

Before:

image


martinpitt:
Looks good, but this isn't specific to quadlets, right? other container images may also not have a command. Can you please add a test?

jelly:
> Looks good, but this isn't specific to quadlets, right? other container images may also not have a command. Can you please add a test?

Sure, will do tomorrow. Basically for images the issue was that the Cmd was `undefined` for containers the whole check was missing.

tomasmatus (approved):
Thanks!

martinpitt (approved):
Cheers!

PR #1967, opened 2025-01-17
test: Simplify selinux check

jelly:


test/check-application
```diff
@@ -153,7 +153,7 @@ class TestApplication(testlib.MachineCase):
 
         # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008249
         self.has_criu = "debian" not in m.image and "ubuntu" not in m.image
-        self.has_selinux = not any(img in m.image for img in ["arch", "debian", "ubuntu", "suse"])
+        self.has_selinux = m.image not in ["arch", "debian", "ubuntu", "suse"]
```
martinpitt, line 156: Ah, that doesn't work, these are substrings, not image names. Now I understand what this wanted to do.

jelly: :facepalm: it's almost weekend :)

PR #1974, opened 2025-01-21
test: Fix firefox scenario

martinpitt:
Actually set `$TEST_BROWSER` so that the /firefox scenario tests what's written on the tin.


tomasmatus (approved):
🦊

PR #1977, opened 2025-01-26
Add screenReaderText, Fix XTerm.js crash, Improve console log messages

martinpitt:
Fixups that I split out from #1976.

See [this log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1976-80759695-20250124-114135-fedora-41/log.html#37) for the `dimensions` crash:

> error: TypeError: Cannot read properties of undefined (reading 'dimensions')

PR #1976, opened 2025-01-23
Automatically start podman.socket

martinpitt:
Always start the system and user `podman.socket` unit on initialization. There really is no reason to explicitly ask the user about it -- we can treat it as "extended cockpit" to just access podman.

There also isn't a reason to enable the socket unit -- starting it on demand is fine from cockpit-podman's perspective.

Now the user will only see the empty state if the service fails, which should be very rare. So turn the Troubleshoot button into a primary one.

We also don't need the "System/User podman is also available" alerts any more -- gaining root privileges auto-starts the system service.

-----

This is a necessary prerequisite for https://issues.redhat.com/browse/COCKPIT-1086 , but useful in its own right -- it removes a big hurdle/annoyance at startup.

 - [x] builds on top of #1977 

## Podman: Automatically start podman.socket

If the system or user's `podman.socket` unit isn't running, cockpit-podman now silently starts it on its own, instead of asking the user about it. There is no need to have the service enabled (i.e. always start it on boot or user login) for cockpit-podman any more.

PR #1978, opened 2025-01-26
Rewrite Application.init() async

martinpitt:
We will soon add another piece of initialization at the top, and this already has too many indentation levels.

 - [ ] builds on top of #1977

### DIFF
--- a/src/ContainerLogs.jsx
+++ b/src/ContainerLogs.jsx
@@ -80,11 +80,14 @@ class ContainerLogs extends React.Component {
     }
 
     resize(width) {
+        if (!this.term?._core?._renderService?.dimensions)
+            return;
         // 24 PF padding * 4
         // 3 line border
         // 21 inner padding of xterm.js
         // xterm.js scrollbar 20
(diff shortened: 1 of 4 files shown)


src/ContainerLogs.jsx
```diff
@@ -80,11 +80,14 @@ class ContainerLogs extends React.Component {
     }
 
     resize(width) {
+        if (!this.term?._core?._renderService?.dimensions)
```
cockpituous, line 83: This added line is not executed by any test.

src/app.jsx
```diff
@@ -200,7 +200,8 @@ class Application extends React.Component {
                     });
                     this.updateContainerStats(system);
                 })
-                .catch(console.log);
+                .catch(e => console.warn("initContainers", system ? "system" : "user",
+                                         "getContainers failed:", e.toString()));
```
cockpituous, line 204: These 2 added lines are not executed by any test.

martinpitt:
Green -- that's good information, I mostly just wanted to test this in isolation as the topmost commit in #1976 breaks the podman service in a weird way.

PR #1981, opened 2025-01-31
Add type check framework

martinpitt:
I'm about to do bigger changes to podman's code base. Before I start that, I'd like to TSify at least the basals, like src/client.js (not necessarily the top-level components).

 - [ ] Land https://github.com/cockpit-project/cockpit/pull/21569 and adjust the cockpit checkout SHA


martinpitt:
@jelly I'm making more typescript things in a branch on top of that, so far it's a nice learning exercise (and found a bug). But you mentioned you had doubts here, so if you don't like this please reject now and I'll put this to rest.

martinpitt:
See https://github.com/cockpit-project/cockpit-podman/pull/1987#issuecomment-2630264907

PR #1983, opened 2025-01-31
test: adjust to new CRIU unsupported error from podman

jelly:
The [new debian-testing](https://github.com/cockpit-project/bots/pull/7379) image now shows a different error, to avoid line limits convert it to a list. In a follow up I intend to cleanup the errors a bit.


martinpitt (approved):
Dankjewel!

PR #1988, opened 2025-02-03
test: Check podman socket going away while cockpit session is running

martinpitt:
This covers the `streamEvents` plus explicit channel watching in
app.jsx. We can use this to guard when to remove the extra
`cockpit.channel()` `close` watching, which is not necessary with the
Python bridge, but still relevant on the C bridge (Debian 12 and RHEL 8).

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -487,7 +487,8 @@ class Application extends React.Component {
                     this.cleanupAfterService(system);
                 });
 
-        // Listen for podman socket/service going away
+        // HACK: Listen for podman socket/service going away; this is only necessary with the C bridge
+        // (Debian 12, RHEL 8). With the Python bridge, the above streamEvents() resolves when the service goes away.
         const ch = cockpit.channel({ superuser: system ? "require" : null, payload: "stream", unix: client.getAddress(system) });
         ch.addEventListener("close", () => {
             console.log("init", system ? "system" : "user", "podman service closed");
--- a/test/check-application
+++ b/test/check-application
@@ -1640,19 +1640,33 @@ WantedBy=multi-user.target default.target
     def testFailingPodmanService(self):
         b = self.browser
 
-        self.execute(True, "systemctl mask podman.service")
-        self.addCleanup(self.execute, True, "systemctl unmask podman.service")
-        self.execute(False, "systemctl --user mask podman.service")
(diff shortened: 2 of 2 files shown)


src/app.jsx
```diff
@@ -483,19 +483,10 @@ class Application extends React.Component {
         client.streamEvents(system, message => this.handleEvent(message, system))
                 .catch(e => console.error("init", system ? "system" : "user", "streamEvents failed:", e.toString()))
                 .finally(() => {
+                    console.log("init", system ? "system" : "user", "podman service closed");
```
cockpituous, line 486: This added line is not executed by any test.

jelly (dismissed):
Keep the cleanups coming :+1:

src/app.jsx
jelly: So I suppose we never test, cleanupAfterService now or before, but can we do that? Wouldn't it just involve `systemctl stop cockpit.socket` while the page is open?

martinpitt: Ah! `testFailingPodmanService` checks exactly that, it stops both user and system service and verifies that the empty state appears. However, it does it synchronously *before* logging into cockpit. I added a test which checks it while cockpit is running, and I see these messages in the console now. So next round should be covered.

cockpituous, line 486: This added line is not executed by any test.

test/check-application
```diff
@@ -1640,19 +1640,33 @@ WantedBy=multi-user.target default.target
...
-        # Troubleshoot action
+        try:
+            self.execute(True, "systemctl mask podman.service")
+            self.execute(False, "systemctl --user mask podman.service")
+            self.login_and_go("/podman")
+
+            # Troubleshoot action
+            b.wait_text("#app .pf-v5-c-empty-state__title", "Podman service failed")
+            b.click("#app .pf-v5-c-empty-state button")
+            b.enter_page("/system/services")
```
jelly, line 1651: This is a minor nit, but we have a `troubleshoot` button which should jump to `/system/services#/podman.socket` where we can also assert the status.

jelly (approved):

test/check-application
martinpitt: The test above already does that, I thought we don't have to do it twice.

PR #2004, opened 2025-02-12
workflows: Move dependabot and reposchutz from Ubuntu 20.04 to 24.04

martinpitt:
20.04 will soon be EOL. "ubuntu-latest" is 24.04 now with a newer podman.

---

I got a warning from GitHub this morning that ubuntu-20.04 workflows will have three warning "brownouts" during March, and cease to exist on April 1.

We moved these workflows to -latest in https://github.com/cockpit-project/cockpit/commit/fa659236791fbdbab229bdaad2e544cc2ad1d2da and everything is fine.

Same as https://github.com/cockpit-project/cockpit-files/pull/948

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -10,8 +10,7 @@ jobs:
       contents: read
       pull-requests: write
     timeout-minutes: 5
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (approved):
Thanks!

Issue #1862, opened 2024-09-26
containers list: Missing right side of row when downloading


garrett:
When downloading images to create a container, the row doesn't extend fully. You can easily see this by noticing the divider line doesn't continue to the edge:

![image](https://github.com/user-attachments/assets/713b7674-1099-4673-becb-48b1d5183df1)

This could be solved either with a colspan of 2 on the last row or by having an actions cell too, even if it's empty.

If we can cancel downloading, it'd make sense to have an actions menu anyway. (If we don't, it'd be fine if it's blank.)

Mash707:
Hi @garrett I am able to reproduce this issue, but I am not able to work it out. I have looked into [Containers.jsx](https://github.com/cockpit-project/cockpit-podman/blob/main/src/Containers.jsx) which seems to be the right place for the fix.
Could you provide some details where I can implement the `colspan`?
I was able to resolve this issue using chrome dev tools.

![Image](https://github.com/user-attachments/assets/33cfb1bd-5fca-4c99-ae39-2dcfdc4d1d69)

![Image](https://github.com/user-attachments/assets/9de272bc-0528-417e-bf0a-b0ac4efa7c32)

I changed the colspan value from 6 to 7 which seems to fix this.

jelly:
That is probably more explained by this:

```
        if (!container.isDownloading) {
            columns.push({
                title: <ContainerActions container={container}
                                         onAddNotification={this.props.onAddNotification}
                                         localImages={localImages}
                                         updateContainer={this.props.updateContainer}
                                         isSystemdService={isSystemdService} />,
                props: { className: "pf-v5-c-table__action" }
            });
        }

```

We don't add this column so the colspan=6 and that causes issues.

PR #2012, opened 2025-02-20
Disable container actions when downloading

jelly:
Not showing the container actions causes the downloading container row to not fully extend so instead disable the container actions so the columns stay the same.

Closes: #1862

---

Screenshot From 2025-02-20 10-37-42

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -45,7 +45,7 @@ import '@patternfly/patternfly/utilities/Accessibility/accessibility.css';
 
 const _ = cockpit.gettext;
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Cheers!

garrett (approved):
This looks like a good approach!

I would have probably made it an empty table cell (`td`) instead of disabled actions, but that's a bit of a quibble. Both solutions (empty cell vs. disabled action) are fine here, and I'm happy with this.

Thanks!! :+1: :rocket:

PR #2014, opened 2025-02-20
test: take a pixel test of a pod

jelly:
Previously we never had a pixel test for a pod.


martinpitt (approved):
Thanks! That will conflict with my #1989 , I'll rebase mine.

PR #2024, opened 2025-02-26
Support connecting to other users

martinpitt:
When the session has superuser capabilities, scan the system for all users
which have some podman containers running (via cgroupfs heuristics) and add
them to the owner selector.

Teach `rest.getAddress()` and the podman.socket startup to connect to the
podman service of an arbitrary UID by assuming that their runtime directory is
`/run/user/$UID`. This is a necessary, but safe enough assumption -- if the
system was modified to use a different one, then we won't support connecting to
that user.

However, we want to be careful and only connect to one extra user at a time: we
don't want connections to pile up (as the parallel events will eventually slow
down the page) or interfere with each other (as these users are isolated for a
reason). So only connect to the service once the user actually switches to that
owner, and disconnect again when switching away.

Fixes #692
https://issues.redhat.com/browse/COCKPIT-1086

----

I am testing this manually by starting a container as root, as user `admin` and as user `builder` (the system user on our bot VMs that runs package builds).

As root in the VM:

```sh
podman run -d --name sys1 --rm docker.io/busybox sleep infinity

loginctl enable-linger builder
su -c 'podman pull quay.io/cockpit/ws' - builder
su -c 'podman run -d --name build1 --rm docker.io/busybox sleep infinity' - builder
```

and the admin container:

```sh
ssh admin@c podman run -d --name adm1 --rm docker.io/busybox sleep infinity
```

Then you should see "builder" in the owner list after logging in, and can switch to it, and away from it:

image

Requirements:
 - [x] #1989
 - [x] #2022 
 - [x] #2020 
 - [x] #2039
 - [x] #2038
 - [x] #2042
 - [x] #2043

TODO:
 - [ ] Check with @garrett about this workflow -- timeout after a month of waiting. This doesn't really touch much UI
 - [x] Should we disconnect from other user when switching Owner to "all"? (right now it stays connected) -- see @jelly 's feedback below, let's do that.
 - [x] More integration tests: create new container, delete container
 - [x] test/fix changing users via URL change

PR #2038, opened 2025-03-07
Fix console junk and cockpit.http() streaming lock-ups

martinpitt:
See my woes in https://github.com/cockpit-project/cockpit-podman/pull/2024#issuecomment-2705655132 . This is a problem on current main already. I started with reproducing this in a test, the first run should make sure it actually fails on all OSes as expected -- it does on commit 987ef61b, like [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2038-987ef61b-20250307-075845-fedora-42/log.html). Only debian-stable and rhel-8-10 are passing (with the C bridge).

The current PR version also includes the fix.

I realize this is rather odd. I put it onto my TODO list to reproduce this issue in cockpit's unit tests and fix that bug properly. But I've been chasing this for three days, and it's not going to be easy with the existing https://docs.python.org/3/library/http.client.html based implementation, we'll have to come up with something creative -- such as generalizing the approach here.

### DIFF
--- a/src/ContainerLogs.jsx
+++ b/src/ContainerLogs.jsx
@@ -149,7 +149,18 @@ class ContainerLogs extends React.Component {
             }
             // First 8 bytes encode information about stream and frame
             // See 'Stream format' on https://docs.docker.com/engine/api/v1.40/#operation/ContainerAttach
-            this.view.writeln(data.slice(8));
+            while (data.byteLength >= 8) {
+                // split into frames (size is the second 32-bit word)
+                const size = data[7] + data[6] * 0x100 + data[5] * 0x10000 + data[4] * 0x1000000;
+                const frame = data.slice(8, 8 + size);
+                // old podman versions just have CR endings, append NL then
+                if (frame[size - 1] === 13)
+                    this.view.writeln(frame);
+                else
+                    // recent podman  versions have CRNL endings
+                    this.view.write(frame);
+                data = data.slice(8 + size);
+            }
         }
     }
 
--- a/src/ContainerLogs.jsx
+++ b/src/ContainerLogs.jsx
@@ -113,19 +113,9 @@ class ContainerLogs extends React.Component {
         this.resize(this.props.width);
 
         const connection = rest.connect(this.props.uid);
-        const options = {
-            method: "GET",
-            path: client.VERSION + "libpod/containers/" + this.props.containerId + "/logs",
-            body: "",
-            binary: true,
-            params: {
-                follow: true,
-                stdout: true,
-                stderr: true,
-            },
-        };
-
-        connection.monitor(options, this.onStreamMessage, true)
+        connection.monitor(client.VERSION + "libpod/containers/" + this.props.containerId +
+                           "/logs?follow=true&stdout=true&stderr=true",
+                           this.onStreamMessage, true)
                 .then(this.onStreamClose)
                 .catch(e => {
                     const error = JSON.parse(new TextDecoder().decode(e.message));
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -494,18 +494,11 @@ class Application extends React.Component {
         this.updatePods(con);
 
         client.streamEvents(con, message => this.handleEvent(message, con))
-                .catch(e => console.error("init uid", uid, "streamEvents failed:", e.toString()))
-                .finally(() => this.cleanupAfterService(con));
-
-        // HACK: Listen for podman socket/service going away; this is only necessary with the C bridge
-        // (Debian 12, RHEL 8). With the Python bridge, the above streamEvents() resolves when the service goes away.
-        const address = rest.getAddress(uid);
-        const ch = cockpit.channel({ unix: address.path, superuser: address.superuser, payload: "stream" });
-        ch.addEventListener("close", () => {
-            console.log("init uid", uid, "podman service closed");
-            this.cleanupAfterService(con);
-        });
-        ch.send("GET " + client.VERSION + "libpod/events HTTP/1.0\r\nContent-Length: 0\r\n\r\n");
+                .catch(e => console.error("uid", uid, "streamEvents failed:", e.toString()))
+                .finally(() => {
+                    console.log("uid", uid, "podman service closed");
+                    this.cleanupAfterService(con);
+                });
     }
 
     componentDidMount() {
--- a/src/rest.js
+++ b/src/rest.js
@@ -19,6 +19,7 @@ function manage_error(reject, error, content) {
 let call_id = 0;
 
 const NL = '\n'.charCodeAt(0); // always 10, but avoid magic constant
+const CR = '\r'.charCodeAt(0); // always 13, but avoid magic constant
 
 const PODMAN_SYSTEM_ADDRESS = "/run/podman/podman.sock";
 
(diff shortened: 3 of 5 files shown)


src/app.jsx
```diff
@@ -494,18 +496,11 @@ class Application extends React.Component {
...
-        // HACK: Listen for podman socket/service going away; this is only necessary with the C bridge
-        // (Debian 12, RHEL 8). With the Python bridge, the above streamEvents() resolves when the service goes away.
-        const address = rest.getAddress(uid);
-        const ch = cockpit.channel({ unix: address.path, superuser: address.superuser, payload: "stream" });
-        ch.addEventListener("close", () => {
-            console.log("init uid", uid, "podman service closed");
-            this.cleanupAfterService(con);
-        });
-        ch.send("GET " + client.VERSION + "libpod/events HTTP/1.0\r\nContent-Length: 0\r\n\r\n");
+                .catch(e => console.error("uid", uid, "streamEvents failed:", e.toString()))
```
cockpituous, line 499: This added line is not executed by any test.

src/rest.js
```diff
@@ -40,6 +40,16 @@ function getAddress(uid) {
...
 }
 
+// split an Uint8Array at \r\n\r\n (separate headers from body)
+function splitAtNLNL(array) {
+    for (let i = 0; i <= array.length - 4; i++) {
+        if (array[i] === 13 && array[i + 1] === 10 && array[i + 2] === 13 && array[i + 3] === 10) {
+            return [array.subarray(0, i), array.subarray(i + 4)];
+        }
+    }
+    return [array, null]; // Return original if not found
```
cockpituous, line 50: This added line is not executed by any test.

src/rest.js
```diff
@@ -104,8 +82,67 @@ function connect(uid) {
...
+                if (headers.match(/^HTTP\/1.*\s+200\s/)) {
+                    // any further message is actual streaming data
+                    ch.removeEventListener("message", onHTTPMessage);
+                    ch.addEventListener("message", onDataMessage);
+
+                    // process the initial response data
+                    if (body)
+                        onDataMessage({ detail: body });
+                } else {
+                    reject(new Error(headers.split('\r\n')[0]));
```
cockpituous, line 109: This added line is not executed by any test.

src/rest.js
```diff
@@ -104,8 +82,67 @@ function connect(uid) {
...
+            // the initial message is the HTTP status response
+            ch.addEventListener("message", onHTTPMessage);
+
+            ch.send("GET " + path + " HTTP/1.0\r\nContent-Length: 0\r\n\r\n");
+        });
+    };
+
     connection.close = function () {
         http.close();
+        raw_channels.forEach(ch => ch.close());
```
cockpituous, line 145: This added line is not executed by any test.

src/app.jsx
```diff
@@ -494,18 +494,11 @@ class Application extends React.Component {
...
```
cockpituous, line 497: This added line is not executed by any test.

src/rest.js
```diff
@@ -40,6 +41,16 @@ function getAddress(uid) {
     throw new Error(`getAddress: uid ${uid} not supported`);
 }
 
+// split an Uint8Array at \r\n\r\n (separate headers from body)
+function splitAtNLNL(array) {
+    for (let i = 0; i <= array.length - 4; i++) {
+        if (array[i] === CR && array[i + 1] === NL && array[i + 2] === CR && array[i + 3] === NL) {
+            return [array.subarray(0, i), array.subarray(i + 4)];
```
cockpituous, line 48: These 2 added lines are not executed by any test.

src/rest.js
```diff
@@ -104,8 +83,67 @@ function connect(uid) {
...
+                const headers = decoder.decode(headers_bin);
+                debug(user_str, "monitor", path, "HTTP response:", headers);
+                if (headers.match(/^HTTP\/1.*\s+200\s/)) {
+                    // any further message is actual streaming data
+                    ch.removeEventListener("message", onHTTPMessage);
+                    ch.addEventListener("message", onDataMessage);
+
+                    // process the initial response data
+                    if (body)
+                        onDataMessage({ detail: body });
```
cockpituous, line 108: This added line is not executed by any test.

src/rest.js
```diff
@@ -104,8 +83,67 @@ function connect(uid) {
...
+                } else {
+                    manage_error(reject, { reason: headers.split('\r\n')[0] }, body);
```
cockpituous, line 110: This added line is not executed by any test.

martinpitt:
@jelly this still needs some refinement. But I'd appreciate a first look from you whether you consider this direction acceptable or completely crazy (alternative is to spend some time fixing this in cockpit proper, but it'll be years before it's available everywhere :cry: )

src/ContainerLogs.jsx
```diff
@@ -149,7 +139,18 @@ class ContainerLogs extends React.Component {
...
             // First 8 bytes encode information about stream and frame
             // See 'Stream format' on https://docs.docker.com/engine/api/v1.40/#operation/ContainerAttach
-            this.view.writeln(data.slice(8));
+            while (data.byteLength >= 8) {
+                // split into frames (size is the second 32-bit word)
+                const size = data[7] + data[6] * 0x100 + data[5] * 0x10000 + data[4] * 0x1000000;
+                const frame = data.slice(8, 8 + size);
+                // old podman versions just have CR endings, append NL then
+                if (frame[size - 1] === 13)
+                    this.view.writeln(frame);
```
cockpituous, line 148: This added line is not executed by any test.

src/app.jsx
cockpituous, line 497: This added line is not executed by any test.

src/ContainerLogs.jsx
martinpitt: Right, this only happens on ubuntu-22.04.

cockpituous, line 148: This added line is not executed by any test.

src/app.jsx
cockpituous, line 497: This added line is not executed by any test.

src/ContainerLogs.jsx
```diff
@@ -149,7 +149,18 @@ class ContainerLogs extends React.Component {
             }
             // First 8 bytes encode information about stream and frame
             // See 'Stream format' on https://docs.docker.com/engine/api/v1.40/#operation/ContainerAttach
-            this.view.writeln(data.slice(8));
+            while (data.byteLength >= 8) {
+                // split into frames (size is the second 32-bit word)
+                const size = data[7] + data[6] * 0x100 + data[5] * 0x10000 + data[4] * 0x1000000;
```
jelly, line 144: TIL, we also know if it is stdout/stderr but don't do anything with it!

src/rest.js
```diff
@@ -104,8 +84,67 @@ function connect(uid) {
         });
     };
 
+    connection.monitor = function(path, callback, return_raw = false) {
+        return new Promise((resolve, reject) => {
```
jelly, line 88: Just for reference and what I recently stumbled on [Promise.withResolvers()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers). You don't have to switch but it is nice if you want to de-indent code no real benefit here

PR #2039, opened 2025-03-07
test: Drop service failure hack from login() helper, robustify cleanupAfterService()

martinpitt:
Three little cleanups, fallout from my multi-user work. But these should be ready to land already.


jelly (approved):
Nice!

PR #2040, opened 2025-03-07
Prefix all unused variables with `_`

jelly:
Modern linters complain about this while eslint finds it acceptable.

### DIFF
--- a/src/ContainerLogs.jsx
+++ b/src/ContainerLogs.jsx
@@ -70,7 +70,7 @@ class ContainerLogs extends React.Component {
         this.connectStream();
     }
 
-    componentDidUpdate(prevProps, prevState) {
+    componentDidUpdate(prevProps) {
         // Connect channel when there is none and container started
         if (!this.state.streamer && this.props.containerStatus === "running" && prevProps.containerStatus !== "running")
             this.connectStream();
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -50,7 +50,7 @@ const ContainerActions = ({ con, container, onAddNotification, localImages, upda
     const isRunning = container.State.Status == "running";
     const isPaused = container.State.Status === "paused";
 
-    const deleteContainer = (event) => {
+    const deleteContainer = () => {
         if (container.State.Status == "running") {
             const handleForceRemoveContainer = () => {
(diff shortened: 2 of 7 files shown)


src/ContainerLogs.jsx
```diff
@@ -70,7 +70,7 @@ class ContainerLogs extends React.Component {
         this.connectStream();
     }
 
-    componentDidUpdate(prevProps, prevState) {
+    componentDidUpdate(prevProps, _prevState) {
```
martinpitt, line 73: I get why we have to do the `_` prefix on the start of function arguments. But on the tail end we can just drop them altogether.

src/Containers.jsx
```diff
@@ -50,7 +50,7 @@ const ContainerActions = ({ con, container, onAddNotification, localImages, upda
     const isRunning = container.State.Status == "running";
     const isPaused = container.State.Status === "paused";
 
-    const deleteContainer = (event) => {
+    const deleteContainer = (_event) => {
```
martinpitt, line 53: Similarly here, let's just make this `()`

martinpitt (changes requested):
Thanks! But let's do one more round while we are at it.

src/Containers.jsx
jelly: Cleaned up everywhere.

src/ImageRunModal.jsx
```diff
@@ -283,7 +283,7 @@ export class ImageRunModal extends React.Component {
 
         try {
             await client.imageExists(con, createConfig.image);
-        } catch (error) {
+        } catch {
```
martinpitt, line 286: Oh, is that valid syntax? TIL!

src/ImageRunModal.jsx
```diff
@@ -305,7 +305,7 @@ export class ImageRunModal extends React.Component {
 
             onDownloadContainer(tempImage);
 
-            client.pullImage(con, createConfig.image).then(reply => {
+            client.pullImage(con, createConfig.image).then(_reply => {
```
martinpitt, line 308: This can also become `()`

martinpitt (approved):
Thanks!

PR #2042, opened 2025-03-10
Fix "Run Image" dialog default owner

martinpitt:
The dialog's Owner: radio button should default to the owner of the image that was selected to run, not statically to "system". In most cases you don't want to run a system container from a user image.

This scenario is not covered anywhere yet, and difficult to put into `testRunImage{User,System}`, so create a new `testRunImageUserWithAuth` test case for this.

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -82,7 +82,9 @@ export class ImageRunModal extends React.Component {
 
         const default_owner = this.props.pod
             ? this.props.users.find(u => u.uid === this.props.pod.uid)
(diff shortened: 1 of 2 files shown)


jelly (dismissed):

martinpitt:
Argh, [regression](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2042-c87e85c1-20250310-142150-fedora-41/log.html#15), one `showImage()` conversion was overzealous -- this was *hiding* the images instead of showing them. Fixed.

jelly (approved):

PR #2043, opened 2025-03-11
Remove actual user name from owner help

martinpitt:
The explanation when a container should run as system or user isn't specific to the particular user name, so drop it.

This avoids confusion with the upcoming support for multiple users.

Note that we already have "User" as translatable string, so this does not actually add a new string.


martinpitt:
@jelly This was the part I backed out from https://github.com/cockpit-project/cockpit-podman/pull/1989#issuecomment-2671284371 -- it's time to reintroduce this now.

jelly (approved):

PR #2046, opened 2025-03-12
Replace "only running" Select with Switch

martinpitt:
We should not have a select with only two options. Replace the "Show" filter with an "Only running" Switch.

Adjust the React state and URL parameters accordingly to reflect that it is actually a boolean value.

Fix the previous `onContainerFilterChanged` (now renamed to `onContainerOnlyRunningChanged`) to not directly manipulate the state, but work on a copy.

----

Before:

running-before

Now:

image
image

@garrett We talked about this the other day. Can you please confirm that this is the direction you intended? Let's also discuss the string -- There are alternatives, like "Show stopped" (but that sounds odd in my ears) or longer text like "Only show running containers" / "Show stopped containers", but I'd prefer a short text (mobile/translations/etc.)

 Thanks!

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -5,9 +5,9 @@ import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { Card, CardBody, CardHeader, CardTitle } from "@patternfly/react-core/dist/esm/components/Card";
 import { Divider } from "@patternfly/react-core/dist/esm/components/Divider";
 import { DropdownItem } from '@patternfly/react-core/dist/esm/components/Dropdown/index.js';
-import { FormSelect, FormSelectOption } from "@patternfly/react-core/dist/esm/components/FormSelect";
 import { LabelGroup } from "@patternfly/react-core/dist/esm/components/Label";
 import { Popover } from "@patternfly/react-core/dist/esm/components/Popover";
+import { Switch } from "@patternfly/react-core/dist/esm/components/Switch";
 import { Text, TextVariants } from "@patternfly/react-core/dist/esm/components/Text";
 import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core/dist/esm/components/Toolbar";
 import { Tooltip } from "@patternfly/react-core/dist/esm/components/Tooltip";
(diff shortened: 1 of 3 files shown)


martinpitt:
[pixel failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2046-f9edaa44-20250312-064551-fedora-41/log.html) are as expected.

martinpitt:
I fixed the vertical alignment, screenshots updated.

garrett:
Thanks for implementing this! What we currently have is awkward and should change. Turning it into a toggle like this is definitely one option (pun intended :wink:) and it's nice to see in this PR.

I'd like to step back and ask:

1. Is there value in only displaying running containers? With labels and sorting, toggling running state (a holdover from the Docker page) might be a bit redundant.

2. Should we consider toggling anything else? Possible options include system, user roles, quadlets, pods, status variations (stopped, restarting, etc.). We should think about what’s actually useful, especially if this kind of thing can't be solved with a text filter or sorting.

I'm not sure whether a toggle to show only running containers is even worthwhile, especially given its prominent placement.

We definitely should avoid using a "form select" widget. What we should use depends on the answers to the questions above: nothing (if it's not really useful), a toggle (like here, if it is useful and there's nothing else useful), or some kind of menu, dropdown, or other select component (if there are more than one types of things to filter.

Some rhetorical questions to consider:

- Should we merge all filtering? (We have filtering at the top as well as on the containers. The split is due to the images.)
- We want to redesign the UI of the page into something like a Machines-style list+detail view instead of an expander table. That might also likely impact filtering. Should we implement changes now or wait for the redesign? (Images would be turned into a card instead of a list at the top, and the actual list of images would be a sub-page. This would let us change the filtering.)

Given all the above, should we defer this for later?

martinpitt:
> Is there value in only displaying running containers? 

I don't personaly run so many containers that I'd depend on it. But my gut feeling is that *stopped* containers are not a big use case. We even have a "prune unused containers" function to clean these up. So for my sake I'm also happy to completely drop this.

> Should we consider toggling anything else? Possible options include system, user roles

We have that, it's the "Owner" select on the top.

> quadlets, pods, 

quadlets maybe. pods are already separated.

> status variations (stopped, restarting, etc.).

filtering for transient states like "stopping" makes little sense IMHO. Persistent states like "stopped" may make more sense to find them more easily. But then we indeed may keep the Select and add "Stop" to it (there are no more sensible entries -- containers are running, or stopped. starting/stopping are transient). The story looks a bit different for quadlets, which have a "failed" state.

> Given all the above, should we defer this for later?

Sure! I don't have much emotional attachment to this. I just did that as you brought it up in your first review of #2024 (which is the one I really care about, and that doesn't otherwise change any React/PF other than more entries in the "Owner:" select).

martinpitt:
Discussed in chat, let's shelve this until we look at the PF6 migration and a bigger redesign.

PR #2049, opened 2025-03-13
test: re-use `getContainerAction` helper

jelly:
Should help when we eventually need to move to PF6 and update all the selectors


martinpitt (approved):
Thanks!

PR #2050, opened 2025-03-13
Fix pod label

martinpitt:
"pod group" makes no sense. A pod is a group of containers, and this is a single pod, not a group of pods.

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -859,7 +859,7 @@ class Containers extends React.Component {
                                                     <CardTitle>
                                                         <Flex justifyContent={{ default: 'justifyContentFlexStart' }}>
                                                             <h3 className='pod-name'>{caption}</h3>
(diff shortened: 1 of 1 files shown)


martinpitt:
@jelly actually, this doesn't even seem to work -- I can't see "pod group" or "pod" (that label) in the container list:

image

Is that just dead code? :thinking:

martinpitt:
Ah, no, it's right there: "web   pod", I'm just blind. I was looking for a badge like "toolbox".

martinpitt:
Pushing [expected pixel diff](https://cockpit-logs.us-east-1.linodeobjects.com/pull-2050-afb081ac-20250313-125800-fedora-41/pixeldiff.html).

tomasmatus (approved):
Fine by me. Unless you want to wait for design review.

PR #2061, opened 2025-03-25
test: generalise bootc check

jelly:
CentOS 9's bootc image no longer exposes `OSTREE_VERSION` in /etc/os-release. Likely due to Fedora CoreOS moving from ostree to composefs. As every image we test is moving to bootc the check whether cockpit-ws should be kept can be simplified to seeing if we booted into a bootc system.


martinpitt (approved):
Thanks!

PR #2062, opened 2025-03-25
test: Robustify admin user SSH key setup

martinpitt:
This makes the admin setup independent from root's, and is easier.

---

See https://github.com/cockpit-project/cockpit-podman/pull/2024/files#r2010660027

Our bots images are already set up that way, but [it is required on TF](https://artifacts.dev.testing-farm.io/f22e32ed-1fde-469f-93c6-2efdddefa4d6/)


jelly (approved):
\o/

PR #2065, opened 2025-03-27
Various fixes (buggy, AI review test)

martinpitt:

PR #2066, opened 2025-03-27
Various fixes (real)

martinpitt:
This is the honest version of #2065.

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -457,14 +457,13 @@ export class ImageRunModal extends React.Component {
                         let imageResults = [];
                         const manifestResult = handleManifestsQuery(reply[0]);
 
-                        for (let i = 1; i < reply.length; i++) {
-                            const result = reply[i];
+                        reply.slice(1).forEach(result => {
                             if (result.status === "fulfilled") {
                                 imageResults = imageResults.concat(JSON.parse(result.value));
                             } else if (!manifestResult && !imageExistsLocally(value, this.props.localImages)) {
                                 changeDialogError(result.reason);
                             }
-                        }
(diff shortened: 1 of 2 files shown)


martinpitt:
The sourcery comment is mixed-good: the ternary operation advice seems bogus to me. You can't efficiently mix a `delete` statement and a property addition in a single ternary statement, at least not in a tasteful way. Replacing the `for` loop is good advice, but not with `.map()`, but with `.forEach()`. I did that now.

src/ImageRunModal.jsx
```diff
@@ -768,17 +767,13 @@ export class ImageRunModal extends React.Component {
                     this.setState({ searchByRegistry: 'all' });
                 }}
                 // Ignore SelectToggle's touchstart's default behaviour
-                onTouchStart={ev => {
-                    ev.stopPropagation();
-                }}
+                onTouchStart={ev => ev.stopPropagation()}
```
cockpituous, line 770: This added line is not executed by any test.

src/ImageRunModal.jsx
```diff
@@ -768,17 +767,13 @@ export class ImageRunModal extends React.Component {
...
                 />
                 <ToggleGroupItem text={_("Local")} key='local' isSelected={this.state.searchByRegistry == 'local'} onChange={(ev, _) => {
                     ev.stopPropagation();
                     this.setState({ searchByRegistry: 'local' });
                 }}
-                onTouchStart={ev => {
-                    ev.stopPropagation();
-                }}
+                onTouchStart={ev => ev.stopPropagation()}
```
cockpituous, line 776: This added line is not executed by any test.

PR #2074, opened 2025-04-01
Almost fully type check application

jelly:


pyproject.toml
```diff
@@ -64,6 +64,7 @@ mark-parentheses = false
 known-first-party = ["cockpit"]
 
 [tool.vulture]
+min_confidence = 80
```
tomasmatus, line 67: should we set this across other projects as well?

tomasmatus (approved):
more types for a greater good

pyproject.toml
jelly: Good question, I shall ask @allisonkarlitskaya about it, but in general adding excludes here is quite annoying

allisonkarlitskaya: A lot of the lower-confidence vulture stuff is still high quality...

jelly: Well not for typed dicts....

```
class ContainerList(TypedDict):
    id: str
    name: str
    image: str
    command: str
    state: list[str] | str
```

Claims `id` / `state` was not used while it is, setting up a vulture suppressions file seems to be too much overhead to me. Alternatively setting this in pyproject.toml with `ignore_names = ["visit_*", "do_*"]` is too much of a :hammer: 

So the only ok solution for me is to bump the confidence level as this was just noise.

allisonkarlitskaya: in fairness, the TypedDict syntax is totally bonkers

see https://github.com/jendrikseipp/vulture/issues/249

PR #2082, opened 2025-04-03
General CSS related cleanups and a small bugfix

jelly:
Went through our CSS to figure out if there are overrides, to remove, document etc.

### DIFF
--- a/src/ContainerIntegration.jsx
+++ b/src/ContainerIntegration.jsx
@@ -12,7 +12,7 @@ const _ = cockpit.gettext;
 
 // ports is a mapping like { "5000/tcp": [{"HostIp": "", "HostPort": "6000"}] }
 export const renderContainerPublishedPorts = ports => {
-    if (!ports)
+    if (Object.keys(ports).length === 0)
         return null;
 
     const ranges = [];
--- a/src/Containers.scss
+++ b/src/Containers.scss
@@ -1,4 +1,4 @@
-@import "global-variables";
+@use "global-variables" as *; /* for $pf-v6-global--breakpoint--sm */
 
 .container-pod {
     .pf-v6-c-card__header {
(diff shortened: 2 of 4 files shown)


src/Containers.scss
```diff
@@ -1,4 +1,4 @@
-@import "global-variables";
+@import "global-variables"; // for $pf-v6-global--breakpoint--sm and .pf-v6-u-hidden-on-sm
```
Venefilyn, line 1: It would probably work with `@use` instead as `@import` is deprecated in Sass
```suggestion
@use "global-variables" as *; // for $pf-v6-global--breakpoint--sm and .pf-v6-u-hidden-on-sm
```

src/ContainerIntegration.jsx
```diff
@@ -12,7 +12,7 @@ const _ = cockpit.gettext;
 
 // ports is a mapping like { "5000/tcp": [{"HostIp": "", "HostPort": "6000"}] }
 export const renderContainerPublishedPorts = ports => {
-    if (!ports)
+    if (Object.keys(ports).length === 0)
         return null;
```
Venefilyn, line 16: Good catch. Keep this in a separate commit when consolidating the commits :D

Venefilyn (dismissed):
LGTM!

src/ContainerIntegration.jsx
jelly: We generally don't squash PR's and merge this as `rebase & merge`

src/Containers.scss
jelly: So hmm seems my comment is then wrong, this is not for .pf-v6-u-hidden-on-sm

Venefilyn: Right the `.pf-v6-u-hidden-on-sm` should work without this import, but this import might still give access to it. Try and see :D

jelly: Right, re-pushed with an updated comment.

Venefilyn (approved):

PR #2083, opened 2025-04-04
Drop ct-table-empty css rule

jelly:
When there are no rows we show the EmptyState component and the Table wouldn't be rendered so `ct-table-empty` is never shown.

### DIFF
--- a/src/Containers.scss
+++ b/src/Containers.scss
@@ -71,11 +71,6 @@
     }
 }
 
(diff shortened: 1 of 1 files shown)


jelly:
Send https://github.com/cockpit-project/cockpit-podman/pull/2088 for the pixel test flake

martinpitt (approved):

PR #2088, opened 2025-04-07
Update pixel tests

jelly:
The prune image dialog pixel tests didn't have both checkboxes checked which caused flakes in CI. The prune image dialog should always have both checkboxes checked by default.

See the failure https://cockpit-logs.us-east-1.linodeobjects.com/pull-2083-3f2bbde9-20250404-152013-fedora-41/log.html

### DIFF
--- a/src/PruneUnusedImagesModal.jsx
+++ b/src/PruneUnusedImagesModal.jsx
@@ -78,9 +78,9 @@ const PruneUnusedImagesModal = ({ close, unusedImages, onAddNotification, users 
 
     const showCheckboxes = unusedOwners.length > 1;
 
-    const onCheckChange = (user, checked) => setDeleteOwners(
-        checked ? deleteOwners.concat([user]) : deleteOwners.filter(u => u !== user)
-    );
+    const onCheckChange = (user, checked) => setDeleteOwners(prevState => {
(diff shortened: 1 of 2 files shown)


src/PruneUnusedImagesModal.jsx
```diff
@@ -103,7 +103,7 @@ const PruneUnusedImagesModal = ({ close, unusedImages, onAddNotification, users
                     <ImageOptions key={user.name}
                                   images={unusedImages.filter(image => image.uid === user.uid)}
                                   name={"deleteImages-" + user.name}
-                                  checked={deleteOwners.includes(user)}
+                                  checked={deleteOwners.some(u => u.uid === user.uid)}
```
martinpitt, line 106: Seems a bit odd -- the objects in `users` shouldn't change while the dialog is open. But it doesn't hurt indeed.

martinpitt (approved):
Dankjewel! I retried the two flakes, looked unrelated.

src/PruneUnusedImagesModal.jsx
jelly: I agree it's weird but it flakes reproducible in the test yet I can't easily reproduce it in a normal dev env.

PR #2089, opened 2025-04-08
Show systemd service links for quadlets

jelly:
Restart/Stop/Start support has been removed so users can't restart a container currently. With a link to the services page they can at least restart it without temporarily losing the container from the view.

---

# Show link to services page in container details

image

### DIFF
--- a/src/ContainerDetails.jsx
+++ b/src/ContainerDetails.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { DescriptionList, DescriptionListDescription, DescriptionListGroup, DescriptionListTerm } from "@patternfly/react-core/dist/esm/components/DescriptionList";
 import { Flex, FlexItem } from "@patternfly/react-core/dist/esm/layouts/Flex";
 
@@ -41,6 +42,17 @@ const ContainerDetails = ({ container }) => {
                     <DescriptionListGroup>
                         <DescriptionListTerm>{_("Command")}</DescriptionListTerm>
                         <DescriptionListDescription>{utils.quote_cmdline(container.Config.Cmd)}</DescriptionListDescription>
+                    </DescriptionListGroup>
+                    }
+                    {Boolean(container.Config?.Labels?.PODMAN_SYSTEMD_UNIT) && (container.uid === 0 || container.uid === null) &&
+                    <DescriptionListGroup>
+                        <DescriptionListTerm>{_("systemd service")}</DescriptionListTerm>
+                        <DescriptionListDescription>
+                            <Button variant="link" isInline onClick={
+                                () => cockpit.jump(`/system/services#/${container.Config?.Labels?.PODMAN_SYSTEMD_UNIT}` + (container.uid === null ? "?owner=user" : ""))}>
+                                {cockpit.format(_("View $0"), container.Config?.Labels?.PODMAN_SYSTEMD_UNIT)}
+                            </Button>
+                        </DescriptionListDescription>
                     </DescriptionListGroup>
                     }
                 </DescriptionList>
--- a/test/check-application
+++ b/test/check-application
@@ -3292,6 +3292,16 @@ PodName={name}
             # Lifecycle operations are not available such as Start/Restart
             b.wait_visible(f"#pod-{podName}-{podOwner}-action-toggle:disabled")
 
+        self.toggleExpandedContainer(service_name)
+
+        b.click(f"#containers-containers tbody tr:contains('{service_name}') button:contains({service_name}.service)")
+        b.switch_to_top()
+        if system:
+            b.wait_js_cond(f'window.location.hash === "#/{service_name}.service"')
+        else:
+            b.wait_js_cond(f'window.location.hash === "#/{service_name}.service?owner=user"')
+        b.wait_js_cond('window.location.pathname === "/system/services"')
+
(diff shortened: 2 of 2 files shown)


src/ContainerDetails.jsx
```diff
@@ -43,6 +44,17 @@ const ContainerDetails = ({ container }) => {
                         <DescriptionListDescription>{utils.quote_cmdline(container.Config.Cmd)}</DescriptionListDescription>
                     </DescriptionListGroup>
                     }
+                    {Boolean(container.Config?.Labels?.PODMAN_SYSTEMD_UNIT) && (container.uid === 0 || container.uid === null) &&
```
martinpitt, line 47: OOI, is that `Boolean()` wrapper for type checking? (we don't currently do that in podman)

src/ContainerDetails.jsx
```diff
@@ -43,6 +44,17 @@ const ContainerDetails = ({ container }) => {
...
+                    <DescriptionListGroup>
+                        <DescriptionListTerm>{_("Systemd service")}</DescriptionListTerm>
+                        <DescriptionListDescription>
+                            <Button variant="link" isInline onClick={
+                                () => cockpit.jump(`/system/services#/${container.Config?.Labels?.PODMAN_SYSTEMD_UNIT}` + (container.uid === null ? "?owner=user" : ""))}>
```
martinpitt, line 52: This was my main fear about the podman multi-user request: it'd soon percolate through all the other pages, like logs, services, terminal, etc. :thinking:

src/ContainerDetails.jsx
```diff
@@ -43,6 +44,17 @@ const ContainerDetails = ({ container }) => {
...
+                                {cockpit.format("View $0", container.Config?.Labels?.PODMAN_SYSTEMD_UNIT)}
```
martinpitt, line 53: String needs i18n. Although I'm actually in favor of just dropping the "View " and showing the unit name only.  There is already a label, and "View" doesn't quite capture it anyway -- you can interact/change the unit as well, and if this needs a verb at all, it sounds more like "Go to.."?

test/check-application
```diff
@@ -3292,6 +3292,17 @@ PodName={name}
             # Lifecycle operations are not available such as Start/Restart
             b.wait_visible(f"#pod-{podName}-{podOwner}-action-toggle:disabled")
 
+        self.toggleExpandedContainer(service_name)
+        testlib.sit()
```
martinpitt, line 3296: only until standup!

test/check-application
```diff
@@ -3434,6 +3445,13 @@ ContainerName=guitarq1
             b.wait_text("#containers-containers [data-label='Owner']", "user: guitar")
             b.wait_visible("#containers-containers .ct-badge-service")
 
+            # No systemd service link is shown for quadlets from other users as
+            # the services page doesn't support anything except the logged in/admin/root user.
+            self.toggleExpandedContainer("guitarq1")
+            sel = "#containers-containers tbody tr:contains('guitarq1')"
+            b.wait_in_text(f"{sel} + tr", "sleep infinity")
+            b.wait_not_present(f"{sel} button:contains('guitarq1.service')")
```
martinpitt, line 3453: It seems a little more robust to check for the absence of the "Systemd service", but matter of taste.

src/ContainerDetails.jsx
```diff
@@ -43,6 +44,17 @@ const ContainerDetails = ({ container }) => {
                         <DescriptionListDescription>{utils.quote_cmdline(container.Config.Cmd)}</DescriptionListDescription>
                     </DescriptionListGroup>
                     }
+                    {Boolean(container.Config?.Labels?.PODMAN_SYSTEMD_UNIT) && (container.uid === 0 || container.uid === null) &&
+                    <DescriptionListGroup>
+                        <DescriptionListTerm>{_("Systemd service")}</DescriptionListTerm>
```
martinpitt, line 49: It's "systemd", not "Systemd". https://systemd.io/ doesn't capitalize it either.

martinpitt (changes requested):
Thanks! I didn't look at the failures (presumably tmt just stumbles over the sit()). Some bikesheds, but mostly looks good!

src/ContainerDetails.jsx
jelly: Ah more a copy-pasta from `const isSystemdService = Boolean(container.Config?.Labels?.PODMAN_SYSTEMD_UNIT);` not type related.

src/ContainerDetails.jsx
jelly: Yes, realized that as well and that is why it is only visible for the "admin session" and "superuser".

src/ContainerDetails.jsx
jelly: :100: carelessly copied from the design.

src/ContainerDetails.jsx
martinpitt: It's fine, I was just curious.

src/ContainerDetails.jsx
garrett: I'm not against considering something else, but PF suggests "View" for this @ https://www.patternfly.org/ux-writing/terminology/

> View (v.)
> 	
> Use to describe accessing another page to see details.

martinpitt: Sure, no strong opinion. Let's just i18n the string.

martinpitt (approved):
Cheers!

PR #2092, opened 2025-04-10
Use the fully qualified registry name in state

SludgeGirl:
When using a registry like `registry.opensuse.org` it would trim off `registry.` for the ui, then later on when it came to actually filtering the results it would also only select `images[opensuse.org]` which would never exist

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -15,10 +15,10 @@ from machine.machine_core import ssh_connection
 
 REGISTRIES_CONF = """
 [registries.search]
-registries = ['localhost:5000', 'localhost:6000']
+registries = ['localhost:5000', 'localhost:6000', 'subdomain.local.localhost:7000']
 
 [registries.insecure]
-registries = ['localhost:80', 'localhost:5000', 'localhost:6000']
+registries = ['localhost:80', 'localhost:5000', 'localhost:6000', 'subdomain.local.localhost:7000']
 """
 
 NOT_RUNNING = ["Exited", "Stopped"]
@@ -273,6 +273,8 @@ class TestApplication(testlib.MachineCase):
         self.execute(True, f"""
             podman run -d -p 5000:5000 --name registry --stop-timeout 0 {IMG_REGISTRY}
             podman run -d -p 6000:5000 --name registry_alt --stop-timeout 0 {IMG_REGISTRY}
+            podman run -d -p 7000:5000 --name registry_subdomain --stop-timeout 0 {IMG_REGISTRY}
+            echo "127.0.0.1 subdomain.local.localhost" >> /etc/hosts
         """)
 
         # Add local insecure registry into registries conf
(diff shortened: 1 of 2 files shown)


jelly:
> When using a registry like `registry.opensuse.org` it would trim off `registry.` for the ui, then later on when it came to actually filtering the results it would also only select `images[opensuse.org]` which would never exist

Does this mean you configure "registry.opensuse.org" in unqualified-search-registries?

SludgeGirl:
> We will also want to add an integration test here so it doesn't regress

Gotcha I'll get one added now

SludgeGirl:
@jelly it's a little clunky so let me know if there's a better way I can handle it

test/check-application
```diff
@@ -1725,6 +1735,11 @@ PodName={name}
         b.wait_not_present(f'.pf-v6-c-menu__item:contains("{IMG_REGISTRY_LATEST}")')
         b.wait_visible(f'.pf-v6-c-menu__item:contains("{IMG_ALPINE_LATEST}")')
 
+        # Test filter with a subdomain.
+        b.set_input_text("#create-image-image input", "my-busybox", blur=False)
+        b.click('.pf-v6-c-toggle-group__text:contains("local.localhost")')
+        b.wait_text(".pf-v6-c-menu__item:not([disabled])", "subdomain.local.localhost:7000/my-busybox")
```
jelly, line 1741: I think we can use the existing registries here for filtering?

image


vs.

image

jelly: I manually tested the test without your fixes and it indeed does not filter on `localhost:6000` or `5000` so that should be good enough.

jelly (changes requested):
Thanks, some small test fixes and then this looks good :+1:

test/check-application
SludgeGirl: I've checked myself against main and it does filter correctly on both of those

image

The only changes we make to the domain which causes this problem is:

https://github.com/cockpit-project/cockpit-podman/blob/28483f585b06634cb186a4cb453c6e0f409a729e/src/ImageRunModal.jsx#L625-L632

So if we just ran the checks against `localhost:6000` or `:5000` we wouldn't really be testing for a regression, it needs to have a subdomain

jelly: I would recheck but adding new extra test coverage doesn't hurt and would be nice! So please keep that in and drop the empty `testBasicFilters` test

https://cockpit-logs.us-east-1.linodeobjects.com/pull-2092-79bbb487-20250415-085646-fedora-41/log.html

jelly (dismissed):

jelly:
And bah ubuntu-2204 has a broken podman release, search returns:

```
Index:  "local.localhost:7000"
Name:  "subdomain.local.localhost:7000/my-busybox"
```

So it just drops off the registry name for no reason. There is no way to guess what is right and I don't want to hold up merging this.

jelly (approved):

PR #2097, opened 2025-04-15
pyproject: set pyright extra import paths

jelly:
Our custom test shebang is not compatible with pyright so we have to like mypy teach it about bots/ our custom testlib.


tomasmatus:
you can ignore the pixeltest fails, new reference was updated with the new container

tomasmatus (approved):

PR #2102, opened 2025-04-22
test: fully type check-application

jelly:
This types the last untyped function `execute` which is a bit of a hassle as it has a first boolean argument which should be the last due to our boolean trap rules.

---

A bit of a yak shave but it should work (tm)


jelly:
centos seems stuck so merging.

PR #2104, opened 2025-04-23
test: fix FBT003 cases in tests

jelly:
Now that we are fully typed since b135d5b4a6ab3cf1ef, fix the last ruff ignore.


martinpitt (approved):
Wow, no ignores any more? :+1:

Issue #2055, opened 2025-03-17
quadlet container service integration


garrett:
We'll eventually want to make sure container + service integration works well, with stopping/starting/restarting containers uses systemd in the background.

For now, one of the easier integration points we can do now is to cross-link containers and services. These are designed to be as simple and unobtrusive as possible, yet provide necessary links between the container and service. 

We'll probably need to adjust this a little when we're eventually working on the redesign for the Podman page (with the list and details pages).

## From container to the service page:

Note that the link is on its own line, just like the image string should be.

![Image](https://github.com/user-attachments/assets/68bfc0a2-ca36-400f-9d6d-0add7f8d0637)

## From the container log to the full service log

This would link to the same full page logs as if you click on all logs from the container page. It's just a simple link under the log widget.

![Image](https://github.com/user-attachments/assets/83243e46-7ce5-4365-a39f-62ea702dac8b)

## From the service page to container:

This would require changes in Cockpit's Services page

![Image](https://github.com/user-attachments/assets/a7a044ba-5ae6-4927-9a90-566d8bc31206)

garrett:
By the way, I did consider other ways to link these pages. But, the containers page already has the key info (start/stop, logs, resources, uptime) or will have it soon. Linking from labels, menus, or adding top-level links (in the top level of the list) felt too clunky or inconsistent. So, I went with the simple link in the details, as it is simple, consistent, and doesn't stand out too much.

jelly:
I've implemented the systemd service link but we are missing a link to the pod service as there is no design for it.

![Image](https://github.com/user-attachments/assets/8c28fbb6-78f5-4b3e-a9f9-129499692d90)

jelly:
Another small change the design needs is to not capitalize "Systemd", see https://systemd.io/

PR #2120, opened 2025-05-01
Show link to logs page for quadlets

jelly:
Quadlets are services managed by systemd and detailed logs including service restarts can only be seen in journald's logs. This functionality is now limited to only system services as Cockpit's log page has no `--user` support.

Related: #2055

We should investigate supporting user journal support in Cockpit, of course we cannot support the other (non-logged in) users.

---

# Link to logs page for quadlets

image

### DIFF
--- a/src/ContainerLogs.jsx
+++ b/src/ContainerLogs.jsx
@@ -19,6 +19,7 @@
 
 import React from 'react';
 
+import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { ExclamationCircleIcon } from '@patternfly/react-icons';
 import { CanvasAddon } from '@xterm/addon-canvas';
 import { Terminal } from "@xterm/xterm";
@@ -165,8 +166,21 @@ class ContainerLogs extends React.Component {
 
     render() {
         let element = <div className="container-logs" ref={this.logRef} />;
-        if (this.state.errorMessage)
+        const { systemd_unit, uid } = this.props;
+
+        if (this.state.errorMessage) {
             element = <EmptyStatePanel icon={ExclamationCircleIcon} title={this.state.errorMessage} />;
+        } else if (uid === 0 && systemd_unit) {
+            element = (
+                <>
+                    {element}
+                    <Button variant="link" isInline className="pf-v6-u-mt-sm" onClick={
+                        () => cockpit.jump(`/system/logs/#/?priority=info&_SYSTEMD_UNIT=${systemd_unit}`)}>
+                        {cockpit.format(_("View $0 logs"), systemd_unit)}
+                    </Button>
+                </>
+            );
+        }
 
         return element;
     }
@@ -175,7 +189,8 @@ class ContainerLogs extends React.Component {
 ContainerLogs.propTypes = {
     containerId: PropTypes.string.isRequired,
     uid: PropTypes.number,
-    width: PropTypes.number.isRequired
+    width: PropTypes.number.isRequired,
+    systemd_unit: PropTypes.string
 };
 
 export default ContainerLogs;
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -452,7 +452,13 @@ class Containers extends React.Component {
                 tabs.push({
                     name: _("Logs"),
                     renderer: ContainerLogs,
-                    data: { containerId: container.Id, containerStatus: container.State.Status, width: this.state.width, uid: container.uid }
+                    data: {
+                        containerId: container.Id,
+                        containerStatus: container.State.Status,
+                        width: this.state.width,
+                        uid: container.uid,
+                        systemd_unit: container.Config?.Labels?.PODMAN_SYSTEMD_UNIT,
+                    }
                 });
                 tabs.push({
                     name: _("Console"),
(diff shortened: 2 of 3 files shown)


src/ContainerLogs.jsx
```diff
@@ -165,8 +166,21 @@ class ContainerLogs extends React.Component {
 
     render() {
         let element = <div className="container-logs" ref={this.logRef} />;
-        if (this.state.errorMessage)
+        const { systemd_unit, uid } = this.props;
+
+        if (this.state.errorMessage) {
             element = <EmptyStatePanel icon={ExclamationCircleIcon} title={this.state.errorMessage} />;
+        } else if (systemd_unit !== undefined && uid === 0) {
```
cockpituous, line 173: This added line is not executed by any test.

cockpituous, line 173: This added line is not executed by any test.

cockpituous, line 173: This added line is not executed by any test.

jelly:
> > error: TypeError: Cannot read properties of undefined (reading 'includes')

src/ContainerLogs.jsx
cockpituous, line 173: This added line is not executed by any test.

src/index.js
```diff
@@ -29,4 +29,8 @@ import './podman.scss';
 document.addEventListener("DOMContentLoaded", function () {
     const root = createRoot(document.getElementById('app'));
     root.render(<Application />);
+    window.onerror = function(message, source, lineno, colno, error) {
+        console.log('Global error caught:', message, source, lineno, colno, error);
```
cockpituous, line 33: These 2 added lines are not executed by any test.

jelly:
The /devel scenario crashes in Cockpit, so we need to naughty this or fix it in Cockpit, wait till a new release and updated image which can take 3 weeks.

src/ContainerLogs.jsx
cockpituous, line 173: This added line is not executed by any test.

cockpituous, line 173: This added line is not executed by any test.

src/index.js
cockpituous, line 33: These 2 added lines are not executed by any test.

src/ContainerLogs.jsx
cockpituous, line 173: This added line is not executed by any test.

src/index.js
cockpituous, line 33: These 2 added lines are not executed by any test.

src/ContainerLogs.jsx
cockpituous, line 173: This added line is not executed by any test.

src/index.js
cockpituous, line 33: These 2 added lines are not executed by any test.

cockpituous, line 33: These 2 added lines are not executed by any test.

src/ContainerLogs.jsx
cockpituous, line 173: This added line is not executed by any test.

src/index.js
cockpituous, line 33: These 2 added lines are not executed by any test.

src/ContainerLogs.jsx
```diff
@@ -165,8 +166,21 @@ class ContainerLogs extends React.Component {
...
+            element = (
+                <>
+                    {element}
+                    <Button variant="link" isInline className="pf-v6-u-mt-sm" onClick={
+                        () => cockpit.jump(`/system/logs/#/?priority=info&_SYSTEMD_UNIT=${systemd_unit}`)}>
+                        {cockpit.format(_("View $0 logs"), systemd_unit)}
+                    </Button>
```
Venefilyn, line 180: Can we make this a link with a `href` instead? It's not great when buttons do navigations when there is no indication of it, and we can change the PF Button to be a link with `href`

src/ContainerLogs.jsx
Venefilyn, line 173: Not sure if it matters with JS but `uid===0` seems like it would be faster as it's a number comparison vs `systemd_unit` being a string comparison, so changing the order of those:
```suggestion
        } else if (uid === 0 && systemd_unit !== undefined) {
```

And can `systemd_unit` be empty string and be acceptable? Otherwise I'd shorten the suggestion even more 
```suggestion
        } else if (uid === 0 && systemd_unit) {
```

src/ContainerLogs.jsx
```diff
@@ -165,8 +166,21 @@ class ContainerLogs extends React.Component {
 
     render() {
         let element = <div className="container-logs" ref={this.logRef} />;
-        if (this.state.errorMessage)
+        const { systemd_unit, uid } = this.props;
+
+        if (this.state.errorMessage) {
             element = <EmptyStatePanel icon={ExclamationCircleIcon} title={this.state.errorMessage} />;
+        } else if (uid === 0 && systemd_unit) {
+            element = (
```
cockpituous, line 174: These 2 added lines are not executed by any test.

src/ContainerLogs.jsx
Venefilyn: Ack. Fine for now then. It is something I'd love for us to fix in the future though

PR #2137, opened 2025-05-22
test: allow testing against a Cockpit Packit copr

jelly:
We can't currently test a Podman PR against a Cockpit PR easily, but as every Cockpit PR creates a Packit rpm build with a copr repository we can use this.

This is for example useful when debugging why clicking on a logs link from Podman throws an exception in the Cockpit logs page.

---

Been using a hack in here https://github.com/cockpit-project/cockpit-podman/pull/2120 but we can easily generalise this.

### DIFF
--- a/test/run
+++ b/test/run
@@ -9,6 +9,11 @@ export RUN_TESTS_OPTIONS=--track-naughties
 TEST_SCENARIO=${TEST_SCENARIO:-}
 [ "${TEST_SCENARIO}" = "${TEST_SCENARIO##firefox}" ] || export TEST_BROWSER=firefox
 
+if [[ "${TEST_SCENARIO}" =~ "copr" ]]; then
+    export TEST_COPR="${TEST_SCENARIO#*copr/}"
+    TEST_SCENARIO="${TEST_SCENARIO%%/copr*}"
+fi
+
 if [ "$TEST_SCENARIO" == "devel" ]; then
     export TEST_COVERAGE=yes
 fi
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -19,4 +19,4 @@ jobs:
           mkdir -p ~/.config/cockpit-dev
           echo "${{ github.token }}" >> ~/.config/cockpit-dev/github-token
           TEST_OS=$(PYTHONPATH=bots python3 -c 'from lib.constants import TEST_OS_DEFAULT; print(TEST_OS_DEFAULT)')
-          bots/tests-trigger --force "-" "${TEST_OS}/updates-testing" "${TEST_OS}/podman-next"
+          bots/tests-trigger --force "-" "${TEST_OS}/updates-testing" "${TEST_OS}/copr/rhcontainerbot/podman-next"
--- a/HACKING.md
+++ b/HACKING.md
@@ -105,3 +105,11 @@ CI machinery do this on a draft pull request.
 Please see [Cockpit's test documentation](https://github.com/cockpit-project/cockpit/blob/main/test/README.md)
 for details how to run against existing VMs, interactive browser window,
 interacting with the test VM, and more.
+
+# Running tests against a Cockpit pull request
+
+The Cockpit testing infrastructure does not support testing a Cockpit pull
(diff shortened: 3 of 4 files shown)


test/run
```diff
@@ -13,6 +13,16 @@ if [ "$TEST_SCENARIO" == "devel" ]; then
     export TEST_COVERAGE=yes
 fi
 
+if [ "$TEST_SCENARIO" == "podman-next" ]; then
+    export TEST_COPR="rhcontainerbot/podman-next"
+fi
+
+TEST_COPR=${TEST_COPR:-}
```
martinpitt, line 20: This looks odd and unnecessary. It's better to not set the variable at all than to an empty value.

test/run
```diff
@@ -13,6 +13,16 @@ if [ "$TEST_SCENARIO" == "devel" ]; then
...
+if [[ "$TEST_SCENARIO" == *"packit"* ]]; then
```
martinpitt, line 21: The quoting here is odd -- the important bit to quote is the `*`, not the "packit" string, no? I.e. should this be `"*packit*"`? (I'm not familiar with these bash specific conditions)

martinpitt (commented):
Thanks, this is nice!

test/run
jelly: bash is weird.. but mm maybe we can do [better](https://stackoverflow.com/a/52671757) ie.

"${TEST_SCENARIO}" =~ "packit"

jelly:
@martinpitt I would like to document this as well

test/run
martinpitt: Ah, if that works, sure. I only know `| grep -q packit` or `[ "${TEST_SCENARIO%packit*}" != "$TEST_SCENARIO" ]`, but I fully admit that this is horrible to read -- the `=~` looks nice!

test/run
```diff
@@ -13,6 +13,15 @@ if [ "$TEST_SCENARIO" == "devel" ]; then
     export TEST_COVERAGE=yes
 fi
 
+if [ "$TEST_SCENARIO" == "podman-next" ]; then
+    export TEST_COPR="rhcontainerbot/podman-next"
+fi
```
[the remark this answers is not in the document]
jelly: Totally fine by me!

test/run
```diff
@@ -13,6 +13,15 @@ if [ "$TEST_SCENARIO" == "devel" ]; then
...
+
+if [[ "${TEST_SCENARIO}" =~ "packit" ]]; then
```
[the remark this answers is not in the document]
jelly: ^_^ yes! We do need some sort of identifier imo which can be `copr` imo!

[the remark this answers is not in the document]
jelly: > However, one more nitpick: This really has nothign to do with packit. It should be "=~ copr", or perhaps more clearly ensure that the string starts with "copr", i.e. == packit* or similar.

Right but one downside is that err we can't also test devel then. Which actually doesn't work in this PR unless I change things around.

martinpitt (approved):
:heart: thanks!

test/run
```diff
@@ -9,6 +9,11 @@ export RUN_TESTS_OPTIONS=--track-naughties
 TEST_SCENARIO=${TEST_SCENARIO:-}
 [ "${TEST_SCENARIO}" = "${TEST_SCENARIO##firefox}" ] || export TEST_BROWSER=firefox
 
+if [[ "${TEST_SCENARIO}" =~ "copr" ]]; then
```
martinpitt, line 12: Ah wait, can this be anchored to the string start? Like this it's a substring match, no?

martinpitt (changes requested):

test/run
jelly: It is a substring match, but if we make it a start string match then well /devel/copr won't work anymore

Now it is made to have that work.

martinpitt: Yeah, ok -- in cockpit we split scenarios with dashes actually, i.e. "storage-devel" or so. Not a biggie, if we ever get a scenario which as "copr" as a substring, we can still fix this :wink:

martinpitt (approved):

Issue #2112, opened 2025-04-27
No longer able to stop/kill individual containers in podman page


TheJman494:
Cockpit version: 337-1
Cockpit-podman version: 104-1
Podman version: 5.4.2-2
OS: Arch

### Issue description
In the Podman interface, I can no longer stop or kill individual containers, when they are not running I only have start/commit, when they are running I have checkpoint/commit.

Containers that are within a pod are behaving normally.

### Please help us out by explaining how to reproduce it.
Update to the latest version on Arch.

### Relevant parts of the system log are also useful:
Logs show no problems.

### In case the issue is clearly visible, screenshots are very helpful
![Image](https://github.com/user-attachments/assets/e0687267-809e-4518-a16b-457dec2476c3)
![Image](https://github.com/user-attachments/assets/ccc31c1d-f6b0-44ee-b3c5-04d1d2f77655)

jelly:
Are these containers based on qaulets or how did you create them?

TheJman494:
Podman-Compose, prior to the re-design this was not an issue no matter how they where created for me.

jelly:
Then it seems that podman-compose also sets PODMAN_SYSTEMD_UNIT which seems strange. Can you post the output of:

`podman inspect --format '{{.Config.Env}}'  $containername` and be careful not to print any sensitive data.

TheJman494:
Plex Container, not in a pod
docker.io/plexinc/pms-docker:latest
```
[LC_ALL=C.UTF-8 CHANGE_CONFIG_DIR_OWNERSHIP=true TERM=xterm PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin container=podman PLEX_CLAIM=XXX HOME=/config LANG=C.UTF-8 TZ=America/Los_Angeles HOSTNAME=XXX]
```

Immich Container, in a pod, `Immich-AIO`, `Immich-Postgres`, `Immich-Redis`
ghcr.io/imagegenius/immich:latest
```
[...]

PR #2142, opened 2025-05-27
Don't flag podman-compose containers as a service container

jelly:
A podman-compose container cannot be managed via systemd but still sets a label PODMAN_SYSTEMD_UNIT with `podman-compose@$project`.

Restarting such a unit isn't even possible with systemd:

systemctl restart podman-compose@root.service
Failed to restart podman-compose@root.service: Unit podman-compose@root.service not found

Closes: #2112

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -169,3 +169,7 @@ export const validationClear = (validationFailed, key, onValidationChange) => {
 
 // This method needs to be outside of component as re-render would create a new instance of debounce
 export const validationDebounce = debounce(500, (validationHandler) => validationHandler());
+
+// Ignore podman-compose containers which like quadlets set PODMAN_SYSTEMD_UNIT.
+// https://github.com/containers/podman-compose/blob/0dcc864fdda280b410ad49ae4fa99740a4770cbb/podman_compose.py#L2263
+export const is_systemd_service = (container_config) => container_config?.Labels?.PODMAN_SYSTEMD_UNIT && !container_config.Labels.PODMAN_SYSTEMD_UNIT.startsWith('podman-compose@');
--- a/src/ContainerDetails.jsx
+++ b/src/ContainerDetails.jsx
@@ -44,7 +44,7 @@ const ContainerDetails = ({ container }) => {
                         <DescriptionListDescription>{utils.quote_cmdline(container.Config.Cmd)}</DescriptionListDescription>
                     </DescriptionListGroup>
                     }
-                    {Boolean(container.Config?.Labels?.PODMAN_SYSTEMD_UNIT) && (container.uid === 0 || container.uid === null) &&
+                    {utils.is_systemd_service(container.Config) && (container.uid === 0 || container.uid === null) &&
                     <DescriptionListGroup>
                         <DescriptionListTerm>{_("systemd service")}</DescriptionListTerm>
                         <DescriptionListDescription>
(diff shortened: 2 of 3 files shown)


tomasmatus (commented):
Looks good to me but maybe it would be nice to cover this by a short test?

src/util.js
```diff
@@ -169,3 +169,7 @@ export const validationClear = (validationFailed, key, onValidationChange) => {
 
 // This method needs to be outside of component as re-render would create a new instance of debounce
 export const validationDebounce = debounce(500, (validationHandler) => validationHandler());
+
+// Ignore podman-compose containers which like quadlets set PODMAN_SYSTEMD_UNIT.
+// https://github.com/containers/podman-compose/blob/0dcc864fdda280b410ad49ae4fa99740a4770cbb/podman_compose.py#L2263
+export const is_systemd_service = (container_config) => container_config?.Labels?.PODMAN_SYSTEMD_UNIT && !container_config.Labels.PODMAN_SYSTEMD_UNIT.startsWith('podman-compose@');
```
Venefilyn, line 175: Just double-checking this in [Podman matrix](https://matrix.to/#/#podman:fedoraproject.org) as I want to avoid checking for something else to determine if it's a `systemd` service if possible. We shouldn't _need_ to rely on `podman-compose` existing or not to determine if it's a quadlet

jelly: So for a reproducer:

```
version: '3.7'
services:
  web:
    image: docker.io/library/nginx:alpine
    ports:
      - "8000:80"
```

`podman-compose up -d` which creates:
```
               "CreateCommand": [
                    "podman",
                    "create",
                    "--name=admin_web_1",
                    "--pod=pod_admin",
                    "--label",
                    "io.podman.compose.config-hash=37791419cd18b7bf02a81b7f96d2b9eca0ea1f0e600055b6bb2352ff98c771dc",
                    "--label",
                    "io.podman.compose.project=admin",
                    "--label",
                    "io.podman.compose.version=1.4.0",
                    "--label",
                    "PODMAN_SYSTEMD_UNIT=podman-compose@admin.service",
                    "--label",
                    "com.docker.compose.project=admin",
                    "--label",
                    "com.docker.compose.project.working_dir=/home/admin",
                    "--label",
                    "com.docker.compose.project.config_files=docker-compose.yml",
                    "--label",
                    "com.docker.compose.container-number=1",
                    "--label",
                    "com.docker.compose.service=web",
                    "--network=admin_default:alias=web",
                    "-p",
                    "8000:80",
                    "docker.io/library/nginx:alpine"
               ],
```

But this doesnt' create a systemd service, to activate that you need:

```
[admin@fedora-42-127-0-0-2-2201 ~]$ podman-compose systemd --action create-unit
# /etc/systemd/user/podman-compose@.service

[Unit]
Description=%i rootless pod (podman-compose)

[Service]
Type=simple
EnvironmentFile=%h/.config/containers/compose/projects/%i.env
ExecStartPre=-/usr/bin/podman-compose up --no-start
ExecStartPre=/usr/bin/podman pod start pod_%i
ExecStart=/usr/bin/podman-compose wait
ExecStop=/usr/bin/podman pod stop pod_%i

[Install]
WantedBy=default.target

WARNING:podman_compose:Could not write to [/etc/systemd/user/podman-compose@.service], use 'sudo'
```

And save that to well ~/ or /etc and then run `podman-compose systemd --action register`. Not sure how well supported or used it is

jelly: The issue is that a `podman-compose` can become a full systemd service so it is kinda tricky..

As podman-compose has no man page...
```
command:
  {help,version,wait,systemd,pull,push,build,up,down,ps,run,exec,start,stop,restart,logs,config,port,pause,unpause,kill,stats,images}
    help                show help
    version             show version
    wait                wait running containers to stop
    systemd             create systemd unit file and register its compose stacks
```

Venefilyn:
> Looks good to me but maybe it would be nice to cover this by a short test?

yeah would be good to have a test to check that it doesn't show when you have podman-compose, as it wouldn't work

PR #2151, opened 2025-06-03
workflows: add a codespell workflow for detecting typos

jelly:


martinpitt (approved):
Cool, we dind't do ayn tpyos in this projetc?

PR #2153, opened 2025-06-04
Support filtering on systemd units

jelly:
Extend the current generic filtering field to also support filtering the systemd service name of a quadlet so we can create links from the services page to cockpit-podman, for example `?name=jellyfin.service`

### DIFF
--- a/src/Containers.jsx
+++ b/src/Containers.jsx
@@ -590,6 +590,21 @@ class Containers extends React.Component {
         this.setState({ showPruneUnusedContainersModal: true });
     };
 
+    filterContainersByText = (lcf, id) => {
+        const container = this.props.containers[id];
+        const systemd_unit_match = container.Config?.Labels?.PODMAN_SYSTEMD_UNIT?.toLowerCase().indexOf(lcf) >= 0;
+        const name_match = container.Name.toLowerCase().indexOf(lcf) >= 0;
+        const image_match = container.ImageName.toLowerCase().indexOf(lcf) >= 0;
+
+        if (container.Pod) {
+            const pod = this.props.pods[utils.makeKey(container.uid, container.Pod)];
+            const pod_match = pod.Name.toLowerCase().indexOf(lcf) >= 0 || pod.Labels?.PODMAN_SYSTEMD_UNIT?.toLowerCase().indexOf(lcf) >= 0;
+            return name_match || systemd_unit_match || image_match || pod_match;
+        } else {
+            return name_match || systemd_unit_match || image_match;
+        }
+    };
+
     render() {
         const Dialogs = this.context;
         const columnTitles = [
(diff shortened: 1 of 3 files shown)


src/Containers.jsx
```diff
@@ -627,8 +627,10 @@ class Containers extends React.Component {
             if (this.props.textFilter.length > 0) {
                 const lcf = this.props.textFilter.toLowerCase();
                 filtered = filtered.filter(id => this.props.containers[id].Name.toLowerCase().indexOf(lcf) >= 0 ||
+                    this.props.containers[id].Config?.Labels?.PODMAN_SYSTEMD_UNIT.toLowerCase().indexOf(lcf) >= 0 ||
                     (this.props.containers[id].Pod &&
-                     this.props.pods[utils.makeKey(this.props.containers[id].uid, this.props.containers[id].Pod)].Name.toLowerCase().indexOf(lcf) >= 0) ||
+                     (this.props.pods[utils.makeKey(this.props.containers[id].uid, this.props.containers[id].Pod)].Name.toLowerCase().indexOf(lcf) >= 0 ||
+                     this.props.pods[utils.makeKey(this.props.containers[id].uid, this.props.containers[id].Pod)].Labels?.PODMAN_SYSTEMD_UNIT?.toLowerCase().indexOf(lcf) >= 0)) ||
```
tomasmatus: these are quite long lines, I think you should put the `this.props.pods[utils.makeKey(this.props.containers[id].uid, this.props.containers[id].Pod)]` into a variable as it is same for both

tomasmatus (changes requested):
Looks good, I only have a small nitpick

src/Containers.jsx
jelly: Uhhh, sure but that means expanding the filter :) We can't put it in `lcf` as it is in a loop

src/Containers.jsx
```diff
@@ -590,6 +590,21 @@ class Containers extends React.Component {
         this.setState({ showPruneUnusedContainersModal: true });
     };
 
+    filterContainersByText = (lcf, id) => {
+        const container = this.props.containers[id];
+        const systemd_unit_match = container.Config?.Labels?.PODMAN_SYSTEMD_UNIT?.toLowerCase().indexOf(lcf) >= 0;
+        const name_match = container.Name.toLowerCase().indexOf(lcf) >= 0;
+        const image_match = container.ImageName.toLowerCase().indexOf(lcf) >= 0;
+
+        if (this.props.containers[id].Pod) {
```
tomasmatus:
```suggestion
        if (container.Pod) {
```

src/Containers.jsx
```diff
@@ -590,6 +590,21 @@ class Containers extends React.Component {
...
+            const pod = this.props.pods[utils.makeKey(this.props.containers[id].uid, this.props.containers[id].Pod)];
```
tomasmatus:
```suggestion
            const pod = this.props.pods[utils.makeKey(container.uid, container.Pod)];
```

tomasmatus (changes requested):
This should be my last request, looks much better now. Thanks!

jelly:
> This should be my last request, looks much better now. Thanks!

Hehe yeah and rightfully so. Thanks!

jelly:
Re-trying ubuntu-2404.

mvollmer (dismissed):
Thanks!

mvollmer (approved):
Thanks!

PR #2172, opened 2025-06-30
Various test fixes

martinpitt:
Broken out from #2170

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -2003,6 +2003,8 @@ PodName={podName}
 
         # with copy&paste you can also set multiple variables
         b.click('.env-form .btn-add')
+        b.focus('#run-image-dialog-env-4-value')
+        b.wait_visible('#run-image-dialog-env-4-value:focus')
         b.set_val('#run-image-dialog-env-4-value',
                   "DURIAN=LEMON TEST_URL=wss://cockpit/?start=1&stop=0")
         # set_val does not trigger onChange so append a space.
--- a/test/check-application
+++ b/test/check-application
@@ -1751,9 +1751,13 @@ PodName={podName}
 
         b.click('button.pf-v6-c-toggle-group__button:contains("Local")')
         b.wait_text(".pf-v6-c-menu__item[disabled]", "No images found")
+        b.wait_visible("button.pf-v6-c-toggle-group__button.pf-m-selected:contains('Local')")
 
         # Local results found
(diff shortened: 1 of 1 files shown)


martinpitt:
OK, "test: Fix testRunImage* copy&paste check" needs to stay in #2170, as with the event synthetization, clicking on the button does not actually focus it. Grr...

martinpitt:
OK, so commit 5d4192cd078e1651dad6b283b4c94c81311b1eb1 works well enough, nice! I still want to address this "DURIAN" race, so re-trying with a backwards compatible approach.

test/check-application
```diff
@@ -1751,9 +1751,13 @@ PodName={podName}
 
         b.click('button.pf-v6-c-toggle-group__button:contains("Local")')
         b.wait_text(".pf-v6-c-menu__item[disabled]", "No images found")
+        b.wait_visible("button.pf-v6-c-toggle-group__button.pf-m-selected:contains('Local')")
 
         # Local results found
         b.set_input_text("#create-image-image input", "registry")
+        b.wait_not_in_text(".pf-v6-c-menu__content", "Local images")
+        # HACK: Wait for "searching in progress" spinner to go away; this does not reflect in the DOM
```
jelly, line 1759: That's this one no?

image

So there is some HTML soup to check....
```
<li class="pf-v6-c-menu__list-item pf-m-disabled" role="none"><button tabindex="-1" class="pf-v6-c-menu__item" disabled="" role="option" aria-selected="false" type="button"><span class="pf-v6-c-menu__item-main"><span class="pf-v6-c-menu__item-text"><div class="pf-v6-l-bullseye"><svg class="pf-v6-c-spinner pf-m-lg" role="progressbar" aria-valuetext="Loading..." viewBox="0 0 100 100" aria-label="Contents"><circle class="pf-v6-c-spinner__path" cx="50" cy="50" r="45" fill="none"></circle></svg></div></span></span></button></li>
```

So theoretically..

```
b.wait_not_present(".pf-v6-c-menu__content svg[aria-valuetext='Loading...']")
```

martinpitt: Ack, thanks! This looks promising. I'll queue that for #2171 or #2170 then, and first see if I can actually catch the spinner with a positive wait. I didn't see any structural change in the inspector, but I may have missed it.

martinpitt: Works, thanks! Pushed to #2171

PR #2175, opened 2025-07-02
test: Fix testCreateContainerValidation for real mouse clicks

martinpitt:
Don't click too fast, wait until the "Add env" button click takes effect. Otherwise it ends up as a triple click on the button.


martinpitt:
I never got back to this, and won't in the forseeable future.

PR #2181, opened 2025-07-07
test: Use fake clicks for the Run Image "Local" button with Firefox

martinpitt:
The "Local button click goes to Nirvana in CI" happens with Firefox as
well. So upgrade the `chromium_fake_mouse` to "specify coordinates"
which activates fake clicks on all browsers.

This is being debugged in #2175, until then avoid the flakes.


martinpitt:
[copr failed](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-podman-2181/srpm-builds/09244902/builder-live.log.gz) due to nodejs being uninstallable in Fedora 42, so the source rpm build failed.

martinpitt:
@jelly Yes -- at least this narrows the scope of the hack to a single click to the Local button, which helps a bit with debugging.

PR #2182, opened 2025-07-08
filter out *-service containers in container list

tomasmatus:


test/check-application
```diff
@@ -2996,6 +3038,15 @@ PodName={name}
             self.execute(f"podman rm -f container{i}", system=True)
 
     def testSpecialContainers(self) -> None:
+        def find_service_container() -> str | None:
+            for container in self.execute("podman ps --all --no-trunc", system=True).strip().split("\n")[1:]:
+                # <sha> <other things> <name>
+                items = container.split()
+                if items[-1].endswith("-service"):
+                    return items[-1]
```
jelly: So the services container is like the the `pause` container, isn't there a better way to find this?

at least `podman ps --all --filter pod=jellyfin-pod` should help, also `--format` can be used:

```
podman ps --all --format '{{.Names}} {{.ID}}'
```

tomasmatus: Good idea, the `--filter name=` even supports regex so that makes it much simpler.

jelly (approved):
Thanks, now go :green_heart:

tomasmatus:
hmm red again, today is not the day to merge

tomasmatus:
@jelly I whipped the CI once more and now all is green ✅

PR #2185, opened 2025-07-11
Bump podman API version to 3.4.0

martinpitt:
We haven't updated this in years, and "1.12" as a version does not make sense. The intention as per Paul Holzinger [1] is to have a podman version there. The oldest one that we still support is 3.4.0 in Ubuntu 22.04 LTS.

[1] https://github.com/containers/podman/pull/24005#issuecomment-3058499967


jelly (approved):
Thanks!

PR #2192, opened 2025-07-16
test: Add ws-container scenario

martinpitt:
Let's deprecate the "native rpm" rhel-8-10 scenario. We won't ever update RHEL 8 to the current upstream versions. Instead, let's make sure that RHEL 8 keeps working with recent cockpit/ws containers or the flatpak in beiboot mode.

Split off the rpm/beiboot specific bits of test image preparation into their own scripts.

---

 - [x] Builds on top of #2191

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,15 @@ TAR_ARGS = --sort=name --mtime "@$(shell git show --no-patch --format='%at')" --
 
 VM_CUSTOMIZE_FLAGS =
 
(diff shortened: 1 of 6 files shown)


martinpitt:
Cool, [works](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2192-a48da865-20250716-071306-rhel-8-10-ws-container/log.html)! :sunglasses: Waiting for the dependent PR and flake fixes to land.

jelly (approved):
Thanks!

PR #2199, opened 2025-07-21
pixels: mock created column in Image table

tomasmatus:
Variable length of relative time when the image was created can cause pixel-test failures.


martinpitt (approved):
Cheers! This is much better.

PR #2200, opened 2025-07-21
pixels: mock ID column in Image table

tomasmatus:


tomasmatus:
[pixels looks good](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2200-fc3d5217-20250721-135200-fedora-42/log.html)

jelly (approved):
Thanks, next time please mention that this is now mocked as the ID's aren't always the same length

PR #2202, opened 2025-07-23
packit: Fix test job indentation

martinpitt:
This is harmless, but still looks weird.


martinpitt:
hmm, C10s-image-mode still doesn't start, so that remains a bug on the packit/TF side. So the wrong indentation doesn't matter after all. let's still clean it up.

jelly (approved):

PR #2209, opened 2025-07-28
test: Fix subuid hack for RHEL

martinpitt:
Stop assuming that the number is preceded by a space; it's a tab in RHEL 9. Just require a word boundary.

Also, with the most recent image refreshes, the `admin` user already doesn't have subuids, so we have to add them back as well.

---

Fixes https://github.com/cockpit-project/bots/pull/8042

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -31,7 +31,8 @@ chmod 600 /root/.ssh/authorized_keys
 
 # HACK: unbreak subuid assignment for new users; see
 # https://bugzilla.redhat.com/show_bug.cgi?id=2382662
-sed -i '/^SUB_.ID_COUNT/ s/ 0/ 65536/' /etc/login.defs
+# https://issues.redhat.com/browse/RHEL-103765
+sed -i '/^SUB_.ID_COUNT/ s/\b0/65536/' /etc/login.defs
 
 # create user account for logging in
 if ! id admin 2>/dev/null; then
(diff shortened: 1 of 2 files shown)


test/browser/browser.sh
```diff
@@ -29,9 +29,14 @@ mkdir -p /root/.ssh
...
-# HACK: unbreak subuid assignment for new users; see
+# HACK: unbreak subuid assignment for current and new users; see
 # https://bugzilla.redhat.com/show_bug.cgi?id=2382662
-sed -i '/^SUB_.ID_COUNT/ s/ 0/ 65536/' /etc/login.defs
+# https://issues.redhat.com/browse/RHEL-103765
+sed -i '/^SUB_.ID_COUNT/ s/\b0/65536/' /etc/login.defs
+if ! grep -q admin /etc/subuid; then
+    usermod --add-subuids 100000-165535 admin
+    usermod --add-subgids 100000-165535 admin
+fi
```
martinpitt: :warning: Don't *ever* do this in production systems. This would have to be dynamic, to avoid uid collision. But there is currently no mechanism or documentation how to *actually* do this in a safe way, so this picks the hack from bugzilla. It's okay for our tests.

martinpitt: Update: We don't actually need it in tmt, as we create the `admin` user after fixing login.defs. We only need it for our own bots VM images.

jelly (approved):
:see_no_evil:

PR #2210, opened 2025-07-28
packit: Enable CentOS 10 image mode

martinpitt:
This only works with the arch suffix, see
https://github.com/packit/packit-service/issues/2834


tomasmatus (approved):
Thanks!

PR #2214, opened 2025-08-01
Initial typing, fix discovered bugs

martinpitt:
This adds typing for the basal podman API. Not yet for the UI, but at least now our editors yell at us for all the untyped stuff.

This is enough to validate the cockpit.http API introduced in https://github.com/cockpit-project/cockpit/pull/21678 and refined in https://github.com/cockpit-project/cockpit/pull/22269 , and already found some bugs.

 - [x] cockpit.http typing: https://github.com/cockpit-project/cockpit/pull/21678
 - [x] Land https://github.com/cockpit-project/cockpit/pull/22269
 - [x] Pull in :arrow_up: , refresh testlib, drop the XXX commit: #2215

PR #2219, opened 2025-08-05
Fix search registries queries

martinpitt:
As per commit ab94b857b654, the podman API returns `registries` as an
object which may contain a "search" list. But checking the length of the
`registries` object is nonsensical -- what we actually want is to check
the length of the `search` list.

This worked through sheer luck: podman does not put a `search` property
into `registries` if the list is empty (even if it is explicitly
configured). But this would break as soon as `registries` grows a second
property.

---

Spotted here: https://github.com/cockpit-project/cockpit-podman/pull/2214#discussion_r2251512668

### DIFF
--- a/src/ImageRunModal.jsx
+++ b/src/ImageRunModal.jsx
@@ -429,9 +429,10 @@ export class ImageRunModal extends React.Component {
         // searching for specific image tags
         // instead only rely on manifests query (requires image:tag name)
         if (!RE_CONTAINER_TAG.test(value)) {
-            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // If there are search registries configured, search in them, or if a user searches for `docker.io/cockpit` let
             // podman search in the user specified registry.
-            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
+            const reg_search = this.props.podmanInfo.registries?.search;
+            if ((reg_search && reg_search.length !== 0) || value.includes('/')) {
                 searches.push(this.activeConnection.call({
                     method: "GET",
                     path: client.VERSION + "libpod/images/search",
--- a/src/ImageSearchModal.jsx
+++ b/src/ImageSearchModal.jsx
@@ -42,7 +42,7 @@ export const ImageSearchModal = ({ downloadImage, users }) => {
     const { registries } = usePodmanInfo();
     const Dialogs = useDialogs();
     // Registries to use for searching
-    const searchRegistries = registries.search && registries.length !== 0 ? registries.search : fallbackRegistries;
+    const searchRegistries = registries?.search && registries.search.length !== 0 ? registries.search : fallbackRegistries;
 
     // Don't use on selectedRegistry state variable for finding out the
     // registry to search in as with useState we can only call something after a


martinpitt:
Ah, I understand this now. This worked through pure happenstance:
```
# echo 'unqualified-search-registries = []' > /etc/containers/registries.conf
# curl --silent --unix-socket /run/podman/podman.sock http://d/v5.0.0/libpod/info|jq .registries
{}

# echo 'unqualified-search-registries = ["foo"]' > /etc/containers/registries.conf
# curl --silent --unix-socket /run/podman/podman.sock http://d/v5.0.0/libpod/info|jq .registries
{
  "search": [
    "foo"
  ]
}
```

src/ImageRunModal.jsx
```diff
@@ -429,9 +429,9 @@ export class ImageRunModal extends React.Component {
         // searching for specific image tags
         // instead only rely on manifests query (requires image:tag name)
         if (!RE_CONTAINER_TAG.test(value)) {
-            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // If there are search registries configured, search in them, or if a user searches for `docker.io/cockpit` let
             // podman search in the user specified registry.
-            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
+            if (this.props.podmanInfo.registries?.search?.length !== 0 || value.includes('/')) {
```
mvollmer: This doesn't work, I am afraid. If `registries` doesn't have a `search` field, `registries?.search` is `undefined`, which is `!== 0`. But we want the condition to be false in that case.

```
>> x = { }
Object {  }
>> x?.search?.length
undefined
>> x?.search?.length !== 0
true 
```

Maybe `registries?.search?.length > 0`?

src/ImageSearchModal.jsx
```diff
@@ -42,7 +42,7 @@ export const ImageSearchModal = ({ downloadImage, users }) => {
     const { registries } = usePodmanInfo();
     const Dialogs = useDialogs();
     // Registries to use for searching
-    const searchRegistries = registries.search && registries.length !== 0 ? registries.search : fallbackRegistries;
+    const searchRegistries = registries?.search?.length !== 0 ? registries.search : fallbackRegistries;
```
mvollmer: Same here.

mvollmer (changes requested):

src/ImageRunModal.jsx
martinpitt: Thanks for spotting! I don't want to rely on `undefined > 0 == false` either. I'd rather make it more explicit with `registries?.search?.length ?? 0`

src/ImageRunModal.jsx
```diff
@@ -429,9 +429,9 @@ export class ImageRunModal extends React.Component {
         // searching for specific image tags
         // instead only rely on manifests query (requires image:tag name)
         if (!RE_CONTAINER_TAG.test(value)) {
-            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // If there are search registries configured, search in them, or if a user searches for `docker.io/cockpit` let
             // podman search in the user specified registry.
-            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
+            if ((this.props.podmanInfo.registries?.search?.length ?? 0) !== 0 || value.includes('/')) {
```
cockpituous, line 434: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2219-fc1b11f0-20250806-074138-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/ImageSearchModal.jsx
```diff
@@ -42,7 +42,7 @@ export const ImageSearchModal = ({ downloadImage, users }) => {
     const { registries } = usePodmanInfo();
     const Dialogs = useDialogs();
     // Registries to use for searching
-    const searchRegistries = registries.search && registries.length !== 0 ? registries.search : fallbackRegistries;
+    const searchRegistries = (registries?.search?.length ?? 0) !== 0 ? registries.search : fallbackRegistries;
```
cockpituous, line 45: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2219-fc1b11f0-20250806-074138-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/ImageRunModal.jsx
mvollmer: The return of the son of Perl! :-) What about `registries.search && registries.search.length > 0`?

martinpitt: Can do as well -- I was mostly trying to avoid having to spell out that long name twice, but I'll do an alias.

mvollmer (approved):
Thanks!

src/ImageRunModal.jsx
```diff
@@ -429,9 +429,10 @@ export class ImageRunModal extends React.Component {
         // searching for specific image tags
         // instead only rely on manifests query (requires image:tag name)
         if (!RE_CONTAINER_TAG.test(value)) {
-            // If there are registries configured search in them, or if a user searches for `docker.io/cockpit` let
+            // If there are search registries configured, search in them, or if a user searches for `docker.io/cockpit` let
             // podman search in the user specified registry.
-            if (Object.keys(this.props.podmanInfo.registries).length !== 0 || value.includes('/')) {
+            const reg_search = this.props.podmanInfo.registries?.search;
+            if ((reg_search && reg_search.length !== 0) || value.includes('/')) {
```
cockpituous, line 435: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2219-12ebb26b-20250806-090255-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

src/ImageSearchModal.jsx
```diff
@@ -42,7 +42,7 @@ export const ImageSearchModal = ({ downloadImage, users }) => {
     const { registries } = usePodmanInfo();
     const Dialogs = useDialogs();
     // Registries to use for searching
-    const searchRegistries = registries.search && registries.length !== 0 ? registries.search : fallbackRegistries;
+    const searchRegistries = registries?.search && registries.search.length !== 0 ? registries.search : fallbackRegistries;
```
cockpituous, line 45: This added line is not executed by any test.  [Details](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2219-12ebb26b-20250806-090255-fedora-42-devel/Coverage/lcov/github-pr.diff.gcov.html)

Issue #2234, opened 2025-08-19
Tests failed on 3ec1322d6bf5e7c45bd6a077a4264f0d1d6db9b9


cockpituous:
The job `fedora-42/copr/rhcontainerbot/podman-next` failed on commit 3ec1322d6bf5e7c45bd6a077a4264f0d1d6db9b9.

Log: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-0-3ec1322d-20250819-023952-fedora-42-copr-rhcontainerbot-podman-next/log.html

PR #2235, opened 2025-08-19
test: properly detach podman exec

jelly:
With podman 5.7.0dev, podman no longer seems to background properly. This behaviour looks a bit sketchy as podman supports properly detaching an exec with `--detach`.

Closes: #2234


allisonkarlitskaya (approved):
I definitely prefer it this way

test/check-application
```diff
@@ -1507,7 +1507,7 @@ Yaml={name}.yaml
             self.assertLessEqual(mem_abs, 300.0)
 
             # Test that the CPU value is updated dynamically
-            self.execute("podman exec -i swamped-crate sh -c 'dd bs=1024 < /dev/urandom > /dev/null &'", system=auth)
+            self.execute("podman exec -d -i swamped-crate sh -c 'dd bs=1024 < /dev/urandom > /dev/null'", system=auth)
```
allisonkarlitskaya: ...but why not drop the `-i`?  Isn't that sort of the opposite of `-d`?

allisonkarlitskaya (dismissed):

PR #2254, opened 2025-09-05
test: drop devel scenario hack

jelly:
The pinnedUnits oops has been included in Cockpit 340 which our TEST_OS_DEFAULT image already has, as it is Fedora 42 and has Cockpit 345 currently.


allisonkarlitskaya (approved):
Yes.

PR #2265, opened 2025-09-15
test: speed up quadlet tests

jelly:
Set the default stop timeout from 20 to 0 seconds which speeds up the test cleanup up by about 20 seconds.


jelly:
Just triggering centos-10 for testing the recent crio issue.

martinpitt (approved):
:fast_forward: dankjewel!

PR #2294, opened 2025-10-17
css: Move `no-masthead-sidebar` to `pf-m-no-sidebar`

jelly:
This is the official class for it so we no longer need our own workaround.

Related-to: https://github.com/patternfly/patternfly/issues/7357


tomasmatus (approved):
failed TF looks like a familiar flakes but I don't see it on weather report

jelly:
I have seen testPullImage fail before on TF on CentOS 9i

PR #2321, opened 2025-11-10
Preparations for rebuilding bundle during distro builds

martinpitt:
Ported from https://github.com/cockpit-project/cockpit-files/pull/1203

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ $(TARFILE): $(DIST_TEST) $(SPEC) packaging/arch/PKGBUILD packaging/debian/change
 		packaging/arch/PKGBUILD packaging/debian/changelog dist/
 
 $(NODE_CACHE): $(NODE_MODULES_TEST)
-	tar --xz $(TAR_ARGS) -cf $@ --exclude .git node_modules
+	tools/node-modules runtime-tar $(NODE_CACHE)
 
 node-cache: $(NODE_CACHE)
 
--- a/packit.yaml
+++ b/packit.yaml
@@ -12,6 +12,7 @@ actions:
   create-archive: make dist
 
 srpm_build_deps:
+  - jq
   - make
   - nodejs
 


Makefile
```diff
@@ -122,6 +122,7 @@ clean:
 	rm -rf dist/
 	rm -f $(SPEC) packaging/arch/PKGBUILD packaging/debian/changelog
 	rm -f po/LINGUAS
+	rm -f metafile.json runtime-npm-modules.txt
```
allisonkarlitskaya, line 125: Could we use the metafile as the stamp?  I'm thinking about what might happen if you do a watch mode build, then ^C and type `make vm`.  I think it would fail to rebuild properly, no?

martinpitt: You mean as a condition for `$(TARFILE)`? It currently uses DIST_TEST aka. dist/manifest.json, which IMHO is the better time stamp. Watch mode doesn't produce metafile.json at all, as that's expensive.

allisonkarlitskaya: that's my point though: because the DIST_TEST already exists then we won't rerun the rule and we'll try to access the metafile and fall on our faces...

martinpitt: > 08:55:19 <lis> if we do ./build.js -w on a fresh checkout then we'll get dist/package.json generated, but not the metafile
> 08:56:16 <lis> then if we type make vm, make will look at the depenencies of the build and conclude that we already have dist/package.json up to date, so no need to rebuild it, then try to continue to building the node-modules tarball, but it will fail because the list of files is missing
> 08:56:28 <lis> we need to pick a stamp that does cause ./build.js to be rerun

martinpitt: Thanks for spotting! I understand and reproduced this now. I added a separate "fixup" commit, as I cherry-picked from cockpit-files, and want to apply the fixup to cockpit-files as well.

martinpitt: https://github.com/cockpit-project/cockpit-files/pull/1232

martinpitt: Argh, this is broken.. We need to ship the stamp file in the packages.

martinpitt: Fixed both here and in the c-files PR.

Makefile
```diff
@@ -14,8 +14,8 @@ APPSTREAMFILE=org.cockpit_project.$(PACKAGE_NAME).metainfo.xml
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 # stamp file to check for node_modules/
 NODE_MODULES_TEST=package-lock.json
-# one example file in dist/ from bundler to check if that already ran
-DIST_TEST=dist/manifest.json
```
jelly, line 18: manifest.json is also like one of the first things we copy :)

martinpitt: What do you mean? Is something broken in this change? :thinking:

jelly: No no, it was more a remark that if you ever cancelled `build.js` we likely would have already copied manifest.json as we copy javascript/sass later.

Makefile
```diff
@@ -14,8 +14,8 @@ APPSTREAMFILE=org.cockpit_project.$(PACKAGE_NAME).metainfo.xml
...
+# build.js ran in non-watch mode
+DIST_TEST=runtime-npm-modules.txt
```
allisonkarlitskaya, line 18: I feel like the metafile would be more neutral, but since this is the one we actually use, I'm good with that :)

Our future node-modules-tar-packer script should ideally be in python and parse the metafile directly, though, so we can toss this secondary output...

PR #2323, opened 2025-11-11
test: use a helper for opening the create container dialog

jelly:
This makes refactoring selectors easier in the future.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -213,9 +213,12 @@ class TestApplication(testlib.MachineCase):
         b = self.browser
         return b.text(f"#containers-containers tbody tr:contains('{container}') > td[data-label={key}] {selector}")
 
-    def openCreateContainerDialog(self, image: str, *, system: bool = True) -> None:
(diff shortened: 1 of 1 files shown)


jelly:
@martinpitt what do you think of this test ballooon to get rid of sizzle?

test/check-application
```diff
@@ -213,6 +213,10 @@ class TestApplication(testlib.MachineCase):
         b = self.browser
         return b.text(f"#containers-containers tbody tr:contains('{container}') > td[data-label={key}] {selector}")
 
+    def openCreateContainerDialog(self, image: str, *, system: bool = True) -> None:
+        tr_sel = "tr[data-row-id^='0-']" if system else "tr[data-row-id^='user-']"
```
martinpitt: You could factorize this further into a `self.row_user_prefix(system)` helper, I figure we use this in other places? If not, make it a local `prefix = '0' if system else 'user'`.

martinpitt: Nevermind, the followup commits rewrite this.

PR #2326, opened 2025-11-13
packaging: Rebuild our bundle during Debian package build

martinpitt:
This is pretty much https://github.com/cockpit-project/cockpit-files/pull/1206 with some -podman specific adjustments.

After this, I'll move the `packaging/build-debian-copyright` script over to cockpit, and pull it in as a shared script. Here it's a verbatim copy of what is in -files, proving that it works for more than just one project.


martinpitt:
This is also the first time we test this stuff on Ubuntu. -files doesn't run there.

PR #2335, opened 2025-11-20
Wait text over visible

jelly:
Just refactoring, not getting us closer to drop sizzle yet :)


martinpitt (commented):
Looks great, but the bots say "Nee" :cry:

PR #2340, opened 2025-11-24
Drop sizzle test dependency :partying_face: 

jelly:


jelly:
I'll investigate the testPullImage thing, it fails way too often

test/check-application
```diff
@@ -1675,7 +1675,8 @@ Yaml={name}.yaml
 
         self.toggleExpandedContainer("swamped-crate", system=True)
         b.wait_visible(".pf-m-expanded button:contains('Details')")
-        b.wait_not_present(f'#containers-containers tr:contains("{IMG_BUSYBOX}") dt:contains("Latest checkpoint")')
+        container_sel = self.getContainerSelector("swamped-crate", system=True)
```
martinpitt, line 1679: How come this doesn't use `.container-latest-checkpoint` ?

test/check-application
```diff
@@ -1230,7 +1230,7 @@ Yaml={name}.yaml
 
             def expectDownloadErrorForNonExistingTag(self) -> Self:
                 title = f"Danger alert:Failed to download image localhost:5000/{self.imageName}:{self.imageTag}"
-                b.wait_visible(f'h4.pf-v6-c-alert__title:contains("{title}")')
+                b.wait_in_text('h4.pf-v6-c-alert__title', title)
```
martinpitt, line 1233: The commit message has it backwards ("prefer wait_visible over wait_in_text") :grin:

Makefile
```diff
@@ -148,11 +148,6 @@ devel-uninstall:
 print-version:
 	@echo "$(VERSION)"
 
-# required for running integration tests
-TEST_NPMS = \
-       node_modules/sizzle \
```
martinpitt, line 153: Nice! Can you please also remove these two from packaging/debian/rules:
```
	# remove mini-node_modules with sizzle
	rm -fr node_modules
```

test/check-application
jelly: Good point, I guess my reasoning was, we don't have to be this specific. But yeah reverted it to `wait_not_present` makes sense here

test/check-application
jelly: welp, good catch!

test/check-application
martinpitt: You are right, the previous version was fine as well. I didn't read the semantics close enough.

PR #2344, opened 2025-11-28
spec: Make node tarball Source1 unconditional

martinpitt:
I originally made this conditional as it would be dead weight for CentOS/RHEL (where we can't/don't rebuild the bundle). But conditional sources are discouraged [1] and also seem to break packit downstream releases [2].

Thanks to Nikola Forró for the hint!

[1] https://docs.fedoraproject.org/en-US/packaging-guidelines/SourceURL/#_do_not_conditionalize_sources
[2] https://src.fedoraproject.org/rpms/cockpit-podman/pull-request/271#

----

@nforro thanks for having a look!

I'll apply the same treatment to c-files after this lands.

### DIFF
--- a/packaging/cockpit-podman.spec.in
+++ b/packaging/cockpit-podman.spec.in
@@ -27,10 +27,8 @@ URL:            https://github.com/cockpit-project/cockpit-podman
 %define rebuild_bundle 1
 %endif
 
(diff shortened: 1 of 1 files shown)


martinpitt:
[testPullImage frequent flake](https://artifacts.dev.testing-farm.io/0c5449d1-d609-46c0-b454-29387cfa5f91/), ignoring as it's not related to the Source1 change.

PR #2345, opened 2025-11-28
workflows: sync cockpit-lib-update workflow with starter-kit

jelly:
The starter-kit workflow never required a deploy_key for cloning the repo as we create the pull request via GitHub token in tasklib.

---

Test run https://github.com/jelly/cockpit-podman/actions/runs/19766633322/job/56640894908


martinpitt (approved):
Thanks!

PR #2350, opened 2025-12-01
packit: Fix for multiple tarballs

martinpitt:
Commit 0e6a2671 forgot to adjust the packit recipe. We missed to update Source1: to the locally built -node tarball. So up to now it was using the tarball from the latest release. That *happened* to work so far, but once we change package.json it's not going to any more.

Also update the upstream release COPR build rule.

----

See [recent build log](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-podman-2349/fedora-42-x86_64/09853036-cockpit-podman/builder-live.log.gz):

> INFO: Downloading cockpit-podman-node-118.tar.xz

Oops!

@jelly spotted that in https://github.com/cockpit-project/cockpit/pull/22634#discussion_r2576991746

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -50,8 +50,8 @@ jobs:
       # HACK: tarball for releases (copr_build, koji, etc.), copying spec's Source0; this
       # really should be the default, see https://github.com/packit/packit-service/issues/1505
       fix-spec-file:
-        - sh -exc "curl -L -O https://github.com/cockpit-project/cockpit-podman/releases/download/${PACKIT_PROJECT_VERSION}/${PACKIT_PROJECT_NAME_VERSION}.tar.xz"
-        - sh -exc "curl -L -O https://github.com/cockpit-project/cockpit-files/releases/download/${PACKIT_PROJECT_VERSION}/cockpit-files-node-${PACKIT_PROJECT_VERSION}.tar.xz"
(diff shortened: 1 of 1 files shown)


packit.yaml
```diff
@@ -8,8 +8,14 @@ downstream_package_name: cockpit-podman
...
 
 actions:
-  post-upstream-clone: make cockpit-podman.spec
-  create-archive: make dist
+  post-upstream-clone:
+    - make node-cache dist
+
+  fix-spec-file:
+    # our locally built tarballs are not on the upstream releases page, point to the local files
+    - sh -c 'sed -i "/^Source1:/ s/https:.*/$(ls *-node*.tar.xz)/" cockpit-*.spec'
```
jelly:
```
[jelle@carbon][~/projects/cockpit-podman]%sed -i "/^Source1:/ s/https:.*/$(ls *-node*.tar.xz)/" cockpit-*.spec
sed: -e expression #1, char 85: unterminated `s' command
```

But I'm a tad bit confused, this only applies to Source1, while both sources come from a https url? But why is this even a problem?

Mental note, we have the same sed in cockpit-files.

martinpitt: Adjusted in the same way as https://github.com/cockpit-project/cockpit-files/pull/1250 , and consistent with cockpit PR.

Issue #2362, opened 2025-12-10
Do not fetch all container logs but only the tail


deRockerTom:
When a container has many logs, it will scroll forever before we can see the current logs.
Old logs will also be removed as the component we use for terminal only displays the last 1000 lines (see https://github.com/xtermjs/xterm.js/blob/master/typings/xterm.d.ts#L253-L258)

Would it be possible to only fetch the last 1000 logs so we don't need to wait forever. It should be possible by just adding the `tail` parameter to the rest call (see https://docs.podman.io/en/latest/_static/api.html?version=latest#tag/containers/operation/ContainerLogsLibpod). I can give it a try if you want, would be my first OSS contribution

Thanks !

PR #2363, opened 2025-12-10
Only fetch last container logs in ContainerLogs component

deRockerTom:
When we open the logs tab of a container, we fetch all container logs, which can take a while if it has many.
Even if we fetch all logs, we'll only display the last 1000 due to the Terminal component props.

In this PR I clarified the max Terminal scrollback size and fetched only the last 1000 logs from the container, so we don't wait a lot with the first fetch

Related to https://github.com/cockpit-project/cockpit-podman/issues/2362

### DIFF
--- a/src/ContainerLogs.jsx
+++ b/src/ContainerLogs.jsx
@@ -35,6 +35,8 @@ import "./ContainerTerminal.css";
 
 const _ = cockpit.gettext;
 
+const LOGS_MAX_SIZE = 1000;
+
 class ContainerLogs extends React.Component {
     constructor(props) {
         super(props);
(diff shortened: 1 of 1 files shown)


src/ContainerLogs.jsx
```diff
@@ -51,7 +53,8 @@ class ContainerLogs extends React.Component {
             disableStdin: true,
             fontSize: 12,
             fontFamily: 'Menlo, Monaco, Consolas, monospace',
-            screenReaderMode: true
+            screenReaderMode: true,
+            scrollback: LOGS_MAX_SIZE,
```
jelly, line 57: Is that needed to be set? You could have longer logs when you keep as the logs keep streaming in

jelly: Ah, I missed, would be nice to have a comment above the constant but lets merge this. https://xtermjs.org/docs/api/terminal/interfaces/iterminaloptions/#optional-scrollback

deRockerTom: Oh indeed I could have let a comment about this. Just saw that you created a new PR to add it, thanks !

PR #2396, opened 2026-01-13
package.json: Fix stylelint rules

jelly:
The effect wasn't that dramatic, we previously just globbed `*scss`.

Thanks to haligh for reporting this!


martinpitt:
needs node_modules refresh

martinpitt (approved):
Cheers!

PR #2434, opened 2026-02-12
build: Update stylelint to 17

mvollmer:
This requires the following changes:

- Remove disabling of scss/at-import-no-partial-leading-underscore. This rule is no longer recognized.

- Switching from stylelint-use-logical-spec to stylelint-use-logical. The former is not yet compatible with stylelint 17.


jelly:
Green!

PR #2435, opened 2026-02-12
systemd-run switchover prepration

jelly:


test/check-application
```diff
@@ -236,11 +237,24 @@ class TestApplication(testlib.MachineCase):
...
-            return self.admin_s.execute(cmd, check=check)
+            # RHEL 8.10 systemd does not support --machine=admin@ for local sessions
+            if m.image == "rhel-8-10":
+                return self.admin_s.execute(cmd, check=check)
+            else:
+                sd_run_admin = "systemd-run --user --machine=admin@"
+                if background:
+                    unit = f'user-{uuid.uuid4()}'
+                    m.execute(f"{sd_run_admin} --unit={unit} /bin/sh -ec '{cmd}'")
+                    m.execute(f"until {sd_run_admin} is-active {unit}; do sleep 1; done")
```
martinpitt, line 253: Isn't that redundant?

> systemd-run will start the service asynchronously in the background and return after the command has begun execution (unless --no-block, --wait, --pipe, or --pty are specified, see below)

PR #2436, opened 2026-02-12
Show description, version and doc labels for images

jelly:
The OpenContainers Annotations Spec specifies what standardised labels may exist on a container image. For now expose the description, documentation link and version in the image details tab. This is particularly useful showing the version of the software you run when using the `latest` tag.

---

# Podman: Show description, version and documentation image labels 

The image details tab now shows a select set of labels from the [opencontainers annotations specification](https://specs.opencontainers.org/image-spec/annotations/). This is particularly useful to find out what version the `latest` tag of an image is.

image

### DIFF
--- a/src/ImageDetails.jsx
+++ b/src/ImageDetails.jsx
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 import React from 'react';
 
+import { Button } from "@patternfly/react-core/dist/esm/components/Button";
 import { DescriptionList, DescriptionListDescription, DescriptionListGroup, DescriptionListTerm } from "@patternfly/react-core/dist/esm/components/DescriptionList";
 
 import cockpit from 'cockpit';
@@ -11,8 +12,36 @@ import * as utils from './util.js';
 const _ = cockpit.gettext;
 
 const ImageDetails = ({ containers, image, showAll }) => {
+    const labels = image?.Labels;
+
+    // https://specs.opencontainers.org/image-spec/annotations/?v=v1.1.1
+    const imageDescription = labels?.["org.opencontainers.image.description"];
+    const imageDocumentation = labels?.["org.opencontainers.image.documentation"];
+    const imageVersion = labels?.["org.opencontainers.image.version"];
     return (
         <DescriptionList className='image-details' isAutoFit>
+            {imageDescription &&
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Description")}</DescriptionListTerm>
+                <DescriptionListDescription data-label="description">{imageDescription}</DescriptionListDescription>
+            </DescriptionListGroup>
+            }
+            {imageVersion &&
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Version")}</DescriptionListTerm>
+                <DescriptionListDescription data-label="version">{imageVersion}</DescriptionListDescription>
+            </DescriptionListGroup>
+            }
+            {imageDocumentation &&
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Links")}</DescriptionListTerm>
+                <DescriptionListDescription data-label="documentation">
+                    <Button variant="link" isInline component="a" href={imageDocumentation} target="_blank noreferrer">
+                        {_("Documentation")}
+                    </Button>
+                </DescriptionListDescription>
+            </DescriptionListGroup>
+            }
             {image.Command &&
             <DescriptionListGroup>
                 <DescriptionListTerm>{_("Command")}</DescriptionListTerm>
--- a/test/check-application
+++ b/test/check-application
@@ -3130,6 +3130,26 @@ Yaml={name}.yaml
             service_container = m.execute("podman ps --all --filter name=.*-service --format {{.Names}}").strip()
             b.wait_not_present(f"#containers-containers tbody tr[data-row-id=\"{service_container}\"]")
 
+        # opencontainer annotations https://specs.opencontainers.org/image-spec/annotations/
+        label_image = "localhost/label-image:latest"
+        tmpdir = self.execute("mktemp -d", system=True).strip()
+        self.execute(f"echo 'FROM {IMG_REGISTRY}\nRUN ls' > {tmpdir}/Containerfile", system=True)
+        doc_label = "--label org.opencontainers.image.documentation='https://cockpit-project.org'"
+        desc_label = "--label org.opencontainers.image.description='The Cockpit container running cockpit-ws'"
+        version_label = "--label org.opencontainers.image.version='355'"
+        m.execute(f"""podman build -t {label_image} {doc_label} {desc_label} {version_label} {tmpdir}""")
+        checkImage(b, label_image, "system")
+        img_id = m.execute(f"podman inspect --format '{{{{ .Id }}}}' {label_image}").strip()
+        b.click(f"#containers-images tr[data-row-id='0-{img_id}'] td.pf-v6-c-table__toggle button")
+        b.wait_text("#containers-images tbody.pf-m-expanded tr .image-details [data-label='version']", "355")
+        b.wait_text("#containers-images tbody.pf-m-expanded tr .image-details [data-label='description']",
+                    "The Cockpit container running cockpit-ws")
+        self.assertEqual(b.attr(
+                            "#containers-images tbody.pf-m-expanded tr .image-details [data-label='documentation'] a",
+                            "href"
+                         ),
+                         "https://cockpit-project.org")
+
     def testCreatePodSystem(self) -> None:
         self._createPod(auth=True)
 


src/ImageDetails.jsx
```diff
@@ -11,8 +12,31 @@ import * as utils from './util.js';
 const _ = cockpit.gettext;
 
 const ImageDetails = ({ containers, image, showAll }) => {
+    const labels = image?.Labels;
```
tomasmatus, line 15: What is the difference between `Labels` and `Annotations`? The spec you [linked](https://specs.opencontainers.org/image-spec/annotations/?v=v1.1.1) talks about annotations. I tried inspecting some containers that I have running on my server and some have these key-value pairs under Labels and some under Annotations. Should we perhaps check both and see what's available?

Eg. the community nextcloud image has it under annotations:
```
          "Labels": null,
          "Annotations": {
               "com.docker.official-images.bashbrew.arch": "amd64",
               "org.opencontainers.image.base.digest": "sha256:c5c7c3f70c20ddbb52d3ae8d9aa8a318cb7518d6092ea48e6311ad15133b84b7",
               "org.opencontainers.image.base.name": "php:8.3-apache-trixie",
               "org.opencontainers.image.created": "2026-01-17T00:12:14Z",
               "org.opencontainers.image.revision": "b91ae50a73884f5b086ad03e61d30e219b376ebb",
               "org.opencontainers.image.source": "https://github.com/nextcloud/docker.git#b91ae50a73884f5b086ad03e61d30e219b376ebb:31/apache",
               "org.opencontainers.image.url": "https://hub.docker.com/_/nextcloud",
               "org.opencontainers.image.version": "31.0.13-apache"
          },
```

but many others indeed have it under `Labels` or mariadb seems to do a mix of both with different values?
```
          "Labels": {
               "org.opencontainers.image.authors": "MariaDB Community",
               "org.opencontainers.image.base.name": "docker.io/library/ubuntu:noble",
               "org.opencontainers.image.description": "MariaDB Database for relational SQL",
               "org.opencontainers.image.documentation": "https://hub.docker.com/_/mariadb/",
               "org.opencontainers.image.licenses": "GPL-2.0",
               "org.opencontainers.image.ref.name": "ubuntu",
               "org.opencontainers.image.source": "https://github.com/MariaDB/mariadb-docker",
               "org.opencontainers.image.title": "MariaDB Database",
               "org.opencontainers.image.url": "https://github.com/MariaDB/mariadb-docker",
               "org.opencontainers.image.vendor": "MariaDB Community",
               "org.opencontainers.image.version": "11.8.4"
          },
          "Annotations": {
               "com.docker.official-images.bashbrew.arch": "amd64",
               "org.opencontainers.image.base.digest": "sha256:4fdf0125919d24aec972544669dcd7d6a26a8ad7e6561c73d5549bd6db258ac2",
               "org.opencontainers.image.base.name": "ubuntu:noble",
               "org.opencontainers.image.created": "2025-11-13T23:29:22Z",
               "org.opencontainers.image.revision": "5ad00a82b5bb2705825c7afb670f6547b1bed316",
               "org.opencontainers.image.source": "https://github.com/MariaDB/mariadb-docker.git#5ad00a82b5bb2705825c7afb670f6547b1bed316:11.8",
               "org.opencontainers.image.url": "https://hub.docker.com/_/mariadb",
               "org.opencontainers.image.version": "11.8.4-noble"
          },
```

jelly: https://xkcd.com/927/

I agree this is absurd... Most of my images had it under `Labels`.

jelly: Thanks for pointing this out! I did some digging and it seems the idea is to move to `Annotations` as replacement. Although slightly?

https://github.com/opencontainers/image-spec/blob/main/annotations.md#back-compatibility-with-label-schema

But looking at the output you posted folks don't seem to include what we use (documentation, url, version) in the `Annotations`. And if they do its not about the application itself. So maybe we should stick with `Labels` for now?

tomasmatus: I don't know honestly, from these two examples it seems like `Annotations` are used to describe the image itself rather than the app running inside but I'd also argue this is too small sample count.

> So maybe we should stick with Labels for now?

yeah, I can agree with this since you said containers you run have it under `Labels` and we'll see if someone shows up with complaints :)

src/ImageDetails.jsx
```diff
@@ -11,8 +12,31 @@ import * as utils from './util.js';
...
     return (
         <DescriptionList className='image-details' isAutoFit>
+            {labels?.["org.opencontainers.image.description"] &&
```
tomasmatus: could you please move these into a variable? I don't like seeing the long `["org.opencontainers.image.description"]` so much. Applies to all 3 ofc.

```const imageDescription = labels?.["org.opencontainers.image.description"]```
and then work with that. Thanks! :)

File: test/check-application [hunk 3]
```diff
@@ -3132,6 +3132,26 @@ Yaml={name}.yaml
...
+        # opencontainer annotations https://specs.opencontainers.org/image-spec/annotations/
+        label_image = "label-image"
+        tmpdir = self.execute("mktemp -d", system=True).strip()
+        self.execute(f"echo 'FROM {IMG_REGISTRY}\nRUN ls' > {tmpdir}/Containerfile", system=True)
+        doc_label = "--label org.opencontainers.image.documentation='https://cockpit-project.org'"
+        desc_label = "--label org.opencontainers.image.description='The Cockpit container running cockpit-ws'"
+        version_label = "--label org.opencontainers.image.version='355'"
+        m.execute(f"""podman build -t {label_image} {doc_label} {desc_label} {version_label} {tmpdir}""")
+        checkImage(b, "localhost/label-image:latest", "system")
+        img_id = m.execute("podman inspect --format '{{ .Id }}' localhost/label-image:latest").strip()
```
tomasmatus: you *could* use the `label_image` variable in these two lines above, or just drop it as it's only used once

src/ImageDetails.jsx
```diff
@@ -11,8 +12,31 @@ import * as utils from './util.js';
...
+            {labels?.["org.opencontainers.image.documentation"] &&
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Links")}</DescriptionListTerm>
+                <DescriptionListDescription data-label="documentation">
+                    <Button variant="link" isInline component="a" href={labels["org.opencontainers.image.documentation"]} target="_blank noreferrer">
+                        {_("Documentation")}
+                    </Button>
+                </DescriptionListDescription>
+            </DescriptionListGroup>
+            }
```
Venefilyn, line 44: Looking at how Quay does it and the [spec itself](https://specs.opencontainers.org/image-spec/annotations/?v=v1.1.1). IMO we should have a map or interface for all the fields we use as there is a lot of room for human error here.

https://github.com/quay/quay/blob/f311b9969ebb66e91e43a2aca5a1382513396290/config.py#L95-L108

So something like this. Added an [index signature](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures) for `OpenContainerImage` as the spec just says these are the pre-defined keys but can be expanded

```tsx
// https://specs.opencontainers.org/image-spec/annotations/?v=v1.1.1
interface OpenContainerImage {
    "org.opencontainers.image.version": string;
    "org.opencontainers.image.created": string;
    "org.opencontainers.image.description": string;
    "org.opencontainers.image.title": string;
    "org.opencontainers.image.documentation": string;
    "org.opencontainers.image.url": string;
    "org.opencontainers.image.source": string;
    "org.opencontainers.image.vendor": string;
    [key: `org.opencontainers.image.${string}`]: unknown
}

type ContainerLabels =
    Partial<OpenContainerImage> & 
    { [key: string]: string };
```

Which we then use as you did as I'm fine with writing it out explicitly.


```suggestion
    const labels: ContainerLabels|undefined = image?.Labels;
    return (
        <DescriptionList className='image-details' isAutoFit>
            {labels?.["org.opencontainers.image.description"] &&
            <DescriptionListGroup>
                <DescriptionListTerm>{_("Description")}</DescriptionListTerm>
                <DescriptionListDescription data-label="description">{labels["org.opencontainers.image.description"]}</DescriptionListDescription>
            </DescriptionListGroup>
            }
            {labels?.["org.opencontainers.image.version"] &&
            <DescriptionListGroup>
                <DescriptionListTerm>{_("Version")}</DescriptionListTerm>
                <DescriptionListDescription data-label="version">{labels["org.opencontainers.image.version"]}</DescriptionListDescription>
            </DescriptionListGroup>
            }
            {labels?.["org.opencontainers.image.documentation"] &&
            <DescriptionListGroup>
                <DescriptionListTerm>{_("Links")}</DescriptionListTerm>
                <DescriptionListDescription data-label="documentation">
                    <Button variant="link" isInline component="a" href={labels["org.opencontainers.image.documentation"]} target="_blank noreferrer">
                        {_("Documentation")}
                    </Button>
                </DescriptionListDescription>
            </DescriptionListGroup>
            }
```

Venefilyn (commented):
I'm fine with the implementation, but think we should be explicit with the image labels to make human error less likely with typing out `org.opencontainers.image.documentation` etc.

It's still not perfect as you can still make mistakes and this will won't catch it. But for making it more type-safe I'd want to split this to a new component where we can filter to only include `org.opencontainers.image`

src/ImageDetails.jsx
jelly: That's a neat trick but the file isn't TypeScript at the moment. We do have some typescript in there but that feels like a follow up.

test/check-application
jelly: Right, we can also combine them easily as building can also input `localhost/label-image:latest`

src/ImageDetails.jsx
Venefilyn: Oh shoot, yeah didn't spot that :D

Venefilyn: New change you made LGTM in that case

PR #2438, opened 2026-02-12
Show container command for inactive quadlets

jelly:


src/detect-quadlets.py
```diff
@@ -29,7 +30,7 @@ def get_name(service_name: str, name: Optional[str]) -> str:
 
 
 def main(generator_dir: Path) -> None:
-    containers: DefaultDict[str, Dict[str, Optional[str]]] = defaultdict(dict)
+    containers: DefaultDict[str, Dict[str, Union[Optional[str], List[str]]]] = defaultdict(dict)
```
martinpitt, line 33: Can you please explain this? It looks like the podman API delivers the command as string, but your quadlet API emulation uses `shlex.split()` to parse that into an array. But in the UI we have to format that back to a string anyway. Why not keep it as a string in the first place?

jelly: It doesn't in the CLI?

```
               "Cmd": [
                    "toolbox",
                    "--log-level",
                    "debug",
                    "init-container",
                    "--gid",
```

jelly: And I checked again, its the same in the API

image

martinpitt: So how come it was just a `str` before? This really smells like you had to adjust this for making it an argv in the quadlet shim, while it was a simple str in the podman API. Otherwise, why the `shlex.split(value)`?

martinpitt: Oooh, now I get it! That type was only about quadlets.py all along.

PR #2439, opened 2026-02-13
workflows: Run browser integration tests

martinpitt:
This is just a proof of concept. It's missing image/npm/pixel git cache and posting to S3.

---

Moreover, this is *slow*. See [this run on my fork](https://github.com/martinpitt/cockpit-podman/actions/runs/21989748043/job/63533286891). The "build package" is slow mostly because of the .qcow2 download (there is no caching). If we ever want to actually *do* that, we need a solution for an image cache for workflows.

But running the integration test (at two tests in parallel) takes 14 mins, whereas on our CI [it takes 5](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2430-b6ed1db5-20260212-042305-fedora-43/log). So this is really only an option for small projects, like starter-kit and cockpit-{files,podman}.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,5 +1,5 @@
 name: CodeQL
-on: [push, pull_request]
+on: pull_request
 
 jobs:
   analyze:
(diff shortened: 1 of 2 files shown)


martinpitt:
@allisonkarlitskaya @jelly this is really just an FYI. I never planned to actually land this, but if there's ever trouble with TF/AWS, this is another arrow in our quiver. Might also make sense to move starter-kit to this approach?

PR #2452, opened 2026-02-28
Hide Quadlet template unit

jelly:
Hide the Quadlet template unit itself but not the instances.

### DIFF
--- a/src/detect-quadlets.py
+++ b/src/detect-quadlets.py
@@ -40,7 +40,8 @@ def main(generator_dir: Path) -> None:
         sys.exit(2)
 
     for entry in entries:
-        if not entry.name.endswith('.service'):
+        # Ignore non service and service template files
+        if not entry.name.endswith('.service') or entry.name.endswith('@.service'):
             continue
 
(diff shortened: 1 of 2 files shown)


src/detect-quadlets.py
```diff
@@ -40,7 +40,8 @@ def main(generator_dir: Path) -> None:
         sys.exit(2)
 
     for entry in entries:
-        if not entry.name.endswith('.service'):
+        # Ignore non service and service template files
+        if not entry.name.endswith('.service') or entry.name.endswith('@.service'):
```
tomasmatus, line 44: do we need to check for `@.service`? Does it actually exist for a quadlet template?

jelly: Yes, if I create a `foo@.container`, it will translate in generating a `foo@.service` and these are template units you can't start but only create an instance of.

tomasmatus: looked into it more, I get it now. Thanks!

tomasmatus (dismissed):
Looks good to me, retrying red ubuntu test.

PR #2500, opened 2026-04-17
Test cleanups

jelly:
Taken from https://github.com/cockpit-project/cockpit-podman/pull/2435

Basically use the `admin_s` helper only where needed as it relies on SSHConnection


jelly:
tumbleweed keeps failing so this needs investigation :eyeglasses:

jelly:
> tumbleweed keeps failing so this needs investigation 👓

Meanwhile fixed in other pull requests.

mvollmer (approved):
Thanks!

PR #2506, opened 2026-04-21
move registries.conf format to v2

Luap99:
Podman 6 is going to drop support for reading the old v1 format in https://github.com/containers/container-libs/pull/773

The v2 format is supported for a long time already so this can just be switched and still works with older versions.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -14,11 +14,23 @@ from lib.constants import DEFAULT_IDENTITY_PUB_FILE
 from machine.machine_core import ssh_connection
 
 REGISTRIES_CONF = """
-[registries.search]
-registries = ['localhost:5000', 'localhost:6000', 'subdomain.local.localhost:7000']
+unqualified-search-registries = ["localhost:5000", "localhost:6000", "subdomain.local.localhost:7000"]
 
-[registries.insecure]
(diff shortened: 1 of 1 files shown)


jelly (commented):
Looks good, but I'll need to check if Ubuntu 2204 supports this. Triggered the CI

jelly:
Ubuntu 2204 is on the way so that shouldn't be a too big issue but it is the oldest podman we test against afaik.

Luap99:
> Ubuntu 2204 is on the way so that shouldn't be a too big issue but it is the oldest podman we test against afaik.

I have not checked what podman version the v2 format was first added into podman but the code is from at least 2018 so I assume podman 3.4.4 (which I guess is ubuntu 2204?) has support for v2 just fine. https://github.com/containers/image/pull/429

jelly:
opensuse-tumbleweed is a known flake...

jelly (approved):

Luap99:
Can this be merged? I am getting close to merge the podman PRs which block the old syntax and then all cockpit tests will start to fail until we have this I think?

jelly:
> Can this be merged? I am getting close to merge the podman PRs which block the old syntax and then all cockpit tests will start to fail until we have this I think?

Yes, will merge it today

PR #2509, opened 2026-04-22
test: tumbleweed testLifeCycleOperations flake

jelly:


jelly:
Tumbleweed has another issue:

```
time="2026-04-22T10:19:11Z" level=warning msg="RunRoot is pointing to a path (/run/user/1000/containers) which is not writable. Most likely podman will fail."
```

Not sure whats going on here, selinux?

jelly:
> Tumbleweed has another issue:
> 
> ```
> time="2026-04-22T10:19:11Z" level=warning msg="RunRoot is pointing to a path (/run/user/1000/containers) which is not writable. Most likely podman will fail."
> ```
> 
> Not sure whats going on here, selinux?

This is still an issue but I can't easily debug it, the tests just keep failing randomly. It would be useful if we had the journal in case that happened.

tomasmatus (approved):
Thanks!

PR #2520, opened 2026-04-29
test: Terminate admin session gently during tear down

mvollmer:
By just waiting for the "user@1000.service" unit to be deactivated after terminating "admin" with loginctl.

The logout takes a long time and seems to hang on stopping "libpod-conmon-UUID.scope" but will eventually timeout after 44 seconds and finish successfully regardless.

Shooting SIGKILL at admin processes (as the regular tearDown method will do) seems to throw a wrench into the logout and processes often end up as "\<defunct\>". Maybe it's not a good idea to kill the user systemd process while it is logging out a session with non-trivial cgroups usage, *shrug*.

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -177,6 +177,35 @@ class TestApplication(testlib.MachineCase):
                 self.execute('for c in $(podman ps -aq); do echo "---- $c ----" >&2; podman logs $c >&2; done',
                              system=auth)
 
+        # cockpit-podman crashes when sessions are terminated
+        # while it is open, so let's log out here.
+        # TODO - find and fix those crashes
+        if self.browser.have_test_api():
+            self.browser.switch_to_top()
+            if self.browser.is_present("#toggle-menu"):
+                self.browser.logout()
+
+        # Gently bring down the admin session. This usually gets stuck
+        # on "conmon" and some "sleep" in a container and we could
+        # just wait for that to timeout and systemd to kill them
+        # forcefully. But we can help a bit by killing those processes
+        # early with SIGKILL.
+        #
+        # The session termination code in super.tearDown() is quite
+        # agressive with shooting "pkill -9" at all processes and that
+        # seems to upset something somewhere to the point where a
+        # number of processes owned by "admin" end up in a permanent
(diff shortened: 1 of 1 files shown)


mvollmer:
Ouch, now I know why we have the `pkill -u $u -9` in there. On fedora-coreos the TimeoutStopSec of user@.service is shorter than that of whatever gets stuck during logout, and it also has `TimeoutStopFailureMode=abort`, which gives us a noise coredump when user@1001.service takes too long to deactivate. 

Hmm, maybe we leave this all as it is and just deal with it. (long machine.execute timeout, ignore core dump.)

mvollmer:
Haha, and now we see the cockpit-podman crashes that happen during session termination. (We didn't see them before because the check for "Oops" happened before the session termination, but they nevertheless are real.)

martinpitt:
This duplicates a lot of stuff from nondestructiveSetup() in common testlib.py. Does `terminate-user` and a wait loop for conmon to go away help by any chance?

mvollmer:
> Does `terminate-user` and a wait loop for conmon to go away help by any chance?

Yes, the basic idea is to just do `loginctl terminate-user admn` and then wait for everything to be gone. This works nicely and does not get pid1 stuck.  Removing the call to `pkill` in `_terminate_sessions` also works, but I didn't want to change testlib just yet, since this seems to be specific to session podman, right?

The problems I am now running into are that conman and the sleep do not react to the initial SIGTERM, and the timeout for systemd to react and send a SIGKILL to those two processes is too long on fedora-coreos, 90 seconds. The timeout for the user@1001.service unit to stop is less than 90 seconds (60s?) and when that is expired, pid1 will send SIGABRT to the user systemd process, which produces a core dump.

This is all as intended by fedora-coreos, so maybe we just deal with it. But it also slow, so maybe selectively killing a few processes is ok.  Killing them all seems to be what triggers the cgroup draining bug.

Whatever we do, I'll see if I can find the Oopses that happen during session termination.

martinpitt:
I sent #2522 with my take on this.

mvollmer:
Let's go with #2522.

PR #2521, opened 2026-04-29
Typescriptify low hanging fruit

jelly:


src/ContainerHeader.tsx
```diff
@@ -8,7 +8,21 @@ import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core/dis
 import cockpit from 'cockpit';
 const _ = cockpit.gettext;
 
-const ContainerHeader = ({ users, ownerFilter, handleOwnerChanged, textFilter, handleFilterChanged }) => {
+// TODO: move to app.tsx when ported to TypeScript
+interface User {
```
tomasmatus, line 12: just a note, the `User` type is more complex but for needs of this file it is fine.

When initiated in app.jsx it contains these props:
`{ con: null, uid: 0, name: _("system"), dbus: null }`

tomasmatus (approved):
Thanks!

src/ContainerHeader.tsx
jelly: Good point! Yeah app.jsx is going to be interesting so we'll figure it out then.

PR #2522, opened 2026-04-29
test: Stop user containers after each test run

martinpitt:
So far we've let testlib.py's nonDestructiveSetup() take care of this through killing user sessions and processes wholesale. But kernel 6.9.11 has a regression [1][2] which locks up the kernel when doing that.

So let's be less brutal and clean up dangling containers orderly.

For that to work without oopsing cockpit, log out first.

[1] https://lore.kernel.org/regressions/afHNg2VX2jy9bW7y@piware.de/T/#u
[2] https://github.com/cockpit-project/bots/pull/8970#issuecomment-4342147158

----

This is my take on #2520 to fix https://github.com/cockpit-project/bots/pull/8970. Locally and on CI it looks promising!

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -177,6 +177,22 @@ class TestApplication(testlib.MachineCase):
                 self.execute('for c in $(podman ps -aq); do echo "---- $c ----" >&2; podman logs $c >&2; done',
                              system=auth)
 
+        # HACK: avoid failing tests on oopses when cleaning up containers below
+        if self.browser.have_test_api():
+            self.browser.switch_to_top()
+            if self.browser.is_present("#toggle-menu"):
+                self.browser.logout()
+
+        # Stop user containers; testlib's _terminate_sessions() kills the users, but this is cleaner
+        # https://lore.kernel.org/regressions/afHNg2VX2jy9bW7y@piware.de/T/#u
+        self.machine.execute("""
+            cd /tmp
+            for u in $(loginctl --no-legend list-users | awk '{ if ($2 != "root") print $1 }'); do
+                username=$(id -nu $u)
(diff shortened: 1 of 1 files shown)


mvollmer:
> This is my take on #2520. Locally it looks promising!

Yeah, #2520 turned into quite a hairball. If this works as well, I am all for it.

mvollmer:
Oh no, crashes when removing the containers from the outside...

martinpitt:
@mvollmer right -- I think we have to combine that with your idea of logging out of cockpit first.

test/check-application
```diff
@@ -177,6 +177,9 @@ class TestApplication(testlib.MachineCase):
                 self.execute('for c in $(podman ps -aq); do echo "---- $c ----" >&2; podman logs $c >&2; done',
                              system=auth)
 
+        if self.browser.have_test_api():
+            self.browser.logout()
```
mvollmer: One test is already logged out when it gets here, actually.

martinpitt: ah right, [here](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2522-132bbf76-20260429-132511-rhel-8-10-ws-container/log.html), as that's rhel-8-10 specific. Fixed.

martinpitt:
But in general this [seems to do it](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2522-132bbf76-20260429-132516-fedora-44-bots%238970/log.html) :tada: If this goes green (enough), I'll clean up the commits.

mvollmer (approved):
Thanks, I prefer this to #2520 since it is less specific and doesn't duplicate _terminate_sessions from testlib.

PR #2525, opened 2026-04-30
move registries.conf into drop in directory

Luap99:
Follow up to[1], with the packaging changes for new config file work[2] we now have vendor overwrites as drop 00-vendor.conf. That file is parsed after /etc/containers/registries.conf and as such the unqualified-search-registries setting is overwritten again.

To ensure the cockpit test file is parsed last put it into the drop in directory as 99-cockpit-podman-test.conf which should also help to make it clear what own/created the file.

This /etc/containers/registries.conf.d directory has been supported for a while already so this works not just with podman 6 but also the previous ones making so not version logic is needed here.

[1] 772987d (move registries.conf format to v2)
[2] https://github.com/containers/container-libs/pull/795

### DIFF
--- a/test/check-application
+++ b/test/check-application
@@ -324,7 +324,8 @@ class TestApplication(testlib.MachineCase):
         """)
 
         # Add local insecure registry into registries conf
-        m.write("/etc/containers/registries.conf", REGISTRIES_CONF)
+        m.execute("mkdir -p /etc/containers/registries.conf.d")
+        m.write("/etc/containers/registries.conf.d/99-cockpit-podman-test.conf", REGISTRIES_CONF)
         m.execute("systemctl stop podman.service")
 
     def createQuadlet(self, name: str, podName: str | None = None, containerName: str | None = None,


Luap99:
looks like packit tests passed here so lets see if I can tests against the containers-common rpm changes

Luap99:
```
  Problem 2: package cockpit-podman-125.11.gbcdf931-1.fc45.noarch from @commandline requires podman >= 2.0.4, but none of the providers can be installed
  - installed package podman-5:5.8.2-1.fc45.x86_64 conflicts with podman < 5:6 provided by containers-common-102:0.68.0~dev-1.20260429142641348389.pr795.735.g22c059f8fc.fc45.noarch from @commandline
  - package containers-common-102:0.68.0~dev-1.20260429142641348389.pr795.735.g22c059f8fc.fc45.noarch from @commandline conflicts with podman < 5:6 provided by podman-5:5.8.2-1.fc45.x86_64 from rawhide
  - package containers-common-102:0.68.0~dev-1.20260429142641348389.pr795.735.g22c059f8fc.fc45.noarch from @commandline conflicts with podman < 5:6 provided by podman-5:5.8.2-1.fc45.x86_64 from testing-farm-tag-repository
```

that is expected as the rpm changes only work with podman. I you do not use podman-next packages then I guess we cannot test this here and it does not seem to be testable on  https://github.com/containers/podman/pull/28601 either.

Anyhow the change here should be safe enough and you just have to trust it then I guess.

martinpitt:
This indeed fixes the regression with podman-next (the upstream cockpit revdeps failures), see [here](https://github.com/containers/podman/pull/28631#issuecomment-4369929403). I triggered the integration tests, might be that the drop-in directory does not yet work on older releases.

martinpitt (approved):
Thanks!

PR #2542, opened 2026-05-18
HACKING: Start debugging tips

martinpitt:
I've used these a lot over the years to create upstream bug reports or do interactive testing.

---

[rendered](https://github.com/martinpitt/cockpit-podman/blob/debug-tips/HACKING.md#debugging-tips)


jelly (approved):
Thanks!

PR #2559, opened 2026-06-10
Revert "test: Stop user containers after each test run"

jelly:
This reverts commit 15c074163aa6bcf1e8dedce5eb8900b8281a00bd.

Stopping containers and logging out Cockpit makes the test failures show a test failure image of the login page and not the actual state during the test.


Venefilyn (approved):
If it works, LGTM

jelly:
> If it works, LGTM

Seems there is an issue still with ubuntu-2604's kernel.

PR #2564, opened 2026-06-18
URL parameter support for the "download images" dialog

ochosi:
This is a POC of adding URL parameter support to the "download images" dialog. With the changes on this branch, you can open the dialog and pre-populate all the fields (owner, registry, search term, tag).

URL format: `#?action=download-image&search=<term>&registry=<registry>&owner=<system|user>&tag=<tag>`

All parameters except 'action' are optional. The dialog automatically:
- Pre-fills search term and triggers search (default: unspecified)
- Selects specified registry (default: unspecified, so "all registries")
- Chooses system or user owner (default: user)
- Pre-fills image tag (default: latest)

I know it convolutes this branch a little, but I also integrated a spinner as a separate commit. This part could be a less controversial and easy-to-merge change. If you like, I'd be happy to extract it into a separate pull request.

### DIFF
--- a/src/ImageSearchDialogOpener.jsx
+++ b/src/ImageSearchDialogOpener.jsx
@@ -1,0 +1,76 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+import React, { useState, useEffect } from 'react';
+
+import cockpit from 'cockpit';
+
+import { ImageSearchModal } from './ImageSearchModal.jsx';
+
+/**
+ * Component that renders the ImageSearchModal when URL parameters are present.
+ * Uses declarative rendering based on URL state (the URL is the single source of truth).
+ *
+ * Supported URL formats:
+ * - #?action=download-image
+ * - #?action=download-image&search=cockpit
+ * - #?action=download-image&registry=docker.io&search=cockpit&owner=user&tag=latest
+ *
+ * Parameters:
+ * - action: 'download-image' (required to open dialog)
+ * - registry: Registry to pre-select (optional)
+ * - search: Search term to pre-fill and auto-search (optional)
+ * - owner: 'system' or 'user' (optional, defaults to 'user')
+ * - tag: Tag to pre-fill (optional, defaults to 'latest')
+ */
+export const ImageSearchDialogOpener = ({ downloadImage, users }) => {
+    const [location, setLocation] = useState(cockpit.location);
+
+    useEffect(() => {
+        const handler = () => setLocation({ ...cockpit.location });
+        cockpit.addEventListener("locationchanged", handler);
+        return () => cockpit.removeEventListener("locationchanged", handler);
+    }, []);
+
+    const { options } = location;
+    const action = options.action;
+
+    // Determine if dialog should be shown
+    const showDialog = action === 'download-image';
+
+    // Don't render until users array is populated with connections to avoid flicker
+    // The users array is initialized with dummy data (con: null) and later updated with real connections
+    if (!showDialog || !users || users.length === 0 || !users[0].con) return null;
+
+    // Parse URL parameters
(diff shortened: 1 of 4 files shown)


ochosi:
Related to https://github.com/osbuild/image-builder-frontend/pull/4540

src/ImageSearchModal.jsx
```diff
@@ -45,21 +57,45 @@ export const ImageSearchModal = ({ downloadImage, users }) => {
     // Registries to use for searching
     const searchRegistries = registries?.search && registries.search.length !== 0 ? registries.search : fallbackRegistries;
 
+    // Handle initial search from URL parameters
+    useEffect(() => {
```
jelly, line 61: For this we also have `useInit()` whcih does not need the eslint-disable.

src/ImageSearchModal.jsx
```diff
@@ -124,13 +160,20 @@ export const ImageSearchModal = ({ downloadImage, users }) => {
         const selectedImageName = imageList[selected].Name;
         if (activeConnection)
             activeConnection.close();
-        Dialogs.close();
+        // Use onExternalClose if provided (URL-based rendering), otherwise use Dialogs.close (modal rendering)
+        if (onExternalClose) {
+            onExternalClose();
+        } else {
+            Dialogs.close();
+        }
```
jelly, line 168: This should still call `Dialogs.close()`

src/ImageSearchModal.jsx
```diff
@@ -26,13 +26,25 @@ import './ImageSearchModal.css';
 
 const _ = cockpit.gettext;
 
-export const ImageSearchModal = ({ downloadImage, users }) => {
+export const ImageSearchModal = ({ downloadImage, users, initialSearchTerm = "", initialRegistry = "", initialOwner = "user", initialTag = "latest", onExternalClose }) => {
```
jelly, line 29: Nit: onExternalClose? That's a bit vague, what about onCloseCallback or something alike, that's what it is.

src/ImageSearchDialogOpener.jsx
```diff
@@ -0,0 +1,76 @@
...
+ * Supported URL formats:
+ * - #?action=download-image
+ * - #?action=download-image&search=cockpit
+ * - #?action=download-image&registry=docker.io&search=cockpit&owner=user&tag=latest
+ *
+ * Parameters:
+ * - action: 'download-image' (required to open dialog)
+ * - registry: Registry to pre-select (optional)
+ * - search: Search term to pre-fill and auto-search (optional)
+ * - owner: 'system' or 'user' (optional, defaults to 'user')
```
jelly, line 21: We support multiple users by uid, so maybe this should act the same although the semantics of it are terrible (ie. uid=0 == system, but owner=user == logged in user, owner=1002 another user running podman.socket

src/ImageSearchDialogOpener.jsx
```diff
@@ -0,0 +1,76 @@
...
+ * - tag: Tag to pre-fill (optional, defaults to 'latest')
+ */
+export const ImageSearchDialogOpener = ({ downloadImage, users }) => {
+    const [location, setLocation] = useState(cockpit.location);
```
jelly, line 25: This feels a bit unwieldy, feels like it doesn't have to be a component and could be a bit more elegant implemented.

jelly (commented):
Sorry for the late review, there are some issues which need to be addressed:

The registry parameter allows setting an arbitrary registry but that won''t be shown in the UI which means will be confusing:

Image

ochosi:
I don't expect to have time to finalize this in the coming weeks, so closing for now. If you like the spinner, feel free to cherry-pick it, it's a tiny commit :)

PR #2620, opened 2026-07-31
Extract inline functions from the render method

jelly:
This makes the render function easier to read and doesn't make React create the same function on every render.


tomasmatus (approved):
Looks good!

PR #2641, opened 2026-08-12
test: Drop Ubuntu 24.04 test workarounds

jelly:
Cockpit no longer tests on Ubuntu 24.04.


tomasmatus (changes requested):
Looks good, but I found one more thing that mentions ubuntu-2404. I think the comment just needs a bit of editing

```
❯ git --no-pager grep -i "ubuntu-2404"
test/check-application:        # Wait for it to properly go away before waitRestart, on Ubuntu-2404 restarting is slow.
```

tomasmatus (approved):
assuming green :)

PR #2652, opened 2026-08-18
create: Fix environment index calculations

### LINKED ISSUE #2637: Envvar pasting crashes if a later entry is deleted
Repro

1. Click Create Container
2. Go to Integration -> Environment variables
3. Click Add Variable twice to create two entries
4. Delete the bottom entry
5. Paste `DURIAN=LEMON TEST_URL=wss://cockpit/?start=1&stop=0` or something like it in the first entry
6. See crash

This happens as `DynamicListForm` soft-deletes entries and clears it out with an `undefined` instead of removing it

mvollmer:
Calling additem() will always add the new row at the end, not after the one that gets pasted into.  We can compute which that is from the length of the state value.

This works also when the state value for the environment variables a sparse array.  Previously, we might have used an index that is actually a hole in the array and then crash.

Fixes #2637

### DIFF
--- a/src/Env.jsx
+++ b/src/Env.jsx
@@ -32,31 +32,34 @@ export function validateEnvVar(env, key) {
     }
 }
 
-const handleEnvValue = (key, value, idx, onChange, additem, _itemCount, companionField) => {
+const handleEnvValue = (key, value, idx, onChange, additem, itemCount, companionField) => {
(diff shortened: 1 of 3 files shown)


Venefilyn (approved):

Venefilyn:
Works great, thanks!

PR #2655, opened 2026-08-20
test: Allow installing updates from a vendor

jelly:
New DNF5 policy does not allow switching vendors if in our example a COPR repository offers a newer version of cockpit-podman. We want to override this behaviour and in reverse dependency testing install the latest version of cockpit-podman (provided by @cockpit/main-builds).

We need the latest version of cockpit-podman in our reverse dependency test against podman so we can update to changes in podman's main repository.

https://fedoraproject.org/wiki/Changes/DisableVendorChangeByDefault

---

This should fix the recent fallout of podman reverse dependency testing failing on Fedora rawhide.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -14,7 +14,7 @@ fi
 main_builds_repo="$(ls /etc/yum.repos.d/*cockpit*main-builds* 2>/dev/null || true)"
 if [ -n "$main_builds_repo" ]; then
     echo 'priority=0' >> "$main_builds_repo"
-    dnf distro-sync -y --repo 'copr*' cockpit-podman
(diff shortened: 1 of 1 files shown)


mvollmer:
> We need the latest version of cockpit-podman in our reverse dependency test against podman so we can update to changes in podman's main repository.

Instead of "latest version", I would say "most recent build of the main branch".  "The latest version" can easily be misunderstood as "the most recent release".

(And the reason we need "the most recent build of the main branch" is because we running the tests from the current head of the main branch.

An alternative would be to run the tests from the exact commit that was built into the cockpit-podman rpm that happens to be installed. I think that would be more robust (no need for a copr) and also remove the remaining race (the "most recent copr build from the main branch" is not necessarily from the same commit as the current head of main when the tests run). But I don't know if that is easily achievable.)

mvollmer (approved):
Thanks!

PR #2674, opened 2026-09-09
workflows: Switch runners to ubuntu-latest

tomasmatus:


jelly (approved):
Thanks, this should be fine as all other projects do the same.

PR #2675, opened 2026-09-10
podman: preserve health row scope and disposition

CompleteDotTech:
Makes container health rows preserve their real Podman owner context and render lifecycle-only dispositions explicitly. A session-user rootless row now carries its verified owner UID/context through inventory, inspect, event, and Quadlet paths; retired and infrastructure rows show their own labels instead of “Health unknown.”

Validation:
- `node --test test/health-state.test.mjs`
- `node --test test/scheduler-refresh.test.mjs`
- `git diff --check`

Source review independently cleared the scope and disposition fixes. This is source-only; live Cockpit/browser validation remains pending.

### DIFF
--- a/src/ContainerDetails.jsx
+++ b/src/ContainerDetails.jsx
@@ -7,6 +7,7 @@ import { Flex, FlexItem } from "@patternfly/react-core/dist/esm/layouts/Flex";
 
 import cockpit from 'cockpit';
 
+import { containerScope } from './health.js';
 import * as utils from './util.js';
 
 const _ = cockpit.gettext;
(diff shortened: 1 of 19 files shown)


jelly:
Please explain in plain english what this does exactly and maybe show a screenshot of the new functionality. From the pull request description I can't make up what this adds to the plugin UI.

Venefilyn:
Does this fix any reported issues so far? What is it solving and why was it an issue before?

The commits themselves are also lacking in descriptions and the codebase throughout is missing explanations for what is going on

There's a lot of code added and I can't tell what this aims to solve that we didn't have before

CompleteDotTech:
Thank you — this was the wrong venue for this work. The change adds extra scope and disposition text to Cockpit Podman's container-health rows so an operator can see whether a result came from the rootful or rootless Podman context and avoid conflating lifecycle state with health state. It was intended as local health-coverage work for our own deployment, not a contribution that Cockpit Podman should carry.

I apologize for opening a large, under-explained PR here and for not providing a clear before/after explanation or screenshot. I am closing it and will recreate the work only in our internal repository.

PR #2678, opened 2026-09-13
Fix image details and resource cleanup after dropping privileges

fivetime:
Fix two cases where resource state causes the Podman page to fail:

- A local manifest list appears in the image listing, but inspecting it returns a member image with a different ID. Key inspect results by the requested image ID so the manifest entry receives its details and the member is not overwritten. Initialize Ports and Env even when Config is absent, avoiding the ImageDetails undefined-length error.
- Dropping administrative access can reject re-initialization of the system connection before its old event stream closes. Remove its resources together with the user entry, including quadlet containers and pods, to avoid rendering rows with no matching user (`User not found for container uid 0`).

The browser regression test creates a manifest list, checks that both it and its member remain visible, opens its details, and removes it.

Validation: ESLint, the repository TypeScript checker (`test/common/typecheck`), and the production build pass using Node.js 22. The modified Python test parses successfully. Full VM/browser integration tests have not been run locally.


src/app.jsx
```diff
@@ -433,7 +433,7 @@ class Application extends React.Component {
 
     cleanupAfterService(con) {
         debug("cleanupAfterService", con.uid, "current owner filter:", this.state.ownerFilter);
-        ["images", "containers", "pods"].forEach(t => {
+        ["images", "containers", "pods", "quadletContainers", "quadletPods"].forEach(t => {
```
jelly, line 436: This is good, and likely the fix for `Error: Assertion failed: User not found for container uid 0` when dropping privileges.

Please put it in a separate commit, with a test in `testQuadletUser` we can `drop_privileges` there and it should survive.

src/app.jsx
```diff
@@ -650,6 +650,10 @@ class Application extends React.Component {
             if (!system || err.problem != 'access-denied')
                 console.warn("init uid", uid, "getInfo failed:", err.toString());
 
+            // On dropping admin privileges the system resources stay in the state until the
+            // previous connection's event stream closes; forget them together with the user,
+            // otherwise their rows render for a uid without a user in the meantime
+            this.cleanupAfterService({ uid });
```
jelly, line 656: I am not entirely convinced by this change, if this fails on an initial `init` there won't be any users state to speak of. If this fails when superuser changes I am not entirely sure what would happen, as this is already handled by streamEvents.

src/client.ts
```diff
@@ -86,6 +86,9 @@ export function resizeContainersTTY(con: Connection, id: string, exec: boolean,
 function parseImageInfo(info: JsonObject): JsonObject {
     const image: JsonObject = {};
 
+    // the details view relies on these lists, so keep them present even without Config
+    image.Ports = [];
+    image.Env = [];
```
jelly, line 91: Alternatively, we fix the view which is imo expected.

```Javascript
image?.Ports?.length !=== 0
```

And
```
    if (image)
        env = <ContainerEnv containerEnv={container.Config.Env} imageEnv={image.Env} />;
```

src/client.ts
```diff
@@ -114,11 +117,14 @@ export function getImages(con: Connection, id?: string) {
 
                 return Promise.all(promises)
                         .then(replies => {
-                            for (const info of replies as JsonObject[]) {
-                                const imageId = info.Id as string;
+                            (replies as JsonObject[]).forEach((info, i) => {
```
jelly, line 120: This needs discussion if it makes sense to expand an image list to multiple or one container. Where have you encountered an image list in the wild?

jelly (changes requested):

